PR #28, opened 2020-02-16
Add BACKEND_API and ELECTRS_API_URL ENV lines

rbrooklyn:
Adds BACKEND_API and ELECTRS_API_URL environment variables which will be populated into backend/mempool-config.json on startup.

This will allow docker users to easily switch to electrs mode and specify a server if they have a need to do so.


Dezzj (commented):
Veiwed

PR #69, opened 2020-04-27
Frontend static directroy is mempoolspace

Emzy:
fixed dist/mempool/ to dist/mempoolspace/


softsimon:
Good catch! But I discovered that it's the project name that is faulty. So I just changed the project name back to just 'mempool' and the current Readme should be correct again.

PR #107, opened 2020-09-13
Bisq markets API

softsimon:
Re-implementing the current legacy Bisq Market PHP API as a NodeJS in the existing Mempool backend along with the Bisq Dao Data API.

The API should contain the same API endpoints give the exact same input parameters aligned with the current API Spec https://markets.bisq.network/api/

Some notable changes from the old API:
* The "basecurrency" param is dropped. BTC is the only base currency
* The "format" param is dropped and the output is always pure JSON
* On error, a correct HTTP status code is now sent to the client, eg 400 Bad Request instead of 200 OK

### DIFF
--- a/production/mempool-config.bisq.json
+++ b/production/mempool-config.bisq.json
@@ -14,7 +14,7 @@
   "TX_PER_SECOND_SPAN_SECONDS": 150,
   "ELECTRS_API_URL": "http://[::1]:3000",
   "BISQ_ENABLED": true,
-  "BSQ_BLOCKS_DATA_PATH": "/bisq/data",
+  "BISQ_BLOCKS_DATA_PATH": "/bisq/data",
   "SSL": false,
   "SSL_CERT_FILE_PATH": "/etc/letsencrypt/live/mysite/fullchain.pem",
   "SSL_KEY_FILE_PATH": "/etc/letsencrypt/live/mysite/privkey.pem"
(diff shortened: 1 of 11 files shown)


production/mempool-config.bisq.json
```diff
@@ -14,7 +14,7 @@
   "TX_PER_SECOND_SPAN_SECONDS": 150,
   "ELECTRS_API_URL": "http://[::1]:3000",
   "BISQ_ENABLED": true,
-  "BSQ_BLOCKS_DATA_PATH": "/bisq/data",
+  "BISQ_BLOCKS_DATA_PATH": "/bisq/data",
```
wiz, line 17: Don't we need some additional new settings here?

production/mempool-config.bisq.json
```diff
@@ -14,7 +14,7 @@
   "TX_PER_SECOND_SPAN_SECONDS": 150,
   "ELECTRS_API_URL": "http://[::1]:3000",
   "BISQ_ENABLED": true,
```
wiz, line 16:
```suggestion
  "BISQ_ENABLED": true,
  "BISQ_BLOCKS_DATA_PATH": "/bisq/seednode-data/btc_mainnet/db/json",
  "BISQ_MARKET_ENABLED": false,
  "BISQ_MARKETS_DATA_PATH": "/bisq/seednode-data/btc_mainnet/db",
```

production/mempool-config.bisq.json
wiz, line 17:
```suggestion
  "BISQ_ENABLED": true,
  "BISQ_BLOCKS_DATA_PATH": "/bisq/seednode-data/btc_mainnet/db/json",
  "BISQ_MARKET_ENABLED": false,
  "BISQ_MARKETS_DATA_PATH": "/bisq/seednode-data/btc_mainnet/db",
```

wiz (approved):
ACK

PR #138, opened 2020-10-12
Update README with repository checkout guidance

lucasmoten:
- Changed repository to clone based on updated organization
- Determine latest release from github api call.  An alternative form being `latestrelease=$(git describe --exact-match --abbrev=0)` to yield the most recent tag name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,9 +33,10 @@ You should see mempool starting up, which takes over an hour (needs 8 blocks). W
 
 ## Checking out release tag
 ```bash
-  git clone https://github.com/mempool-space/mempool.space
-  cd mempool.space
-  git checkout v1.0.0 # put latest release tag here
+  git clone https://github.com/mempool/mempool
+  cd mempool
+  latestrelease=$(curl -s https://api.github.com/repos/mempool/mempool/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")')
+  git checkout $latestrelease
 ```
 
 ## Bitcoin Core (bitcoind)
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ You should see mempool starting up, which takes over an hour (needs 8 blocks). W
 ```bash
   git clone https://github.com/mempool/mempool
   cd mempool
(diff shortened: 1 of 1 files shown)


wiz:
Hey lucas, thanks for the PR. After we release v2.0 we will definitely have to rewrite the docs, and I'm actually working on a proper installation script, so the docs will probably change completely for that too.

For now, I agree we should update the repo URL and use latest release tag (currently v1.0.1), but I tried your command, but it seems that only works on GNU grep, so I'll suggest changing it to a more portable implementation.

For example on FreeBSD:
```
curl -s https://api.github.com/repos/mempool/mempool/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")'
grep: The -P option is not supported
```
And on macOS:
```
 % curl -s https://api.github.com/repos/mempool/mempool/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")'
usage: grep [-abcDEFGHhIiJLlmnOoqRSsUVvwxZ] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
```

README.md
```diff
@@ -33,9 +33,10 @@ You should see mempool starting up, which takes over an hour (needs 8 blocks). W
 
 ## Checking out release tag
 ```bash
-  git clone https://github.com/mempool-space/mempool.space
-  cd mempool.space
-  git checkout v1.0.0 # put latest release tag here
+  git clone https://github.com/mempool/mempool
+  cd mempool
+  latestrelease=$(curl -s https://api.github.com/repos/mempool/mempool/releases/latest | grep -oP '"tag_name": "\K(.*)(?=")')
```
wiz, line 38:
```suggestion
latestrelease=$(curl -ssL https://api.github.com/repos/mempool/mempool/releases/latest|grep tag_name|head -1|cut -d '"' -f4)
```

wiz (approved):
ACK

PR #143, opened 2020-10-23
update legend state after chart update

TechMiX:
Hello,

In the Graphs tab, after changing the time span from the default, if you click on the legends on the side, the chart goes back to the default time span. It's because the legend plugin was not updated after the chart was updated.

This pull request fixes this, so that you can work with the legends after changing the time span.


softsimon:
Tested it and it seems to be working. Thanks a lot for the contribution!

PR #154, opened 2020-11-13
Fix graph legend bug on resize

TechMiX:
I spotted this weird bug on the graph page that if you inactive some legends and then resize the page, all the legends will get disappeared.

This PR will fix it.


softsimon:
Seems to work!

PR #187, opened 2020-11-22
Angular Universal Milestone 1

softsimon:
For the first milestone.. these features should be complete:

- [x] Merge new init-api to master
- [x] Go through all pages and make sure they are pre-renderable (fix document, window, setTimeout etc)
- [x] Support the "generated config" for the correct network dropdown menu
- [x] Make API url's configurable on the frontend AND server.ts universal backend
- [x] Add TransferHttpCacheModule and enable transfer of data to remove the "flickering" and double loading
- [ ] Make sure angular-universal is mergable to master without any conflicts with the default version
- [ ] NoScript tags must be handled in both versions

Not included:
- [ ] Investigate if it's possible to pre-render the chartist graph on the backend
- [ ] Support and proxy other networks than mainnet
- [ ] Make unique URLs for tabs and paginations
- [ ] Make nginx pre-cache the pre-rendered content for top 10 or 20 most popular pages every 5 seconds to remove SSR lag


wiz:
angular frontend flow

wiz (changes requested):
It seems SSR got broken when i18n support was merged, so I pushed some fixes and got it running on mempool.ninja with SSR enabled. However the i18n locale selector doesn't work because `en-US` is hard-coded in `server.ts` now, so we need to implement a way to serve different templates depending on URL prefix of `/sv` or whatever locale, and also the site feels quite slow which I suspect might be due to the AU cache missing. Please take a look at .ninja and the fixes I pushed.

wiz:
The solution I ended up going with for AU SSR with i18n is that server.ts is no longer the entrypoint, instead we run server.run.ts which creates one instance of the app in server.ts for each locale, and maps those apps below their `/<locale>` URL namespace to a master express app instance - it seems to work okay but the build time is significantly longer now as the SSR build is now also localized. Going to test it a bunch more before merging.

wiz (approved):
ACK - SSR works well enough for SEO and static works same as before, seems good enough to merge

PR #251, opened 2020-12-03
i18n all the remaining strings.

softsimon:
Located and added i18n on almost all the remaining strings.

* Titles
* Tooltips
* Bisq and Liquid
* SEO titles


frontend/src/app/components/block/block.component.ts
```diff
@@ -95,7 +95,7 @@ export class BlockComponent implements OnInit, OnDestroy {
       tap((block: Block) => {
         this.block = block;
         this.blockHeight = block.height;
-        this.seoService.setTitle('Block: #' + block.height + ': ' + block.id);
+        this.seoService.setTitle($localize`:@@block.component.block:Block` + ' ' + block.height + ': ' + block.id);
```
wiz, line 98: All the parameters need to be part of the string to be translated, so that translators can move the parameters around to the front or middle or back

wiz (approved):
ACK

PR #262, opened 2020-12-05
add 'sat' string for i18n

TechMiX:


wiz:
will merge in a few days we are doing a strings freeze now

TechMiX:
This doesn't add any new string, it just use an already existing translated string.

PR #311, opened 2021-01-27
Very basic desktop notifications implementation

tomat:
Minimum viable implementation

Probably needs:

- [x] Tweak the source English strings and add i18n tags
- [x] Maybe centralize notification API for re-use
- [x] Better placement of button
- [x] Actually design the button or make it not a button or whatever

For now hides notification when tab becomes visible, and does not send notifications if the tab is already visible. Does not ask any permissions until button is pressed. Browsing away from the address page will disable notifications.


wiz:
Cool. For starters we probably want to tweak the source English strings and add i18n tags

tomat:
I assume I don't need to regenerate the xlf files myself? Got a bunch of unrelated diffs there when running `i18n-extract-from-source` 😄

tomat:
Should be good to go, seems "desktop notifications" is the canonical term so went with that

wiz:
Closing due to stalled, can re-open this again in the future if necessary

Issue #272, opened 2020-12-16
Improvements to locale selector


wiz:
Just some ideas for now, but the locale selector needs to have:

* Country (automatically populated from GeoIP-DB lookup, but can be overriden)
* Currency (automatically populated from country, but can be overridden)
* Language (automatically set from browser accept-language header, but can be overridden)
* Fallback for NoScript/NoJS or SEO when JS is disabled, to a simple and properly tagged `<a href="/ja/" hreflang="ja">Japanese</a>` style links

For some design inspiration, here is how fs.com does it which is kinda cool:

<img width="300" alt="Screen Shot 2020-12-17 at 5 57 40" src="https://user-images.githubusercontent.com/232186/102405650-d323a780-402c-11eb-8bfa-2473b6f945ab.png">

PR #313, opened 2021-01-28
Currency chooser

tomat:
Currently uses the default currency for the chosen language, only added SEK to backend for now

- [x] Add all currencies we should have support for to backend (check which currencies are default on the supported languages, and make sure they actually have price data)
- [x] Add currency chooser somewhere (next to language chooser?) to make that a separate choice
- [x] Fix so language is never used to select the default value, or to deduce when to unset the `currency` cookie
- [x] Fix so default is USD for now, regardless of `lang`

Some considerations;
1. How do we store the currency choice? Language seems to use cookie `lang`, so maybe cookie `currency`? But only set it when you explicitly choose a currency, not just when changing languages, so the language currency is still default if you never touch the currency chooser
- Using `currency`, and set it when choice is not USD
2. Do we clear the `currency` cookie when changing currency to the default of the current language?
- No, never check against language, just clear it when choosing USD


tomat:
Now stores currency choice as cookie `currency`, and removes that cookie if you choose the default currency for your language.

wiz:
Cool. I'd like to automatically detect the user's country using a geoip db lookup, and then we can set the currency from the combined locale of language-country

wiz:
For the UI layout, I wrote some ideas in #272 which might be interesting

tomat:
> Cool. I'd like to automatically detect the user's country using a geoip db lookup, and then we can set the currency from the combined locale of language-country

Yeah I would consider that out of scope for this PR, the currency defaults to whatever locale is used, so if we add a geoip-detector for `lang` then the currency will be automatically set to the currency of that locale. May need a few tweaks for EUR since there are multiple locales for EUR (right now `fr` is used) but I didn't want to show duplicates in the dropdown so just commented them out for now.

tomat:
> * [ ]  Fix so EUR is correctly default selected for other languages than `de`
> * [ ]  Make the language chooser have "Language / Currency" pairs instead of just currency to follow the locale formatting

Marking as draft until I've fixed these

wiz:
>  if we add a geoip-detector for lang then the currency will be automatically set to the currency of that locale

Well it's important to clarify that language, country, and currency are all completely unrelated things, and each should have their own pull-down box selector... as for their default values,

* the default language is requested by the user's browser specifically using the `Accept-Language` HTTP header so we don't need to guess that
* the default country can be determined from a geoip db lookup
* the default currency can be determined from a hard-coded mapping of the country to its fiat currency

I think it would be easy to configure our nginx to set these 3 variables as HTTP cookies, so for now if you want to make the pull-down boxes for those 3 as separate things they should integrate nicely when I get the nginx.conf working :)

tomat:
So since there's no geoip db lookup yet, how should default country be determined? Deduced from language?

Also, what will country decide in terms of actual content/formatting on the site?

wiz:
> how should default country be determined? Deduced from language?

no, that's not how the world works heh. but I can configure nginx to set a country cookie based on geoip db lookup quite easily.

tomat:
Any suggestions on how we can do this PR until a country chooser is available? Just default to USD and call it a day?

tomat:
Now entirely independent from language or any other choice, defaults to USD, and the cookie is reset if USD is chosen again

wiz:
Closing due to stalled, can re-open this again in the future if necessary

PR #321, opened 2021-02-02
Update dockerization

bguillaumat:


wiz:
This is not the correct way to implement Dockerfile support

PR #322, opened 2021-02-02
Update dockerization

bguillaumat:
Every needed files are in the subdir (Backend/Frontend/MySQL) to be compliant with docker hub


wiz (approved):
utACK

PR #343, opened 2021-02-14
Make clear protection timeout configurable.

### LINKED ISSUE #335: Bitcoind: Consider re-enabling mempool clear protection
When rebooting, the Bitcoin Core mempool might be empty for 10+ minutes before it's back again so the mempool may be cleared and rebroadcasted.

Maybe enable clear protection as default only at launch?

softsimon:
fixes #335


wiz (approved):
ACK - tested OK on mempool.ninja

PR #347, opened 2021-02-17
Fix style on block hover

Czino:
I found that when you hover over the block there is a small blue line over the s of "sat".

This is just a minor style bug fix


softsimon:
Thanks!

PR #362, opened 2021-02-25
Update docker part

bguillaumat:
This PR handle:
- The update of the cache dir in config.ts 32290d1e0da2102acf289ae7dc671c4a07f9671f
- The CLEAR_PROTECTION_MINUTES
- The activation of TLS for electrum via env var
- README update

@wiz do you think it would be better to rename ELECTRS to ELECTRUM to be more compliant with all the ELECTRUM possibility?


wiz:
Hey @bguillaumat - thanks for another great PR. Yeah, now that we have more understanding of the RPI performance after running it for a few months, the mempool clear protection could be increased to 30 minutes, @softsimon what do you think? As for the other settings they look good. IIRC we used ELECTRS because it was written for https://github.com/romanz/electrs but if it works on other electrum servers maybe that makes sense... not too familiar with those tbh.

softsimon (approved):

softsimon:
* 20 minutes is a good default for mempool protection
* We already name the backend config setting `BACKEND: "electrum"` and the settings block is called `ELECTRUM` ... so I guess it can make sense to update the Docker variables as well there.

bguillaumat:
Ok @softsimon I've updated the var to `ELECTRUM` to be compliant with the others setting

softsimon:
utACK

wiz (approved):
utACK

PR #369, opened 2021-03-03
Decode hex into utf8 instead of ascii

junderw:


wiz (approved):
ACK - review is OK since IE11 not supported by Angular anyway, and tested OK using https://mempool.ninja/tx/7b29f30462307bda0a73d99df8d0ceb9ca4868174f97ecdb6b0fab4a3ad68049

PR #374, opened 2021-03-04
Show fee tiles on mobile

junderw:
It was setting display to none when the screen width was being set to <= 767.98px (So basically mobile)

This makes the screen a tad cramped. But since these tiles have a useful function now, trading appearance for usability seems ok to me.


junderw:
The bugfix for 350+ might not work... I am not too familiar with chartist, and I couldn't get mempool to build properly.

wiz (approved):
ACK @ 4fbab41cc873416d3a8382495f953775471910e4
Tested OK on mempool.ninja/graphs

Issue #190, opened 2020-11-23
Graph legend seems to be shifted by one


softsimon:
Got this bug report that seems to suggest that the legend labels are shifted by one

> It looks like there is an issue on the mempool,space graps#2h It burns 20-30 span, while actually it should be 30-40

> I mean it shows they got confirmed or mined (the height of the given span 20-30 is decreasing) but actually I have not confirmed transaction from the 30-40 span.

jambolo:
You can verify this bug by comparing the lowest fee range with transactions removed by the latest block shown in the graph to the lowest fees in the that block's transactions.

wiz:
Funny, I always just figured this was just a quirk of your implementation, never thought to report it as a bug

jambolo:
I believe the problem is in mempool-graph.component.ts. ngOnInit() creates a legend whose first element is "1 sat/vb", but handleNewMempoolData() returns a series array that doesn't match for a non-liquid graph:

```javascript
 handleNewMempoolData(mempoolStats: OptimizedMempoolStats[]) {
    ...
    // Only Liquid has lower than 1 sat/vb transactions
    if (this.stateService.network !== 'liquid') {
      finalArrayVbyte.shift();
    }
    ...
```
I believe a solution is to also remove the first element of the the legend array for a non-liquid graph.

jambolo:
Fix like this?
```javascript
    if (this.showLegend) {
      legendNames = [1, 2, 3, 4, 5, 6, 8, 10, 12, 15, 20, 30, 40, 50, 60, 70, 80, 90, 100, 125, 150, 175, 200,
            250, 300, 350, 400].map((sat, i, arr) => {
              if (sat === 400) {
                return '350+';
              }
              if (i === 0) {
                return '0 - 1';
              }
              return arr[i - 1] + ' - ' + sat;
            });
      // Only Liquid has lower than 1 sat/vb transactions
      if (this.stateService.network !== 'liquid') {
        legendNames.shift();
      }
      this.mempoolVsizeFeesOptions.plugins.push(
        Chartist.plugins.legend({ legendNames: legendNames })
      );
    }
```

PR #376, opened 2021-03-05
Fix mempool-graph legend off-by-one

jambolo:
Resolves issue #190 as noted.

### DIFF
--- a/frontend/src/app/components/mempool-graph/mempool-graph.component.ts
+++ b/frontend/src/app/components/mempool-graph/mempool-graph.component.ts
@@ -78,22 +78,22 @@ export class MempoolGraphComponent implements OnInit, OnChanges {
     };
 
     if (this.showLegend) {
-      this.mempoolVsizeFeesOptions.plugins.push(
-        Chartist.plugins.legend({
-          legendNames: [1, 2, 3, 4, 5, 6, 8, 10, 12, 15, 20, 30, 40, 50, 60, 70, 80, 90, 100, 125, 150, 175, 200,
+      const legendNames: string[] = [1, 2, 3, 4, 5, 6, 8, 10, 12, 15, 20, 30, 40, 50, 60, 70, 80, 90, 100, 125, 150, 175, 200,
             250, 300, 350, 400].map((sat, i, arr) => {
               if (sat === 400) {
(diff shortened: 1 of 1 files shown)


jambolo:
Note that the commit message lists the wrong issue number.

junderw:
This only fixes the color issue by virtue of lowering the legend count to 26.

This is not off-by-one, but rather modulo constraint mismatch.

Currently mempool sets 26 classes for dynamic legend css classes, a through z.

There are 27 legends. so index 26 (the 27th one) is modulo 26, which 26 % 26 is 0, so the css class for `a` (aka the first one, currently 1 sat/vB) is used for the last legend.

However, I worry about the treatment of the 1 sat/vB data. Currently bitcoin is showing 45 MvB @ 1sat/vB (the legend you're removing) and 15 MvB @ 1 - 2... I wonder if the data will shift over 1 so that `1 - 2` will now say 45 MvB with this change even though it's not correct.

The issue of the wrapping class names still exists for Liquid...

So if 45 MvB is actually for 1 - 2 and not 1... and 15 MvB (currently 1 - 2) is actually 2 - 3 then this is a o-b-o bug... but the legend coloring bug still exists.

---

1. We need to check that this fix is actually fixing a data shift of some kind (where the wrong data is in the wrong legend) if we were to merge as is.
2. The legend coloring issue will still remain for liquid if so.

junderw:
スクリーンショット_2021-03-06_00-21-05

junderw:
Here's the list of series names... I tried adding `aa` to this list but it didn't fix it.

https://github.com/mempool/mempool/blob/v2.1.2/frontend/src/styles.scss#L209

junderw:
@wiz can we deploy this to ninja?

jambolo (commented):
Note that the commit message lists the wrong issue. I believe that can be fixed when the PR is merged.

wiz (approved):
ACK - tested OK on node220.bitcoin.wiz.biz/graphs

wiz:
> @wiz can we deploy this to ninja?

actually it's not possible to fully test on .ninja because it only has a 300MB so I deployed to node220 which has a larger mempool which contains 1 sat/vB transactions - looks good

PR #380, opened 2021-03-07
Fix bug where legend state is reverted when clicking 1-2 after a 1-minute auto-update.

junderw:


junderw:
Deployed to mempool.ninja and was not able to reproduce the bug.

wiz (approved):
ACK

Issue #384, opened 2021-03-12
Chime too loud for tx confirmation


junderw:
I swear I was about to have a heart attack.

PR #385, opened 2021-03-12
Lower volume for sound effects

junderw:
My old ears can't handle it!


junderw:
Fixes #384

wiz (approved):
utACK

softsimon:
utACK

PR #395, opened 2021-03-17
CPFP support

### LINKED ISSUE #5: Implement CPFP support in the backend
Currently CPFP is not supported so low fee transactions are incorrectly sorted into the back. 

A child transaction that rely on an unconfirmed transaction, their combined fee per size should be used to place them together in the sorted list of mempool transactions.

### LINKED ISSUE #353: Lack of CPFP calculation enables potential manipulation of fee estimates to high values
### Description

Due to a lack of CPFP calculation in the mempool backend, it might be possible for someone to manipulate the mempool backend fee estimates very high

#### Version

v2.1.2

### Steps to reproduce

1) Create a very huge low fee transaction and intentionally get it stuck in the mempool
2) CPFP the above transaction with a very high fee, say 9000 sats/vByte
3) Because the parent TX is so large, this still isn't enough to raise the average fee for the unconfirmed mempool chain high enough so that the chain gets mined into a block
3) Repeat this over and over until you have hundreds and hundreds of 9000 sats/vByte transactions stuck in the mempool

### Expected behaviour

Fee estimation is not affected by this manipulation

### Actual behaviour

Due to the lack of CPFP average fee calculation in the mempool backend, and the way that the mempool backend suggests fees to users using the median fee in the next projected block, the fee estimates could be manipulated much higher by a malicious person

### LINKED ISSUE #360: Upper bound of feerate range should maybe show effective feerate
### Description

[Two](https://twitter.com/ksedgwic/status/1364356751001149441) people [asked](https://twitter.com/flying_dude/status/1364550229715013635) about this on Twitter in the past 24h:

image

The preview of the next block states that the next block will include transactions of the feerates 88–8,428 sat/vB.

The transaction with the highest feerate in the mempool is https://blockchair.com/bitcoin/transaction/337de3cdcf0c0a9527bc4e7c8577afe31bcecd4d54ee72f07283476ce7dabe8a, which does indeed pay 8,428 sat/vB. However, this is actually a 110 vB CPFP transaction trying to bump an 82 kvB parent transaction that paid 10 sat/vB. In result, the effective feerate for the former transaction is that of its ancestor set, i.e. 21.38 sat/vB for the two transactions together.

### Expected behaviour

I was wondering whether it might make sense to present the future blocks' feerate ranges in effective feerates rather than in absolute feerates. I realize that this may be non-trivial, though.

This could potentially be achieved by running `getblocktemplate` for the next block, and using the transaction package feerates of the transactions in the block. Since dependent transactions need to be sorted in topological order in the block, transaction packages may be efficiently discoverable in a linear pass. The subsequent block's upper bound could then perhaps be inductively calculated from the prior subsequent block's lower bound.

### Notes

TBH, not sure if this is worth the effort, or whether this would even be a good way to approach it. I just wanted to put it out there because the feerate range seems to raise questions among users.

softsimon:
fixes #5
fixes #353
fixes #360


backend/src/api/common.ts
```diff
@@ -71,4 +71,63 @@ export class Common {
...
+    }
+
+    tx.effectiveFeePerVsize = totalFees / (totalWeight / 4);
+    tx.cpfpChecked = true;
+
+    return {
+      ancestors: tx.ancestors,
+      bestDescendant: tx.bestDescendant || null,
+    };
+  }
```
glozow, line 103: My sincerest apologies for being nosy, I'm just a big CPFP nerd and fan of mempool. Is this your algorithm for finding the effective feerate of transactions?

softsimon: @glozow Correct. This is my humble CPFP algorithm. If you know ways of improvement it's highly appreciated. The trick is to have an algorithm is also quick since the mempool and all the CPFP numbers are recalculated every ~2 second.

glozow: The most accurate would be `max(basefeerate, highest ancestorfeerate of all descendants)` because the mining code sorts by highest ancestorfeerate and then puts all of the transaction's direct ancestors in before it. I'm not familiar with the code / know how long this would take but if you're able to query the node, an RPC call `getmempoolentry(txid)` and `res['fees']['ancestor']/res['vsize']` would give you the ancestor feerate.

softsimon: The problem is, it would take too long time to fetch a `getmempoolentry` on all the mempool transactions,, every 2 second. So it has to be calculated naively in the local copy of our NodeJS backend in-memory mempool.

glozow: Yeah fosho, that makes sense.

PR #398, opened 2021-03-18
Shuffle mempool transactions before saving disk cache.

### LINKED ISSUE #397: Shuffle transactions before saving to disk cache to prevent uneven files and possible OOM
This is how uneven the cache files are saved. The biggest problem is the first file taking 300MB in this case which can possibly cause OOM error in the future.

Screen Shot 2021-03-18 at 15 47 28

softsimon:
fixes #397

End result, nice and even cache files:

Screen Shot 2021-03-18 at 15 58 11


softsimon:
Ok I tried this fix on my Raspberry pi 8 GB ram with "full" mempool ( 292 MB / 300 MB ).
I added a timer and it only took `0.021 seconds ` for the shuffle script to execute and there were no OOM error crashes indicating it was all just object references moving around, no extra memory allocation. Was able to restart and restore from cache successfully.

PR #403, opened 2021-03-21
docker: add Makefiles for easier docker deploy

RandyMcMillan:


GNUmakefile
```diff
@@ -0,0 +1,61 @@
...
+all: init
+	@echo 'all'
+	docker-compose up --force-recreate --always-recreate-deps
+	@echo ''
+#######################
+.PHONY: torproxy
+torproxy:
+	@echo 'torproxy'
+#REF: https://hub.docker.com/r/dperson/torproxy
+	bash -c 'docker run -it -p 8118:8118 -p 9050:9050 -p 9051:9051 -d dperson/torproxy'
```
wiz, line 59: Why is this needed? It seems to be not used for anything, and it adds a trusted third party dependency.

docker/docker-compose.yml
```diff
@@ -0,0 +1,51 @@
...
+    command: "./wait-for-it.sh db:3306 --timeout=720 --strict -- ./start.sh"
+    volumes:
+      - ./data:/backend/cache
+    environment:
+      RPC_HOST: "127.0.0.1"
+      RPC_PORT: "8332"
+      RPC_USER: "mempool"
+      RPC_PASS: "mempool"
+      ELECTRUM_HOST: "127.0.0.1"
+      ELECTRUM_PORT: "50002"
```
wiz, line 29: It seems this assumes you already have romanz/electrs running on the host machine. Maybe we should add romanz/electrs to this docker-compose template, perhaps as a fragment option?

GNUmakefile
RandyMcMillan: I will take it out :)

docker/docker-compose.yml
RandyMcMillan: thanks for the feed back - i will add this.

RandyMcMillan: We also want to avoid conflicts with other bundled docker implementations (like umbrel) so it will be an optional command in make - i will add some notes in documentation as well.

.gitignore
```diff
@@ -1 +1,6 @@
 sitemap
+data
+docker-compose.yml
+mysql
```
wiz, line 4: Is this needed? Looks like `mysql` will go in `data/`

docker/electrs/Dockerfile
```diff
@@ -0,0 +1,34 @@
+ARG VERSION=v0.8.6
```
wiz, line 1: Looks like the most recent release is v0.8.9, any reason not to use that instead?

.gitignore
```diff
@@ -1 +1,6 @@
...
+electrs
```
wiz, line 5: Also looks like `electrs` will go in `data/` unless I'm missing something?

wiz (approved):
utACK

PR #408, opened 2021-03-24
fixes to the electrum Dockerfile

knorrium:
minor fixes to the script:

- made it verbose for easy debugging
- don't assume that the keys have already been imported
- fixed the path to the home directory
- saving the tarball to `/tmp` explicitly


wiz:
wow, amazing to see a PR from you :heart:

wiz (approved):
utACK

PR #417, opened 2021-04-01
Use SRI

junderw:


junderw:
Deployed to ninja.

Confirmed that only one page is broken.

Bisq.

If anyone has any idea why only Bisq is broken by this (logically, it would mean that the minified JS for bisq only is somehow modified after Angular builds it.) any ideas are welcome.

junderw:
Maybe this is the culprit?

https://github.com/mempool/mempool/blob/0b04dbbca2d608c33b40fb39d492ce5e221153c0/frontend/src/app/app-routing.module.ts#L270

junderw:
This is the only path that uses import...

junderw:
ok, removing the import statement and just using Promise resolve works fine.

Tested all pages on all networks and no integrity errors.

however, I wonder why, pre-PR, bisq was downloading a 76kB minified JS file, whereas post-PR does not download anything and yet it still works.

Also, none of the other JS files increase any more than a few 100 bytes, nowhere near the 76kB file which is now not used or downloaded at all...

@softsimon any ideas why?

wiz (approved):
ACK - tested OK on mempool.ninja but should probably also test using SSR as there might be other random issues this causes

frontend/src/app/app-routing.module.ts
```diff
@@ -16,6 +16,7 @@ import { DashboardComponent } from './dashboard/dashboard.component';
 import { LatestBlocksComponent } from './components/latest-blocks/latest-blocks.component';
 import { ApiDocsComponent } from './components/api-docs/api-docs.component';
 import { TermsOfServiceComponent } from './components/terms-of-service/terms-of-service.component';
+import { BisqModule } from './bisq/bisq.module';
```
softsimon, line 19: This causes the BisqModule to be pre-loaded and included in the main.js bundle removing the benefits of LazyLoading. The Bisq bundle is 350 kb (and growing) so would like to keep this lazy loaded.

softsimon (changes requested):

frontend/src/app/app-routing.module.ts
junderw: I was unable to see any increase in initial bandwidth. I don't think you've properly removed the Bisq module from the main js file to begin with... so if that is expected behavior there is a bug.

1. visit mempool space and check the size of all the JS files downloaded in the sources tab.
2. visit the Bisq page and notice the extra 76kB minified JS file downloaded.
3. visit mempool ninja and check the size of all the JS files. They are not significantly different in size.
4. visit Bisq and notice nothing is downloaded.

junderw: I saved (Ctrl+S) each webpage. home, then bisq, for both ninja and space.

```
$ du | sort -k2
11148	.
2816	./ninja
2696	./ninja-bisq
2532	./ninja-bisq/トランザクション - mempool - Bisq Explorer_files
2532	./ninja/mempool - Bitcoin Explorer_files
2836	./space
2796	./space-bisq
2632	./space-bisq/トランザクション - mempool - Bisq Explorer_files
2556	./space/mempool - Bitcoin Explorer_files
```

_files directory (all the assets needed to show the site offline) for space is 76 (kB) difference, but the ninja site is less than 1kB difference... meaning the bisq JS is already in the main JS bundle for space as-is.

In which case you have a bug.

junderw: Due to the bug. This PR actually decreases the JS downloaded... so in a way, it's a bugfix.

The goal would be to decrease space home from 2556 to 2480 and have Bisq be 2556, but right now you're downloading 2 copies of the same JS. This PR only downloads the JS in the main JS file (which currently contains BisqModule on space)

junderw:
Looks like the issue with differential loading causing SRI errors was fixed in `v10.2.0` of `@angular/cli`

Looking at the package-lock.json of the frontend folder, there should be no version of the cli older than that installed.

@wiz could you try deleting node_modules and then running `npm ci` instead of `npm install` (ci will install from the package-lock file) and maybe the issue will resolve?

junderw:
The Fix PR: https://github.com/angular/angular-cli/pull/19093

If you check the 10.2.0 release tag and view the commit history from there, this PR should be in 10.2.0

We currently are installing (according to package-lock) 10.2.1

junderw:
or maybe you have angular-cli installed globally with an older version?...

or maybe the angular cli bugfix didn't actually fix the bug?

wiz:
sorry but I don't think that's related, the `ng` command is mapped to `./node_modules/@angular/cli/bin/ng` in `package.json`, and both `ng` and the absolute path to it return this same output

Screen Shot 2021-04-02 at 12 20 24

junderw:
https://github.com/angular/angular-cli/issues/19460

Apparently the issue above is not fixed for people using localization. (us)

So to summarize:

1. Lazy loading is currently broken in master (My PR which disabled lazy loading is 76kB less data than current `mempool.space` and doesn't download anything when switching to Bisq)
2. Lazy loading + SRI + localize is broken in current angular.
3. If we fix mempool lazy loading we can not use SRI until it is fixed in angular.
4. If we don't fix lazy loading, we can use SRI.

Why don't we use SRI and disable Lazy Loading, then if we can figure out how to fix lazy loading we can disable SRI (single line change)

junderw:
Either way, disabling lazy loading right now will lower the payload size by 76kB until we can fix it... so even if we don't enable SRI, we should disable lazy loading.

wiz:
Surprisingly the lighthouse performance score is higher on ninja with this PR applied

Screen Shot 2021-04-06 at 16 32 20

softsimon:
> [angular/angular-cli#19460](https://github.com/angular/angular-cli/issues/19460)
> 
> Apparently the issue above is not fixed for people using localization. (us)
> 
> So to summarize:
> 
>     1. Lazy loading is currently broken in master (My PR which disabled lazy loading is 76kB less data than current `mempool.space` and doesn't download anything when switching to Bisq)
> 
>     2. Lazy loading + SRI + localize is broken in current angular.
> 
>     3. If we fix mempool lazy loading we can not use SRI until it is fixed in angular.
> 
>     4. If we don't fix lazy loading, we can use SRI.
> 
> 
> Why don't we use SRI and disable Lazy Loading, then if we can figure out how to fix lazy loading we can disable SRI (single line change)

Ok, if this is the case I agree we should do like you are suggesting, disable Lazy loading for now and enable SRI.

But while testing this there is still something I am not able to wrap my head around. I am trying to confirm that the lazy loading is not working on master, but I am not able to. If I search for Bisq-bundle only code like `this.bisqApiService.getBlock$` on mempool.space, I can only find it in the lazy-loaded bisq-bundle, and not in any of the main bundles. And trying this PR with disabled lazy loading, the main non compressed bundle is ~30 kb bigger.


Screen Shot 2021-04-06 at 11 47 56

wiz:
How about we try upgrading to Angular 11 (#429) and fixing the lazy loading issue before proceeding with this PR?

junderw:
> How about we try upgrading to Angular 11 (#429) and fixing the lazy loading issue before proceeding with this PR?

I think the choices are:

1. Merge this, disable lazy loading (since it's broken anyways), and start a new PR that disables SRI and tries to fix lazy loading.
2. Don't merge this, and wait until SRI + localization + lazy loading fix is done in angular and upgrade to that angular version. (the fix isn't done yet)... and at the same time we are waiting try to fix lazy loading in separate PR.

So your idea of fixing lazy loading first means you want to wait for SRI until it is fixed in angular (no clue when it will be fixed / released)?

junderw:
IMO we should do 1 and then tag a release.

That way this release:

1. Removes the lazy-load-bug (where we just extra load unneeded JS)
2. Adds SRI

Then we can look into lazy-load bug, and if we can't fix it, then hey, we have SRI during the long battle of trying to fix LL.

If we fix lazy-loading, we can disable SRI in the same PR. Then next release will have no-SRI and proper LL.

And who knows, maybe by the time we fix LL, a fix for SRI+LL+Localization will be merged and released in Angular, giving us an excuse to upgrade Angular B-D

wiz (changes requested):
Since we're upgrading to Angular 11, this will be necessary
```
Support for camel case arguments has been deprecated and will be removed in a future major version.
Use '--subresource-integrity' instead of '--subresourceIntegrity'.
```

wiz:
@junderw could you take another look at this please

junderw:
Looks like angular-cli 17.3.7 fixed the only bug, so it should work well with localization.

junderw:
Looks like the Bisq issue also is irrelevant now.

So the next step would be to deploy this frontend somewhere and see if our browsers like it.

frontend/generate-config.js
```diff
@@ -105,6 +107,22 @@ function readConfig(path) {
...
 function writeConfig(path, config) {
   try {
+    const sriHash = crypto.createHash('sha384').update(config).digest('base64');
+    console.log(`SRI hash for ${path}: sha384-${sriHash}`);
+
+    let fileContent = fs.readFileSync(SRC_INDEX_FILE_PATH, 'utf8');
+    // There are two variants, depending on the filename path given to this function.
+    // config.js or customize.js
+    // We need to decide which one to replace based on the path argument.
+    const scriptFileName = path.split('/').pop();
```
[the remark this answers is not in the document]
junderw: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/mempool/mempool/pull/417#discussion_r2637514828)

frontend/generate-config.js
```diff
@@ -105,6 +107,22 @@ function readConfig(path) {
...
+    // The format of the script tag is:
+    // <script src="/resources/config.js"></script>
+    const regex = new RegExp(`(<script\\s+src=["']/resources/${scriptFileName}["'])(\\s*></script>)`);
+    fileContent = fileContent.replace(regex, `$1 integrity="sha384-${sriHash}" crossorigin="anonymous"$2`);
```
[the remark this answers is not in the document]
junderw: Hallucination. The current Regex only allows for spaces between the <> and the matching stuff, so if there is already integrity, it will not match. So the behavior of appending infinite integrity attributes does not occur.

Not to mention we are copying the index.html from the template at the top of this file, so the file will always be fresh.

frontend/generate-config.js
```diff
@@ -105,6 +107,22 @@ function readConfig(path) {
...
+    fs.writeFileSync(SRC_INDEX_FILE_PATH, fileContent, 'utf8');
+    console.log(`Updated SRI hash in ${SRC_INDEX_FILE_PATH}`);
+
+    // Write the config file itself to the given path
     fs.writeFileSync(path, config, 'utf8');
   } catch (e) {
     throw new Error(e);
```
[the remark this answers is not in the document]
junderw: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/mempool/mempool/pull/417#discussion_r2637514832)

PR #422, opened 2021-04-02
Catch getMempoolInfo errors gracefully to not break general main loop

### LINKED ISSUE #411: Crash bug: if minfee node becomes unresponsive, backend gets totally stuck
```
Mar 29 23:02:05 [14190] WARN: runMainLoop error: ETIMEDOUT. Retrying in 60 sec.
Mar 29 23:02:05 [14190] DEBUG: {"code":"ETIMEDOUT"}
Mar 29 23:04:05 [14190] WARN: runMainLoop error: ETIMEDOUT. Retrying in 60 sec.
Mar 29 23:04:05 [14190] DEBUG: {"code":"ETIMEDOUT"}
Mar 29 23:06:05 [14190] WARN: runMainLoop error: ETIMEDOUT. Retrying in 60 sec.
Mar 29 23:06:05 [14190] DEBUG: {"code":"ETIMEDOUT"}
Mar 29 23:08:05 [14190] WARN: runMainLoop error: ETIMEDOUT. Retrying in 60 sec.
Mar 29 23:08:05 [14190] DEBUG: {"code":"ETIMEDOUT"}
Mar 29 23:10:05 [14190] WARN: runMainLoop error: ETIMEDOUT. Retrying in 60 sec.
Mar 29 23:10:05 [14190] DEBUG: {"code":"ETIMEDOUT"}
```

softsimon:
**NOTE**
The issue reported is not a "crash bug" as errors are handled, and I don't think the backend gets "totally stuck", it just halted the main loop.

This fix handles getMempoolInfo errors separately and will now throw a warn on first error if using a separate RPC server config, as opposed to after 5 tries for general main loop errors.

fixes #411

### DIFF
--- a/backend/src/api/mempool.ts
+++ b/backend/src/api/mempool.ts
@@ -14,7 +14,7 @@ class Mempool {
   private inSync: boolean = false;
   private mempoolCache: { [txId: string]: TransactionExtended } = {};
   private mempoolInfo: IBitcoinApi.MempoolInfo = { loaded: false, size: 0, bytes: 0, usage: 0,
-                                                    maxmempool: 0, mempoolminfee: 0, minrelaytxfee: 0 };
+                                                    maxmempool: 300000000, mempoolminfee: 0.00001000, minrelaytxfee: 0.00001000 };
   private mempoolChangedCallback: ((newMempool: {[txId: string]: TransactionExtended; }, newTransactions: TransactionExtended[],
     deletedTransactions: TransactionExtended[]) => void) | undefined;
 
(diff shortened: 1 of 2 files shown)


wiz:
Isn't the main issue that this is using `await` so it gets stuck in the loop here?

softsimon:
If anything in the loop crashes it will retry after X seconds... This make sense since if the first task doesn't work the second shouldn't work either when it's the same backend .... The exception in this case was using a RPC Minfee server, but the main backend was still working .. so aborting and retrying was probably not correct.

wiz (approved):
utACK - looks safe, let's try running this for a while and see how it works

PR #433, opened 2021-04-08
Build mempool-js to root www folder.

softsimon:


wiz (changes requested):
I don't see the reason to use `mempool.min.js`, just `mempool.js` is best IMO

PR #436, opened 2021-04-09
FIX: align project contributor handle text.

### LINKED ISSUE #435: GitHub profile handle text is not aligned at About page.
### Description

In the https://mempool.space/about there is a text misalignment when showing the GitHub profile handle.

#### Version

v2.2-dev (dee614f6)

### Steps to reproduce

Go to https://mempool.space/about and you can see the misalignment here:

image

### Expected behaviour

The `username` should be align at the middle of the component.
image

### Actual behaviour

The `username` is not at the center of the profile component.
image

MiguelMedeiros:
FIX #435


wiz (approved):
ACK - much better, thanks for fixing this!

PR #449, opened 2021-04-15
Syslog is now configurable with minimum priority and facility option.

### LINKED ISSUE #135: Log backend messages to syslog if configured
Add new variable to backend config, something like this
```
"SYSLOG_SOCKET": "127.0.0.1:514" // udp
```

softsimon:
fixes #135

### DIFF
--- a/backend/mempool-config.sample.json
+++ b/backend/mempool-config.sample.json
@@ -39,6 +39,13 @@
     "USERNAME": "mempool",
     "PASSWORD": "mempool"
   },
+  "SYSLOG": {
+    "ENABLED": true,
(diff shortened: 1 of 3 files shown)


wiz (changes requested):
NACK - tested on mempool.ninja using default configuration (no syslog block at all), but the syslog daemon still receives `debug` level messages instead of the default minimum priority of `info`

softsimon:
Please try again :)

wiz (approved):
ACK - tested OK on mempool.ninja using various configurations

wiz:
Thanks for fixing this :heart:

PR #452, opened 2021-04-17
add new configuration and targets to point the frontend to staging

knorrium:
This PR provides a quick way to point a local frontend development environment to `mempool.ninja` so we can take in contributions that don't require any backend changes faster.

Ideally this would use Angular's environment support but I wanted to unblock @MiguelMedeiros quickly.

Routing to `signet` does not work because that is not working on `mempool.ninja` either, so I didn't include the routes in the config file.


softsimon:
This is good when doing frontend development only. Tested and working.

MiguelMedeiros:
> This PR provides a quick way to point a local frontend development environment to `mempool.ninja` so we can take in contributions that don't require any backend changes faster.
> 
> Ideally this would use Angular's environment support but I wanted to unblock @MiguelMedeiros quickly.
> 
> Routing to `signet` does not work because that is not working on `mempool.ninja` either, so I didn't include the routes in the config file.

Thanks @knorrium !!

PR #454, opened 2021-04-17
fix native scrolling on iOS devices

knorrium:
Currently on iOS devices you can't scroll to top and sometimes you'll notice that the page will look frozen or broken.

This change restores the native scrolling and should fix that. 

Tested on Desktop and iOS. Untested on Android phones.


wiz (approved):
ACK - tested OK on node100, scrolling feels smoother and doesn't break anything from what I can see. IIRC we added this line because a scrollbar appeared when scrolling the blocks but maybe that was fixed when we upgraded our deps a while back

PR #455, opened 2021-04-18
preserve the user's network selection on the about page

knorrium:
The navigation bar was always routing to `/about`, which ends up switching back to the default network.

We will now respect the user's selection.

liquid
bisq
testnet

### DIFF
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -124,6 +124,10 @@ const routes: Routes = [
             path: 'api',
             component: ApiDocsComponent,
           },
+          {
+            path: 'about',
+            component: AboutComponent,
+          },
         ],
       },
       {
@@ -185,6 +189,10 @@ const routes: Routes = [
           {
             path: 'api',
             component: ApiDocsComponent,
+          },
+          {
+            path: 'about',
+            component: AboutComponent,
           },
         ],
       },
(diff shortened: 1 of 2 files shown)


wiz (changes requested):
NACK - Simon and I intentionally decided a while back to only have one About page, and since 99% of users only use Mainnet this seemed like a reasonable thing to do. This PR title is "preserve the user's network selection" but what this PR actually does is creates multiple About pages on separate URLs, which isn't really what we want. If you can somehow cleanly "preserve the user's network selection" instead of making multiple About pages that might be cool, but as-is having one About page per network isn't what we want for the site :)

knorrium:
@wiz technically, it’s the same page since it’s using the same component and it follows the same pattern of going to /network/page.

I can look into using localStorage, but that might end up as a bigger refactor if we support both preference based navigation and /network/* links.

wiz:
SInce >99% of users only use mainnet I don't think it's worth it

MiguelMedeiros:
utACK

I like the idea! 
It's so annoying when you are on `bisq` network and then you go back to `mainet` when you click on `about` nav-link! :/

wiz:
Alternatively, maybe clicking on the About page should switch the user to a new "network" in the pull-down selector that's not actually a network but maybe a documentation section

wiz:
Let's close this for now, and discuss it again later :P

PR #456, opened 2021-04-18
persist graph window preference between reloads

### LINKED ISSUE #365: Remember /graphs time period setting at new load
### Description

When I open /graphs I would like to have my previous time period preselected

#### Version

 v2.2-dev (b8963095) 

### Steps to reproduce

1. Open /graphs, select 1M
2. Go back to home
3. Open /graphs again, time period is again default (2H)

### Expected behaviour

1. Open /graphs, select 1M
2. Go back to home
3. Open /graphs again, time period is 1M

### Screenshots
N/A

#### Device or machine
Linux/Firefox

#### Additional info
Love Mempool!

knorrium:
fixes #365 

Initial support for storing the time period as a user preference. It currently forces a reload but could be improved later on.

### DIFF
--- a/frontend/src/app/components/statistics/statistics.component.html
+++ b/frontend/src/app/components/statistics/statistics.component.html
@@ -14,29 +14,28 @@
       <div class="card mb-3" *ngIf="mempoolStats.length">
         <div class="card-header">
           <i class="fa fa-area-chart"></i> <span i18n="statistics.memory-by-vBytes">Mempool by vBytes (sat/vByte)</span>
-
-          <form [formGroup]="radioGroupForm" style="float: right;">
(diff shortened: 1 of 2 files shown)


softsimon (changes requested):
Concept ACK, but setting the active class using classes is not the way to do it since the dynamic component is able to do that automatically.

The change that should work is:

Remove all the
```[class.active]="graphWindowPreference === ```

and in the component move the formBuilder setup and set the default selection the cache variables

```
this.radioGroupForm = this.formBuilder.group({
      dateSpan: this.graphWindowPreference
    });
```

knorrium:
@softsimon thanks for the feedback, I've made the changes you suggested

PR #474, opened 2021-04-26
Fix crash issues related to new unconfirmed transactions in bitcoind …

### LINKED ISSUE #391: The watch mempool TX push feature breaks when using bitcoind as backend
When subscribing to a transactionID using `'watch-mempool'` will not run "extend transaction" so the transaction will lack the prevout transactions transaction page will throw error.

softsimon:
This PR solves a few issues related to opening recently broadcasted transactions when running Bitcoind:

* The `prevouts` wasn't appended when opening a transaction that is unconfirmed but not yet was into the backend mempool, causing frontend to crash
* A very recent transaction doesn't contain CPFP data causing frontend TX to have the TX location arrow missing. Added a retry every 2 sec until backend has caught up on the CPFP checking on the transaction.

fixes #391


wiz (approved):
utACK - I don't have a bitcoind setup to test this so I'll take your word for this

PR #475, opened 2021-04-27
Fixing incorrect difficulty adjustment call

### LINKED ISSUE #471: Incorrect Difficulty Adjustment Percentage
### Description

The estimated difficulty adjustment percentage is probably calculated by dividing the average time per block of this difficulty epoch by the expected amount of time (600 seconds), instead of the other way around. 
### Expected behaviour

-100*(1-600/720) = -16,7%

### Actual behaviour

-100*(720/600-1) = -20%

softsimon:
fixes #471


wiz (approved):
ACK - tested OK seems fine

PR #478, opened 2021-04-27
Localize electrum limit error.

### LINKED ISSUE #442: Error messages on the address explorer are not localized
### Description

Two error messages on the address explorer are not localized

- `The number of transactions on this address exceeds the Electrum server limit`
- `Consider viewing this address on the official Mempool website instead:`

#### Version

v2.1.2 (846d7e90)

### Steps to reproduce

Attempt to load an address with a long list of transactions on an instance running electrum with low transaction limits on a non-en-US locale.

### Expected behaviour

The error messages are translated

### Actual behaviour

The messages are in English

### Screenshots

Screen Shot 2021-04-10 at 9 12 59 PM

softsimon:
fixes #442

### DIFF
--- a/frontend/src/app/components/address/address.component.html
+++ b/frontend/src/app/components/address/address.component.html
@@ -114,10 +114,13 @@
     <div class="text-center">
       <span i18n="address.error.loading-address-data">Error loading address data.</span>
       <br>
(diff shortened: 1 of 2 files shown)


wiz (approved):
utACK - untested because no romanz/electrs server

PR #481, opened 2021-04-28
Add animated css logo.

MiguelMedeiros:
### Description

New concept for animated css logo.

### Expected behaviour

https://user-images.githubusercontent.com/5798170/116329973-df6dfd00-a7a2-11eb-9e03-eecccd2cd20b.mp4

### JS Fiddle

https://jsfiddle.net/miguelmedeiros/7nt8dy6b/4/


AaronDewes:
Having the logo animated for ever doesn't seem good to me, I'd use this as a loading animation, but I'm very bad at designing, so my feedback probably isn't good.

softsimon:
I don't think we want our logo to be animating and moving. If it was to represent the full mempool it would always be full which is not good either.

MiguelMedeiros:
> Having the logo animated for ever doesn't seem good to me, I'd use this as a loading animation, but I'm very bad at designing, so my feedback probably isn't good.

I think it's a good idea to remove the infinite animation and limit it to reproduce only 2 or 3 cycles.

MiguelMedeiros:
> I don't think we want our logo to be animating and moving. If it was to represent the full mempool it would always be full which is not good either.

The main idea was to simulate that the mempool logo was "alive" (like a living organism, changing all the time).

wiz:
I agree it would be cool to use this for a loading animation, but currently we don't have anything that takes a while to load

PR #499, opened 2021-05-06
FIX unify the units css.

### LINKED ISSUE #490: UX: Unify the units
The units, including BTC, MB, MWU etc should have the same design language.
Also maybe the (1 min ago) should be in the same style.

Screen Shot 2021-05-03 at 17 17 15

MiguelMedeiros:
Fix #490 

## Fix Preview

image

image

image


MiguelMedeiros:
Fix #499

softsimon:
I found 3 different sizes used for these grey small units/symbols around the site,, 10px, 11px and 12px.

To unify I think it should be the same everywhere, and `12px` is the best size, because sometimes it looks too small, especially when displaying `(23 minutes ago)`

PR #503, opened 2021-05-08
Improve docker images and add a makefile for local builds

pox:
Please consider this PR a suggestion for a couple of improvements. If you like any of them, please let me know and I'll change as requested.

1. The backend image doesn't have a proper PID entrypoint. Just a CMD. Because of this, for example, when you try to kill it with `docker-compose down` it hangs for a while (it's not forwarding signals properly. So I've added [TINI](https://github.com/krallin/tini). If you prefer another handler let me know. Also, if you prefer to just add TINI (it's just a `sh` file) to the repo, instead of downloading it from an external source, I'd be happy to do that as well.
2. I've added a simple Makefile for building locally. I couldn't find the instructions for how to do so and had to dig into the GitHub CI yaml. This makes it easier for those who want to build from source and run on their own instead of pulling for docker hub.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+.PHONY: backend init frontend up down build
+.DEFAULT_GOAL := build
+
+compose := sudo docker-compose
+
+tag ?= latest
+
+init:
+	docker/init.sh $(tag)
+
+frontend backend: init
+	sudo docker build -f docker/$@/Dockerfile -t mempool/$@:$(tag) $@
+
+build: frontend backend
+
+up:
+	$(compose) up -d
+
+down:
+	$(compose) down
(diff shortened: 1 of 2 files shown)


wiz (changes requested):
Hi @pox - sorry for the late response. Yes, we would prefer you to copy the file from the project instead of adding it as a third party dependency to the Dockerfile. Please preserve the original copyright/license and other relevant legal notices at the top of anything you copy from that project. Also I see there is a merge conflict now, so it would be cool if you can resolve that one as well.

Thank you for fixing this!

wiz:
@knorrium since it's been over a month did you want to take over this PR ?

knorrium:
@wiz I'm going to close this PR as it has merge conflicts. we can make these improvements when introducing the other Docker changes.

PR #509, opened 2021-05-11
update the default value for ITEMS_PER_PAGE to match production settings

knorrium:
Fixes an issue with pagination on the block viewer when running locally but pointing to staging, as 10 is not a multiple of 25 and we would get a 400 error.


wiz:
I think the production configuration files are in production/ and have this value set accordingly, but IIRC the reason for the non-production config default being lower is because 99% of users don't use blockstream/electrs backend but romanz/electrs backend which is slower, so to improve page load time we only get 10 instead of 25

knorrium:
@wiz thanks for the context, I'll close this PR and refactor the config generator as a separate effort.

Issue #493, opened 2021-05-04
Website is not scrolling to the top when switching pages


softsimon:
### Description

Website is not scrolling to the top when switching pages

### Steps to reproduce

Use iPhone/iOS
1. Scroll to the bottom of the dashboard page
2. Click on a new incoming transaction

### Expected behaviour

The transaction should be opened and site scroll position should be at the top

### Actual behaviour

The transaction page is loaded but the site scroll position is somewhere in the middle

PR #511, opened 2021-05-11
Improve scrolling between routes

knorrium:
Potential fix for #493 

Uses `scrollPositionRestoration` and `anchorScrolling` to improve scrolling and state between routes. The block viewer page was also scrolling to top after clicking on the `Details` button because it's technically a route change, so setting anchors will keep the position relatively the same between changes.

Reference: https://angular.io/api/router/ExtraOptions#scrollPositionRestoration


wiz (approved):
ACK - tested OK on ninja, if I click on Terms of Service from the bottom of the About page it brings me to the top 👍

PR #515, opened 2021-05-12
Round up fee estimations

softsimon:


wiz (approved):
ACK - tested OK on ninja

PR #521, opened 2021-05-13
Bisq dashboard: Change color between red/green when price changes

softsimon:
So how it works is:
* Price starts out in green
* If price goes down, the color flips to red
* If price goes up, it flips back to green


wiz (approved):
ACK - tested OK looks cool

PR #524, opened 2021-05-18
New separate sponsorship page

softsimon:


wiz (approved):
ACK - tested OK on ninja

PR #529, opened 2021-05-18
Align scriptmessage text to the left.

### LINKED ISSUE #512: Coinbase tag should be left aligned
Coinbase tag should be left aligned at all resolutions, and the container should only be as long as the text.

Screen Shot 2021-05-11 at 18 16 07

MiguelMedeiros:
Fix #512 .

image


MiguelMedeiros:
Change scriptmessage box to only fills the text width.

image

Issue #461, opened 2021-04-19
Add new API documentation.


MiguelMedeiros:
### Description

Add new API documentation page with live `mempool-js` examples.

wiz:
We'd like to include this as part of the upcoming v2.2.0 release - do you have a rough idea when it will be ready? :)

MiguelMedeiros:
I'm finishing the new responsive layout today and I´ll start the API docs tomorrow! 
I´ll give you a better prediction as soon as I start! :)

MiguelMedeiros:
I'll start the API docs today! :)

PR #544, opened 2021-05-29
New API docummentation.

MiguelMedeiros:
New API documentation.


MiguelMedeiros:
Fix #461

wiz:
Great work, I like the syntax highlighting but I don't think it's worth adding a new npm dep for that - if we want to do syntax highlighting probably best to just make our own CSS and not add a new dependency

MiguelMedeiros:
> Great work, I like the syntax highlighting but I don't think it's worth adding a new npm dep for that - if we want to do syntax highlighting probably best to just make our own CSS and not add a new dependency

Ok, done!

- [x] Removed the highlight.js dependency.
- [x] Made our own styles for `code-template-component`.

softsimon (approved):
I approve 👍

PR #546, opened 2021-05-30
Cosmetic changes to the navbar and menu items

knorrium:
The dropdown menu was setting the `mainnet` class to the Bisq menu item.

This PR bundles a few changes to make it easier for upcoming automation work and also setting an active style for Mainnet.

### DIFF
--- a/frontend/src/app/components/bisq-master-page/bisq-master-page.component.scss
+++ b/frontend/src/app/components/bisq-master-page/bisq-master-page.component.scss
@@ -56,7 +56,7 @@ nav {
 }
 
 .mainnet.active {
-  background-color: #653b9c;
+  background-color: #fb8c00;
 }
 
 .liquid.active {
@@ -69,6 +69,10 @@ nav {
 
 .signet.active {
   background-color: #6f1d5d;
+}
+
+.bisq.active {
+  background-color: #653b9c;
 }
 
 .dropdown-divider {
--- a/frontend/src/app/components/bisq-master-page/bisq-master-page.component.scss
+++ b/frontend/src/app/components/bisq-master-page/bisq-master-page.component.scss
@@ -56,23 +56,23 @@ nav {
 }
 
 .mainnet.active {
-  background-color: #fb8c00;
+  background-color: rgb(251 140 0 / 55%);
 }
 
 .liquid.active {
-  background-color: #116761;
+  background-color: rgb(17 103 97 / 55%);
 }
 
 .testnet.active {
-  background-color: #1d486f;
+  background-color: rgb(29 72 111 / 55%);
 }
 
 .signet.active {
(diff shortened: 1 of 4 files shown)


softsimon:
With this change, the "active" Bisq menu option loses it's color to the default dark. Is this the intention?

`.mainnet.active` should be changed to `.bisq.active` in the css as well to not change the color.

Screen Shot 2021-05-31 at 18 12 59

softsimon (changes requested):
With this change, the "active" Bisq menu option loses it's color to the default dark. Is this the intention?

`.mainnet.active` should be changed to `.bisq.active` in the css as well to not change the color.

Screen Shot 2021-05-31 at 18 12 59

knorrium:
@softsimon good catch, it was not intentional. 

Looking at this in more detail, the styles for Bisq and Mainnet are being shared because of a conflict between `master-page.component.scss` and `styles.scss`.

What if we change the colors a bit and keep `nav-bar.active` and `nav-item.active` more consistent? If we don't want to change the default color for mainnet, we could pick something else for Bisq:

Screen Shot 2021-05-31 at 8 26 06 PM

Screen Shot 2021-05-31 at 8 26 19 PM

knorrium:
cc @MiguelMedeiros for visibility on the css conflicts

MiguelMedeiros:
> cc @MiguelMedeiros for visibility on the css conflicts

Maybe we should remove the css code from `/src/styles.scss`:
image

And add them in `/src/app/components/master-page-component.scss`:
(**idea: add 55% opacity to navbar.active - more readable text**)

```scss

.mainnet.active {
  background-color: rgb(251 140 0 / 55%);
}

.liquid.active {
  background-color: rgb(17 103 97 / 55%);
}

.testnet.active {
  background-color: rgb(29 72 111 / 55%);
}

.signet.active {
  background-color: rgb(111 29 93 / 55%);
}

.bisq.active {
  background-color: rgb(101 59 156 / 55%);
}

.navbar-nav.mainnet > .active {
  background-color: #fb8c00 !important;
}

.navbar-nav.liquid > .active {
  background-color: #116761 !important;
}

.navbar-nav.testnet > .active {
  background-color: #1d486f !important;
}

.navbar-nav.signet > .active {
  background-color: #6f1d5d !important;
}

.navbar-nav.bisq > .active {
  background-color: #653b9c !important;
}
```

## preview
navbar-colors

knorrium:
@MiguelMedeiros it looks much better after changing the opacity, thanks for the tip!

@softsimon do you have any feedback on the colors?

softsimon:
The tab buttons on mainnet should have this color, not yellow.

knorrium:
I'll have a different PR up with more changes that will just make the app more testable with no visual differences.

PR #559, opened 2021-06-08
Refactor about page layout.

MiguelMedeiros:


softsimon (changes requested):
I think the margins are a little bit too big.

wiz:
Sorry, what is the point of this PR? Seems to be breaking many things and margins etc. for no apparent benefit?

frontend/src/app/components/about/about.component.html
```diff
@@ -1,290 +1,204 @@
...
-        </a>
-      </ng-template>
-    </ng-container>
-    <br><br>
-    
+    <div class="wrapper">
+      <ng-container *ngIf="sponsors$ | async as sponsors; else loadingSponsors">
+        <ng-template ngFor let-sponsor [ngForOf]="sponsors">
+          <a [href]="'https://twitter.com/' + sponsor.handle" target="_blank" rel="sponsored" [title]="sponsor.handle">
+            <img class="image" [src]="sponsor.imageUrl" />
```
wiz, line 55: This is not correct and introduces a bug

MiguelMedeiros: reverted:
``
to:
``

PR #576, opened 2021-06-13
Lower scroll threshold to fix load more not triggering.

### LINKED ISSUE #575: Auto load more on scroll doesn't trigger on address page
If I go to an address page with more than 25 transactions and scroll to the bottom, it doesn't trigger the load more function.

Example: https://mempool.space/address/1E6vTBe9KLCh5ZqEkLoV2Fh5syXVHxHkna
Browser: Chrome

**Technical details**
If I debug the `      console.log(percentageScrolled);` variable the value never goes above 89, so it never gets triggered.

softsimon:
fixes #575

### DIFF
--- a/frontend/src/app/components/transactions-list/transactions-list.component.ts
+++ b/frontend/src/app/components/transactions-list/transactions-list.component.ts
@@ -78,7 +78,7 @@ export class TransactionsListComponent implements OnInit, OnChanges {
     const scrollTop = document.documentElement.scrollTop;
     if (scrollHeight > 0){
       const percentageScrolled = scrollTop * 100 / scrollHeight;
-      if (percentageScrolled > 90){
+      if (percentageScrolled > 80){
         this.loadMore.emit();
       }
     }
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/transactions-list/transactions-list.component.ts
```diff
@@ -78,7 +78,7 @@ export class TransactionsListComponent implements OnInit, OnChanges {
     const scrollTop = document.documentElement.scrollTop;
     if (scrollHeight > 0){
       const percentageScrolled = scrollTop * 100 / scrollHeight;
-      if (percentageScrolled > 90){
+      if (percentageScrolled > 80){
```
MiguelMedeiros, line 81: I've done some testing and it might be better if we let the data refresh take place if the scroll percentage is greater than 60% of the page. This way the scrolling is smoother, without having to wait for the loading placeholder.

`if (percentageScrolled > 60){`

MiguelMedeiros (commented):
I've done some testing and it might be better if we let the data refresh take place if the scroll percentage is greater than 60% of the page. This way the scrolling is smoother, without having to wait for the loading placeholder.

PR #578, opened 2021-06-15
Parse TXID, hash or address strings from search bar and allow searchi…

### LINKED ISSUE #415: Allow for searching for TXID:OUTPUTNUM
### Description

My LND node often gives channel points, which are TXID:OUTPUTNUM . Every time I look for those, mempool.space fails.

#### Version

web version / v2.2-dev (21bb879)

### Steps to reproduce

Enter a channel point in search box or URL (/tx/chanpoint)

### Expected behaviour

This can be done in 2 levels:
* Strip/rewrite everyting after `:`
* If #413 is implemented, highlight the actual output

### Actual behaviour

I get `Invalid hex string` error

#### Additional info

In the future a link to the relevant channel would be nice.

### LINKED ISSUE #450: Feature request: parse TXID or address strings out of what gets pasted into search box
It would be nice if you can paste a URL like  `https://blockstream.info/txid/bd993e6ff9c10df92a58230ba7c70af3f756aa5163cfb2377ece31a23fd14156` into our search box and have it trim out the TXID, address, or other relevant data automatically. You probably only need to regex for a few address types and TXID hex string in the frontend code, so I guess it's easy to add? :)

softsimon:
1. You can now post random URLs in the search bar and it will parse out addresses, transaction IDs and block hash
2. Transaction ids can now contain the Output index like so: TXID:OUTPUTINDEX and it will highlight the selected index. (see screen shot)

Screen Shot 2021-06-14 at 19 09 32

fixes #450
fixes #415


wiz:
@knorrium can you help us test this ? seems like a good job for a QA engineer :)

frontend/src/app/components/search-form/search-form.component.ts
```diff
@@ -96,21 +96,26 @@ export class SearchFormComponent implements OnInit {
...
+        const matches = this.regexBlockhash.exec(searchText);
+        this.navigate('/block/', matches[0]);
+      } else if (this.regexBlockheight.test(searchText)) {
         this.navigate('/block/', searchText);
       } else if (this.regexTransaction.test(searchText)) {
+        const matches = this.regexTransaction.exec(searchText);
         if (this.network === 'liquid') {
-          if (this.assets[searchText]) {
-            this.navigate('/asset/', searchText);
+          if (this.assets[matches[0]]) {
```
knorrium, line 109: @softsimon I don't think we will ever reach this since we are comparing assets with a tx regex?

knorrium (changes requested):

knorrium:
Screen Shot 2021-06-15 at 3 51 29 PM

Can we scope down which rows can be highlighted instead of highlighting per `tr` index?

frontend/src/app/components/search-form/search-form.component.ts
```diff
@@ -23,9 +23,9 @@ export class SearchFormComponent implements OnInit {
   searchForm: FormGroup;
   @Output() searchTriggered = new EventEmitter();
 
-  regexAddress = /^([a-km-zA-HJ-NP-Z1-9]{26,35}|[a-km-zA-HJ-NP-Z1-9]{80}|[bB]?[a-z]{2,5}1[ac-hj-np-z02-9]{8,87})$/;
-  regexBlockhash = /^[0]{8}[a-fA-F0-9]{56}$/;
-  regexTransaction = /^[a-fA-F0-9]{64}$/;
+  regexAddress = /([a-km-zA-HJ-NP-Z1-9]{26,35}|[a-km-zA-HJ-NP-Z1-9]{80}|[bB]?[a-z]{2,5}1[ac-hj-np-z02-9]{8,87})/;
+  regexBlockhash = /[0]{8}[a-fA-F0-9]{56}/;
```
knorrium, line 27: @softsimon Liquid block hashes do not follow this pattern, so they end up being treated as transactions instead (see `bfed8e58a220f6459920a4d7ffc1e822df619a0191068e195845aa237972337e`).

I'd refactor the code below to be more context aware, maybe extracting the logic to a function called `parseNetworkRegex(network, searchText)`.

PR #581, opened 2021-06-15
Add addresses endpoint

TKone7:


softsimon (changes requested):
There are some problems with this PR.

1. Could you please specify in detail what this PR does and why, what is the use case etc.
2. It seems like the commits are branched from some weird commit hash that includes a lot of commits from Wiz, so it is not mergable. Maybe redo-the commits and PR in a new standalone branch forked from latest master?
3. There are too many excessive `logger.info()` and `logger.debug()` which should not be needed
4. The api endpoint "check" is too ambiguous. What is it checking? Maybe `check-addresses` is better.

PR #584, opened 2021-06-15
fix the block viewer for liquid

knorrium:
`toString()` on `decimal2hex` was throwing an exception because `nonce` and `bits` are not available on Liquid blocks, `Difficulty` is also not available, so remove it as well


softsimon:
Good find!

PR #593, opened 2021-06-25
refactor config generator and make missing git non-fatal

knorrium:


knorrium:
I'll include this with the docker fixes

Issue #480, opened 2021-04-27
Add number of blocks and/or countdown timer to next difficulty adjustment


softsimon:

MiguelMedeiros:
Hey @softsimon can you link this issue with this PR https://github.com/mempool/mempool/pull/602 ? Thanks!

PR #602, opened 2021-07-01
UI/UX - New component for difficult adjustment.

### LINKED ISSUE #518: Difficulty adjustment displays -100% after adjustment
**Expected**:

Maybe 0% ?

**Actual**:

Screen Shot 2021-05-13 at 13 03 21

MiguelMedeiros:
Fix #518 and #480 .

## Tasks
- [x] Add next difficulty blocks.
- [x] Add next difficulty target date.
- [x] Add next difficulty total progress.
- [x] Add adjustment difficulty avg min per block.
- [x] Add i18n strings.
- [x] Add skeleton loader.
- [x] Bugfix difficulty adjustment validation.
- [x] Trigger next difficulty calculation every second.

## Observations
We need to re-translate the `shared.block` and `shared.blocks` strings.

## Preview images

New component layout.
image

Responsive layout.
difficulty-adjustment-responsive

Skeleton layout.
difficulty-adjustment-skeleton


frontend/src/app/dashboard/dashboard.component.ts
```diff
@@ -22,6 +22,12 @@ interface EpochProgress {
   green: string;
   red: string;
   change: number;
+  progress: string;
+  remainingBlocks: number;
+  newDifficultyHeight: number;
+  colorAdjustments: string;
+  timeAvg: string;
+  remaingTime: number;
```
knorrium, line 30: typo: `remainingTime`

frontend/src/app/dashboard/dashboard.component.html
```diff
@@ -200,12 +200,56 @@ <h5 class="card-title" i18n="dashboard.incoming-transactions">Incoming transacti
...
-        <div class="progress-text"><ng-template [ngIf]="epochData.change > 0">+</ng-template>{{ epochData.change | number: '1.0-2' }}%</div>
+      <div class="difficulty-adjustment-container" *ngIf="(difficultyEpoch$ | async) as epochData; else loadingDifficulty">
+        <div class="item">
+          <h5 class="card-title" i18n="difficulty-box.next-difficulty">Next Difficulty</h5>
+          <div class="card-text">
+            <ng-container *ngTemplateOutlet="epochData.remainingBlocks === 1 ? blocksSingular : blocksPlural; context: {$implicit: epochData.remainingBlocks }"></ng-container>
+            <ng-template #blocksPlural let-i i18n="shared.blocks">{{ i }} <span class="shared-block">blocks</span></ng-template>
+            <ng-template #blocksSingular let-i i18n="shared.block">{{ i }} <span class="shared-block">block</span></ng-template>
+          </div>
+          <div class="symbol">~{{ epochData.remaingTime | date:'yyyy-MM-dd HH:mm' }}</div>
```
knorrium, line 211: typo: `remainingTime`

frontend/src/app/dashboard/dashboard.component.ts
```diff
@@ -108,38 +114,67 @@ export class DashboardComponent implements OnInit {
...
+            green: green + '%',
+            red: red + '%',
+            change: difficultyChange,
+            progress: base.toFixed(2),
+            remainingBlocks,
+            timeAvg: timeAvgMins.toFixed(0),
+            colorAdjustments,
+            blocksInEpoch,
+            newDifficultyHeight: block.height + remainingBlocks,
+            remaingTime: remainingBlocsMilliseconds + nowMilliseconds
```
knorrium, line 174: typo: `remainingTime`

frontend/src/app/dashboard/dashboard.component.scss
```diff
@@ -213,4 +213,91 @@
...
+.small-bar {
+  height: 8px;
+  top: -4px;
+}
+
+.difficulty-adjustment-container {
+  display: flex;
+  flex-direction: row;
+  justify-content: space-around;
+  .shared-block {
```
knorrium, line 227: @MiguelMedeiros it looks like you're adding a span with this class to components that don't need it (footer, transaction and mempool-blocks).

jaonoctus (approved):
ACK

frontend/src/app/dashboard/dashboard.component.ts
MiguelMedeiros: Thanks! :+1: 
Fixed here: https://github.com/mempool/mempool/pull/602/commits/3463188e45cc3d0e323bdc8248063993803fd47a

frontend/src/app/dashboard/dashboard.component.ts
MiguelMedeiros: Thanks! :+1: 
Fixed here: https://github.com/mempool/mempool/pull/602/commits/3463188e45cc3d0e323bdc8248063993803fd47a

frontend/src/app/dashboard/dashboard.component.html
MiguelMedeiros: Thanks! :+1: 
Fixed here: https://github.com/mempool/mempool/pull/602/commits/3463188e45cc3d0e323bdc8248063993803fd47a

frontend/src/app/dashboard/dashboard.component.scss
MiguelMedeiros: I talked to @softsimon about it. We decided to add the html tag (span) with a new class to handle css separately in each component. But to work the translation correctly I needed to change all the components that use the translation of "shared.blocks" and "shared.block". So we can use the same strings without creating new ones.

wiz:
Need to have the word "difficulty" in the title IMO

wiz:
Need the word "difficulty" in mobile layout IMO

wiz:
@softsimon what do you think of: instead of copy/pasting the entire `time-since` component and duplicating it into a new `time-until` component, why not simply make a `time` component which can take negative or positive time deltas

softsimon:
> @softsimon what do you think of: instead of copy/pasting the entire `time-since` component and duplicating it into a new `time-until` component, why not simply make a `time` component which can take negative or positive time deltas

No, I don't think so. They are different and doesn't share much code.

wiz:
Maybe add a "difficulty adjustment" label above the box, with a "fee estimates" label above the left side box to match, and tweak individual labels to something like this?

Screen Shot 2021-07-03 at 19 48 19

softsimon:
Personally I love and use the "red" and "green" bar that appear visualizing how much behind or ahead the progress is. @MiguelMedeiros removed with the reason that it is confusing and nobody understands it.

I vote to return it, I don't believe it causes extra confusion. What does @wiz and @knorrium think?

Screen Shot 2021-07-05 at 15 38 08

Screen Shot 2021-07-05 at 15 38 36

jaonoctus:
As a long-time user, I never understood this progress bar until I read the thread. I think many other people will agree as well that it is quite confusing.

MiguelMedeiros:
What about now @wiz ?

image

image

image

image


difficulty-adjustment-title

softsimon:
My preference. Centered titles with some padding:
Screen Shot 2021-07-09 at 22 09 46

MiguelMedeiros:
@softsimon much better! Thanks!

difficulty-adjustment-responsive-final

softsimon (changes requested):
We need to refactor all the i18n strings that has to do with time periods (day, hour, minutes etc) so we can re-use the same strings for the current 3 use cases for example:

```
$localize`:@@time-until.year:In ~ (i18n string year)

$localize`:@@time-ago.year:(i18n string year) ago

$localize`:@@time-after.year:After (i18n string year)
```

PR #608, opened 2021-07-05
Only display Trade Mark sign on english language.

softsimon:


MiguelMedeiros:
Reviewed. 
But I have a question about the `tm` that appears at the bottom of the about page: 
Should we remove it as well (if it's not `en-US` )?

image

PR #612, opened 2021-07-06
Layout - Latest Transaction table overflow.

MiguelMedeiros:
## Description
Sometimes the amount of the transactions causes a table overflow.

## Current
table-overflow

## Fixed
Reduced the amount to 4 decimal cases and fixed the table layout.

table-overflow-fixed


softsimon:
tested ACK

PR #613, opened 2021-07-07
UI/UX - Scroll by clicking on the pagination button.

MiguelMedeiros:
## Description
When you click on a pagination button the page should scroll to the transactions title.

## Actual behaviour
pagination-scroll-current

## Expected behaviour
pagination-scroll

### DIFF
--- a/frontend/src/app/components/block/block.component.ts
+++ b/frontend/src/app/components/block/block.component.ts
@@ -177,12 +177,13 @@ export class BlockComponent implements OnInit, OnDestroy {
     const start = (page - 1) * this.itemsPerPage;
     this.isLoadingTransactions = true;
     this.transactions = null;
+    target.scrollIntoView(); // works for chrome
 
     this.electrsApiService.getBlockTransactions$(this.block.id, start)
      .subscribe((transactions) => {
         this.transactions = transactions;
         this.isLoadingTransactions = false;
-        target.scrollIntoView();
+        target.scrollIntoView(); // works for firefox
       });
   }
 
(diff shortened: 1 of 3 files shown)


softsimon:
Great, but because of my OCD I have a suggested change: Reducing the scrollTop with few pixels creating a padding.

What it looks like now:

Screen Shot 2021-07-09 at 17 50 08

What I'd like the scroll position to be when changing page (approximately):

Screen Shot 2021-07-09 at 17 50 48

frontend/src/app/components/block/block.component.ts
```diff
@@ -172,12 +172,13 @@ export class BlockComponent implements OnInit, OnDestroy {
       halvenings--;
     }
   }
-
   pageChange(page: number) {
     const start = (page - 1) * this.itemsPerPage;
     this.isLoadingTransactions = true;
     this.transactions = null;
 
+    document.querySelector('#block-tx-title').scrollIntoView();
```
softsimon, line 180: Access the DOM using ViewChild .. the angular way instead of the global querySelector

https://stackoverflow.com/q/38944725

MiguelMedeiros: Done!

MiguelMedeiros:
> Great, but because of my OCD I have a suggested change: Reducing the scrollTop with few pixels creating a padding.
> 
> What it looks like now:
> 
> Screen Shot 2021-07-09 at 17 50 08
> 
> What I'd like the scroll position to be when changing page (approximately):
> 
> Screen Shot 2021-07-09 at 17 50 48

Done!

softsimon (changes requested):
1. The scrollIntoView doesn't seem to work in Firefox, so this feature only works on Chrome
2. The scroll position is a bit too far up now as the Details button is visible, not like the screen shot I posted in previous post

PR #615, opened 2021-07-08
UI/UX - Add skeleton for blocks preload.

### LINKED ISSUE #168: Replace the "waiting for blocks" with a skeleton blockchain loader

MiguelMedeiros:
Fix #168.

## Tasks

- [x] Remove old `waiting blocks` pre-loader.
- [x] Add new blocks skeleton pre-loader.
- [x] Add e2e testing for blocks skeleton pre-loader.

## Preview
skeleton-blocks-preload

### DIFF
--- a/frontend/cypress/integration/mainnet/mainnet.spec.ts
+++ b/frontend/cypress/integration/mainnet/mainnet.spec.ts
@@ -16,6 +16,17 @@ describe('Mainnet', () => {
       cy.wait(1000);
     });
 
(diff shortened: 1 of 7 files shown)


softsimon (changes requested):
I think the amount of loading mempool blocks should be the same as the amounts that will be displayed on screen when the loading is done.

There is some code to calculate how many will fit the screen in the end.

`    const blocksAmount = Math.max(2, Math.floor(window.innerWidth / 2 / (this.blockWidth + this.blockPadding)));`

PR #627, opened 2021-07-13
UI/UX - Fix arrow position.

### LINKED ISSUE #625: UX: CSS arrows are broken in Firefox
Screen Shot 2021-07-12 at 17 00 12

MiguelMedeiros:
Fix #625.

## Preview
image


softsimon (changes requested):
The distance between the arrow and address has increased. Should be the same on both sides, and the arrow design is no longer even in shape.
Before:
Screen Shot 2021-07-13 at 12 03 31



After:
Screen Shot 2021-07-13 at 12 01 36

PR #628, opened 2021-07-13
Expose the difficulty adjustment information in an API #622

### LINKED ISSUE #622: Expose the difficulty adjustment information in an API
Some people have asked to be able to get the difficulty adjustment estimation from our API.

1. Backend: Create a new NodeJS endpoint to retrieve the difficulty adjustment estimate information
2. Frontend: Add the new API information to the API doc page

The API should return an object containing the estimated change, how many % of the period has progressed, how many remaining blocks and if possible also the estimated date. The code for this calculations already exists in the frontend (this Pull Request https://github.com/mempool/mempool/pull/602 which is not merged yet in the time of writing ), but it needs to be in the backend as well for the API.

priyanshiiit:
Fix #622 

**Tasks**
---
- [x] estimated change
- [x] % of period progressed
- [x] remaining blocks  
- [x] estimated date
- [x]  API information in frontend

**Preview Images**
---
Screenshot from 2021-07-12 19-07-30

API Endpoint Output
Screenshot from 2021-07-12 18-02-07


softsimon (changes requested):
- Rename the new endpoint to `difficulty-adjustment`
- On the API docs page the `GET /api/difficulty-adjustment` should be a clickable link similar to most of the other api docs
- I think we can remove both the `now` and `date` as API users can expect the time to be now and just check their local clock.
- `DATime` can be removed as this is the previous difficulty adjustment date, we are only interested in the next one
- I believe remainingBlockSeconds is the remaining time until adjustment. It should be renamed to simply `remainingTime` and be the unix time in seconds and rounded
- const blockHeight=await bitcoinApi.$getBlockHeightTip(); should be replaced by blocks.getCurrentBlockHeight();
- There seems to be something wrong with the difficulty calculation. Not sure what.. but the API tells me 94% done .. but we're only about 80% in the period.

priyanshiiit:
Implemented the changes. Please review.

PR #643, opened 2021-07-19
Update nginx configuration for cache settings and new locales

wiz:
* Consolidates add_header statements into single top-level section
* Updates locales to match frontend/src/app/app.constants.ts
* Re-orders locales to match locale selector for easier checking

### DIFF
--- a/nginx-mempool.conf
+++ b/nginx-mempool.conf
@@ -5,26 +5,35 @@
 
 	index index.html;
 
+	# enable browser and proxy caching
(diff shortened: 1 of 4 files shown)


Emzy (approved):
ACK code review
Looks good.
I'm not sure if "Vary Accept-Language" is really needed, because we have different URLs for languages, but it's not a problem to have it anyway.

Emzy:
ACK tested production
I have the production configs running on my instance mempool.emzy.de

PR #650, opened 2021-07-20
Bugfix: Change mempool block time precision.

MiguelMedeiros:
### Description

The blocks of mempool were rounding up and it was making a confusion about the precision of the block time.

### Actual Behaviour
image
`Math.round`

### New Behaviour
image
`Math.floor`


softsimon:
I still don't really like that it will very often show 9 minutes until the next block .. just because difficulty adjustment is positive by a few single %. 

Not sure what the best solution is, but to me at least +5% positive adjustment is needed for 9 min to be a realistic suggestion. In reality +10% is needed for 10 min to become 9 min.

softsimon (changes requested):
Make sure the transaction ETA date is matching the mempool block time that is now based on difficulty estimate.

Screen Shot 2021-07-22 at 00 23 44

PR #654, opened 2021-07-24
Doc: Add new endpoints to api documentation.

MiguelMedeiros:
Tasks:

- [x] Remove `Difficulty` tab and `Fee` tab and create a new one called `General` tab.
- [x] Reorder tabs.
- [x] Add `block header` examples.
- [x] Add `difficulty adjusment` examples.


softsimon (changes requested):
1. The endpoint should be `GET /api/v1/difficulty-adjustment` with `/v1/` since it's a NodeJS backend API. On production `/api` goes to esplora electrum server so it doesn't work.
2. Maybe move the GET Mempool Blocks from General to the Mempool tab?

PR #655, opened 2021-07-24
Add previous adjustment retarget.

### LINKED ISSUE #640: Add previous difficulty adjustment visible on dashboard
Myself and many people are always interested in the previous adjustment, so it's more important than showing the current block speed (that will be displayed on the mempool blocks).

Something like this:

Screen Shot 2021-07-18 at 12 13 12

MiguelMedeiros:
Fix #640.

### Preview
image

### DIFF
--- a/frontend/src/app/dashboard/dashboard.component.html
+++ b/frontend/src/app/dashboard/dashboard.component.html
@@ -215,8 +215,8 @@
           </div>
           <div class="item">
             <h5 class="card-title" i18n="difficulty-box.estimate">Estimate</h5>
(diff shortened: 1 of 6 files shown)


softsimon (changes requested):
1. We should not use the `getDifficultyAdjustment$()` API from the dashboard causing more API calls. I have added the variable `previousRetarget` to the initial WebSocket data the same way `lastDifficultyAdjustment` is from the websocket. The frontend also need to handle the switching of periods, updating the local previousRetarget observable
2. Previous retarget should have %-sign

PR #675, opened 2021-07-31
Fix concurrent asynchronous calls.

### LINKED ISSUE #633: feeRange is undefined causes blockchain to not be displayed
This error shows up sporadically, I believe since the new blockchain skeleton loader.

`feeRange is undefined` is shown in the console.

MiguelMedeiros:
Fix #633.


softsimon:
Since `isLoading$` is never used anymore, you should remove them from all the steps, blockchain.component.ts/html etc

softsimon:
Also remove this check `*ngIf="block.feeRange !== undefined"` since this was not the cause of the problem.

PR #696, opened 2021-08-05
UI/UX: Fix Blockchain skeleton preloader.

### LINKED ISSUE #688: UX: When switching networks, the skeleton loader is not visible

MiguelMedeiros:
Fix #688 .

### DIFF
--- a/frontend/cypress/integration/mainnet/mainnet.spec.ts
+++ b/frontend/cypress/integration/mainnet/mainnet.spec.ts
@@ -41,20 +41,12 @@ describe('Mainnet', () => {
         cy.visit('/');
         cy.waitForSkeletonGone();
 
-        changeNetwork("testnet");
-        changeNetwork("signet");
-        changeNetwork("liquid");
+        cy.changeNetwork("testnet");
+        cy.changeNetwork("signet");
+        cy.changeNetwork("liquid");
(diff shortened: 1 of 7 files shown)


frontend/cypress/integration/mainnet/mainnet.spec.ts
```diff
@@ -12,30 +12,50 @@ describe('Mainnet', () => {
...
+    it('loads skeleton when changes between networks', () => {
+        cy.visit('/');
+        cy.waitForSkeletonGone();
+
+        changeNetwork("testnet");
+        changeNetwork("signet");
+        changeNetwork("liquid");
+    });
+
+    function changeNetwork (network: "testnet"|"signet"|"liquid" ){
```
knorrium, line 49: @MiguelMedeiros thanks for adding a test! Custom commands that are shared across specs should go into `support/commands.ts` like this:

https://github.com/mempool/mempool/blob/master/frontend/cypress/support/commands.ts#L49-#L53

see https://docs.cypress.io/api/cypress-api/custom-commands for reference

MiguelMedeiros: Fixed here:
https://github.com/mempool/mempool/pull/696/commits/b4525b05f6d4fb422d4aeed0f446081a23e56986

PR #702, opened 2021-08-07
Resume tx tracking when network goes offline.

### LINKED ISSUE #609: Confirmed transactions may still show as unconfirmed
Sometimes transactions can go from unconfirmed to confirmed without the website getting this information, probably due to a websocket connection error.

The website/backend should be able to handle all cases of connection error and reconnect and push the updated TX status to the client.

In this case:

```[Log] CloseEvent (main-es2020.a72ffada6431bf54b29d.js, line 3)
[Log] WebSocket error, trying to reconnect in 10000 seconds (main-es2020.a72ffada6431bf54b29d.js, line 3)
[Error] WebSocket network error: The operation couldn’t be completed. Socket is not connected
```

Same transaction on 2 browser windows. The websocket is alive and working.

softsimon:
fixes #609

### This PR fix the following 2 cases that is broken currently:

A). Start tracking a unconfirmed transaction. Let your websocket connection disconnect... and back online again.. TX tracking should now be resumed and TX confirmation will be notified as expected

B) Start tracking a unconfirmed transaction. Let your websocket connection disconnect. While disconnected, your transaction is mined in a block. Resume websocket connection. The client is now notified and transaction moves to confirmed into the new block.

Worth noting that this should be tested both on `esplora` and `bitcoind` backend as the TX tracking behavior have been affected and behave differently depending on backend.

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -53,18 +53,25 @@ class WebsocketHandler {
               if (parsedMessage['watch-mempool']) {
                 const tx = memPool.getMempool()[client['track-tx']];
                 if (tx) {
-                  if (config.MEMPOOL.BACKEND !== 'esplora') {
+                  if (config.MEMPOOL.BACKEND === 'esplora') {
+                    response['tx'] = tx;
+                  } else {
+                    // tx.prevouts is missing from transactions when in bitcoind mode
                     try {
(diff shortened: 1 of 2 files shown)


MiguelMedeiros (approved):

MiguelMedeiros:
Tested ACK.

PR #720, opened 2021-08-15
Fix update-config to work with string values

knorrium:
String values like used by the new `BASE_MODULE` config key were not working


knorrium:
I'll bundle this with the new proxy changes.

Issue #514, opened 2021-05-12
Transaction fee (sats/vB) should display decimals depending on 


softsimon:
Currently fee rate on transactions etc are rounded to 1 decimal.

It should be like this:

>= 100 no decimals
>= 10 one decimal
< 10 two decimals

wiz:
significant digits

PR #722, opened 2021-08-16
Updated significant digits of transaction fee

priyanshiiit:
**Fix** #514 

**Fee rate updated as:**
>`>=100 no decimal`
`<100 && >=10 one decimal`
`< 10 two decimals`

Preview
---
Peek 2021-08-16 18-50


softsimon (changes requested):
Works well!

One nit-pick with the code styling in the Pipe to be consistent and for increased readability and reduce bugs. All if-statements should have { } brackets and some added spaces after if-statement, brackets and between operators.

Like this:

```
if (fee >= 100) {
  return fee.toFixed(0);
} else if (fee < 10) {
  return fee.toFixed(2);
}
return fee.toFixed(1);
```

is much more readable than
```
if(fee>=100)
return fee.toFixed(0);
else if(fee<10)
return fee.toFixed(2);
return fee.toFixed(1);
```

priyanshiiit:
**Thanks for review!**
Made the changes.

wiz:
Excellent

PR #724, opened 2021-08-17
Only display "Load all" if there are at least 3 more items to load

### LINKED ISSUE #631: There is no point with "Load more" if there is only 1 more address
Example transaction:

https://mempool.space/tx/df06f37266b7696808ddc9d920e24da2f8c63096b93157a801d7aaff0a621187

There is a "load more" button even though there is only 1 more address to load, which is unnecessary.

**Solution**:
Only display load more if there is at least 3 more items to load.

rishkwal:
Resolves #631 
## Description
It does not show "Load all" unless there are at least 3 more addresses to load.
## Screenshots
Below transaction has 11 outputs and all of them are displayed.
Screenshot 2021-08-17 at 3 29 17 PM
Below transaction has 37 outputs, 10 are displayed with "Load all" option
Screenshot 2021-08-17 at 3 28 58 PM


softsimon (changes requested):
This works well! 

But can you do the same thing for the list of inputs as well as the outputs. 
I don't have a example transaction, but I did some mock code.

Screen Shot 2021-08-17 at 15 41 54

PR #733, opened 2021-08-18
Bugfix: Fix bisq dashboard tables overflow.

MiguelMedeiros:


softsimon:
Getting some overflow with iPhone X as screen size

Screen Shot 2021-08-19 at 02 12 41

MiguelMedeiros:
> Getting some overflow with iPhone X as screen size
> 
> Screen Shot 2021-08-19 at 02 12 41

I will replace trading view charts with my new charts PR (so try refreshing the page after resizing).
I've changed the margin of the nav-bar item so they can fit properly.

Issue #600, opened 2021-06-30
High Memory Consumption


Cristian-Bejan:
I noticed today that mempool.space consumed a lot of memory in Chrome Version 91.0.4472.114 (Official Build) (64-bit) for Windows - around 4GB. It's not the first time and seems to happen independent of the Chrome version and when the tab has been opened for a long time. A refresh fixes the issue.
![Mempool-Memory](https://user-images.githubusercontent.com/60579945/124008951-af7ee800-d9e5-11eb-965b-882f9563acb0.PNG)

wiz:
yeah... this memory leak is a long-standing issue that I believe got introduced when we made a change to the chart library for inverting it - IIRC when disabling the charts on the dashboard it doesn't reproduce

Cristian-Bejan:
Thanks for clarifying that @wiz . Didn't notice that it's possible to disable charts. Nice! Learning something everyday 😎

wiz:
@knorrium has been working hard on this investigation and found many leaks so far in the chart library

softsimon:
Yes, the legacy chart library is most likely to blame. It was the cause last time I investigated this.

We are planning to replace the library in the next version update so will start working on it soon. #94

wiz:
Fixed by #755

PR #738, opened 2021-08-21
Feature: New charts library.

### LINKED ISSUE #94: Charts: Investigate and replace the current chartist library
To fix various bugs and limitations like #70, but also add new features like hover, we need to replace the current deprecated Chartist chart library that is used throughout the site to display mempool and fee data. The graph and colors should not differ too much from the current design.

Consider the following libs:

http://dygraphs.com/
https://d3js.org/

The charting libs should also be wrapped in angular components for seamlessness and tree shaking ability.

MiguelMedeiros:
Fix #94 and #600 .

## Tasks

- [x] Create new component `incoming-transactions-graph`;
- [x] Refactor component `mempool-graph`;
- [x] Refactor component `fee-distribution-graph`;
- [x] Add `incoming-transactions-graph` to `dashboard`;
- [x] Add `incoming-transactions-graph` to `statistics`;
- [x] Add `incoming-transactions-graph` to `television`;
- [x] Add `mempool-graph` to `dashboard`;
- [x] Add `mempool-graph` to `statistics`;
- [x] Add `mempool-graph` to `television`;
- [x] Remove `chartist.component`;

## Preview
preview-dashboard


Dahershiser:
👍 I'm a lil confused but I'll work on it

wiz:
This is really cool. Initial thoughts:
* You need to display the date in addition to the time when viewing graphs >2h period, for example check https://mempool.ninja/graphs#1y and you will see it only displays the time
* Same goes for the y-axis labels at the bottom when viewing graphs >2h period, only the time is displayed
* These 2 cases apply to both charts
* It would be cool to more clearly communicate what the rate of incoming transactions is somehow, newbies are often confused by this

wiz:
Sorry maybe I'm missing something but how do you toggle on/off the colors tiers on the stacked chart?

wiz:
Also missing the button to invert the chart colors. And you should invert the order of the overlay legend to have the order of colors so they always match how they are stacked in the graph.

wiz:
Might be better without the crosshairs and just focus on polishing the overlay legend

MiguelMedeiros:
## TASKS
- [x] Add total sum to mempool chart.
- [x] Add zoom tools.
- [x] Add different theme for charts `big` and `small` (default).
- [x] Fix date format on mouseover:
    - [x] 2h/24h shows date format `HH:mm`;
    - [x] Default date format `MM/dd HH:mm`.
- [x] Fix animations on graphs page.
- [x] Fix overflow tv page.
- [x] Remove `crosshair` on mouseover, changed to `line`.
- [x] Fix custom tooltip styles:
    - [x] Add highlight tooltip label color when mouseover that fee range;
    - [x] Style symbol, value, total value an title with proper color and font-size;
    - [x] Add different background-color to pages with `big` theme chart.
- [x] Remove inverted button (will add in a future PR).
- [x] Remove fee range labels (will add in a future PR).
- [ ] Add e2e testing.

Xekyo:
Hi Miguel, I am a heavy user of mempool charts and I'm really excited for the chart makeover! :nerd_face:
wiz invited me to comment on this PR.

Especially, I frequently post screenshots of mempool charts to comment on the current mempool state. Given that the chart is almost exclusively interesting in the context of the time it was recorded at, I'd prefer if even the 2h/24h chart had the date.

I noticed that the order of the layers in the chart (high feerate at the top) is reverse of the order of the legend (low feerate at the top). It would feel more natural to me if they were both in the same order, i.e. if the legend were also to have the lowest feerate at the bottom.

I often want to know _both_ the transaction weight in a specific band and the total transaction weight at and above this feerate. The former is for example useful to observe which bands are growing or where big batches of consolidation transactions get added to the mempool. The latter is more useful to determine how long it takes until transactions at specific feerates may be included in blocks. I surmise that a 3-column legend would be rather wide, but would it be possible to have both the specific band's volume and the sum of all transactions down to that feerate per band? If that were deemed too aesthetically unpleasing, would it perhaps be possible to show the two values for the band that the cursor is hovering over in an area of the chart that doesn't show data?

To minimize the size of the legend, perhaps the rows for which there are no transactions with that feerate or higher could be not rendered. I.e. if the highest band with transactions were 30-40, 40-50 would show zero/zero and the ones above would not show up in the legend.

image
I noticed that especially when tracking significant parts of the chart like the instant a block was found, it can be hard to determine where that significant point exactly lies in the chart because of the buttons obscuring the curve in the chart. Perhaps these buttons could be a little less opaque?

wiz:
> the staging environment is reporting 20 MvB, while the main page is showing 6.5 MvB.

it's a bug which miguel is already working on 😅

wiz:
@Xekyo mempool.ninja has been updated with latest fixes

wiz:
- On the main dashboard, we are missing the y-axis on the incoming transactions chart
- The x-axis labels would probably look better being 2 lines, with a line break between the date and time
- Looks like the labels on the y-axis are getting truncated
Screen Shot 2021-09-08 at 7 51 49

wiz:
And lastly, need a y-axis slider to hide fee rate tiers, as currently works on the old chart library

wiz:
Oh, and some users want to invert the order of the fee rate tiers, we have this feature on the current site

Xekyo:
image

1) Occasionally, there is a line break introduced in the headline of the legend, I think this is probably unintended.
2) The third column in the legend gives the percentages that each band has relative to the total. Thinking more about it, I'd prefer it were the sum of the virtual size of this band and all bands paying a higher feerate. (As on [jochen-hoenicke.de/queue](https://jochen-hoenicke.de/queue):

| Current: %  | Suggested: Sum of vsize | Comment |
| ------------- | ------------- | ------------- |
|  0.0 %  |  0 | |
|  0.2 %  | 0.02 MvB  |
| 1.0 % |  0.12 MvB | 0.1 MvB + 0.02 from band with higher feerates |
|  0.2 %  |  0.14 MvB | 0.02 MvB + 0.12 MvB from higher feerates
| 1.0 % |  0.24 MvB  | 0.1 MvB + 0.14 MvB from higher feerates
| ... |  ...  | ... |

3) The disks that are drawn at the intersection of the cursor line and the graph occlude the graph where it cannot be seen. Perhaps it would b better only show a small dot at the intersections or otherwise make it less covering?

4) I like the bottom giving the details of the current band as is!

Xekyo:
Just a quick follow-up to substantiate the motivation for my proposal for the third column: one of the main insights users wish to glean from a mempool chart is _"How much transaction data has a higher priority than mine?"_. The easiest way to make that available is by summing all transactions' vsizes above a given feerate as I propose for the third column.

wiz:
@Xekyo updated mempool.ninja with a WIP commit just now

wiz (approved):

Issue #465, opened 2021-04-20
Bitcoind: Lazy load inputs to speed up loading


softsimon:
When loading a transaction, block transaction, and the address page, it can be very slow due to the need to lookup all the previous input transactions.

Load up to 5 of them, and let the frontend know that there are more to lazy load, and when clicking load more, a new API endpoint is called to load remaining inputs.

Examples of "slow" (more than 12 inputs) block transactions pages:
Mainnet: https://mempool.space/block/0000000000000000000ef31f7271b9ed0fbc5c56c233241ab16201769d533191  (go to page 7 or page 8 with Bitcoind mode)

Signet: https://mempool.space/signet/tx/6fb077083adfa6115cc6ebda3e6dfef3813d11f50fc1972375ccf25e60cada24

PR #739, opened 2021-08-21
Bitcoind : Lazy Load of Transaction Inputs to Speed Up Loading

priyanshiiit:
**Fix** #465 

**Preview**
---
Peek 2021-08-21 20-46


softsimon (changes requested):
This PR/feature should not break the current behavior with displaying 10-12 items and then Load More (X items). 

The backend should keep the same list of inputs (vin) so that the frontend knows how many it is, but they should be flagged or handled in the frontend that the prevout is missing. So when pressing Load more .. instead of the X amount of items, a skeleton loader should be visible instead for the same amount of items, because we now know the items being loaded.

When the user presses Load more, a backend API should request only the missing Vin inputs data. 

The default scenario is not using bitcoind as backend, all inputs will always be loaded in that default scenario, so the frontend should only handle the above scenario when it detects that some of the inputs/load more inputs are missing prevout/fee.

wiz:
Closing due to stalled, feel free to re-open if you guys are still working on this

PR #749, opened 2021-08-27
Use block cache when searching or opening a recent block.

softsimon:
fixes #715

This PR improves the following things:

- When opening / reloading the page straight to a block, median fee is no longer missing
- When searching for a block height or hash that is in recent cache, use the cache for instant opening


MiguelMedeiros (approved):

MiguelMedeiros:
Tested ACK.

MiguelMedeiros (approved):

Issue #740, opened 2021-08-21
Make price feed update configurable


softsimon:
Current default is every minute (1000 * 60), but set default to every 1 hour, and make it configurable.

softsimon:
Fixed in https://github.com/mempool/mempool/pull/751

PR #751, opened 2021-08-28
Made Price feed update configurable

priyanshiiit:
**Fix** #740 

**Task**
Previously default was 1 minute (1000 * 60), changed default to 1 hour, and made it configurable.


softsimon (changes requested):
Since it's an interval a config name like `PRICE_FEED_UPDATE_INTERVAL` is probably better

Also better if the config is expressed second.. so do * 1000 to the variable in the setInterval function, so the default is 3600

priyanshiiit:
Hi @softsimon!
Thanks for the review.
I've made the requested changes.

PR #765, opened 2021-09-04
Updated address regex to handle all types of addresses.

### LINKED ISSUE #301: Add support for Bech32 bc1 addresses being case-insensitive
### Description

Bech32 (Segwit) bitcoin addresses are case-insensitive, per https://archive.vn/tzxCb#selection-909.18-909.92
The search box should accommodate for this, but does not currently.

### Steps to reproduce

Search on example address "BC1Q0QFZUGE7VR5S2XKCZRJKCCMXEMLYYN8MHX298V".

### Expected behaviour

Should match "bc1q0qfzuge7vr5s2xkczrjkccmxemlyyn8mhx298v" and show results.

### Actual behaviour

Tool does NOTHING, not even an error(?!).

### Screenshots

Screenshot from 2021-01-19 05-31-47

#### Device or machine

Ubuntu 20.04, Firefox

### LINKED ISSUE #750: Liquid: Search field must handle all types of Liquid addresses and transaction IDs

softsimon:
fixes #301
fixes #750

This PR fixes the following bugs:
* All Liquid address types are now accepted as valid in the search box.
 Examples:
 1. lq1qqgddwun5n54jq8za4rkddalngqtcc6skcelg7mquc40ayscr285ksaxs3a72pf6knc4zrx7pk7ggwueu0lkdw009cvvwhqel2
 2. VJL8eyC5Hux6khmKNLDKTqhkihUkupeDhU8DjbFctmRWs2b5PwE2VnuqEUmdPYwXTPTSjWrkvxJJvUbH
 3. GuHku7sQRYQqSqvyWGd2Vobq3WLXBpEzZU
* Uppercase Bc1 addresses now works in the search box.
Example: BC1Q0QFZUGE7VR5S2XKCZRJKCCMXEMLYYN8MHX298V

Uppercase bc1 addresses also gets converted to lower case in the UX.

Make sure that tx-tracking still works in the address view, especially for uppercase segwit addresses. Note that tracking of some Liquid address types still doesn't work.

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -82,7 +82,11 @@ class WebsocketHandler {
           if (parsedMessage && parsedMessage['track-address']) {
             if (/^([a-km-zA-HJ-NP-Z1-9]{26,35}|[a-km-zA-HJ-NP-Z1-9]{80}|[a-z]{2,5}1[ac-hj-np-z02-9]{8,100}|[A-Z]{2,5}1[AC-HJ-NP-Z02-9]{8,100})$/
               .test(parsedMessage['track-address'])) {
-              client['track-address'] = parsedMessage['track-address'];
+              let matchedAddress = parsedMessage['track-address'];
+              if (/^[A-Z]{2,5}1[AC-HJ-NP-Z02-9]{8,100}$/.test(parsedMessage['track-address'])) {
+                matchedAddress = matchedAddress.toLowerCase();
+              }
+              client['track-address'] = matchedAddress;
             } else {
               client['track-address'] = null;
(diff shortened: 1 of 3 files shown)


wiz:
Shouldn't the regex patterns be different for each network, i.e. mainnet, testnet, signet, liquid, and bisq ?
Also, it would be nice if you can break up the complex regex patterns into chunks with comments to explain what is going on like this: https://github.com/wiz/bisq/blob/7f92d12cc90e9f0156d81c9c252470c291420945/desktop/src/main/java/bisq/desktop/util/GUIUtil.java#L1120

wiz (approved):

PR #768, opened 2021-09-06
Liquid: Display unconfidential address and fix tracking

### LINKED ISSUE #761: Liquid: Support tracking of confidential addresses

softsimon:
fixes #761

Loading a liquid confidential address like: lq1qqgddwun5n54jq8za4rkddalngqtcc6skcelg7mquc40ayscr285ksaxs3a72pf6knc4zrx7pk7ggwueu0lkdw009cvvwhqel2

It will now look like this, and tracking should now work since the unconfidential address is used.

Screen Shot 2021-09-06 at 10 18 30


wiz (changes requested):
1) Regex fails to match legacy non-segwit confidential liquid addresses such as  `VJL85kr4chhQBjKTzf8TE3nWnR1QaU93MzyjgzYkxAJiYrvajR5mDrJ9gUHvrd9DUozQzs88WWMpMdMp`
2) When using `esplora` backend, requesting `/api/v1/validate-address/VJL85kr4chhQBjKTzf8TE3nWnR1QaU93MzyjgzYkxAJiYrvajR5mDrJ9gUHvrd9DUozQzs88WWMpMdMp` returns 404

backend/src/index.ts
```diff
@@ -251,6 +251,7 @@ class Server {
         .get(config.MEMPOOL.API_URL_PREFIX + 'address/:address/txs', routes.getAddressTransactions)
         .get(config.MEMPOOL.API_URL_PREFIX + 'address/:address/txs/chain/:txId', routes.getAddressTransactions)
         .get(config.MEMPOOL.API_URL_PREFIX + 'address-prefix/:prefix', routes.getAddressPrefix)
+        .get(config.MEMPOOL.API_URL_PREFIX + 'validate-address/:address', routes.validateAddress)
```
wiz, line 254: Shouldn't this be wrapped in a "if liquid enabled" statement?

softsimon:
>     1. Regex fails to match legacy non-segwit confidential liquid addresses such as  `VJL85kr4chhQBjKTzf8TE3nWnR1QaU93MzyjgzYkxAJiYrvajR5mDrJ9gUHvrd9DUozQzs88WWMpMdMp`
> 
Which regex? How do I reproduce the problem?

wiz:
Just search for the address I gave, the regex doesn't match so it doesn't try to lookup the unconfidential address for it.

wiz (approved):

PR #775, opened 2021-09-10
Set canonical URLs for new 3 site structure

wiz:


frontend/src/app/components/app/app.component.ts
```diff
@@ -39,7 +39,13 @@ export class AppComponent implements OnInit {
   ngOnInit() {
     this.router.events.subscribe((val) => {
       if (val instanceof NavigationEnd) {
-        this.link.setAttribute('href', 'https://mempool.space' + (this.location.path() === '/' ? '' : this.location.path()));
```
softsimon, line 42: May I suggest a code change to something like this to void the repeating setAttribute

```
      if (val instanceof NavigationEnd) {
        let href = 'https://mempool.space';
        if (this.stateService.env.BASE_MODULE === 'liquid') {
          href = 'https://liquid.network';
        } else if (this.stateService.env.BASE_MODULE === 'bisq') {
          href = 'https://bisq.markets';
        }
        this.link.setAttribute('href', href + (this.location.path() === '/' ? '' : this.location.path()));
      }
```

softsimon (approved):

PR #782, opened 2021-09-15
Fix RTL issues

TechMiX:
During some updates such as 1e9f131a2a30302b791612b924fbb7c0d13d8d6c , 9dae7020c8b8a4b3cf9676a073cbb0de5cbd21bc and 8208bbf0b7e398a3bcce5e64d58348728ff956fe some RTL hints (like the LRM control character) were removed. This PR brings them back and also add some more justifications to the `.rtl-layout` class in `styles.scss` file.


wiz (approved):
Tested ACK @ 2c496e9a50cf031b6b04ccb1ac8c8e0f65853480

Thanks for fixing this, I guess we forgot to test v2.2.2 on RTL locales, let's try to be more careful in the future not to break RTL :) @MiguelMedeiros @knorrium

PR #794, opened 2021-09-17
Pulled from transifex

softsimon:


wiz:
you really want to update these 3 strings right now? it seems kinda insignificant heh

for existing locales I think we would wait until closer to release time when there is actual changes to be done

wiz (approved):

Issue #718, opened 2021-08-15
Implement new dashboard widgets for Liquid


wiz:
- [ ] Monitor status of hard fork for dynafed activation

similar to taproot activation lock-in
it happens in 144 block epochs
get version number, look at bit 25(&10000000000000000000000000).
if 75% signal in an epoch
range is from [%144=1 to %144=0]

- [ ] Monitor status of functionaries

identify functionary ID from coinbase transaction bytes
reference implementation is inside `parser.zip`
tag each block by who "mined" it using functionary ID
also on blocks page table indicate which functionary
want to see a "mining dashboard" which shows if functionaries are offline
based on last XXXX blocks mined, need a main dashboard widget to show 15/15 functionaries online

- [x] Monitor status of L-BTC in circulation and BTC pegged-in

use liquid bitcoin asset data to get the two numbers
display on main dashboard as a new widget like 3000 BTC pegged in / 3000 L-BTC in circulation
numbers will be slightly off due to human error where some L-BTC was burned, but >100% reserves

- [ ] Monitor status of UTXOs owned by federation

refer to screenshots sent privately that identify UTXOs in federation 11 of 15 multisig wallet
refer to code in `parser.zip` to idenfiy UTXOs
make a new page with table that lists all Liquid UTXOs and list their locktime expiration

wiz:
The first one might not make sense for mempool, need to consider alternatively forking the taproot.watch website code for the dynafed hard-fork lock-in monitoring

softsimon:
Closing to create new separate issues

https://github.com/mempool/mempool/issues/1025
https://github.com/mempool/mempool/issues/1026

PR #799, opened 2021-09-18
Liquid L-BTC widgets (Backend)

softsimon:
This is part of the Liquid dashboard widgets collection of tasks #718 .

This PR includes a Elements sidechain parsing script that automatically runs and updates the local database, currently with the L-BTC peg in/out.

Note the new SQL tables (`mariadb-structure.sql`) needed for this to work, when Liquid is enabled on the backend.

The first run will take many hours, and afterwards it will just parse the latest blocks coming in.

This PR only covers the backend part and only provides the data needed for the dashboard widget.

This PR also renames the `CORE_RPC_MINFEE` config to `SECOND_CORE_RPC` because the second node might not just be used for getting minfee. In Liquid mode, the main Core node will be Elements but the Second node will be Bitcoin Core mainnet. I removed minfee node enabled and replaced with a new config: `MEMPOOL: USE_SECOND_NODE_FOR_MINFEE` boolean.


wiz:
> In Liquid mode, the main Core node will be Elements but the Second node will be Bitcoin Core mainnet.

I guess you figure it's fine to obtain the Liquid network's `getmempoolinfo` RPC from the primary core daemon since the Liquid mempool will never be >300MB

softsimon:
This is the only log message the new Elements parser script is outputting.

Screen Shot 2021-09-20 at 02 31 58

wiz (approved):
Tested ACK
```
lmempool@localhost [lmempool]> select sum(amount) from elements_pegs;
+--------------+
| sum(amount)  |
+--------------+
| 321505778047 |
+--------------+
1 row in set (0.002 sec)

lmempool@localhost [lmempool]> select count(*) from elements_pegs;
+----------+
| count(*) |
+----------+
|     4129 |
+----------+
1 row in set (0.002 sec)
```

PR #800, opened 2021-09-19
Always use local hostname for API examples.

softsimon:
This PR does:

* location.hostname is now always used for API examples to click to view api works even in local environments
* Corrected URL to the `/api/tx` endpoint
* Corrected URL to the `/api/v1/cpfp` endpoint


wiz (approved):
utACK

PR #806, opened 2021-09-22
Upgrading to Angular 12 and NgBootstrap 10

softsimon:
Let's test so nothing got broken on the site due to this upgrade.
NgBootstrap had to be upgraded to fit with Angular 12 and make sure that nothing broke using these components. (Typeahead, accordion)


frontend/angular.json
```diff
@@ -157,7 +156,13 @@
             ],
             "scripts": [
               "generated-config.js"
-            ]
+            ],
+            "vendorChunk": true,
+            "extractLicenses": false,
+            "buildOptimizer": false,
+            "sourceMap": true,
+            "optimization": false,
```
wiz, line 164: why disable optimization?

frontend/angular.json
```diff
@@ -145,7 +145,6 @@
             "main": "src/main.ts",
             "polyfills": "src/polyfills.ts",
             "tsConfig": "tsconfig.app.json",
-            "aot": true,
```
wiz, line 148: why disable AOT compilation?

frontend/angular.json
```diff
@@ -270,7 +276,9 @@
           "options": {
             "outputPath": "dist/mempool/server",
             "main": "server.ts",
-            "tsConfig": "tsconfig.server.json"
+            "tsConfig": "tsconfig.server.json",
+            "sourceMap": true,
+            "optimization": false
```
wiz, line 281: why disable optimization?

wiz (changes requested):
Apparently there was an option in Angular 11.1 that is now enabled by default in Angular 12 called "Critical CSS inlining" which causes CSS to be loaded asynchronously, so the initial page renders faster now, but looks really weird until a split second later when the CSS loads. To resolve this we need to configure the "Critical CSS inlining" using a webpack plugin. More information on these blog articles:

https://blog.ninja-squad.com/2021/05/12/angular-cli-12.0/
https://blog.ninja-squad.com/2021/01/21/angular-cli-11.1/

wiz (approved):
Tested ACK @ bbee2dcb5bfb92e6764f8a332b718750255c7d90

Disabled the `inlineCritical` for now, will implement support for this in a separate PR later

PR #807, opened 2021-09-22
Enable inline critical CSS for proper Angular 12 optimization

wiz:


softsimon (changes requested):
Since `@import "~bootstrap/scss/bootstrap";` is still present in styles.scss .. Bootstrap is now included twice increasing the total bundle size.

softsimon:
Closing until there is a proper solution.

Issue #543, opened 2021-05-29
Preimage broadcast API


softsimon:
- check every OP_RETURN output for 32 byte long sequences (each output may have up to 2 of them due to size limitations)

- hash each found sequence (it is assumed to be a preimage)

- have a key-value db of SHA256_hash(chunk32) -> chunk32

- have an API to return chunk32 by hash(chunk32), if exists

btcontract:
A relevant test which shows the flow: https://github.com/btcontract/plugin-hosted-channels/blob/master/src/test/scala/fr/acinq/hc/app/PreimageBroadcastCatcherSpec.scala#L37-L47.

PR #808, opened 2021-09-23
preimage api first commit

pseudozach:
This PR resolves https://github.com/mempool/mempool/issues/543 to add a new preimage API.


wiz:
what's a preimage API and why would we want this?

btcontract:
@wiz it's a technique to prove that LN payment routed through [hosted channel](https://lightning-wallet.com/posts/scaling-ln-with-hosted-channels/) has been completed in time.

Having an explorer supporting this would be a nice redundant way to obtain such preimages and make running hosted channels safer. 

This had been discussed with @softsimon.

PR #813, opened 2021-09-25
Fix for fee rounding not using locale

### LINKED ISSUE #796: Significant digits doesn't take localization into account
Since FeeRoundingPipe now uses `toFixed()` instead of the ` | number` pipe, fee rounding doesn't take localization into account.

Minimum fee is `1.00 sat/vB` when it should be displayed with a comma in Europe for example `1,00 sat/vB`

softsimon:
fixes #796

All display of fee and the fee rounding should now take the locale into account.

Locale is determined by the language selected on the site, to test, change language on the site from default english `/` to for example Swedish `/sv`

The minimum fee on the dashboard for example should show as `1.00` on default english-us but `1,00` on the Swedish site.


wiz (approved):
Tested ACK @ 8cb1c5c88cb25350e6dfde59c57d76d4478fecd7

PR #817, opened 2021-09-25
UX/UI: Add filter and invert features to mempool fee chart.

MiguelMedeiros:
## Tasks
- [x] Add invert button.
- [x] Add filtering legends.


wiz:
Can you please split these features into separate PRs? The invert button is the priority, but since the mempool is empty now, we can take our time later designing the filtering fee tiers feature.

PR #818, opened 2021-09-25
UI/UX: Add inverted feature to mempool fee chart.

MiguelMedeiros:
### Preview
interted-chart


wiz:
Please move all changes unrelated to the invert button feature, such as changing the fee rate tiers, the colors in the mouseover legend box, etc. into their own separate PRs.

PR #821, opened 2021-09-26
Adding POST /tx API to bitcoind mode

### LINKED ISSUE #777: POST /api/tx doesn't work
### Description

`POST /api/tx` returns `Cannot POST /api/v1/tx` as response with Bitcoin core (from polar) + electrum as backend

#### Version

v2.2.2 [251a1af4]

### Steps to reproduce

Use the `POST /api/tx` api

### Expected behaviour

Posting the transaction

### Actual behaviour

`Cannot POST /api/v1/tx`

softsimon:
fixes #777

In the docs, this API URL is `POST /api/tx` but since this is a NodeJS backend API the endpoint is exposed from `/api/v1/tx/` . Is this a problem? Will this be redirected using Nginx on peoples installations to just /api/tx or do we need to expose this directly under /api/tx? @wiz


wiz:
If no electrum backend then nginx will route `/api/tx` to `/api/v1/tx` so should work fine

wiz (approved):
utACK

PR #822, opened 2021-09-26
Fix tooltip mempool chart hover selection.

MiguelMedeiros:
### Preview
inverted-chart-selection-fix


MiguelMedeiros:
The size column is not reversed properly when you invert the chart.
Hold on, I will open a new PR.

PR #823, opened 2021-09-26
Fix tooltip mempool chart hover selection.

MiguelMedeiros:
### Preview
inverted-chart-selection-fix-2


wiz (approved):
Tested ACK @ 461296e0
Seems better but still a bit buggy when moving mouse between sections - please polish more in a separate PR later :)

PR #825, opened 2021-09-26
Fix the focus effect on the mempool graph.

### LINKED ISSUE #786: Moving mouse over some area of the chart cause glitch looking blinking.
It's hard to show with anything else than a video but it's like there's invisible lines and area in between colors that very quickly switch from highlighting a single color to showing everything.

image

MiguelMedeiros:
Fix #786.

### Before
blink-before

### After
blink-after


wiz (approved):
Tested ACK @ 9e0f9840

PR #826, opened 2021-09-26
Fix for stuck Bisq transaction page when filtering

### LINKED ISSUE #540: [Bisq] Transactions screen can hang in a loading state forever
### Description

When filtering Bisq transactions, if we 

#### Version

v2.2.0-dev [edad15da]

### Steps to reproduce

1) Go to the Bisq Transactions screen
2) Click on the `Filter` button
3) Click on `Unselect all`
4) Click on any transaction
5) Hit the back button on the browser

### Expected behaviour

The transactions list is shown on an empty state with a message to change the filter criteria.

### Actual behaviour

The transactions list is stuck in the loading state until we change the filters again.

### Screenshots

6F0E5269-528F-41D0-BAC7-F9B6749C7F59

#### Additional info

`Unselect all` should remove the `types` parameter instead of setting it to empty (`https://mempool.space/bisq/transactions?types=&page=1`)

softsimon:
fixes #540

Since I did a refactor here, make sure to test every possible way the Filter feature and the browser history behavior is working when browsing the Bisq transactions page.


wiz (approved):
Tested ACK @ f0462114

PR #830, opened 2021-09-28
Change limit fee values.

MiguelMedeiros:
In the last meeting, @knorrium suggested that we decrease the range size to a smaller number, in order to decrease the amount of null data.

If you approve this PR, I'll create another one by increasing the font size and making better use of the space left over.

## Preview
Graph and TV page:
image

Dashboard page: 
image

### DIFF
--- a/frontend/src/app/components/statistics/statistics.component.html
+++ b/frontend/src/app/components/statistics/statistics.component.html
@@ -44,7 +44,7 @@
             <app-mempool-graph
               dir="ltr"
               [template]="'advanced'"
-              [limitFee]="500"
+              [limitFee]="200"
               [height]="500"
               [left]="65"
               [data]="mempoolStats"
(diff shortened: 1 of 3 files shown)


wiz:
This PR doesn't make any sense. Just because the mempool is empty right now doesn't mean we can hide those fee rate tiers. Only a few months ago we saw 350+ sat/vB transactions.

MiguelMedeiros:
> This PR doesn't make any sense. Just because the mempool is empty right now doesn't mean we can hide those fee rate tiers. Only a few months ago we saw 350+ sat/vB transactions.

If it doesn't make any sense, why do we limit the rates to `50` at dashboard page and `500` at graph and TV pages?
Should we remove the limit and let it go up until 2000 vbytes?

wiz:
I saw that you limited the main dashboard to 150 sat/vB (not 50 sat/vB) which is okay for now, but as soon as the mempool gets full we will need to fix that bug - all high fee transaction fee rate tiers should be displayed somehow

PR #831, opened 2021-09-28
Change total sum column to fixed color.

MiguelMedeiros:
## Description
The idea here is to leave the columns with the summed values in the same color.

The intention is to convey the idea that we are adding values up to that range. 
If we leave different colors, it can give the wrong idea that the value is not a sum, but the absolute value of that range.

I took the opportunity to add a background color to highlight the color value.

### Actual behaviour
image

### New behaviour
image


wiz (approved):
Tested ACK @ b2140c2a

Issue #780, opened 2021-09-15
Chart legend sometimes highlights wrong color


wiz:
It seems that depending on where you hover your mouse over the chart, the wrong color inside the legend is sometimes displayed. In this screenshot, red is highlighted on the chart but grey is highlighted in the legend
<img width="2528" alt="Screen Shot 2021-09-15 at 13 34 48" src="https://user-images.githubusercontent.com/232186/133371271-ab5cb4f2-16dc-44dd-9ccf-1daef03d2daa.png">

wiz:
Fixed by #780

wiz:
Fixed by #837

PR #837, opened 2021-09-28
Remove circle symbols when hovering the series.

### LINKED ISSUE #829: Disks at cursor line occlude part of chart under investigation
### Description

A key place of interest in a mempool chart are the times at which a block was found. E.g. this local peak in today's chart: 
image

While hovering over the chart, there are disks drawn over the intersections of the cursor line and the borders of feerate bands:
image

Since the chart has sharp drops wherever blocks are found, and when hovering over such a time of interest to get the specific band depths, the mentioned disks occlude the most interesting part of the graph: 
image

Even when zooming in, the disks make it hard to see how much each band was affected by the newly issued block.
image

#### Version

node206.mempool.space (v2.3.0-dev) [d0a85091] 

### Steps to reproduce

Stare at the chart and move the pointer.

### Expected behaviour

The chart is clearly visible without occlusion especially at the timestamp currently being investigated.

### Actual behaviour

The disks in combination with the low contrast between some of the band colors and the background make it hard to tell whether one is actually hovering exactly over the peak, how sharp the drop was, and generally disrupt the chart.

MiguelMedeiros:
Hey @Xekyo! 

I've managed to find a way to get the highlighted stack when you hover the mempool chart! :)

I've removed the circle buttons and catched the mouse movements checking if they are hovering the stacked area, selecting the fee range in the tooltip. 

I think I finally got a cool user experience! What do you think?

Fix #829 and #780.

## Tasks
- [x] Remove circle symbols when hovering the series.
- [x] Fix selected index when hovering the series.

## Preview
hover-without-circles

### DIFF
--- a/frontend/src/app/components/mempool-graph/mempool-graph.component.ts
+++ b/frontend/src/app/components/mempool-graph/mempool-graph.component.ts
@@ -66,8 +66,10 @@ export class MempoolGraphComponent implements OnInit, OnChanges {
   }
 
   onChartReady(myChart: any) {
-    myChart.on('mouseover', 'series', (serie: any) => {
-      this.hoverIndexSerie = serie.seriesIndex;
+    myChart.getZr().on('mousemove', e => {
+      if (e.target !== undefined) {
+        this.hoverIndexSerie = e.target.parent.parent.__ecComponentInfo.index;
+      }
     });
   }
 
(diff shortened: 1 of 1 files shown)


Xekyo:
Wonderful, looks great, thank you!

PR #838, opened 2021-09-29
Fix parse numbers localized.

### LINKED ISSUE #836: Graph: Parse numbers localized
replace the use of `parseFloat(2)` to use the `formatNumber(item.value, this.locale, '1.2-2')` and it will be localized

MiguelMedeiros:
Fix #836 .


softsimon (approved):

wiz (commented):
NACK - this breaks the CSS values of the total progress sum bar when using Swedish locale
@softsimon did you even test this using different locales before merging?
Screen Shot 2021-09-30 at 2 57 14

PR #840, opened 2021-09-29
Fix confirmations button positioning.

### LINKED ISSUE #815: UI/UX: Confirmations button positioning has changed
Before:
Screen Shot 2021-09-25 at 16 34 10

Currently:
Screen Shot 2021-09-25 at 16 35 22

MiguelMedeiros:
Fix #815 .

**Observation:** Maybe we should check with @TechMiX about the RTL translations.

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -8,9 +8,9 @@
         <span class="d-none d-lg-inline">{{ rbfTransaction.txid }}</span>
       </a>
     </div>
-  
+
     <div>
-      <div class="title float-left">
+      <div class="title">
         <h1 i18n="shared.transaction">Transaction</h1>
       </div>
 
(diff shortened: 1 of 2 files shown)


frontend/src/app/components/transaction/transaction.component.html
```diff
@@ -8,9 +8,9 @@
         <span class="d-none d-lg-inline">{{ rbfTransaction.txid }}</span>
       </a>
     </div>
-  
+
     <div>
-      <div class="title float-left">
```
TechMiX, line 13: We should keep this to maintain the RTL hint

frontend/src/app/components/transaction/transaction.component.html
```diff
@@ -22,7 +22,7 @@ <h1 i18n="shared.transaction">Transaction</h1>
         <app-clipboard [text]="txId"></app-clipboard>
       </div>
 
-      <div class="container-buttons float-right">
+      <div class="container-buttons">
```
TechMiX, line 25: Add text-right to classes for and RTL hinting.

frontend/src/app/components/transaction/transaction.component.html
MiguelMedeiros: I've solve this by adding the classes under the `.rtl-layout`.

frontend/src/app/components/transaction/transaction.component.html
MiguelMedeiros: I've solve this by adding the classes under the `.rtl-layout`.

wiz (approved):
Tested ACK @ d0a85091

PR #843, opened 2021-09-29
Fix total percentage bar value.

MiguelMedeiros:
Fix total percentage bar value.


wiz (approved):
Tested ACK @ 87eeef5d

PR #845, opened 2021-09-30
Add mark line to mempool chart.

### LINKED ISSUE #844: Chart: (Re)add the vertical dashed line at 1 MvB
When users are in inverted graph mode, a white dash line should indicate the 1 MvB level. 

We don't want people to have to draw it them selves like this: https://twitter.com/murchandamus/status/1443239923033726980

MiguelMedeiros:
Fix #844.

### Avanced template
image

### Widget template
image

### DIFF
--- a/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
+++ b/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
@@ -133,7 +133,7 @@ export class IncomingTransactionsGraphComponent implements OnInit, OnChanges {
             symbol: 'none',
             lineStyle: {
               color: '#fff',
-              opacity: 0.75,
+              opacity: 1,
               width: 2,
             },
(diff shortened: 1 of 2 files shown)


wiz (changes requested):
NACK - line should only be visible when chart is inverted

This is good:
Screen Shot 2021-10-01 at 3 52 49

This is bad:
Screen Shot 2021-10-01 at 3 57 01

softsimon:
What wiz said but opposite. Only when the higher fee starts from the bottom the line works and makes sense.

Also the white line has different color/shading from the transaction per second graph which is more visible on the dashboard. It should look the same.

MiguelMedeiros:
### Tasks
- [x] Fix dotted line style at transaction chart.
- [x] Remove dotted line from inverted mempool chart.

### Preview
inverted-line

wiz (approved):
Tested ACK @ 5c9d44e9

PR #846, opened 2021-09-30
Remove vbytesPipe from series data.

MiguelMedeiros:
Optimize mempool charts series data.
We don't need the `vbytesPipe`, since we already formatting the data on the yAxis formatter option.


wiz (approved):
Tested ACK @ 83cce0c3

PR #847, opened 2021-10-01
Add mempool chart filtering.

### LINKED ISSUE #834: Chart filtering

MiguelMedeiros:
Fix #834 .

### Preview
filtering3


wiz (changes requested):
I think this should be moved into a drop-down box, using the "filter" button like how the bisq.markets transactions table can be filtered - we don't want to lose that nice full-width chart that you just made :)

LamarisSmiley (approved):

softsimon (commented):
I suggest:

1. Replace the "Fees Range" button text with a filter icon
Screen Shot 2021-10-06 at 11 39 46

2. Clicking a second time on the dropdown should close it

wiz (changes requested):
Miguel, this PR should not hide fee rate tiers above 200 sats/vB - we've discussed this many times already.
 
Before:
Screen Shot 2021-10-06 at 17 03 19

After:
Screen Shot 2021-10-06 at 17 03 33

frontend/src/app/components/television/television.component.html
```diff
@@ -10,7 +10,8 @@
         [template]="'advanced'"
         [limitFee]="500"
         [height]="600"
-        [left]="60"
+        [left]="75"
```
wiz, line 13: Why does adding a filter feature require modifying the TV view?

MiguelMedeiros:
@wiz @softsimon as you suggested:

### Tasks
- [x] Add toggle button to dropdown menu.
- [x] Revert left margin from tv page.
- [x] Change text dropdown filter to icon.
- [x] Change dropdown inactive item color.
- [x] Revert 500 limit rate.

### Preview
filter-done

PR #848, opened 2021-10-01
Update README preview image

softsimon:


wiz:
Have we finalized the UI for v2.3.0 release? I was waiting to do this until we stopped iterating on the graphs but I guess we might be ready to ship now?

softsimon:
This is the preview for 2.2.1 which is present but not updated in the README

wiz (approved):
utACK

PR #853, opened 2021-10-05
Use MEMPOOL_BLOCKS_AMOUNT config in the frontend

### LINKED ISSUE #852: Frontend config MEMPOOL_BLOCKS_AMOUNT is not being used or not working

softsimon:
fixes #852


wiz (approved):
Tested ACK @ 822c840e
Seems to correct the issue on liquid.place - just concerned about any possible regression on mainnet with a full mempool

PR #855, opened 2021-10-05
Display mempool graph on the Liquid dashboard

softsimon:
Screen Shot 2021-10-05 at 14 59 35


wiz (approved):
Tested ACK @ acf6fd9d
Looks good on liquid.place and mempool.ninja

PR #868, opened 2021-10-08
UI/UX: Fix mempool chart tooltip at dashboard component.

MiguelMedeiros:
Fix mempool chart at dashboard page.

### Before
image

### After
image


softsimon:
tested AGAIN ACK @ 9bf475bc

PR #898, opened 2021-11-01
Localization: Add localize strings at echarts tooltip.

MiguelMedeiros:
Add localize strings at echarts tooltip: `Range`, `Size` and `Sum`.

### Preview
image


softsimon (changes requested):
This adds a duplicate translation string for `Size` ,, so reuse the current string id:

```$localize`:@@7faaaa08f56427999f3be41df1093ce4089bbd75:Size`;```

softsimon (approved):
utACK

Issue #842, opened 2021-09-29
Human-readable x-axis label intervals


Xekyo:
### Description

The labels on the x-axis of the chart have various random time intervals. I suspect that there is a fixed number per space and it just splits the time frame according to the count. This makes it hard to quickly find specific intervals such as "last three hours", since the start of the day etc. Would it be possible that instead of having a fixed number of labels on the x-axis, the count were variable and it preferred specific times, e.g. if looking at a graph of the last day, each hour or 90 minutes, when looking at a shorter time interval chunks of 30, 15, or even 5 minutes, so that there is a reasonable number of labels on the x-axis but they're generally in time intervals that are human-pleasing?

![image](https://user-images.githubusercontent.com/4060799/135324755-701079cc-4556-45e4-8d7f-e584b6ea5709.png)
In this 6-month graph for example, labels are spaced at intervals of 6 days, 7h and 12 minutes. It would have been more human readable to have just slightly fewer labels but spaced at one week.

PR #900, opened 2021-11-01
UI/UX: make human readable x-Axis date labels.

### LINKED ISSUE #899: Human readable x-Axis date labels.
### Description
Make echarts human readable x-Axis date labels.

### Expected Behaviour
2 hours chart: **15 mins interval**
2h-span

24 hours chart: **2 hours interval**
24h-span

1 week chart: **1 day interval**
1w-span

1 month chart: **1 week interval**
1m-span

3 months chart: **1 week interval**
3m-span

6 months chart: **1 month interval**
6m-span

1 year chart: **1 month interval**
1y-span

MiguelMedeiros:
Fix #899.

### DIFF
--- a/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
+++ b/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
@@ -108,21 +108,21 @@ export class IncomingTransactionsGraphComponent implements OnInit, OnChanges {
           formatter: (value: string) => {
             const date = new Date(value);
             if (this.template !== 'advanced') {
-              return `${date.toLocaleTimeString(this.locale, { hour: 'numeric', minute: 'numeric' })}`;
+              return date.toLocaleTimeString(this.locale, { hour: 'numeric', minute: 'numeric' });
             }
             switch (this.windowPreference) {
               case '1w':
-                return `${date.toLocaleDateString(this.locale, { month: 'short', weekday: 'short', day: 'numeric' })}`;
+                return date.toLocaleDateString(this.locale, { month: 'short', weekday: 'short', day: 'numeric' });
               case '1m':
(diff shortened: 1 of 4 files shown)


wiz (changes requested):
Tested f5bd9cc9 and using the new angular localized timestamps look good - however, I found the following issues:

* This PR should fix issue #842, but for some reason you created a new issue #899 with the exact same title which is very strange and confusing. To be clear, this PR should fix #842 which for the avoidance of doubt includes aligning the labels with the axis on midnight which requires forking the chart library as we discussed.

* The new `getAxisLabelInterval()` method added in this PR doesn't seem to take the width of the window into account at all, and just returns a static number. This won't work for smaller window sizes since the labels will get squished together, and it doesn't look good on large window sizes since the labels will become too far apart. I think the default implementation of this method takes the width into account and determines what can reasonably fit - not sure why it's necessary to override it, but if it is necessary to re-implement this method then the same size width evaluation needs to be added.

* This PR seems to change the font style and layout of the mouseover tooltip legend, which if intentional should probably be in a separate PR.... unless I am missing something, I don't see how the x-axis labels are related to the mouseover legend.

MiguelMedeiros:
Hey @wiz, thanks for the feedback!

I've managed to make it work with our forked npm packages :partying_face: : 
- **mempool/echarts** 
- **mempool/npx-echarts**.

I got stuck a little on the mobile screen, but now I think I've got a way to deal with the xAxis labels when you resize the screen, so they don´t get clustered.

I need a little more time, though! I will send this PR tomorrow morning.

wiz:
Okay cool, can't wait to see it - btw on the 1Y chart I see this issue, looks like it's not getting aligned properly on months with 31 days?
Screen Shot 2021-11-20 at 7 16 36

MiguelMedeiros:
I will close this one.

Opened a new PR: https://github.com/mempool/mempool/pull/946

PR #910, opened 2021-11-09
Display lower <1 s/vB fee rate tiers for Liquid

### LINKED ISSUE #859: Display lower <1 s/vB fee rate tiers for Liquid Network
Since transactions fees are much lower on Liquid than Mainnet, it makes sense to use different fee rate tiers on that mempool graph. However, we would need to make it clear that Liquid uses different fee rate tiers using a different color scheme, such as using shades of the liquid blue color.

We should also probably experiment to see which tiers make most sense using the actual data, but as an example here are some tiers which might make sense for Liquid:

10+
8 - 10
6 - 8
5 - 6
4 - 5
3 - 4
2 - 3
1 - 2
0.8 ~ 1.0
0.6 ~ 0.8
0.4 ~ 0.5
0.3 ~ 0.4
0.2 ~ 0.3
0.1 ~ 0.2
0.0 ~ 0.1

softsimon:
fixes #859

This PR does:

* Liquid fee ranges 0.1, 0.2, 0.3 sat/vB etc are now stored in the 1, 2, 3 respective database fields
* The frontend adapt and display 1/10 the size of fee levels 
 
Screen Shot 2021-11-09 at 23 55 33
e

### DIFF
--- a/backend/src/api/statistics.ts
+++ b/backend/src/api/statistics.ts
@@ -3,6 +3,7 @@ import { DB } from '../database';
 import logger from '../logger';
 
 import { Statistic, TransactionExtended, OptimizedStatistic } from '../mempool.interfaces';
+import config from '../config';
 
 class Statistics {
   protected intervalTimer: NodeJS.Timer | undefined;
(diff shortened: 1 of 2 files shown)


knorrium:
@softsimon did you test this on the main dashboard as well to make sure `limitFeesTemplate` is correct based on the view?https://github.com/mempool/mempool/blob/26a540a57c899ce0634d7c9816557d9d644d2c86/frontend/src/app/components/mempool-graph/mempool-graph.component.ts#L88-L91

softsimon:
> @softsimon did you test this on the main dashboard as well to make sure `limitFeesTemplate` is correct based on the view?
> 
> https://github.com/mempool/mempool/blob/26a540a57c899ce0634d7c9816557d9d644d2c86/frontend/src/app/components/mempool-graph/mempool-graph.component.ts#L88-L91

Yes, the main dashboard is affected as well.

MiguelMedeiros:
Tested ACK.

Suggestion: Maybe you could make the decimal cases of the labels fixed to 1 for all labels, example: 
`0.8 - 1.0`,  `1.0 - 1.2`, etc

---

BASE_MODULE='liquid'.
### Preview
image

---

@knorrium I've also tested with the `BASE_MODULE='mempool'`.

### Preview
image

MiguelMedeiros (approved):

wiz (approved):
Tested ACK @ 26a540a5
Seems okay, but often the fee rate tier has < 0.01 MvB of transactions and gets rounded off to 0 which we probably want to fix in the future...

PR #912, opened 2021-11-10
Add special blocks animation: fireworks.

### LINKED ISSUE #573: Make taproot activation block glow
On block 709632, taproot will be activated. This will be an event similar to the halving or segwit activation and should be visible on Mempool with a green glowing/blinking mempool block and mined block, and activation should have fireworks spewing for 8 blocks until it's gone out of sight.

We did just like this for this halving so old code should exist for this.

This code should remain so it can be easily configured and adapted for new events. Including all the coming halving events.

MiguelMedeiros:
Proposal to fix #573 .

### Preview
fireworks-taproot


knorrium:
IMO this looks better than confetti in #913

frontend/src/app/app.constants.ts
```diff
@@ -129,3 +129,13 @@ export const languages: Language[] = [
    { code: 'vi', name: 'Tiếng Việt' },      // Vietnamese
    { code: 'zh', name: '中文' },            // Chinese
 ];
+
+
+export const specialBlocks = {
+  '708998': {
```
softsimon, line 135: If I'm not mistaken.. 709632 is the actual block height of taproot activation.

softsimon (changes requested):
What I'd like to see is the upcoming taproot activation block or halving block arrive through the mempool. This feature was in my old halving code. The block could be bouncing in the mempool just like when it's a mined block.

frontend/src/app/app.constants.ts
MiguelMedeiros: Ops, force pushed with my test block! Will change to 709632!

softsimon (approved):

softsimon:
Tested ACK at 390e704

PR #913, opened 2021-11-10
Add special blocks animation: confetti.

### LINKED ISSUE #573: Make taproot activation block glow

MiguelMedeiros:
Proposal to fix #573 .

### Preview
confetti-taproot


softsimon:
We will go with fireworks to save on dependencies :)

PR #922, opened 2021-11-12
Update mainnet partial address tests

knorrium:
- Split case-insensitive autocomplete tests into bech32 and bech32m addresses
- Fix the bech32m address used


softsimon (approved):
uTACK .. tests passes so.

PR #939, opened 2021-11-17
Fix: mempool empty block position.

### LINKED ISSUE #938: [Liquid] Blocks not displayed properly on the dashboard
Screen Shot 2021-11-16 at 9 35 45 PM

Leaving a tab open for too long and coming back to it later can cause rendering issues like more than a dozen of mined blocks being displayed and the mempool block moving to the wrong location.

MiguelMedeiros:
Fix #938.

---
### Preview
empty-block-liquid


softsimon:
There are 2 things I think can be improved here so I made a separate PR.

1. There are some duplicate lines of code now
2. A empty mempool block doesn't show the golden countdown block
3. 
https://github.com/mempool/mempool/pull/940

PR #940, opened 2021-11-17
Fix for empty mempool block position

### LINKED ISSUE #938: [Liquid] Blocks not displayed properly on the dashboard

softsimon:
fixes #938


wiz (approved):
Tested ACK @ 15a0644b
Seems ok on mainnet, testnet, signet, and liquid

PR #943, opened 2021-11-18
Delete nginx logs after 10 days and explain this in our Privacy Policy

wiz:
Screen Shot 2021-11-18 at 17 09 40


softsimon (approved):
utACK . .. text looks ok

PR #946, opened 2021-11-20
Fix human readable dates with mempool npm packages

### LINKED ISSUE #842: Human-readable x-axis label intervals

MiguelMedeiros:
Fix #842.

NPM dependencies packages PRs:
https://github.com/mempool/ngx-echarts/pull/1
https://github.com/mempool/echarts/pull/1


frontend/src/app/components/mempool-graph/mempool-graph.component.ts
```diff
@@ -100,12 +99,136 @@ export class MempoolGraphComponent implements OnInit, OnChanges {
...
+            if (this.windowPreference === '1w') {
+              if (labelDate.getMinutes() % 60 === 0 &&
+                  labelDate.getHours() % 6 === 0) {
+                labelDates.push(labelDate.toISOString());
+              }
+            }
+
+            if (this.windowPreference === '1m') {
+              if (labelDate.getMinutes() % 60 === 0 &&
+                  labelDate.getHours() % 12 === 0) {
```
wiz, line 139: It looks like you should be targeting midnight instead of noon here
Screen Shot 2021-11-21 at 1 29 24

frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
```diff
@@ -99,12 +242,48 @@ export class IncomingTransactionsGraphComponent implements OnInit, OnChanges {
       },
       xAxis: {
         type: 'category',
+        axisTick: {
+          alignWithLabel: true,
+          lineStyle: {
+            width: 0,
```
wiz, line 248: Why are you setting the tick line width to 0, are you trying to hide it for some reason?

frontend/src/app/components/mempool-graph/mempool-graph.component.ts
```diff
@@ -304,12 +446,47 @@ export class MempoolGraphComponent implements OnInit, OnChanges {
...
+                  return date.toLocaleTimeString(this.locale, { hour: 'numeric', minute: 'numeric' });
+                case '24h':
+                  return date.toLocaleTimeString(this.locale, { hour: 'numeric' });
+                case '1w':
+                  return date.toLocaleDateString(this.locale, { month: 'short', day: 'numeric', weekday: 'short' });
+                case '1m':
+                case '3m':
+                case '6m':
+                case '1y':
+                  return date.toLocaleDateString(this.locale, { month: 'short', day: 'numeric' });
```
wiz, line 482: It feels like the syntaxes you were using in PR 900 looked nicer for the 6M chart, could you explain why you changed it to these instead? I'm not sure we want to use the same syntax for these 4 timespans

frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
MiguelMedeiros: I put it again.

wiz (changes requested):
quick test of x-axis labels using PC set to GMT +9 offset:
2h -> labels missing?
24h -> OK, labels are on midnight
1w -> labels are on 5am instead of midnight
1m, 3m -> labels are on 6am instead of midnight
6m, 1y, 2y, 3y -> labels are on 9am instead of midnight

wiz:
A few more things:
* To reproduce the 2h issue I reported in previous comment, open https://mempool.ninja/graphs in incognito and the labels are missing
* The layout of the x-axis labels in the homepage graph seems wrong (see screenshot below)
* The TV layout mouseover legend is significantly changed
Screen Shot 2021-11-23 at 6 03 03

wiz:
I just realized the way you are doing things in this PR causes the timestamps on the chart to become incorrect..

https://user-images.githubusercontent.com/232186/143152452-00bd6ab7-baf4-4e28-af22-8a5ebf7434c2.mov

We will have to consider another fix for this issue later on.

Issue #965, opened 2021-12-02
Create documentation section


hunicus:
Lately I've been brainstorming approaches for docs for Mempool. I'm making this issue to explain what I have in mind, start a conversation, and collect feedback.

## Docs Contents

What should Mempool docs actually cover? After going through Telegram chat history, browsing GitHub, etc...I suggest the following outline for docs hosted at /docs for master-page only (mainnet, testnet, and signet networks only).

This outline doesn't mention API docs since they already exist in their own section on the website at /api.

```
— Introduction
    — What Mempool is 
    — What Mempool is not
— Frequently Asked Questions
— Understanding Mempool (combination of Mempool-specific concerns + general Bitcoin workings)
    — Explain graphs (how to read, what they mean)
    — Clarify memory usage, purging
    — Explain meanings of various sizes (memory usage v aggregate on-disk v aggregate block-weight)
    — Explain difficulty adjustment, how prediction is calculated, hash rate, fluctuations, effects on fees, etc.
    — Explain what mempool actually is, why your mempool is different from my mempool, etc.
    — Explain fee estimates, how they're calculated, what they mean
    — Random stuff
        — SegWit, s/b vs s/vb, why block sizes are what they are right now
        — Why there are sometimes empty blocks, or blocks with very few transactions (in past and/or future)
        — Might be cool to have links/tags on certain momentous blocks (halvenings, taproot activation, etc) and link to docs with explanation
— Using Mempool
    — Making transactions that don't get stuck
    — Troubleshooting transactions
        — Stuck transactions, fee-bumping
        — Missing transactions
    — Hosting your own instance of Mempool for ultimate privacy and control
        — Umbrel
            — Notes + Troubleshooting
        — RaspiBlitz
            — Notes + Troubleshooting
        — myNode
            — Notes + Troubleshooting
— Contributing to Mempool
    — Reporting bugs and requesting new features
    — Development (setting up dev environment, etc)
    — Translating (transifex)  
    — Sponsorships
    — Other (?)
```

## Design

The Mempool website has a very clean and simple aesthetic to it, despite the enormous amount of data it offers. It's elegant. I think docs should maintain that elegance and fit in the existing site seamlessly.

[...]

PR #966, opened 2021-12-02
Create docs page with faq and api docs

hunicus:
This pull request is the (small) start of a docs section for the Mempool site. You can see the rest of my proposal [here](https://github.com/mempool/mempool/issues/965).

Summary of changes:
- add new page at /docs for bitcoin network pages only (mainnet, testnet, and signet)
- add 'FAQ' tab (future tabs are outlined [here](https://github.com/mempool/mempool/issues/965))
- add questions and answers covering most common questions from users (gathered from surveying Telegram, GitHub, and Keybase activity from over the past few months)

Handling i18n is left for later once the direction of this /docs page is better known.

This PR was opened in draft status because it is probably not ready to merge:

#### ~1. Not tested with other networks~ (initial roll-out to mempool.ninja looked fine; more tests will be done before merging final changes)

My local mempool instance is only configured to be in mainnet mode. Obviously this doesn't matter for the FAQ content itself since it's just static text, but I can't be 100% sure if/how the `/docs` menu item will show when in testnet or signet mode.

#### ~2. Unsure about colors/design~ (solved with move to accordion)

What do you think about the colors/design? I took colors from the existing color palette, but I'm not sure they go well together.

#### ~3. FAQ link list~ ([this will be implemented](https://github.com/mempool/mempool/pull/966#issuecomment-986260855))

Should the FAQ list have a list of question at the top that allow the user to jump to the answer below? Like wikis often do?

#### 4. Curious about others' thoughts on how this FAQ page fits in with other plans

[See this issue](https://github.com/mempool/mempool/issues/965). If there are serious objections, then it might become necessary to take these changes back to the drawing board.

I'd appreciate any advice/guidance to help address the issues mentioned above.

### DIFF
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -81,6 +81,10 @@ let routes: Routes = [
       },
       {
         path: 'docs',
+        redirectTo: 'docs/faq'
+      },
+      {
+        path: 'docs/:tab',
         component: DocsComponent
       },
       {
@@ -233,6 +237,10 @@ let routes: Routes = [
       },
       {
         path: 'docs',
+        redirectTo: 'docs/faq'
+      },
+      {
+        path: 'docs/:tab',
         component: DocsComponent
       },
       {
@@ -293,6 +301,10 @@ let routes: Routes = [
           },
           {
             path: 'docs',
+            redirectTo: 'docs/faq'
+          },
+          {
+            path: 'docs/:tab',
             component: DocsComponent
           },
         ],
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -69,7 +69,7 @@ let routes: Routes = [
       },
       {
         path: 'api',
-        component: ApiDocsComponent,
+        redirectTo: 'docs/api'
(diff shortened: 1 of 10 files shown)


wiz:
Wow, thanks for the excellent first PR. The FAQ page looks great, please DM me on Keybase as I would like to go over the text with you. Also keep in mind since this code is served by both the official mempool.space instance and unofficial instances like mempool.bitcoin.ninja we need to wrap things like hard-coded Tor onion hostname in an "if mempool.space official" code block so it only gets displayed on the official site

hunicus:
Working on updates per chat with wiz. 

These include:
- moving faq to accordion so that items are collapsed by default
- make faq items linkable so direct answers can be sent to users
- move api documentation at /api to /docs

hunicus:
At this point, all content is in this PR.

Remaining items:
- fragment linking for faq and api items
- api section layout (a side-nav setup was discussed, but accordion style [which is currently implemented] might be better?)
- should faq be included in docs section for liquid and bisq sites?

hunicus:
Closing this PR since the scope got too big. Opening a new one that just moves API docs from /api to /docs/api.

PR #978, opened 2021-12-07
Update Docker scripts to Node 16.13.1

knorrium:


wiz (approved):
ACK

wiz:
Actually upon further discussion we decided to use NodeJS 16.10.0 which has NPM 7.x

PR #980, opened 2021-12-07
Set Dockerfiles to use Node v16.10.0

knorrium:
https://hub.docker.com/layers/node/library/node/16.10.0-buster-slim/images/sha256-9ecdeb8c39d83fadf1c5d01ebb625ac0a65ec3faaf6c5b997b36d72ad2c56e66?context=explore


wiz (approved):
ACK

PR #981, opened 2021-12-07
Move npm deps in package.json to fix `npm install --prod`

wiz:
To minimize the amount of npm deps required for production use, I'd like to fix `npm install --prod` which looks like these few deps just need to moved from dev deps to normal deps, but requires testing


knorrium (approved):

softsimon:
utACK

wiz:
@softsimon you have to actually approve it :P

PR #982, opened 2021-12-07
Adding Core Contributors on the About page

softsimon:
The new updated Service backend is required for the new Core Contributors separation to be visible.
It should work seamlessly even with old service backend, and old mempool about pages should handle the new service backend as well.

Screen Shot 2021-12-07 at 18 00 56


wiz (approved):
Tested ACK @ 1594fed8

PR #983, opened 2021-12-07
Fix string for Project Staff on About page

wiz:


frontend/src/app/components/about/about.component.html
```diff
@@ -174,7 +174,7 @@ <h3 i18n="about.contributors">Project Contributors</h3>
     </div>
 
     <div class="maintainers" *ngIf="contributors.core.length">
-      <h3 i18n="about.core_contributors">Core Contributors</h3>
+      <h3 i18n="about.core_contributors">Project Staff</h3>
```
softsimon, line 177: Could you also change the i18n identifier so it matches with project staff,, instead of core_contributors

wiz:
```suggestion
      <h3 i18n="about.project_staff">Project Staff</h3>
```

softsimon (approved):
utACK

PR #986, opened 2021-12-07
Fixing broken Liquid peg-in and peg-out links

softsimon:
fixes #977

* Fixes broken Peg-Out link. Example TXID: `ecf6eba04ffb3946faa172343c87162df76f1a57b07b0d6dc6ad956b13376dc8` 
* Fixes broken Peg-In link  Example TXID `fe764f7bedfc2a37b29d9c8aef67d64a57d253a6b11c5a55555cfd5826483a58`

PR #987, opened 2021-12-08
Add tests for Liquid peg in and peg out addresses

knorrium:
These should fail when `BASE_MODULE` is set to `liquid` until #986 is merged


wiz:
@knorrium ok it was merged so please rebase onto current master

PR #988, opened 2021-12-08
Pin node version on e2e GHA

knorrium:
Use the official GHA to setup node and cache dependencies on the Cypress GHA


softsimon (approved):
utACK

PR #993, opened 2021-12-08
Upgrading frontend to Angular 13

softsimon:


wiz (approved):
Tested ACK @ ed9b46ba
Everything builds/works fine using NodeJS v16.10.0

Issue #992, opened 2021-12-08
Time gets cut off at the bottom of some charts


darthginger:
### Description
Time gets cut off at the bottom on these charts:

1. Dashboard mempool
2. Dashboard incoming transactions
3. TV view

No cut off on:

1. Graphs page

#### Version

7c89dde5
### Steps to reproduce
Always happening.

### Expected behaviour
No cutoff of text to the bottom side.

### Screenshots
![image](https://user-images.githubusercontent.com/8257956/145238062-5b5148e1-2fd1-413d-98be-5ae9c19c71af.png)

#### Device or machine
Desktop 27" monitor 1440p, Windows 10, Brave 1.32.113, same issue on Edge and Firefox

#### Additional info
Not happening on my phone

softsimon:
Closed in https://github.com/mempool/mempool/pull/995

PR #995, opened 2021-12-08
Fix graph issue in #992

hunicus:
Seems to be only happening on Windows. Proposed changes fixed the issue on Edge, Chrome, and Firefox for Windows while not causing any issues on Firefox and Chromium on Linux.

Reviewers should check graphs at `/` and `/tv`.


softsimon (approved):
utACK

PR #998, opened 2021-12-09
Pulling from Transifex

softsimon:


wiz (approved):
Tested ACK @ a5c7f152

PR #1000, opened 2021-12-09
Add Liquid peg in and peg out tests

knorrium:
- Disable Chrome security to allow cross site navigation (localhost -> mempool.space)
- Add tests for the peg in and peg out addresses


wiz:
🏆 for 1000th issue

softsimon (approved):
utACK ... tests doesn't fail so

PR #1003, opened 2021-12-11
Automated database creation and migration

### LINKED ISSUE #1002: Backend database migration feature
Requirements: 

* Stop erasing MySQL data on upgrading/re-installation
* Store database structure version in the database
* Detect version and run required migration scripts
* Add `statistics` migration script

softsimon:
fixes #1002

Removed the database structure SQL-file and created a fully automated database creation and migration service. A new table called "state" will keep various information like the current mempool table schema version and the current Liquid pegs sync header.

When the backend starts up it will:
1. Assume that a database exist with the name configured
2. Check if the base table exists (`statistics`), if not, it will create all base tables (just that one actually)
3. It will check if the new "state" table exist and check for the current schema version. It will run available migrations up to the latest version.

The base version is 0 and the 2.3 release will bump the database schema version to 1... so when the backend detects that a '`statistics`' table already exists, but not the new '`state`', it will assume version 0 and run the migration to 1 which includes creating the new tables (`state, elements_pegs`) and migrating the statistics data (if not on Liquid backend).

This pave way for very easy future database schema updates just by adding more queries to the migration service and bump the schema version. Example:

Bump the version:
```
private static currentVersion = 2;
```

Add to method `getMigrationQueriesFromVersion`:
```
if (version < 2) {
   queries.push(`CREATE TABLE version_2_table (id int(11) NOT NULL)`);
}
```


wiz:
This PR isn't ready for testing yet since this is only the backend component, of course this feature requires updates to backend Dockerfile and the documentation etc. so I will make those changes and push to this PR branch

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+import { DB } from '../database';
+import logger from '../logger';
+
+class DatabaseMigration {
+  private static currentVersion = 1;
+  private queryTimeout = 120000;
+
+  constructor() { }
+
+  public async $initializeOrMigrateDatabase(): Promise<void> {
```
knorrium, line 11: @softsimon I don't like how this method has multiple responsibilities. IMO it should be split into `initializeDatabase` and `migrateDabase` so we know exactly what's being done.

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (!await this.$checkIfTableExists('statistics')) {
+      const connection = await DB.pool.getConnection();
+      for (const query of this.getInitializeTableQueries()) {
+        await connection.query<any>({ sql: query, timeout: this.queryTimeout });
```
knorrium, line 15: @softsimon shouldn't this be in a try-catch? what happens if one of the queries fail? we'll probably want to rollback and/or exit early.

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+      }
+      connection.release();
+      logger.info(`Initial database tables has been created`);
```
knorrium, line 18:
```suggestion
      logger.info(`Initial database tables have been created`);
```

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    if (await this.$checkIfTableExists('state')) {
+      const databaseSchemaVersion = await this.$getSchemaVersionFromDatabase();
+      if (DatabaseMigration.currentVersion > databaseSchemaVersion) {
+        const connection = await DB.pool.getConnection();
+        for (const query of this.getMigrationQueriesFromVersion(databaseSchemaVersion)) {
+          await connection.query<any>({ sql: query, timeout: this.queryTimeout });
```
knorrium, line 26: @softsimon same as the previous comment on try-catch

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+        }
+        connection.release();
+        this.$updateToLatestSchemaVersion();
+        logger.info(`Database schema has been migrated to the latest version (${DatabaseMigration.currentVersion})`);
```
knorrium, line 30:
```suggestion
        logger.info(`Database schema has been migrated from ${databaseSchemaVersion} to the latest version: (${DatabaseMigration.currentVersion})`);
```

knorrium: @softsimon I think it's important to log what version we are migrating from as well

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+      }
+    } else {
+      const connection = await DB.pool.getConnection();
+      for (const query of this.getMigrationQueriesFromVersion(0)) {
+        await connection.query<any>({ sql: query, timeout: this.queryTimeout });
+      }
+      connection.release();
+      this.$updateToLatestSchemaVersion();
+      logger.info(`Database schema has been migrated to the latest version (${DatabaseMigration.currentVersion})`);
```
knorrium, line 39:
```suggestion
      logger.info(`Database schema has been migrated from ${databaseSchemaVersion} to the latest version: (${DatabaseMigration.currentVersion})`);
```

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+    }
+  }
+
+  private async $getSchemaVersionFromDatabase(): Promise<number> {
```
knorrium, line 43: @softsimon all these methods should have some better error handling

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+    const connection = await DB.pool.getConnection();
+    const query = `SELECT number FROM state WHERE name = 'schema_version';`;
```
knorrium, line 45: @softsimon SQL queries should be imported as constants so it's easier to maintain and track  vs sql changes

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+    const query = `SELECT COUNT(*) FROM information_schema.tables WHERE table_schema = '${config.DATABASE.DATABASE}' AND TABLE_NAME = '${table}'`;
+    const [rows] = await connection.query<any>({ sql: query, timeout: this.queryTimeout });
+    connection.release();
+    return rows[0]['COUNT(*)'] === 1;
+  }
+
+  private getInitializeTableQueries(): string[] {
+    const queries: string[] = [];
+
+    queries.push(`CREATE TABLE IF NOT EXISTS statistics (
```
knorrium, line 69: @softsimon see how clean this would be if the queries are extracted from here

knorrium (changes requested):

backend/src/api/database-migration.ts
```diff
@@ -0,0 +1,176 @@
...
+    queries.push(`ALTER TABLE statistics ADD PRIMARY KEY (id);`);
+    queries.push(`ALTER TABLE statistics MODIFY id int(11) NOT NULL AUTO_INCREMENT;`);
+
+    return queries;
+  }
+
+  private getMigrationQueriesFromVersion(version: number): string[] {
+    const queries: string[] = [];
+
+    if (version < 1) {
```
knorrium, line 127: @softsimon IMO the the version should also be mapped to semver somewhere so we know that 0 = < 2.3.0 and 1 >= 2.3.0

knorrium (changes requested):

wiz:
btw don't merge this, let me do it since I have to manually do SQL commands on the production servers to avoid corrupting the database

softsimon:
So since we already ran the `statistics` data migration SQL on the production servers, we need to do the other new SQL queries manually as well, for the migration not to be run twice and ruin the data.

How to upgrade production:
1. Run the new SQL-queries below in order
2. Upgrade and restart the backend

Liquid backend only:
1. Copy the `block` height number from the now deprecated `last_elements_block` table and add it to the new `state` table either in the INSERT `last_elements_block` value query or update the row, before restarting the backend.

Queries:
```
CREATE TABLE IF NOT EXISTS state (
        name varchar(25) NOT NULL,
        number int(11) NULL,
        string varchar(100) NULL
      ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

```
INSERT INTO state VALUES('schema_version', 0, NULL);
```
```
INSERT INTO state VALUES('last_elements_block', 0, NULL);
```

wiz (approved):
Tested ACK @ 91e30fbc
Tested ACK @ `v2.3.0-dev3` tag (from `wiz/database-migration-feature` branch) using @knorrium/umbrel#mempool_230 fork on Raspberry Pi 
Works well on Umbrel using Docker, and Works well upgrading SQL database on production server, took about 40 seconds:
```
Dec 23 21:46:10 [31725] INFO: Database connection established.
Dec 23 21:46:50 [31725] INFO: Database schema have been migrated from version 0 to 1 (latest version)
```

wiz:
Upgraded prod servers manually using this command:
```
for i in mempool mempool_testnet mempool_signet mempool_liquid;do echo "
CREATE TABLE IF NOT EXISTS state ( name varchar(25) NOT NULL, number int(11) NULL, string varchar(100) NULL) ENGINE=InnoDB DEFAULT CHARSET=utf8;
INSERT INTO state VALUES('schema_version', 1, NULL);
" | mysql -u root -p $i;done
echo "INSERT INTO state VALUES('last_elements_block', (SELECT block FROM last_elements_block), NULL);"|mysql -u root -p mempool_liquid
```

PR #1004, opened 2021-12-11
Improve x-axis labels and graph data ticks

### LINKED ISSUE #969: Improvements to historical mempool chart x-axis labels
## Background

Recently we migrated the project to the Apache echarts graph library, the primary reason being so that we could implement a mouseover legend, which appears when you hover over the historical mempool data, as well as it being needed for new charts and graphs we want to implement in the upcoming mining dashboard.

The migration to echarts is nearly complete, but one major task remaining is polishing the x-axis labels, as originally reported in #842 by Murch who noticed they were incorrectly positioned, i.e. if you hover over "May 30" it would not be exactly at midnight on May 30. Therefore the remaining work is to fix the x-axis labels so they are both localized and accurate.

Previously a WIP of this was attempted in PR #946 but it was not completed. Therefore this issue is to complete the WIP PR, which requires implementing changes in the forked echarts npm modules (https://github.com/mempool/echarts/pull/1 and https://github.com/mempool/ngx-echarts/pull/1) and other changes in mempool as necessary.

### Part 1: Localizing the label date/time strings

This first part is pretty straightforward, as it just uses the Angular code to localize the timestamps. Of course, different date formats are appropriate for different timestamps, such as the following:

2 hour
Screen Shot 2021-12-03 at 9 43 59

24 hour

Screen Shot 2021-12-03 at 9 44 15

1 week
Screen Shot 2021-12-03 at 9 44 29

1 month
Screen Shot 2021-12-03 at 9 44 45

3 month
Screen Shot 2021-12-03 at 9 45 12

6 month
Screen Shot 2021-12-03 at 9 45 24

1 year
Screen Shot 2021-12-03 at 9 45 35

### Part 2: Fix inconsistent tick and label positions

After localizing the timestamps, the following work must be done:

1) The x-axis labels becomes too large to fit and therefore need to be correctly spaced on the size of the user's browser display width:

Screen Shot 2021-11-23 at 6 15 13

2) On chart timespans greater than 1 week, the position of the "ticks" is not exactly on midnight of the date labels, and other places that are expected. This is the main complaint of the issue. To correct this, custom tick position is probably required, which might require modifying the mempool backend to return such customized time points, or further changes inside our fork of the echarts library, or both.

3) Testing. After implementing all of the above, the new version should be consistent with current production mempool.space data and match completely - lots of testing of all functionality, including "zoom in" on the chart etc. in various timeframes will be necessary.

jorisvial:
fixes #969 

Changes list:
* Switch graph x-axis type to `time` to leverage native echart features
* The field ``added`` from `/api/v1/statistics/` API is now returned as timestamp (number). In the future it makes more sense if other people want to use the API directly
* Query historical data using mysql native datetime features
* Deleted unused component code `FeeDistributionGraphComponent`
* Apply proper tooltip title datetime formatting based on the time scale choosen by the user
* In the dashboard, force the time frame to `2h` so that graphs shows proper tooltip

Things to improve:
* Refactor copy/paste for tooltip formatting
* If you are not in UTC, there is an issue with x-axis labels not matching data ticks on time frames of 1 month+, therefore you cannot exactly align the vertical cursor with all labels. This is "passable" UX but would be better if fixed
* In the mempool graph component, there is 3 dimensional array `const finalArray: number[][][]`. It would make sense to reduce this complexity for easier maintenance in the future.


jorisvial:
Here are some number I've ran on my machine regarding the following commit in this PR https://github.com/mempool/mempool/pull/1004/commits/c9f5002dc2b659aaf1157e6026a39594093f3a8b

Measured time in seconds to query the historical mempool data on my laptop, using mysql `profiling=1`
* CPU: Intel(R) Core(TM) i5-7300HQ CPU @ 2.50GHz
* RAM: 16G
* SSD drive

Without avg():
24h: 0.02956841
1w:  0.08247139
1m:  0.24980947
3m:  2.31804638
6m:  2.52517240
1y:  2.96404507
2y:  3.83326954
3y:  4.71079904

With avg():
24h: 0.04001340 -> **applied** avg() in the source code
1w:  0.09009741 -> **applied** avg() in the source code
1m:  0.39601776 -> **applied** avg() in the source code
3m:  2.92609427 -> **applied** avg() in the source code
6m:  3.91138025 -> **applied** avg() in the source code
1y:  5.32411039 -> **not applying** avg() in the source code
2y:  8.05741048 -> **not applying** avg() in the source code
3y:  10.38970137 -> **not applying** avg() in the source code

wiz (changes requested):
Wow, what an amazing first contribution! Looks great, just a few minor change requests:

- When loading for the first time in an incognito, the mouseover legend title is "undefined"
- Mobile devices have x-axis labels squished together on some timespans
- 1W timespan could use the month on the left side of x-axis
- consider doubling the data points for 1W and 3Y to improve resolution
- needs more testing by @knorrium

knorrium:
To echo what @wiz said, this is a great first contribution!

I've noticed a few things that could be improved before we merge (mostly with the responsive design):

1) It's very hard to read the labels on smaller screens, I wonder if we could push down every other label so there's no overlap, see how it looks on an iPhone 11 sized screen:

Screenshot from 2021-12-11 19-10-14

2) On the full graph, when selecting a 24h or bigger time span, the dates are just numbers which is a bit confusing and makes the graph less useful when sharing - think of making the graph self contained so if someone tweets it today, all the information to know when the screenshot was taken is in the graph itself, without needing extra context:

Screenshot from 2021-12-11 19-11-31
Screenshot from 2021-12-11 19-11-49

3) Similar to 2, when we select a 6m+ time span, we only see short month names, it would be good to see the year (i.e, Dec '21, localized of course) so the graph has more context.

Screenshot from 2021-12-11 19-12-30


4) I think the zoom component was not localized and it's hard to read on smaller screens. I'm not sure how feasible this is but it might be better to show the localized date and under the component instead of inside:

Screenshot from 2021-12-11 19-16-54

wiz:
And we don't need it for this PR but in the future please sign your commits :)

jorisvial:
Fixed:
- [X] When loading for the first time in an incognito, the mouseover legend title is "undefined"
- [X] Mobile devices have x-axis labels squished together on some timespans
- [X] 1W timespan could use the month on the left side of x-axis (added in all resolution actually based on @knorrium feedback)
- [X ] consider doubling the data points for 1W and 3Y to improve resolution

wiz (approved):
Tested ACK @ 28d3f190

PR #1005, opened 2021-12-12
fixed connection refused on local network

antonilol:
when hosting mempool on my local network it gave me a 'connection refused' except for `curl localhost` on the host. i search the internet for fixes and this one did the trick for me

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -144,7 +144,7 @@ http {
 	}
 
 	server {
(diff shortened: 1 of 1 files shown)


wiz:
I'd prefer to simply improve the documentation that explains how to open this up if the user wants to - everyone has their own specific use cases and binding to all interfaces by default might not be the best idea for security / privacy reasons.

PR #1010, opened 2021-12-14
Liquid icons api

softsimon:
So .. place the `icons.json` in the same folder as the `mempool-config.json`

and there is 2 new APIs
List all icons
`/api/v1/liquid/icons`
Get a specific image from an asset id
`/api/v1/liquid/icon/6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d`


wiz:
@softsimon could you please add API documentation for the 2 new APIs this PR adds

wiz:
> So .. place the icons.json in the same folder as the mempool-config.json

shouldn't the sync-assets.js script do this automatically?

wiz:
Unfortunately the liquid API docs examples you added that reference https://liquid.network/liquid.js don't work because liquid.js doesn't seem to exist - guess it was never finished. Perhaps we should hide those (in a separate PR)

wiz:
Did you add the API methods to @mempool/liquid.js ? will also need to publish a v2.3.0 npm module for that

softsimon:
* Updated `@mempoool/mempool-j`s with the new API endpoints
* Updated frontend build to use the new build that produces a `liquid.js`, `bisq.js` along with the `mempool.js`
* Backend has a new Sync Assets feature that requires a new configuration. So for the Liquid environment, configure the following external asset for automatic download when backend restarts..
```

{
  "MEMPOOL": {
    "EXTERNAL_ASSETS": [
      "https://raw.githubusercontent.com/Blockstream/asset_registry_db/master/icons.json"
    ]
```

backend/.gitignore
```diff
@@ -1,7 +1,10 @@
 # See http://help.github.com/ignore-files/ for more about ignoring files.
 
-# production config
-mempool-config.json
+# production config and external assets
+*.json
+!mempool-config.sample.json
```
wiz, line 5:
```suggestion
!mempool-config.json
```

frontend/package.json
```diff
@@ -34,7 +34,10 @@
     "sync-assets": "node sync-assets.js && rsync -av ./dist/mempool/browser/en-US/resources ./dist/mempool/browser/resources",
     "sync-assets-dev": "node sync-assets.js dev",
     "generate-config": "node generate-config.js",
-    "build-mempool.js": "tsc | browserify -p tinyify ./node_modules/@mempool/mempool.js/lib/index.js --standalone mempoolJS > ./dist/mempool/browser/en-US/mempool.js",
+    "build-mempool.js": "npm run build-mempool-js && npm run build-mempool-liquid-js && npm run build-mempool-bisq-js",
+    "build-mempool-js": "tsc | browserify -p tinyify ./node_modules/@mempool/mempool.js/lib/index.js --standalone mempoolJS > ./dist/mempool/browser/en-US/mempool.js",
+    "build-mempool-bisq-js": "tsc | browserify -p tinyify ./node_modules/@mempool/mempool.js/lib/index-bisq.js --standalone bisqJS > ./dist/mempool/browser/en-US/bisq.js",
+    "build-mempool-liquid-js": "tsc | browserify -p tinyify ./node_modules/@mempool/mempool.js/lib/index-liquid.js --standalone liquidJS > ./dist/mempool/browser/en-US/liquid.js",
```
wiz, line 40: The `tsc` command doesn't send any output, and the `browserify` command doesn't receive any input, so not sure why the `|` character is used:
```suggestion
    "build-mempool-js": "tsc && browserify -p tinyify ./node_modules/@mempool/mempool.js/lib/index.js --standalone mempoolJS > ./dist/mempool/browser/en-US/mempool.js",
    "build-mempool-bisq-js": "tsc && browserify -p tinyify ./node_modules/@mempool/mempool.js/lib/index-bisq.js --standalone bisqJS > ./dist/mempool/browser/en-US/bisq.js",
    "build-mempool-liquid-js": "tsc && browserify -p tinyify ./node_modules/@mempool/mempool.js/lib/index-liquid.js --standalone liquidJS > ./dist/mempool/browser/en-US/liquid.js",
```

frontend/src/app/components/docs/api-docs.component.ts
```diff
@@ -693,6 +675,60 @@ export class ApiDocsComponent implements OnInit {
...
+        codeTemplate: {
+          curl: `/api/v1/asset/%{1}/icon`,
+          commonJS: `
+        const { %{0}: { assets } } = mempoolJS();
+
+        const asset_id = '%{1}';
+        const assetIcon = await assets.getAssetIcon({ asset_id });
+
+        document.getElementById("result").textContent = JSON.stringify(assetIcon, undefined, 2);
+        `,
```
wiz, line 716: The API example for the asset icon doesn't require JS at all, it should simply be something like this:
```
<img src="https://liquid.place/api/v1/asset/6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d/icon">
```

frontend/package.json
```diff
@@ -70,7 +73,7 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.35",
     "@fortawesome/free-solid-svg-icons": "^5.15.3",
     "@juggle/resize-observer": "^3.3.1",
-    "@mempool/mempool.js": "^2.2.4",
+    "@mempool/mempool.js": "2.2.6",
```
wiz, line 76: This should be v2.3.0-dev1 etc. until the final release of v2.3.0 so it matches the backend.

wiz (changes requested):
Works well, but a few minor changes requested

frontend/src/app/components/docs/api-docs.component.ts
softsimon: Not possible at the moment without refactoring the API Docs component to support this.

wiz (approved):
Tested ACK @ 533653e5
Works well using example https://liquid.place/api/v1/asset/6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d/icon

PR #1011, opened 2021-12-14
Fixing broken RBF alert

softsimon:
fixes #516

Fixing the CSS flexbox code so that elements are positioned properly. Screenshot provided with Dev Tools visuals enabled.

Desktop:
Screen Shot 2021-12-15 at 01 14 35

Mobile:
Screen Shot 2021-12-15 at 01 16 01

PR #1015, opened 2021-12-15
Add tests for RBF txs

knorrium:
Checks layout issues on RBF transactions using new websocket mocks.

Overlapping confirmations and alert on smaller screens:

pr-before1

Wrong alert size on desktop:

pr-before2

Both tests pass after merging #1011:

pr-after


softsimon (approved):
ACK

PR #1024, opened 2021-12-17
User can drag the blockchain blocks horizontally with the mouse

### LINKED ISSUE #907: Have a way to scroll the blocks chain to the right.
image

Currently it's possible by dragging the text select toward the right.

Not optimal, what about a > button on the right side like so.

image

nymkappa:
Closes #907 

Tested on Brave, Firefox and Safari (Desktop)
Test on DuckDuckGo and Safari (Mobile iOS)

ezgif-4-dedea358d7

Feature specs/checklist:
- [X] User can drag blocks with the followings actions: mousedown -> mousemove -> mouseup
- [X] When the mouse leaves the blocks container, drag does not stop
- [X] User can still click properly on blocks to inspect them (there is no conflict between the drag and click actions)
- [X] Disable all element selections events in the blocks container. When user move the mouse all over the page while dragging blocks, no element are selected.
- [X] It does not break mobile interactions

Feel free to update this checklist if you see anything else.
Thank you.

### DIFF
--- a/frontend/src/app/components/start/start.component.ts
+++ b/frontend/src/app/components/start/start.component.ts
@@ -1,8 +1,7 @@
-import { Component, OnInit } from '@angular/core';
+import { Component, ElementRef, HostListener, OnInit, ViewChild } from '@angular/core';
 import { WebsocketService } from 'src/app/services/websocket.service';
 import { StateService } from 'src/app/services/state.service';
 import { specialBlocks } from 'src/app/app.constants';
-import { takeLast } from 'rxjs/operators';
 
 @Component({
   selector: 'app-start',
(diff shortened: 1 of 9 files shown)


frontend/src/app/components/start/start.component.ts
```diff
@@ -50,4 +52,27 @@ export class StartComponent implements OnInit {
       });
   }
 
+  onMouseDown(event: MouseEvent, blockchainContainer: Element) {
+    this.mouseDragStartX = event.clientX;
+    this.blockchainScrollLeftInit = blockchainContainer.scrollLeft;
+    this.blockchainContainer = blockchainContainer;
```
softsimon, line 58: You can add `@ViewChild('blockchaincontainer') blockchainContainer: ElementRef;` to the class and you don't have to pass the element to the function and set it here.

softsimon (approved):
tested ACK

PR #1027, opened 2021-12-17
Revamp docs layout

hunicus:
These changes migrate docs from a simple-but-limited `ngb-accordion`-based layout to something more flexible.

This PR is big but doesn't add or remove any content that is on the site right now.

Results:
- anchor links make every api endpoint linkable
- custom layouts allow for expansion later on as/when needed (e.g., search)
- new layouts show more info on screens with more space
- new layouts should also accommodate faqs and other docs content nicely

### DIFF
--- a/frontend/src/app/components/docs/api-docs-nav.component.html
+++ b/frontend/src/app/components/docs/api-docs-nav.component.html
@@ -1,0 +1,63 @@
+<p *ngIf="network.val !== 'bisq' && network.val !== 'liquid'">General</p>
+<a *ngIf="network.val !== 'bisq' && network.val !== 'liquid'" [routerLink]="['./']" fragment="get-difficulty-adjustment">GET Difficulty Adjustment</a>
+
+<p *ngIf="network.val === 'bisq'">Markets</p>
+<a *ngIf="network.val === 'bisq'" [routerLink]="['./']" fragment="get-market-currencies">GET Market Currencies</a>
+<a *ngIf="network.val === 'bisq'" [routerLink]="['./']" fragment="get-market-depth">GET Market Depth</a>
+<a *ngIf="network.val === 'bisq'" [routerLink]="['./']" fragment="get-market-hloc">GET Market HLOC</a>
+<a *ngIf="network.val === 'bisq'" [routerLink]="['./']" fragment="get-markets">GET Markets</a>
+<a *ngIf="network.val === 'bisq'" [routerLink]="['./']" fragment="get-market-offers">GET Market Offers</a>
+<a *ngIf="network.val === 'bisq'" [routerLink]="['./']" fragment="get-market-ticker">GET Market Ticker</a>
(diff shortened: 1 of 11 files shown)


frontend/src/app/components/docs/api-docs-nav.component.html
```diff
@@ -0,0 +1,63 @@
+<p *ngIf="network.val !== 'bisq' && network.val !== 'liquid'">General</p>
+<a *ngIf="network.val !== 'bisq' && network.val !== 'liquid'" [routerLink]="['./']" fragment="get-difficulty-adjustment" (click)="collapseItem.toggle()">GET Difficulty Adjustment</a>
+
+<p *ngIf="network.val === 'bisq'">Markets</p>
```
softsimon, line 4: Consider replacing whole blocks of identical *ngIf rows with a single wrapper like:
```
<ng-template [ngIf]="network.val === 'bisq'">
   <a> ... </a>
   <a> ... </a>
   <a> ... </a>
</ng-template> 
```

frontend/src/app/components/docs/api-docs-nav.component.html
```diff
@@ -0,0 +1,63 @@
...
+<a *ngIf="network.val !== 'bisq'" [routerLink]="['./']" fragment="get-block-status" (click)="collapseItem.toggle()">GET Block Status</a>
+<a [routerLink]="['./']" fragment="get-block-tip-height" (click)="collapseItem.toggle()">GET Block Tip Height</a>
+<a *ngIf="network.val !== 'bisq'" [routerLink]="['./']" fragment="get-block-tip-hash" (click)="collapseItem.toggle()">GET Block Tip Hash</a>
+<a *ngIf="network.val !== 'bisq'" [routerLink]="['./']" fragment="get-block-transaction-id" (click)="collapseItem.toggle()">GET Block Transaction ID</a>
+<a *ngIf="network.val !== 'bisq'" [routerLink]="['./']" fragment="get-block-transaction-ids" (click)="collapseItem.toggle()">GET Block Transaction IDs</a>
+<a *ngIf="network.val !== 'bisq'" [routerLink]="['./']" fragment="get-block-transactions" (click)="collapseItem.toggle()">GET Block Transactions</a>
+<a *ngIf="network.val !== 'bisq'" [routerLink]="['./']" fragment="get-blocks" (click)="collapseItem.toggle()">GET Blocks</a>
+<a *ngIf="network.val === 'bisq'" [routerLink]="['./']" fragment="get-blocks" (click)="collapseItem.toggle()">GET Blocks</a>
```
softsimon, line 41: If GET Blocks are going to show both when network is Bisq and not, better just use one line with no *ngIf

wiz (approved):
Tested ACK @ a2a85469
Looks good, tested on all networks and permalinks to API docs are great. The only thing I would fix is changing from mainnet to signet or testnet, it should stay on the /docs page - but that can be done in a separate PR.

PR #1036, opened 2021-12-21
Refactor docs

hunicus:
These changes move all API documentation data to a single separate file, so that future changes are easier and less error-prone. 

There is probably more that can be optimized, but my main goal here was to prevent repetition and time-consuming updates.

I opened this PR early as a draft in case anyone has suggestions about the approach.


hunicus:
I've marked this pull request as ready to review. I'm aware @knorrium mentioned he might have some suggestions for further improvements and @softsimon wanted to see i18n done better.

This PR is already pretty big...should we do further docs refactoring in new PRs, instead of adding to this one?

wiz (changes requested):
Few things:
* Need to fix merge conflict due to new Liquid Assets icon API being added in #1010
* Localized URLs don't open the specific question, for example https://node206.mempool.space/ja/docs/api/rest#get-transaction-merkle-proof
* Since it's a major refactor, I think we should ask @knorrium and @softsimon to test / approve this PR as well

hunicus:
> * Localized URLs don't open the specific question, for example https://node206.mempool.space/ja/docs/api/rest#get-transaction-merkle-proof

They work for me. I've tested on Chromium and Firefox desktop and Chrome on Android. Can you provide additional details on the setup it didn't work for you on?

hunicus:
Also I'm adding the new endpoints from #1010 here to resolve merge conflicts but getting error when clicking ES Module tab for get-asset-icon. Looks like it's not working in prod either (see https://liquid.network/docs/api/rest#get-asset-icon).

hunicus:
This PR is now out-of-date and needs significant updates beyond a simple rebase, so I'll close it. I'll either re-open or re-submit when I've got something ready to review.

PR #1038, opened 2021-12-22
Switch the 24h chart to 1 min data ticks

nymkappa:
Fix feedback from Murch regarding the lacking of resolution in the 24h mempool chart when blocks are mined


wiz (approved):
Tested ACK @ e4a9fd06

PR #1041, opened 2021-12-25
Display legal notices for third party software on About page

wiz:
Made a scrollable textarea for all the 3rd party licenses and legal notices, how does this look?

Screen Shot 2021-12-25 at 10 49 14


wiz:
Oops, forgot to push fix tor mobile layout so will revert to draft

wiz:
@softsimon any suggestions how to make this look less ugly?

wiz:
hmm ok will think about this later

PR #1057, opened 2021-12-29
Tweak hover effect on profile photos on About page

wiz:


softsimon (approved):

softsimon:
tested ACK at 766803de

PR #1060, opened 2021-12-30
Fixing misplaced Unknown text after the Coinbase

softsimon:


wiz (approved):
Tested ACK @ 40fb1792

PR #1062, opened 2021-12-30
Fix backend support for 'liquidtestnet' network

softsimon:
Since we added the new '`liquidtestnet`', all cases where we detect liquid network also needs to check for the liquid testnet network

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -109,7 +109,8 @@ export class Common {
       totalFees += tx.bestDescendant.fee;
     }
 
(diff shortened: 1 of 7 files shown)


knorrium:
@softsimon I would perhaps change the check to use indexOf(“liquid”) to simplify things or if you feel like refactoring the code to something like isLiquid() which abstracts both checks

wiz (approved):
Tested ACK @ 350b4922
Required clearing backend cache to fix issue, but works well

softsimon:
> @softsimon I would perhaps change the check to use indexOf(“liquid”) to simplify things or if you feel like refactoring the code to something like isLiquid() which abstracts both checks

Yeah.. I agree... was thinking the same.

PR #1065, opened 2021-12-30
Block navigation routing fix

softsimon:


wiz (approved):
Tested ACK @ 4f21fc0d

PR #1079, opened 2022-01-04
Liquid support to the Graph fee filter dropdown

### LINKED ISSUE #1072: Liquid: Graph fee filter is in a different scale than the legend
The changes to the Liquid fee scale were not done to the filter dropdown:

Screenshot from 2021-12-30 20-45-11

softsimon:
fixes #1072

Displaying fee levels on Liquid networks just like we do on the graph.

Due to the refactor in this PR, double check so the fee levels match when filtering and reversing order on both Liquid and Bitcoin mainnet.

Screen Shot 2022-01-04 at 04 37 29


wiz (approved):
Tested ACK @ e1275c62
Seems OK on mainnet/liquid using desktop/mobile

PR #1110, opened 2022-01-10
Cap extreme vbytes_per_second values

nymkappa:
* [x] Apply AVG() on vbytes_per_second so we can see trends
* [x] Cap extreme vbytes_per_second values so the y-axis scale is too big and we can still see details
  * For 2h/24h/1w chart - Cap to 10x of the median value
  * For longer timeframe - Cap to 4x of the median value
  * Those ratios can be changed easily here https://github.com/nymkappa/mempool/commit/45542d5f06888617f7388db802115c325496b430#diff-42a88bf2b0c6898fd6e3a11fa981beb6b26bb7b292e34c033cbfbd313ce5c77cR172


wiz (approved):
Tested ACK @ 45542d5f

PR #1120, opened 2022-01-11
Updates to the docker-compose setup

knorrium:
- allow docker-compose to override any backend settings
- add new `docker-compose.yml` file
- updated README with new instructions and troubleshooting steps


wiz (approved):
utACK

PR #1121, opened 2022-01-11
Wrap migration with transactions

nymkappa:
- [X] `database-migration.ts` now uses transactions for executing atomic query series
- [X] Replace `CAST as FLOAT` with `CAST as DOUBLE` to support MariaDB version 10.2.X
- [X] Manually check if `statistics.added` is INDEX or not and create it if needed (compatible with MariaDB 5.5+)
- [X] Refactor `database-migration.ts`


wiz (approved):
Tested ACK @ 2944f0b80563ae743729c027f95f41990d652ee4

PR #1124, opened 2022-01-11
Bumping minor Angular version and major ngBootstrap

softsimon:


wiz (approved):
Tested ACK @ 2f5cad9d

PR #1127, opened 2022-01-12
fix directory issue in rsync command

antonilol:
the command that's now in the readme create the locale and resource folders in `/var/www/browser/`

now it corresponds with https://github.com/mempool/mempool/blob/master/nginx-mempool.conf#L4

after fix
```
$ ls /var/www/mempool/browser/
ar  cs  en-US  fa  fr  hi  it  ka  mk  nl  pt         ro  sl  th  uk  zh
ca  de  es     fi  he  hu  ja  ko  nb  pl  resources  ru  sv  tr  vi
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ Install mempool dependencies from npm and build the frontend static HTML/CSS/JS:
 Install the output into nginx webroot folder:
 
 ```bash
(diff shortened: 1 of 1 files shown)


wiz:
Thanks for the PR but @knorrium just rewrote our entire README

antonilol:
i see the rsync command is fixed so no problem

PR #1130, opened 2022-01-12
Removing statistics cache and setting headers

softsimon:


wiz (approved):
Tested ACK @ 29581f32

PR #1133, opened 2022-01-13
Disable graph touch interaction in dashboard on mobile so we can scroll properly

nymkappa:
- [X] Can't interact with graphs in the dashboard on mobile only
- [X] Always hide tooltip on vBytes/sec chart on mobile for consistency with the mempool chart which already does not show it
- [X] Can still zoom in/out in `/graphs` view (and therefore the scrolling UX is as before, not so great)
  - For a later PR, need to improve mempool chart performance on mobile, it's terribly slow right now (very hard to use the zoom)


wiz (approved):
Tested ACK @ 5f2350b76308c76531b7ba5cab4b9636d440e113
Works well, my only thought is that isMobile() should probably be refactored into a shared/common typescript file somewhere

PR #1139, opened 2022-01-14
Order by native `statistics.added` field for better query performances

nymkappa:
Replace `ORDER BY added` by `ORDER BY statistics.added` to improve query speed


nymkappa:
> SELECT *, UNIX_TIMESTAMP(added) as added FROM statistics ORDER BY added DESC LIMIT 120
120 rows (2.235 s)
> SELECT *, UNIX_TIMESTAMP(added) as added FROM statistics ORDER BY added DESC LIMIT 120
120 rows (0.002 s)

wiz (approved):
Tested ACK @ 8797ef26

PR #1141, opened 2022-01-14
Remove unused fields from statistics queries since we don't use them in the front end

nymkappa:
Remove `id`, `unconfirmed_transactions` and `tx_per_second` from the statistics queries since we don't use them in the front end


wiz (approved):
Tested ACK @ v2.3.0-dev [5b39ad21]
Seems OK and 20% less data returned:

### Before
Screen Shot 2022-01-14 at 19 32 42

### After

Screen Shot 2022-01-14 at 19 32 08

softsimon (approved):
utACK

PR #1147, opened 2022-01-15
Utilize gettxout to display spent/unspent

### LINKED ISSUE #1088: Use gettxout to flag transaction outputs as spent/unspent
Use the `gettxout` API https://developer.bitcoin.org/reference/rpc/gettxout.html to flag transaction outputs as spent/unspent.

Currently we only display grey arrows due to the missing UTXO API.
Although it's not possible to know the TXID spending transaction, we can at least display red/green if they are spent or unspent.

Screen Shot 2022-01-05 at 15 54 29

softsimon:
fixes #1088

Unlike with esplora the red spent is not clickable, but at least we are showing if UTXOs are spent or not, and we get rid of the 501 error message.

Before:
Screen Shot 2022-01-15 at 22 10 35

Screen Shot 2022-01-15 at 22 07 29

After:
Screen Shot 2022-01-15 at 22 07 23


wiz (approved):
utACK but will test as part of v2.3.0 release testing

PR #1152, opened 2022-01-17
Move difficulty adjustment code to separate module

nymkappa:
Split difficulty adjustment code into a separate module so we can re-use it in the mining dashboard


wiz:
Why is the difficulty adjustment code in the frontend at all?
Shouldn't it be in the backend and exposed by the websocket and REST APIs ?

nymkappa:
> Why is the difficulty adjustment code in the frontend at all? Shouldn't it be in the backend and exposed by the websocket and REST APIs ?

I agree. This would be an useful API endpoint for mining service providers and miners.

wiz:
I suppose it just needs to be added to the websocket since the REST API endpoint exists at https://mempool.space/testnet/api/v1/difficulty-adjustment

Did you want to close this PR and do that instead? I don't see why there is a totally duplicate implementation of the difficulty adjustment code in the frontend at all, unless there is some reason I am missing? @softsimon

softsimon (approved):
tested ACK @ fcca9113

PR #1156, opened 2022-01-17
Add 0 as lower bound for block's fees

Xekyo:
I noticed that block 501726 didn't collect a mining reward and tested it
on a few block explorers. On mempool, it currently shows up as "Total
fees: -₿12.5". Negative fees don't make sense, so I'm introducing a
lower bound of 0 for the displayed value.
image

-----

This is my first code contribution, it seems very likely that I didn't catch all the relevant places or broke something else. Please let me know what else needs to be done or whether you'd rather take over the PR. I'm also not sure whether there should be a test for this, and it seems a bit arduous to sync an instance to block 502k to see whether it was fixed, so ideas how to confirm that the issues was resolved would be welcome. ;)

### DIFF
--- a/frontend/src/app/components/block/block.component.ts
+++ b/frontend/src/app/components/block/block.component.ts
@@ -76,7 +76,7 @@ export class BlockComponent implements OnInit, OnDestroy {
 
         if (block.id === this.blockHash) {
           this.block = block;
-          this.fees = block.reward / 100000000 - this.blockSubsidy;
+          this.fees = Math.max(0, block.reward / 100000000 - this.blockSubsidy);
         }
       });
 
@@ -147,7 +147,7 @@ export class BlockComponent implements OnInit, OnDestroy {
         }
         this.setBlockSubsidy();
         if (block.reward !== undefined) {
-          this.fees = block.reward / 100000000 - this.blockSubsidy;
+          this.fees = Math.max(0, block.reward / 100000000 - this.blockSubsidy);
         }
         this.stateService.markBlock$.next({ blockHeight: this.blockHeight });
         this.isLoadingTransactions = true;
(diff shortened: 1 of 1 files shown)


nymkappa:
You can try using `npm run start:local-prod` and it will get the data from mempool.space API. If you get some error about asset syncing, get rid of the `npm run sync-assets-dev` in the `start:local-prod` command in `package.json`.

It works fine for me.
image

Xekyo:
Thanks @nymkappa!

wiz:
@Xekyo thanks for the PR, might need some more tweaking, take a look at https://mempool.ninja/block/0000000000000000004b27f9ee7ba33d6f048f684aaeb0eea4befd80f1701126

Xekyo:
Mh, seems that this line would produce an inaccurate amount when the miner collects less than the block subsidy:   https://github.com/mempool/mempool/blob/7ce1cc5103ac3b989ca309a41b556f93a8da9f27/frontend/src/app/components/block/block.component.html#L92

Perhaps it would need to be turned around altogether:

"Subsidy + fees:" should be calculated from the outputs of the coinbase transaction as done in https://github.com/mempool/mempool/blob/3a67bc64259d1cdfcc861240c7aa02f57f2f8359/frontend/src/app/components/block/block.component.ts#L167, and "total fees" should be calculated via `fees = Math.max(0, reward - subsidy`.

nymkappa:
The core issue is that the full block reward is assumed to be claimed by the miner, which is incorrect. It is set here and used later on for display: https://github.com/mempool/mempool/blob/3a67bc64259d1cdfcc861240c7aa02f57f2f8359/frontend/src/app/components/block/block.component.ts#L214

Let's get rid of this and instead compute things properly:
* Fetch the total block subsidy (`block_subsidy`) using the coinbase tx
* Sum the fee paid for all tx (`fee_sum`) in the block (using the diff between old and new utxos)
* Minus the `fee_sum` the from `total block subsidy` => Gives the block reward and should ~match 50/25/12.5/6.5/etc...
* Display all of this nicely :)

Let me know if you want me to take a look at it.

Xekyo:
@nymkappa: Your approach sounds good to me. If you want to take over the PR, I don't mind at all.

PR #1158, opened 2022-01-18
dont use hardcoded genesis coinbase and block hash

### LINKED ISSUE #1128: genesis block doesn't load on regtest
### Description

when clicking on the genesis block in regtest mempool is stuck loading and the log says
`DEBUG: <regtest> getBlockTransactions error: Block not found`

image

https://github.com/mempool/mempool/discussions/1023#discussioncomment-1953509

antonilol:
fixes #1128

i generalized the way the coinbase of the genesis block is detected (first tx of first block instead of hardcoded tx hash and block hash)

if i did something wrong in indentation / tab style feel free to edit

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api.ts
+++ b/backend/src/api/bitcoin/bitcoin-api.ts
@@ -21,11 +21,6 @@ class BitcoinApi implements AbstractBitcoinApi {
       return this.$addPrevouts(txInMempool);
     }
 
-    // Special case to fetch the Coinbase transaction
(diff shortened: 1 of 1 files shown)


nymkappa:
Other than my previous comment, I've tested it on mainnet/testnet/regtest and it works fine.

antonilol:
On mainnet, testnet and signet it works fine because they have the same genesis block. Regtest has a different one. I dont know if it is a new one for every regtest.

softsimon (approved):
tested ACK @ fac40b15 on Signet

PR #1160, opened 2022-01-18
outputs of genesis coinbase are always unspent

### LINKED ISSUE #1159: genesis coinbase looks spent (romanz/electrs)
### Description

because gettxout for the genesis coinbase returns nothing it looks like it was spent 😳

antonilol:
fixes #1159

forces the outspend arrow of the genesis coinbase to green (`spent=false`)

maybe even make it gray because it is unspendable  image


softsimon (approved):
tested ACK @ 63939ddb

PR #1162, opened 2022-01-19
Mining dashboard (2/2) - Dashboard PoC

nymkappa:
This PR depends on https://github.com/mempool/mempool/pull/1162

image

Mining page proof of concept.

* Add `/pools` page to show
  * Epoch data
  * Mining pool ranking
    * Hashrate distribution 1 day
    * Number of block mined
    * Logo (need to figure out where we can get the assets, or ask permission to another platform to use theirs)
  * Can switch the timespan. Only 1 day shows hashrate data
* When Core is fully sync, mempool automatically starts indexing blocks
  * height, hash, block timestamp, size, weight, tx count, raw coinbase tx, difficulty, mining pool
  * *fee data will be added in a following PR* 
* Hide `/blocks` from the menu


wiz:
Feedback so far:
* Remove difficulty adjustment for now
* Add (50%) percentages in the pie chart titles
* The block count for all miners on All is only 603K and doesn't match block height of 719K

nymkappa:
> Feedback so far:
> 
> * Remove difficulty adjustment for now
> * Add (50%) percentages in the pie chart titles
> * The block count for all miners on All is only 603K and doesn't match block height of 719K

Fixed the first two.

I've also noticed the weird behavior for `All` but I can't reproduce it properly. I think it happens when API responses arrives in a different order from the user click actions (I need to block the timespan switch while the data are loading).
I also noticed some odd result while indexing blocks.

I'll look into those two situations. If you have more info let me know.

softsimon (commented):
A rough calculation based on the speed I am getting, it takes about 6-8 days to do the block sync with coinbase transactions on my Raspberry Pi 8 GB.

Although I don't think adds so much extra strain to the device, we should we have a configurable setting that can have a default value of only going back 3 days worth of blocks when doing the indexing and go from there.

nymkappa:
@softsimon if we decide not to index all blocks then we wont be using this new data for the /block page and APIs.

nymkappa:
> A rough calculation based on the speed I am getting, it takes about 6-8 days to do the block sync with coinbase transactions on my Raspberry Pi 8 GB.
> 
> Although I don't think adds so much extra strain to the device, we should we have a configurable setting that can have a default value of only going back 3 days worth of blocks when doing the indexing and go from there.

Added `INDEXING_BLOCKS_AMOUNT` config option, default set to 432 (3 days worth of mining at 10min/block)

backend/src/index.ts
```diff
@@ -136,6 +140,16 @@ class Server {
       }
       await blocks.$updateBlocks();
       await memPool.$updateMempool();
+
+      const blockchainInfo = await bitcoinClient.getBlockchainInfo();
+      if (this.blockIndexingStarted === false
```
softsimon, line 145: Maybe move this block of code into the `blocks.ts` file so we can clean up `index.ts` from blockIndexingStarted and bitcoinClient ?

backend/src/index.ts
```diff
@@ -136,6 +138,8 @@ class Server {
       }
       await blocks.$updateBlocks();
       await memPool.$updateMempool();
+      blocks.$generateBlockDatabase();
```
softsimon, line 141: We need to wrap this in `if (cluster.isMaster) ` so that only one of the nodejs threads are doing the block indexing.

Same thing when a new block is found. I have not confirmed whether this is an issue or not at the moment

wiz: We aren't using the cluster feature anymore, so I'm fine with deprecating it for now and re-visiting this later - as you said, the only thing that really needs to be clustered is probably the websocket

backend/src/repositories/PoolsRepository.ts
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Get basic pool info and block count
+   */
+  public async $getPoolsInfo(interval: string = '100 YEARS'): Promise<PoolInfo[]> {
+    const connection = await DB.pool.getConnection();
+    const [rows] = await connection.query(`
+      SELECT COUNT(height) as blockCount, pool_id as poolId, pools.name as name, pools.link as link
+      FROM blocks
+      JOIN pools on pools.id = pool_id
+      WHERE blocks.blockTimestamp BETWEEN DATE_SUB(NOW(), INTERVAL ${interval}) AND NOW()
```
wiz, line 34: please don't use unfiltered user input in the SQL query

wiz (changes requested):
NACK for using unfiltered user input in SQL query

backend/src/repositories/PoolsRepository.ts
nymkappa: Embarrassing issue fixed in a2bc7298d346dd0a2f818bbda0deecc78d34e413

backend/src/routes.ts
```diff
@@ -531,6 +532,18 @@ class Routes {
     }
   }
 
+  public async $getPools(req: Request, res: Response) {
+    try {
+      let stats = await miningStats.$getPoolsStats(req.query.interval as string);
+      // res.header('Pragma', 'public');
+      // res.header('Cache-control', 'public');
+      // res.setHeader('Expires', new Date(Date.now() + 1000 * 300).toUTCString());
```
wiz, line 540: I assume you want to uncomment this now

frontend/mempool-frontend-config.sample.json
```diff
@@ -12,6 +12,7 @@
   "NGINX_PORT": "80",
   "BLOCK_WEIGHT_UNITS": 4000000,
   "MEMPOOL_BLOCKS_AMOUNT": 8,
+  "INDEXING_BLOCKS_AMOUNT": 432,
```
wiz, line 15: Shouldn't this be 1100?

frontend/server.ts
```diff
@@ -6,7 +6,6 @@ import * as express from 'express';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as domino from 'domino';
-import { createProxyMiddleware } from 'http-proxy-middleware';
```
wiz, line 9: Why was this removed?

backend/src/api/mempool.ts
```diff
@@ -118,11 +118,11 @@ class Mempool {
...
-          if (diff > 0) {
-            logger.debug('Fetched transaction ' + txCount + ' / ' + diff);
-          } else {
-            logger.debug('Fetched transaction ' + txCount);
-          }
+          // if (diff > 0) {
+          //   logger.debug('Fetched transaction ' + txCount + ' / ' + diff);
+          // } else {
+          //   logger.debug('Fetched transaction ' + txCount);
+          // }
```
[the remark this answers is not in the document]
nymkappa: Yes, temporary dev thing I forgot, nice catch!

frontend/mempool-frontend-config.sample.json
nymkappa: Actually, wrong file. It should be in backend only.

frontend/server.ts
nymkappa: This import is not used anywhere in the file so I removed it.
Is there some kind of magic injection thingy with this import or can I get rid of it?

backend/src/routes.ts
nymkappa: I've set the cache to 1 min

frontend/server.ts
nymkappa: I confirm this import is not used. Ref on how it is used: https://github.com/chimurai/http-proxy-middleware#tldr

nymkappa:
@wiz @softsimon I have rebased this branch, you need to reset your branch at `1a22923cd8e5a21bf0c164fb11dcc74162bcc807` and pull again

wiz (approved):
Tested ACK @ v2.4.0-dev [9df49037]
Very nice for a first PoC, but needs more work on the following:
* Missing call to SEO service to set page title, etc.
* Method of finding mining pool logo images can be improved
* API endpoints should not use ?interval=1d but simply /1d
* Need to add relevant API endpoints to nginx-cache-warmer script
* And of course, we need to design a nicer UI now that the building blocks are in place :)

Excellent work!

PR #1163, opened 2022-01-19
Mining dashboard (1/2) - Import mining pools into the database - Increment db schema to 3

nymkappa:
This PR is a prelude to the new mining dashboard https://github.com/mempool/mempool/pull/1162.

When mempool is started, the database migration will run and increment the db schema version to 3. It will create a new table `pools` and parse + import https://github.com/mempool/mempool/blob/master/frontend/cypress/fixtures/pools.json into the database.

Note: The `PoolsParser` script does not yet support updating. A future PR will allow updating the `pools` table when needed.

### DIFF
--- a/backend/src/api/pools-parser.ts
+++ b/backend/src/api/pools-parser.ts
@@ -3,10 +3,10 @@ import { DB } from '../database';
 import logger from '../logger';
 
 interface Pool {
-  name: string,
-  link: string,
-  regexes: string[],
-  addresses: string[],
+  name: string;
+  link: string;
+  regexes: string[];
+  addresses: string[];
 }
 
 class PoolsParser {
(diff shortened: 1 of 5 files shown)


backend/src/api/pools-parser.ts
```diff
@@ -0,0 +1,118 @@
...
+    // but that will come in a later version
+    let [rows] = await connection.query<any>({ sql: 'SELECT count(id) as count from pools;', timeout: 120000 });
+    if (rows[0].count !== 0) {
+      logger.info('Pools table already contain data, updating it is not yet supported, skipping.');
+      connection.release();
+      return;
+    }
+
+    logger.info('Open ../frontend/cypress/fixtures/pools.json');
+    const fileContent: string = readFileSync('../frontend/cypress/fixtures/pools.json','utf8');
```
softsimon, line 30: Add `https://raw.githubusercontent.com/btccom/Blockchain-Known-Pools/master/pools.json` to the `EXTERNAL_ASSETS` config so it will be updated to the latest on startup. Then just read `./pools.json`

wiz: that JSON file should be in the default configuration

nymkappa: > Add `https://raw.githubusercontent.com/btccom/Blockchain-Known-Pools/master/pools.json` to the `EXTERNAL_ASSETS` config so it will be downloaded on startup. Then just read `./pools.json`

I think it's better to keep a copy in our repo and update from time to time for now. Downloading it automatically on everyone's Core node server is not be a good idea IMO.

Also, are `EXTERNAL_ASSETS` automatically re-downloaded if they change?

knorrium: you shouldn't be reading from that file anyway as it's intended to be used as a mock for the e2e tests

nymkappa: Fixed! :)

wiz (changes requested):
NACK @ v2.4.0-dev [8d1cc404]

Screen Shot 2022-01-20 at 22 20 10

This PR breaks all networks except mainnet. Obviously there are no mining pools for testnet, signet, liquid, etc. so it seems you are missing an "if network is mainnet" check around your code

wiz:
Also crashes if the MySQL server becomes unavailable, currently this is handled gracefully with automatic reconnection but this PR crashes due to an unhandled promise

Screen Shot 2022-01-20 at 22 38 03

wiz (approved):
Tested ACK @ v2.4.0-dev [1322298a]

PR #1165, opened 2022-01-19
load blocks with height under INITIAL_BLOCKS_AMOUNT + more regtest fixes

antonilol:
makes mempool more regtest friendly

mempool couldn't load more than one block if the block height is lower than `INITIAL_BLOCKS_AMOUNT` from the config

- [x] network in config and dropdown
- [x] genesis block visibility + arrow
- [x] dashboard blocks issue (https://github.com/mempool/mempool/pull/1165#issuecomment-1017808054)
- [x] the same as above but with TXs
- [ ] code examples, dont know about this one, maybe copy it from the signet
- [ ] color scheme/style and icon, input needed here


antonilol:
is there a more efficient/faster frontend build method for development?
i now use the one from the readme (but then as one-liner)
`npm run build && sudo rsync -av --delete dist/ /var/www/ && sudo systemctl restart nginx`

nymkappa:
> is there a more efficient/faster frontend build method for development? i now use the one from the readme (but then as one-liner) `npm run build && sudo rsync -av --delete dist/ /var/www/ && sudo systemctl restart nginx`

`npm run start`

antonilol:
fixed it!

antonilol:
most of the fixes here are for bugs only seen on regtest but also present on the others

e.g. to see the bug i fixed in e8986e5 you must scroll all the way down on https://mempool.space/blocks

antonilol:
everything should be good by now except 2 things, will be fixing it later

- if the Blocks list in not completely filled (less than 6 blocks) and you switch to a network without the backend running, the list stays and when going back the blocks are added again and again
image
- no regtest code examples
image

antonilol:
which colors do you want for regtest? I copied css from signet and symlinked the icon to the signet one for now

PR #1171, opened 2022-01-20
Fix Liquid proxy settings

knorrium:


softsimon (approved):
tested ACK @ 35e69f2e

PR #1182, opened 2022-01-24
Wrap statistics db ops with try/catch

### LINKED ISSUE #1181: Mempool crashes if we cannot connect to the database anymore
The backend crashes if the MySQL server becomes unavailable. @wiz mentioned that
> currently this is handled gracefully with automatic reconnection but this PR crashes due to an unhandled promise

However this is not the case right now on master branch. To reproduce, simply run the backend with master branch, and shutdown mysql service, and wait for a few seconds.

Screen Shot 2022-01-20 at 22 38 03

_Originally posted by @wiz in https://github.com/mempool/mempool/issues/1163#issuecomment-1017514578_

nymkappa:
Fixes #1181

### DIFF
--- a/backend/src/api/statistics.ts
+++ b/backend/src/api/statistics.ts
@@ -53,12 +53,16 @@ class Statistics {
     memPoolArray = memPoolArray.filter((tx) => tx.effectiveFeePerVsize);
 
     if (!memPoolArray.length) {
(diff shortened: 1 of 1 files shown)


softsimon (approved):
utACK

softsimon:
tested ACK @ a805c86

PR #1185, opened 2022-01-24
fix 0 sat/vB not displaying

### LINKED ISSUE #1115: show fee rate of 0 sat/vB TXs
### Description

https://mempool.space/testnet/block/000000001efffa5b45762dd0be070a9fa949df27344f2969828e038a5c3f595b

the second TX has 0 sat/vB fee rate, it is not displayed unlike the third one which has 134 sat/vB fee rate

i think it is not showing because showing fee rate for coinbase TXs is useless, but coinbase TXs are not the only TXs with 0 sat/vB.

non coinbase mainnet example (F2Pool paying out there miners in their own blocks, 0 sat/vB):

[it](https://mempool.space/tx/11fa91a1a0a17373d3217f55a6a98ee637a58702c12251975532a7423358448c) shows after the coinbase of [block 717977](https://mempool.space/block/0000000000000000000ad0561808f0953e13434b614e53593e9142713807cee0)

0 sat/vB TXs now are sorted at the top (above the highest fee TXs) probably to get the coinbase at the top

### Screenshots

image

antonilol:
fixes #1115 

now it displays fee for every TX except coinbase, even if it's zero

image


softsimon (approved):
tested ACK @ 590170a0

PR #1187, opened 2022-01-25
Remove useless autocommit=0 in db migration script

nymkappa:


wiz (approved):
Tested ACK, fixes the issue we observed

Issue #1117, opened 2022-01-11
TV view is not responsive


nymkappa:
### Description

In the `/tv` view, I need to use the brower native zoom feature to see the whole page (75% zoom).

#### Version

https://github.com/mempool/mempool/commit/d809e85ddec801db392090205f0af1eac4cc8e81

### Expected behaviour

When you load the `/tv` page, all elements should be displayed regardless of the native browser zoom, screen resolution, etc...

### Screenshots

![image](https://user-images.githubusercontent.com/9780671/148874545-d787e1ea-e5a4-400a-b83d-33d7c1d102b2.png)

#### Device or machine

Brave Browser with 100% zoom - Linux Mint - 1080p screen
![image](https://user-images.githubusercontent.com/9780671/148874844-8d75712d-ff13-4e3e-946e-add27571e634.png)

hunicus:
The TV view in question was removed in https://github.com/mempool/mempool/pull/5791.

PR #1191, opened 2022-01-26
TV view improvements (1/2) - Improve layout

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1117. Now you should be able to see both the mempool chart and blocks by default and most screen setups. It's definitely not perfect but much better than now.

Hide the blocks when the screen height is less than 650px.

image


wiz:
Wait, why are the blocks on the top lol

nymkappa:
Re-opening here https://github.com/mempool/mempool/pull/1203

PR #1194, opened 2022-01-27
Change Citadel link to core

AaronDewes:
The Citadel dashboard repo will soon be replaced, and this is consistend with other projects like Umbrel where the link leads to the main repo.


wiz (approved):
utACK

PR #1199, opened 2022-01-28
Let backend set its own HTTP cache time headers for APIs

wiz:
```
curl -si https://mempool.ninja/api/v1/statistics/24h
```
gotta check expires headers for all api endpoints


Emzy (approved):
untested ACK

PR #1201, opened 2022-01-29
Merge the mempool.js repo into the main mempool repo

wiz:
Currently we have the problem where if we add new APIs or other functionality into the mempool.js codebase, we can't easily ship that together with the mempool frontend, since we currently have to release the mempool.js by shipping it as an npm module, and then bumping the version of the mempool.js npm dep in the package.json of the mempool frontend.

By merging the mempool.js repo into the main mempool repo this will allow us to greatly simplify everything by modifying the frontend build process to just browserify the mempool.js code directly and ship it as part of the frontend build, and we can remove the npm dep of mempool.js from the frontend package.json and we'll only need to ship npm modules when we're actually ready to release them.


wiz:
@knorrium could you takeover this PR and fix the frontend build system so that instead of mempool.js being an npm dep of the frontend, the frontend build simply builds from the mempool.js folder in same repo, and also change the target from mempool.js to liquid.js or bisq.js depending on the base_module in the frontend config

PR #1203, opened 2022-01-31
Fix TV View layout

nymkappa:
I rewrote the html/css for the mempool/blockchain blocks so that it does not use absolute positioning anymore.
This was required to fix the TV view layout.

I'm terrible a css so please let me know if it's fully broken and what need to be changes.

This PR should not create regression in the dashbaord:
* Blocks position and width
* Arrow positioning must still be correct (transactions and blocks arrow)
* Must not break other network specific rendering

I tired to use Cypress for checking the rendering on different device, but the rendering seems incorrect. Please try on your different devices (including mobile) and please leave a screenshot if the rendering looks broken.

The basic things to check are the following, settings can be tweaked.
* If the viewport is less than 500px, blocks are not shown at the bottom and the chart takes all the space
* If the viewport is between 500px and 700px, we show everything, but we set the zoom factor to 0.7
* If the viewport is more than 700px tall, we show everything with a zoom factor of 1

### DIFF
--- a/frontend/cypress/integration/mainnet/mainnet.spec.ts
+++ b/frontend/cypress/integration/mainnet/mainnet.spec.ts
@@ -178,7 +178,7 @@ describe('Mainnet', () => {
           cy.viewport('macbook-16');
           cy.visit('/');
           cy.waitForSkeletonGone();
-          cy.get('.blockchain-blocks-0 > a').click().then(() => {
+          cy.get('.blockchain-blocks-0 > a').click({force: true}).then(() => {
             cy.get('[ngbtooltip="Next Block"] > .ng-fa-icon > .svg-inline--fa').should('not.exist');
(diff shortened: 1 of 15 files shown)


nymkappa:
I've squashed the branch. Please `git reset --hard 631864d24bd3500c7d5eefe80c93ea7b2e231668` before reviewing.

wiz (commented):
Tested NACK @ v2.4.0-dev [631864d2]

Refactoring the TV layout and related components seems like a large project, and requires lots of testing on various screen sizes... I'm a bit reluctant to merge this PR since the current TV view is mostly working fine and this PR seems to introduce a lot of issues from quick testing...

nymkappa:
Closing for now as this refactoring deserves more attention than this PR provided :-)

Issue #680, opened 2021-08-02
UX: When Bitcoin Core is still synchronizing, display something nicer than no blocks


softsimon:
When the Bitcoin Core is still synchronizing and there is no data, the Mempool screen looks like this which is confusing.

We do have the "Backend is synchronizing" message but in this case, maybe it should cover the blockchain or the whole dashboard, until there is 8 blocks at least.

![mempool-in-sync](https://user-images.githubusercontent.com/8561090/127856744-7142cfad-6567-4d90-ba92-2ebf1fb4cc49.jpg)

PR #713, opened 2021-08-13
Added loader while synchronising backend

priyanshiiit:
**Fix** #680 

**Preview Screencast**
---
https://user-images.githubusercontent.com/43826321/129402944-9c0f78ca-3729-422a-94bb-1420daaf31da.mp4

PR #1205, opened 2022-02-01
Bitcoin loading svg icon

Draichi:
## Screenshot

Recording 2022-02-01 at 02 42 05

This is a suggestion on how to show something cool to the user while he waits

Fix: #680 

## What needs to be done

The logic in this component is a copy of https://github.com/mempool/mempool/pull/713. Although there is an error on the `map` function when used in `start.component.ts`:
image

```
Cannot find name 'map'. Did you mean 'Map'?ts(2552)
lib.es2015.collection.d.ts(36, 13): 'Map' is declared here.
```

Couldn't find a solution yet


frontend/package.json
```diff
@@ -83,6 +83,7 @@
     "domino": "^2.1.6",
     "echarts": "^5.1.2",
     "express": "^4.17.1",
+    "gsap": "^3.9.1",
```
wiz, line 86: Concept NACK for adding new trusted third party npm dependencies

wiz:
Thanks for the PR, but to be honest, I don't think this animation matches the design style of the project...

softsimon:
This should be the dashboard loading icon:

https://jsfiddle.net/miguelmedeiros/7nt8dy6b/2/

wiz:
Another contributor of ours made an animated version of the mempool logo entirely in CSS that we had considered using for this purpose, maybe you can look at integrating that instead? And you don't need any new npm dependencies

softsimon:
It seems like the `specialEvent` feature is deleted in this PR and the `mempoolLoadingStatus$` is not working.. So this is not really a working PR ready for review.

If you instead use the animated mempool logo and solve the loading status detection, it would be something we would consider merging.

PR #1207, opened 2022-02-02
Set proper mining pool page title - Rename "Name" to "Pool"

nymkappa:
Self explanatory, mini PR 👍🏻 

Fixes `Missing call to SEO service to set page title, etc.
` from https://github.com/mempool/mempool/pull/1162#pullrequestreview-865932360


wiz (approved):
Tested ACK @ v2.4.0-dev [ae9afe2f]

PR #1208, opened 2022-02-02
Replaced /mining/pools?interval=X by /mining/pools/X

nymkappa:
Fixes `API endpoints should not use ?interval=1d but simply /1d` from https://github.com/mempool/mempool/pull/1162#pullrequestreview-865932360

Cache warmer has been updated accordingly.


softsimon (approved):
concept ACK

wiz (approved):
Tested ACK @ v2.4.0-dev [e48a2f73]

PR #1209, opened 2022-02-02
Add coinbase.hex and pool info in blocks data stream

nymkappa:
This PR simply adds two new information in the websocket blocks data stream:
* Added `hex` in `BlockExtended` interface
* Added `miner` (type `PoolTag`)

image


nymkappa:
After discussing this change with @softsimon, I will apply this change in a blocks data structure refactoring instead

PR #1213, opened 2022-02-04
Fetch conversion rates over Tor

### LINKED ISSUE #730: Add SOCKS proxy config option to route HTTP requests thru Tor
Currently the `price.bisq.wiz.biz` price API is hard-coded into mempool from the early days. We should improve this:

1) Add a new backend configuration option for price server URL
2) Add a new backend configuration option for SOCKS proxy, to be passed to axios

This will allow us to configure the Tor HS hostname `wizpriceje6q5tdrxkyiazsgu7irquiqjy2dptezqhrtu7l2qelqktid.onion` to be used on Umbrel and Raspiblitz instead of hitting it over clearnet, which will improve self-hosted home user privacy.

knorrium:
Fix #730 

- Adds a new configuration key to specify a SOCKS5 proxy to use when querying the conversion rates service (with a mempool/vX User-Agent)
- Update the docs, reference config files and Docker scripts with the new key

Tested with a Dockerized Tor Proxy: `docker run --rm --name tor -p 127.0.0.1:50090:9050 osminogin/tor-simple`

Still untested on Umbrel and Raspiblitz


wiz:
Awesome, thanks so much for working on this - can't wait to test it on Umbrel

wiz:
Looks good, but we should probably change all instances of the word "PRICENODE" to "PRICE_DATA_SERVER" since the price data is not served by a "node" on a p2p network, but instead a plain old server

wiz (approved):
Tested ACK @ v2.4.0-dev [4afeb399]
Works well using both clearnet and tor onion.
Excellent work sir.

PR #1215, opened 2022-02-04
Proxy all /api and /api/v1 requests to local nodejs server

softsimon:
Before multiple requests like "/api/block-height" and "/api/address-prefix" wasn't proxying.

This fix proxies all `/api/**` and `/api/v1/**` to the local server.


nymkappa:
Tested ACK - `/api/block-height` and `/api/address-prefix` works fine on my local now. Very nice.

wiz (approved):
utACK

PR #1216, opened 2022-02-04
Added `Pool` and `Reward` info to the blocks page

nymkappa:
image


nymkappa:
Note: As it is right now, re-orgs are not supported. A fix will be provided in a later PR. See https://github.com/mempool/mempool/issues/1221

softsimon (approved):
UI needs some polish but until then:

Code utACK

wiz (changes requested):
Tested NACK @ v2.4.0-dev [9b6289f1]

Loading `/blocks` page crashes with the following error in JS console:
```
ERROR TypeError: Cannot read properties of undefined (reading 'pool')
```

nymkappa:
Closing as this will only be used for historical reference. A future PR (building on top of https://github.com/mempool/mempool/pull/1235 if it is merged) will an updated version of the block list page. It is not a priority for now.

PR #1218, opened 2022-02-04
Local dev proxy working with all base modules

softsimon:
The default local dev proxy now works for all requests when using all 3 `BASE_MODULE` frontend configurations, mempool, liquid and bisq.


wiz (approved):
utACK

PR #1219, opened 2022-02-04
Adding a "mixed" dev proxy

softsimon:
This is a middle ground between `npm run start:local-prod` and `npm run start`

The new `npm run start:mixed` proxy will generally redirect all
`/api/` endpoints to Mempool.space or Liquid.network depending on Base Module
`/api/v1/` will redirect to the local NodeJS.

This is very helpful during development when there is a lack of a proper Esplora backend setup.
It makes it easier to browse around and develop on pages like Liquid Assets.

mempool, liquid and bisq as Frontend Base Module is supported.

### DIFF
--- a/frontend/angular.json
+++ b/frontend/angular.json
@@ -218,6 +218,10 @@
               "proxyConfig": "proxy.conf.local.js",
               "verbose": true
             },
+            "mixed": {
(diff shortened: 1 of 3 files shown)


wiz (approved):
utACK

PR #1220, opened 2022-02-05
Duplicated db blocks insertion attempts are expected

nymkappa:
Adding one extra mysql query to check if a block already exists would be very inefficient. There are situation where duplicated block insertion are possible (for example, if you clear your cache after indexing blocks, or according to different timing when restarting the backend), and those are non critical so we don't need to log an error for this.


knorrium (approved):
Tested ACK

softsimon (approved):
tested ACK at 4b871468

PR #1225, opened 2022-02-06
Add EXTERNAL_ASSETS defaults to the Docker start script and the README

knorrium:
No value was set in the Docker start script so the mining dashboard would index all blocks as `Unknown`


softsimon:
utACK

wiz (approved):
utACK

PR #1228, opened 2022-02-06
Run tests in Staging

knorrium:
- Add mining pool resources route
- Update Cypress to run against the staging backends


softsimon (approved):
utACK

PR #1232, opened 2022-02-08
Remove debug return which break the UX the first time we open mempool

nymkappa:
Right now on master, if you clear your `cache/` folder and restart the backend, blocks will come one by one every couple of seconds.

This one line PR actually restore the previous behavior, showing `INITIAL_BLOCKS_AMOUNT` right from the start with zeroed fee range.


softsimon (approved):
utACK

PR #1233, opened 2022-02-08
sync-assets: Download pool logos from github

nymkappa:
Added a new command `downloadMiningPoolLogos` in `sync-assets.js` file. It fetches the file list from https://api.github.com/repos/mempool/mining-pools/contents/ and download all the files to the `resources/mining-pools` folder.

It is then used by the pool ranking page `/mining/pools`.


wiz (approved):
Tested ACK @ v2.4.0-dev [c704bfed]

PR #1235, opened 2022-02-08
Added /api/v1/blocks-extras endpoint

nymkappa:
This PR adds two news endpoints `/api/v1/blocks-extras` and `/api/v1/blocks-extras/:height`. This endpoint is not yet used anywhere in the front end, and it does not replace any existing node backend endpoint.

This PR adds two new optional field to the `ExtendedBlock` interface. `pool` and `coinbaseHex`. Those fields are only available in the response if the network is `bitcoin`, `testnet` or `signet` AND `config.database.enabled` flag is set to `true`.

**Before testing each scenarios** make sure to set your `schema_version` to 4 in the database, delete your blocks db table content, and wipe the `cache/` folder. Then restart the node backend.

Feel free to test other scenarios as well as I'm limited locally.

Therefore, three scenarios must be tested (`curl localhost:4200/api/v1/blocks-extras | jq`)
* First scenario: `config.database.enabled` set to false
  * Should only return the following fields in the `extras` sub object, with possible zeroed values:
    * `medianFee`, `feeRange`, `reward`, `coinbaseTx`, `matchRate`
* Second scenario: `config.database.enabled` set to true
  * Should return the following fields in the `extras` sub object, with possible zeroed values:
    * `medianFee`, `feeRange`, `reward`, `coinbaseTx`, `matchRate`, `pool.name` `pool.id` and ~`coinbaseHex`~
* Third scenario: Liquid or Bisq
  * Should behave like in the first scenario, but it's unclear is this new API will be used for non Bitcoin networks 

Since new fields are added to the block data stucture, the cached blocks in the node backend are also impacted, therefore it is still possible that the front end may be impacted (it should not of course).
Things to double check are:
* Blocks hyperlink and data in the dashboard
* Arrow right/left action in the dashboard
* Blocks hyperlink and data in the /blocks list page
* Recent blocks detailed page in /block page


nymkappa:
Quick benchmark regarding the block indexing process. It took 22961 seconds (~6 hours 22 minutes) to index all blocks and their coinbase transaction on my machine at an average of 31 blocks per seconds.

Benchmark hardware specs:
* Intel(R) Core(TM) i5-7300HQ CPU @ 2.50GHz (6144 KB cache)
* 16GB memory
* SSD drive

softsimon (approved):
utACK

wiz (approved):
Tested ACK @ v2.4.0-dev [ba36ab31]

PR #1239, opened 2022-02-09
Refactor docs

hunicus:
This PR is a redo of #1036, which was closed since it got out-of-date with the Liquid and Liquid testnet updates.

Benefits:
- Move all endpoint data to one file to make updating and adding new endpoints less of a pain
- Allow for descriptions that vary by network, which is a need I encountered when auditing docs for accuracy

Todo:
- Audit docs for accuracy (this new structure makes that much easier)

I really wanted to get rid of a lot of the 'wrapper' functions, empty code sample items, and other things but those fixes weren't obvious to me so I leave them for future PRs. I think these changes at least make the docs more manageable for all involved and will allow me to get on with fixing higher-priority issues like accuracy, layout, etc.


frontend/src/app/components/docs/api-docs-data.ts
```diff
@@ -0,0 +1,4821 @@
...
+    commonJS: [],
+    curl: [],
+    response: ``,
+  },
+  codeSampleBisq: {
+    esModule: [],
+    commonJS: [],
+    curl: [],
+    response: ''
+  },
```
nymkappa: Is there any way to have default values automatically set when rendering instead of having those empty objects everywhere? That would cleanup lot of unnecessary code in this file.

PR #1245, opened 2022-02-11
Display top featured assets on Liquid dashboard

softsimon:
Screen Shot 2022-02-12 at 00 15 11


wiz:
Looks better but after seeing the latest screenshot it looks like the ticker should be after the supply amount

wiz (approved):
Tested ACK @ v2.4.0-dev [[e816f536](https://github.com/mempool/mempool/commit/e816f536)]

PR #1248, opened 2022-02-13
implement /api/mempool for home users (romanz/electrs backend)

antonilol:
what about the `fee_histogram`? 
what is it/what is it used for/is it needed?


softsimon:
fee_histogram is Esplora specific fee estimation thing, we might not need it in the nodejs api.

Do we need this api point to exist at all? Maybe adding fee_histogram with empty array to please the interface

antonilol:
i like to make a widget for on my desktop with some mempool info (e.g. vsize, candidate block median fee) and this api had the information with HTTP (i know i can get mempool info via websocket) but didnt work on my local hosted mempool.

softsimon (approved):
tested ACK

PR #1250, opened 2022-02-14
Provide a way to completely disable block indexing and mining menu

nymkappa:
* Added a new frontend config variable `MINING_DASHBOARD`
  * If set to `true` then the mining menu will be shown
  * If set to `false` the previous blocks list menu will be shown
* Created a a backend util function `Commong.indexingEnabled()` to cleanup the existing check code


nymkappa:
@softsimon I've rebased on master

softsimon (approved):
tested ACK

PR #1252, opened 2022-02-15
Remove duplicated tests

nymkappa:
I must have messed up a conflict resolving at some point and I've noticed we have duplicated tests on master. I've reverted the bogus commit and re-applied only the fix that were supposed to be pushed.


softsimon (approved):
utACK

PR #1253, opened 2022-02-16
Index more block data

nymkappa:
This PR adds the following fields to the `blocks` table. Also, block indexing save this new data in the database.
* `version`: `integer unsigned`
* `bits`: `integer unsigned`
* `nonce`: `bigint unsigned`
* `merkle_root`: `varchar(65)`
* `previous_block_hash`: `varchar(65)`

It also set proper data type for existing fiels:
* `height`: `int(11)` => `integer unsigned`
* `tx_count`: `int(11)` => `smallint unsigned`
* `size`: `int(11)` => `integer unsigned`
* `weight`: `int(11)` => `integer unsigned`
* `difficulty`: `bigint(20)` => `double`

Database schema is incremented to version `6`. If you have trouble running the migration, let me know, or if you don't mind loosing currently indexed data:
* Drop `blocks` table
* Drop `pools` table
* Set `state.schema_version = 2`
* Re-start node backend

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -6,7 +6,7 @@ import logger from '../logger';
 const sleep = (ms: number) => new Promise(res => setTimeout(res, ms));
 
 class DatabaseMigration {
-  private static currentVersion = 5;
+  private static currentVersion = 6;
   private queryTimeout = 120000;
   private statisticsAddedIndexed = false;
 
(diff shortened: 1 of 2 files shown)


backend/src/api/database-migration.ts
```diff
@@ -90,6 +91,29 @@ class DatabaseMigration {
         await this.$executeQuery(connection, 'DROP table IF EXISTS blocks;');
         await this.$executeQuery(connection, this.getCreateBlocksTableQuery(), await this.$checkIfTableExists('blocks'));
       }
+      if (databaseSchemaVersion < 5 && isBitcoin === true) {
```
nymkappa, line 94: Note that I moved this query here because it's an `ALTER` and therefore cannot be part of a `TRANSACTION`

wiz:
Screen Shot 2022-02-16 at 15 32 44

wiz (approved):
Tested ACK @ v2.4.0-dev [0c1fa2b4]

softsimon (approved):
utACK

PR #1257, opened 2022-02-17
Change pool ranking pie chart colors

nymkappa:
This mini PR replaces the default apache echarts pie color with our fee range colors

image


wiz (approved):
utACK

PR #1258, opened 2022-02-17
When blocks need re-indexing, truncate the table

nymkappa:
Mini PR to delete all data in the blocks table when the table schema is updated (database schema version 5 and 6).


wiz (approved):
utACK

PR #1268, opened 2022-02-20
Inline all Enterprise Sponsor logos on About page

wiz:


softsimon (approved):
tested ACK

wiz:
Thanks for the CSS fix

PR #1269, opened 2022-02-21
Created hashrate chart component

nymkappa:
image

This PR adds a new indexing loop as well as a new hashrate chart component. We generate one data point per day (as it does not make much sense to have lower resolution due to the luck factor in mining). The indexing starts after the block indexing is completed. It will go back in time as far as we have blocks in our database. It cannot be disabled right now but does not require much resources nor time to run. A pi should be able to generate the data points in a few hours.

Also, it will generate one new data point every 24 hours as soon as the node backend is running. This should be discussed in more details in the future (we could use a cronjob instead for example).

**Note**

Once the hashrate indexing is complete, we set the current timestamp into `state.last_hashrates_indexing` table. This is used to only run the loop once every 24 hours. If you need to re-index:
* `npm run start --reindex` will erase all blocks and hashrate indexed data

**Performances**

On my laptop, I'm generating the daily hashrate data (for the whole network) at ~3 days/sec. It will be slower once we index hashrate for individual mining pools as well.

### DIFF
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -164,12 +164,17 @@ class Server {
     }
   }
 
-  runIndexingWhenReady() {
+  async runIndexingWhenReady() {
     if (!Common.indexingEnabled() || mempool.hasPriority()) {
       return;
     }
-    blocks.$generateBlockDatabase();
-    mining.$generateNetworkHashrateHistory();
+
+    try {
+      await blocks.$generateBlockDatabase();
+      await mining.$generateNetworkHashrateHistory();
+    } catch (e) {
+      logger.info(`Unable to run indexing right now, trying again later. ` + e);
+    }
   }
 
   setUpWebsocketHandling() {
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -173,7 +173,7 @@ class Server {
       await blocks.$generateBlockDatabase();
       await mining.$generateNetworkHashrateHistory();
     } catch (e) {
-      logger.info(`Unable to run indexing right now, trying again later. ` + e);
+      logger.err(`Unable to run indexing right now, trying again later. ` + e);
     }
   }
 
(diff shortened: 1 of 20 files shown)


wiz:
> If you need to re-index hashrates:
> Delete all rows in hashrates table
> Set statelast_hashrates_indexing to NULL or 0

There should simply be a "reindex" featured called via command line option or other privileged way, we should never have to hack the DB and run manual queries like this

nymkappa:
> > If you need to re-index hashrates:
> > Delete all rows in hashrates table
> > Set statelast_hashrates_indexing to NULL or 0
> 
> There should simply be a "reindex" featured called via command line option or other privileged way, we should never have to hack the DB and run manual queries like this

Fixed in https://github.com/mempool/mempool/pull/1269/commits/6436095a7dfffafd46a55e238c2281edaf1c2836.
You can use `npm run start --reindex` and all indexed data will be truncated and re-indexed.

softsimon:
Could you add `[routerLinkActiveOptions]="{exact: true}"` to the Mining dashboard menu link in master-page, otherwise it is lid up when we are on sub pages like hash rate

nymkappa:
> Could you add `[routerLinkActiveOptions]="{exact: true}"` to the Mining dashboard menu link in master-page, otherwise it is lid up when we are on sub pages like hash rate

Fixed in https://github.com/mempool/mempool/pull/1269/commits/413cf3ccaafa68c7c2f40250d051b6ac5feaea87

backend/src/index.ts
```diff
@@ -164,12 +164,17 @@ class Server {
...
       return;
     }
-    blocks.$generateBlockDatabase();
-    mining.$generateNetworkHashrateHistory();
+
+    try {
+      await blocks.$generateBlockDatabase();
+      await mining.$generateNetworkHashrateHistory();
+    } catch (e) {
+      logger.info(`Unable to run indexing right now, trying again later. ` + e);
```
wiz, line 176: since this is an error this should be an ERR priority syslog

wiz (approved):
Tested ACK @ v2.4.0-dev [[976017db](https://github.com/mempool/mempool/commit/976017db)]

PR #1272, opened 2022-02-21
Tweak new charts design

nymkappa:
- I applied the existing tooltip color and style to all the new mining charts
- Added fancy coloring to the hashrate and difficulty chart (kinda experimental, we can play around with color palettes here)
- Added data zoom component to the difficulty chart for consistency

### Widget
image
### Full page
image

### Widget
image
### Full page
image

### DIFF
--- a/backend/src/repositories/BlocksRepository.ts
+++ b/backend/src/repositories/BlocksRepository.ts
@@ -274,7 +274,7 @@ class BlocksRepository {
     }
 
     query += ` GROUP BY difficulty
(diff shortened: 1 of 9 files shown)


nymkappa:
Note: The marker color in the tooltip does not match the chart color, need to fix it somehow (show the correct color or use a fixed color or remove it)

wiz (approved):
Tested ACK @ v2.4.0-dev [[e770520f](https://github.com/mempool/mempool/commit/e770520f)]

PR #1274, opened 2022-02-22
Show block reward in the mining dashboard

nymkappa:
image

## UX issue ?

It feels weird that when you click on a block from the mining dashboard, the display changes. I feel like we should also show the mining reward if the previous page was the mining dashboard. What do you think?

https://user-images.githubusercontent.com/9780671/155051130-32c111c1-fe3f-4d8e-a9ed-46c286cd2115.mp4


wiz (approved):
Tested ACK @ v2.4.0-dev [[e748775f](https://github.com/mempool/mempool/commit/e748775f)]

This works well, but of course further work on the UI is needed, as kappa said we should probably keep the block reward BTC displayed when you click on the block, etc. - need to consider this further. Also curious to hear what @softsimon suggests to further improve the UI design.

PR #1275, opened 2022-02-22
Merge hashrate and difficulty components - Cleanup mining design overall

nymkappa:
### Hashrate and difficulty
* Hashrate and difficulty components have been merged at `/mining/hashrate` (difficulty component has been deleted completely)
  * You can show/hide graph by clicking on the legend
  * Tooltip  show both data value at the same time
  * Data zoom is still available
  * **Note**: Performances are not great when showing `All` . Will need to improve this in the future
  * Difficulty adjustment history table is shown below the "full screen" chart

### Mining dashboard
* Click on widget title to open the detailed sub page
* Make widgets smallers
* Got rid of the extra space
* Tweak the pie chart to make it fit nicely in different layouts

### Desktop
image
### Mobile
image


wiz (approved):
Tested ACK @ v2.4.0-dev [[807ef228](https://github.com/mempool/mempool/commit/807ef228)]

Excellent work, looks great. I'm a little concerned about the performance of the SQL query since it looks quite complex but it seems to work well.

nymkappa:
> Tested ACK @ v2.4.0-dev [[807ef228](https://github.com/mempool/mempool/commit/807ef228)]
> 
> Excellent work, looks great. I'm a little concerned about the performance of the SQL query since it looks quite complex but it seems to work well.

If lower end hardware cannot run it smoothly, we can always index the result in the database. I'm not sure mysql can cache this query natively either.

PR #1277, opened 2022-02-22
Remove block subsidy from mempool blocks

nymkappa:
* Removed block subsidy from mempool blocks
* Show one more decimal to avoid situation where
  * Mined blocks only have 6.25 BTC reward (no fees?)
  * Mempool blocks with quite mempool shows 0.00 BTC for a long time (no fees?)

image


wiz (approved):
Tested ACK @ v2.4.0-dev [[6c5253a7](https://github.com/mempool/mempool/commit/6c5253a7)]

Feels like we will need to make a similar change on the mining pool details page and possibly also the blocks page, to show the fees (without the subsidy included)

PR #1281, opened 2022-02-22
Fixed incoming transactions bar on the status page

### LINKED ISSUE #1106: Status page: Tweal Tx vBytes per second bar for low volume networks
For networks like Signet or Liquid where there's a low volume of transactions, you can't read the actual value in the progress bar:

Screenshot from 2022-01-08 21-58-30

softsimon:
fixes #1106

Deprecated the old "vBytes per second" string and used the same code as on the Dashboard to fix the Text overlap bug.

Screen Shot 2022-02-22 at 17 33 21


wiz (approved):
Tested ACK @ v2.4.0-dev [d05d7f1e]

PR #1285, opened 2022-02-24
Create pools hashrate dominance chart

nymkappa:
* Data schema has been incremented to 8
  * Table `hashrates` is now indexing individual mining pools weekly historical hashrates
    * Re-indexing should be automatic after the migration is completed. If not, try to use the `--reindex=hashrates` option.
    * It takes a couple of hours to re-index hashrates on a performant laptop
* New api endpoint has been added at `/api/v1/mining/hashrate/pools/:interval`
* New component `HashrateChartPools` has been added and can be seen in the mining dashboard as well at `/mining/hashrate/pools`
  * _It only shows the history of the current top 15 mining pools as it is quite challenging to show it for all mining pools at once in terms of data size and performance_  
* Some colors have been hardcoded for today's top mining pools and they are shared between components

image
image


wiz:
> Re-indexing should be automatic after the migration is completed. If not, try to use the --reindex=hashrates option

Re-indexing should be automatic after the migration is completed. If not, it's a bug, please report it as such.

There, fixed it for you ;)

backend/src/repositories/HashratesRepository.ts
```diff
@@ -49,10 +52,43 @@ class HashratesRepository {
...
+        AND hashrates.type = 'weekly'
+        AND pool_id IN (${topPoolsId})`;
+    } else {
+      query += ` WHERE hashrates.type = 'weekly'
+        AND pool_id IN (${topPoolsId})`;
+    }
+
+    query += ` ORDER by hashrate_timestamp, FIELD(pool_id, ${topPoolsId})`;
+
+    console.log(query);
```
softsimon, line 79: Remove `console.log`s

nymkappa: Fixed in https://github.com/mempool/mempool/pull/1285/commits/434b60ef8b95be505ebf0238ad47413ddb87fd9f

softsimon (approved):
utACK

PR #1286, opened 2022-02-25
Change blockchain blocks bottom color based on total block fees

nymkappa:
An example using 0.195 BTC as the maximum block total fees (in this PR, the maximum is defined as 5 BTC). The higher the block total fees amount, the brighter the blue bottom color becomes.

image


wiz:
Let's hold off on this PR for now until we have more time to play with the colors and stuff

Issue #1296, opened 2022-03-05
Migration doesn't work from a clean slate


softsimon:
```
Mar  5 12:45:56 [75272] INFO: Database connection established.
/Users/simon/repos/mempool/backend/dist/index.js:92
                throw new Error(e instanceof Error ? e.message : 'Error');
                      ^

Error: Table 'mempool.state' doesn't exist
    at Server.startServer (/Users/simon/repos/mempool/backend/dist/index.js:92:23)
```

Cause:

`        await this.$resetHashratesIndexingState();`
Is executed before
`        await databaseMigration.$initializeOrMigrateDatabase();`
when the state table hasn't been created.

PR #1297, opened 2022-03-05
Only reset hashrate state flag after database migration - Fix weekly …

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1296

Also fixes a hashrate indexing issue which was not showing the network hashrate + difficulty using default 1100 blocks indexing config.

image


softsimon (approved):
tested ACK @ 5f6c1c6c

PR #1298, opened 2022-03-05
Make sure to reset hashrates indexing flags upon error

nymkappa:
The diff looks big, but I actually just wrapped the hashrates indexing into a `try {} catch {}` so that `this.hashrateIndexingStarted` is properly set back to `false` upon error, so another attempt will be done in the next main loop.


softsimon (approved):
tested ACK

PR #1300, opened 2022-03-05
Pause block indexing chunk upon error and retry later

nymkappa:
Block indexing indexes blocks by chunk of 10,000. Upon indexing error (if core crashes) the log will be spammed with thousands of messages.
Also reset the `this.blockIndexingStarted = false;` flag upon error so we can retry later.


softsimon (approved):
utACK

PR #1303, opened 2022-03-06
Mining dashboard polishing

nymkappa:
* Split pie chart and pool dominance
  * Removed `(XX%)` in the pie chart label in the widget
  * Added `Pools luck` , `Pools count`, `Blocks` count for the past week 
* Added difficulty adjustment table
* Added next difficulty adjustment estimation + halving countdown
* Top left widget is temporary. While it shows data based on the last 8 blocks, we may want to display more useful info in the future in there.
* Re-added the `View more` links for now, we will iterate on those later
* Disabled mobile chart interaction in the dashboard

image


softsimon (approved):
tested ACK ... We'll polish the loading indications later

PR #1304, opened 2022-03-06
Address page highlight and transfer value

### LINKED ISSUE #1074: highlight address in transactions
### Description
on mempool.space/address/`address`, highlight `address` in every transaction so it is more clear which transactions sent something to that address or something was sent from that address

#### Additional info

example:

image

i know this color is too bright choose another one

### LINKED ISSUE #1287: On the address page, the total transferred button should reflect the address
If we are watching an address for example `bc1qkd5az2ml7dk5j5h672yhxmhmxe9tuf97j39fm6`

We know here that the 81 BTC is just change. 

It would be more informative if every TX show the sum actually deposited or withdrawed from the address in question. In this same button, or a new button in a different color to not cause confusion between different transaction pages.

image

### LINKED ISSUE #1294: Highlight the receive address when looking at an address
People might want to check the amount they get from a receive transaction.

image

softsimon:
* Highlight the current address
* Display the value transferred instead of the total when viewing an address

fixes #1294
fixes #1074
fixes #1287

I am not totally pleased with the use of colors. And I didn't want to use a blue button that will confuse with the "total amount", so would like to hear some feedback.

Screen Shot 2022-03-06 at 13 48 21

### DIFF
--- a/frontend/src/app/components/address/address.component.html
+++ b/frontend/src/app/components/address/address.component.html
@@ -62,7 +62,7 @@
       </h2>
     </div>
 
-    <app-transactions-list [transactions]="transactions" [showConfirmations]="true" (loadMore)="loadMore()"></app-transactions-list>
+    <app-transactions-list [transactions]="transactions" [showConfirmations]="true" [address]="address.address" (loadMore)="loadMore()"></app-transactions-list>
 
     <div class="text-center">
       <ng-template [ngIf]="isLoadingTransactions">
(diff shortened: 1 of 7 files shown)


nymkappa:
Address highlight suggestion: No background, but bold font instead
image

nymkappa:
The darker color of the confirmations / amount button at the bottom of a tx should only be enabled on hover. Right now if you click on it, it will stay darker when the mouse leave the button IMO.
image

nymkappa:
For the received vs sent amount tag, I agree it would be good for those to have their own style. A suggestion, we can play around with the background/border combo to keep the same color code but have more variety:
image
image

PR #1318, opened 2022-03-08
Replace by fee storage

### LINKED ISSUE #1222: Keep RBF-replaced transactions for a while
Current behavior:

1. Track transaction
2. RBF it
3. A nice dialog shows up guiding the user to the new TXID
4. Reload the page

Result:
"Transaction not found - Waiting for it to appear in the mempool...". 

Expected behavior:
Display the same nice RBF info dialog that guides the user to the new TXID

Solution:
When a transaction is replaced, store the TXID of the old TX being replaced in a short term memory
When user trying to load/track, the RBF notification will now still remain and guide the user to the new TXID

softsimon:
fixes #1222

Added a temporary key value index storage for all replaced transactions. 

Now this dialog will be displayed instead of "Waiting for Transaction" when visiting replaced transactions.

Screen Shot 2022-03-08 at 14 50 18

### DIFF
--- a/backend/src/api/mempool.ts
+++ b/backend/src/api/mempool.ts
@@ -8,6 +8,7 @@ import { IBitcoinApi } from './bitcoin/bitcoin-api.interface';
 import loadingIndicators from './loading-indicators';
 import bitcoinClient from './bitcoin/bitcoin-client';
 import bitcoinSecondClient from './bitcoin/bitcoin-second-client';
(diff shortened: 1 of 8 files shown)


nymkappa (approved):
Tested ACK @https://github.com/mempool/mempool/commit/8267344cdc0239425a4750cc19ca5c69101d3d91

PR #1320, opened 2022-03-08
regtest fixes

antonilol:
#1165 but removed the separate network + routes


frontend/src/app/components/block/block.component.ts
```diff
@@ -217,12 +217,8 @@ export class BlockComponent implements OnInit, OnDestroy {
       this.blockSubsidy = 0;
       return;
     }
-    this.blockSubsidy = 50;
-    let halvenings = Math.floor(this.block.height / 210000);
-    while (halvenings > 0) {
-      this.blockSubsidy = this.blockSubsidy / 2;
-      halvenings--;
-    }
+    const halvings = Math.floor(this.block.height / (this.network === 'regtest' ? 150 : 210000));
```
softsimon, line 220: `this.network` will never be `regtest`

backend/src/api/bitcoin/bitcoin-api.ts
```diff
@@ -41,7 +41,12 @@ class BitcoinApi implements AbstractBitcoinApi {
 
   $getBlockHeightTip(): Promise<number> {
     return this.bitcoindClient.getChainTips()
-      .then((result: IBitcoinApi.ChainTips[]) => result[0].height);
+      .then((result: IBitcoinApi.ChainTips[]) => {
+        const activeTip = result.find(tip => tip.status === 'active');
+        if (activeTip) {
+          return activeTip.height;
+        }
```
softsimon, line 48: Does `if (activeTip)` mean that it might be undefined? In that case we should return a default value.

frontend/src/app/components/block/block.component.ts
antonilol: oops left that in there

backend/src/api/bitcoin/bitcoin-api.ts
antonilol: bitcoin core always has an active chaintip, right?
my IDE said it can be undefined because it doesn't know that

antonilol: i can add a non null assertion
like `return activeTip!.height;`

softsimon: Ok, I am just making sure to reduce any possibility of error because I am not sure if there will always be an active tip or not.

antonilol: src/rpc/blockchain.cpp:1783
`if (active_chain.Contains(block)) { status = "active" }` 

always 1 will be active since a block must be in a chain

backend/src/api/blocks.ts
```diff
@@ -268,17 +268,19 @@ class Blocks {
         this.lastDifficultyAdjustmentTime = block.timestamp;
         this.currentDifficulty = block.difficulty;
 
-        const previousPeriodBlockHash = await bitcoinApi.$getBlockHash(blockHeightTip - heightDiff - 2016);
-        const previousPeriodBlock = await bitcoinApi.$getBlock(previousPeriodBlockHash);
-        this.previousDifficultyRetarget = (block.difficulty - previousPeriodBlock.difficulty) / previousPeriodBlock.difficulty * 100;
-        logger.debug(`Initial difficulty adjustment data set.`);
+        if (blockHeightTip - heightDiff - 2016 >= 0) {
```
nymkappa, line 271: Is this new check to avoid computing the difficulty adjustment for the first 2016 blocks? If that's the case, it would be clearer to just have:
> `if (blockHeightTip > 2016)`

antonilol: Yes, mainly because $getBlockHash needs a positive integer, so i just copied that argument from there to compare it to 0, for clarity.

Is `blockHeightTip >= 2016` also correct?

nymkappa:
The arrow display of the genesis block is broken

image

backend/src/api/blocks.ts
nymkappa: > Is blockHeightTip >= 2016 also correct?
Yes I think it's correct

antonilol: ok, changing this

frontend/src/app/dashboard/dashboard.component.ts
```diff
@@ -153,6 +156,9 @@ export class DashboardComponent implements OnInit {
     this.transactions$ = this.stateService.transactions$
       .pipe(
         scan((acc, tx) => {
+          if (acc.find((t) => t.txid == tx.txid)) {
```
softsimon, line 159: What does this fix again? Was there a problem where transaction IDs got duplicated in the latest transactions list? Why would there be duplicate txids?

softsimon, line 159: What does this fix again? Was there a problem where transaction IDs got duplicated in the latest transactions list? Why would there be duplicate txids?

antonilol: If you have less than 6 txs in that list and switch to a network with the backend offline, the txs stay there. When switching back, the txs were appended again. This was also for the "recent blocks".

antonilol: 1st one of https://github.com/mempool/mempool/pull/1165#issuecomment-1017808054

softsimon (approved):

Issue #1157, opened 2022-01-18
Refactor: Move difficulty adjustment code to the backend


softsimon:
Why is the difficulty adjustment code in the frontend at all?
Shouldn't it be in the backend and exposed by the websocket and REST APIs ?

_Originally posted by @wiz in https://github.com/mempool/mempool/issues/1152#issuecomment-1014185002_

PR #1322, opened 2022-03-08
fix and improve block time estimations

### LINKED ISSUE #1051: testnet block mine time estimation
### Description
on testnet, the difficulty is set to 1 when no block is mined in 20 minutes (to keep it going). when the last block was mined 12 minutes ago, you can know for certain the next one will be mined in at least 8 minutes

### Expected behaviour
mempool block mine time estimate says 'In ~8 minutes'

### Actual behaviour
In ~12 minutes

### Screenshots
image

antonilol:
image

testnet has been going like this for at least 3000 blocks (around 1 per minute) but the difficulty adjustment panel says 1694 blocks will be mined in 1 minute

fixes this ^
fixes #1051

and set the mempool block time estimations relative to the chaintip instead of `now`, because it is statistically more likely for a block to be mined when the tip is mined 10 minutes ago than just now

Questions and comments are welcome!

### DIFF
--- a/frontend/src/app/components/difficulty/difficulty.component.ts
+++ b/frontend/src/app/components/difficulty/difficulty.component.ts
@@ -66,14 +66,7 @@ export class DifficultyComponent implements OnInit {
             }
           }
 
-          const timeAvgDiff = change * 0.1;
-
-          let timeAvgMins = 10;
-          if (timeAvgDiff > 0) {
-            timeAvgMins -= Math.abs(timeAvgDiff);
-          } else {
-            timeAvgMins += Math.abs(timeAvgDiff);
-          }
+          const timeAvgMins = blocksInEpoch ? diff / blocksInEpoch / 60 : 10;
 
           const timeAvg = timeAvgMins.toFixed(0);
           const remainingTime = (remainingBlocks * timeAvgMins * 60 * 1000) + (now * 1000);
--- a/frontend/src/app/components/mempool-blocks/mempool-blocks.component.html
+++ b/frontend/src/app/components/mempool-blocks/mempool-blocks.component.html
@@ -26,7 +26,7 @@
                 <app-time-until [time]="(1 * i) + now + 61000" [fastRender]="false" [fixedRender]="true"></app-time-until>
               </ng-template>
               <ng-template #timeDiffMainnet>
-                <app-time-until [time]="(timeAvg * i) + now + timeAvg" [fastRender]="false" [fixedRender]="true" [forceFloorOnTimeIntervals]="['hour']"></app-time-until>
+                <app-time-until [time]="(timeAvg * i) + lastBlockTime + timeAvg" [fastRender]="false" [fixedRender]="true" [forceFloorOnTimeIntervals]="['hour']"></app-time-until>
               </ng-template>
             </div>
             <ng-template #mergedBlock>
--- a/frontend/src/app/components/mempool-blocks/mempool-blocks.component.html
+++ b/frontend/src/app/components/mempool-blocks/mempool-blocks.component.html
@@ -26,7 +26,7 @@
                 <app-time-until [time]="(1 * i) + now + 61000" [fastRender]="false" [fixedRender]="true"></app-time-until>
               </ng-template>
               <ng-template #timeDiffMainnet>
-                <app-time-until [time]="(timeAvg * i) + lastBlockTime + timeAvg" [fastRender]="false" [fixedRender]="true" [forceFloorOnTimeIntervals]="['hour']"></app-time-until>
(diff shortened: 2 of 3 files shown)


nymkappa:
> and set the mempool block time estimations relative to the chaintip instead of now, because it is statistically more likely for a block to be mined when the tip is mined 10 minutes ago than just now

I'm not sure I understand your point. The probability never changes, it's always ~10 minutes (considering the next difficulty change is estimated to be ~0%).

It does not matter if the last block was mined 1 minute ago or 1 hour ago, the probability never changes. That's a fundamental property in proof of work.

I believe this https://github.com/mempool/mempool/pull/1322/files#diff-ac766b5089eb75889ef48e31a242c62f42d8cd0292988e4567b7a350442502afR29 is not correct.

nymkappa:
Could you explain your [reasoning behind change the way `timeAvg` is computed](https://github.com/mempool/mempool/pull/1322/files#diff-24419c1fae319a1ebec9c0c514a5080796d8dbae69f60df93df1031d05a0f234L136-L149) as well? it seems like a big change so it shoudl be clarified.

I think the change for testnet is fine, but it should not change the whole logic for computing the next block time.

nymkappa:
I think the difficulty estimation does not work properly on master branch for regtest.  Before the first difficulty adjustment estimation (first few blocks), it shows `2 weeks` before next difficulty adjustment
image

But as soon as the adjustment is first estimated, even though it is expected the difficulty to drop, the time increases. It's opposite of what it's supposed to show.
image

antonilol:
> > and set the mempool block time estimations relative to the chaintip instead of now, because it is statistically more likely for a block to be mined when the tip is mined 10 minutes ago than just now
> 
> I'm not sure I understand your point. The probability never changes, it's always ~10 minutes (considering the next difficulty change is estimated to be ~0%).
> 
> It does not matter if the last block was mined 1 minute ago or 1 hour ago, the probability never changes. That's a fundamental property in proof of work.
> 
> I believe this https://github.com/mempool/mempool/pull/1322/files#diff-ac766b5089eb75889ef48e31a242c62f42d8cd0292988e4567b7a350442502afR29 is not correct.

the *probability* doesn't change, but it is *statistically* more likely for a block to mined then.

antonilol:
> Could you explain your [reasoning behind change the way `timeAvg` is computed](https://github.com/mempool/mempool/pull/1322/files#diff-24419c1fae319a1ebec9c0c514a5080796d8dbae69f60df93df1031d05a0f234L136-L149) as well? it seems like a big change so it shoudl be clarified.
> 
> I think the change for testnet is fine, but it should not change the whole logic for computing the next block time.

timeAvg was timeAvgMins but it was used as timeAvgMins * 60, so i changed it to seconds because other variables are in seconds too. 

timeAvg is the average time between 2 blocks in seconds, so the unit is *time per block* and is calculated as time / blocks (`diff / blocksInEpoch`)

previously, timeAvgMins could vary from 10+7.5 to 10-30, resulting in a negative estimation when when the estimated adjustment is greater than 200%

nymkappa:
> > > and set the mempool block time estimations relative to the chaintip instead of now, because it is statistically more likely for a block to be mined when the tip is mined 10 minutes ago than just now
> > 
> > 
> > I'm not sure I understand your point. The probability never changes, it's always ~10 minutes (considering the next difficulty change is estimated to be ~0%).
> > It does not matter if the last block was mined 1 minute ago or 1 hour ago, the probability never changes. That's a fundamental property in proof of work.
> > I believe this https://github.com/mempool/mempool/pull/1322/files#diff-ac766b5089eb75889ef48e31a242c62f42d8cd0292988e4567b7a350442502afR29 is not correct.
> 
> the _probability_ doesn't change, but it is _statistically_ more likely for a block to mined then.

No. Every mined blocks are completely independant events. This is what mainnet mempool blocks would look like with your change, this is not correct.

image

Current implementation is correct

image

antonilol:
> Every mined blocks are completely independent events.

True

image

and i agree this doesn't look ok

antonilol:
`timeOffset` is there for testnet

nymkappa:
Ignore my last comment, actually the difficulty adjustment calculation must be refactored before we try to fix it for testnet/regtest/etc.

There is an open issue for that, @softsimon will work on it when he can https://github.com/mempool/mempool/issues/1157

Can you close this PR and make a new one for fixing the testnet next block time estimation only? (`timeOffset`)

antonilol:
the timestamp of the regtest genesis block is `2011-02-03 00:16 (11 years ago)`, so 276 blocks were mined in 11 years. the other 1740 would be mined in 11/276*1740=69.3 years. so it kinda makes sense

PR #1324, opened 2022-03-09
Fix skeleton for difficulty adjustment table

nymkappa:
### Before

Screenshot from 2022-03-09 11-44-06

### After

image


softsimon (approved):
tested ACK

PR #1326, opened 2022-03-09
Fix pool detail page on Firefox

nymkappa:
#### Does not work on Firefox
image
#### Chromium layout
image

#### Desktop

image

#### Mobile
image


softsimon (approved):
tested ACK

PR #1327, opened 2022-03-09
Always show scrolbar to avoid horizontal jumpy UI

nymkappa:
Need to test on different browser/OS


softsimon (approved):
tested ACK on Mac OS with Firefox, Chrome and Safari ... There are no visible changes or side effects.

PR #1330, opened 2022-03-09
Index more data using getblockstats core RPC

nymkappa:
Now indexing the following additional data:

- blockExtended.extras.feeRange
- blockExtended.extras.totalFees
- blockExtended.extras.avgFee
- blockExtended.extras.avgFeeRate
- blockExtended.extras.maxFee
- blockExtended.extras.maxFeeRate
- blockExtended.extras.minFee
- blockExtended.extras.minFeeRate
- blockExtended.extras.subsidy
- blockExtended.extras.medianFeeValue `// This is the actual median fee amount, different from our own medianFee field which represent the median fee rate and is not available during back indexing`

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -218,6 +218,7 @@ class Blocks {
             break;
           }
           ++indexedThisRun;
+          ++totaIndexed;
           const elapsedSeconds = Math.max(1, Math.round((new Date().getTime() / 1000) - timer));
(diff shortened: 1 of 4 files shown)


softsimon (changes requested):
Store minimum and maximum in the feeRange.
Ignore median fee value, remove that column.
Store the 50% percentile value from feeRange in the median_fee table.

softsimon (approved):
tested ACK

PR #1331, opened 2022-03-09
Show current indexing progress in charts without data

nymkappa:
This PR simply show the oldest block we've currently indexed, so the user has a rough idea of where the indexing process is. The date is refreshed every 60 seconds (_considering the query isn't cached)_.

https://user-images.githubusercontent.com/9780671/157534861-c42ef447-8c27-4b63-925f-a273a9f390ec.mp4


katesalazar:
do every node operator want this info to be available?

(not for this PR)
can you track indexing progress and print some
estimated time to indexed?

nymkappa:
The currently WIP mining dashboard can be disabled through the config file.
We'll provide a better UI for tracking the indexing progress in the near future yes.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

softsimon (approved):
tested ACK

Works for now but we should move to a websocket push based %-loading indicator https://github.com/mempool/mempool/issues/1334

PR #1333, opened 2022-03-10
Index weekly hashrates using last Monday midnight - Fix charts tooltip

nymkappa:
Weekly hashrates are currently indexing from last midnight, which means that everytime we restart the server, we may add a new entry even though the most recent one is younger than 7 days.

This PR fixes this issue by doing weekly indexing from last monday midnight.

I also fixed the tooltip labeling formatting for hashrates charts using toLocaleDateString function.


softsimon (approved):
utACK

Issue #1049, opened 2021-12-27
Mining Dashboard (Phase 1): Frontend support for new blocks and pool APIs


softsimon:
### Main dashboard
1. Consume the new websocket mining data 
2. When loading an old block, fetch the new mining data replacing the local mining pool JSON

### New mining dashboard on /blocks

1. List latest blocks using new API endpoint with all old values including new data like new `pool name` with icon, and `coinbase` 

<img width="1243" alt="Screen Shot 2021-12-27 at 14 09 21" src="https://user-images.githubusercontent.com/8561090/147461335-67577fb7-6586-400f-a7f6-19b0cad9de95.png">

softsimon:
Closed with https://github.com/mempool/mempool/pull/1335

PR #1335, opened 2022-03-11
Create new /mining/blocks page

nymkappa:
First implementation of https://github.com/mempool/mempool/issues/1049

Consumes the `/blocks-extra` API endpoint to create a new blocks list page at `/mining/blocks`. There is some work to do on the pagination navigation (you can't "go to page" etc...) and also filters should be added in the future.

image

image


softsimon (approved):
tested ACK at a893e873]

PR #1341, opened 2022-03-11
Pin actions by SHA and set permissions for workflow

naveensrinivasan:
- Pinned dependencies https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies
- Restricting permissions for github actions https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions


knorrium (approved):

naveensrinivasan:
A friendly ping. What does it need to happen to merge this?

PR #1351, opened 2022-03-12
Fix duplicate hashrate data points in "difficulty vs hashrate" chart

nymkappa:
* Increments DB schema version to 14 to fix unique constraint by making pool_id = 0 instead of NULL
* Use wrapper around getBlock() to call Bitcoin Core RPC instead of esplora backend, due to https://github.com/Blockstream/esplora/issues/370
* Use wrapper around DB.getConnection() so we can always set session time zone to UTC upon connecting

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -6,7 +6,7 @@ import logger from '../logger';
 const sleep = (ms: number) => new Promise(res => setTimeout(res, ms));
 
 class DatabaseMigration {
-  private static currentVersion = 13;
+  private static currentVersion = 14;
   private queryTimeout = 120000;
   private statisticsAddedIndexed = false;
 
@@ -77,7 +77,7 @@ class DatabaseMigration {
     await this.$setStatisticsAddedIndexedFlag(databaseSchemaVersion);
 
     const isBitcoin = ['mainnet', 'testnet', 'signet'].includes(config.MEMPOOL.NETWORK);
-    const connection = await DB.pool.getConnection();
+    const connection = await DB.getConnection();
     try {
       await this.$executeQuery(connection, this.getCreateElementsTableQuery(), await this.$checkIfTableExists('elements_pegs'));
       await this.$executeQuery(connection, this.getCreateStatisticsQuery(), await this.$checkIfTableExists('statistics'));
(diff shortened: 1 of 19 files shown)


wiz (changes requested):
```
Mar 12 17:01:08 [65317] DEBUG: MIGRATIONS: Running migrations
/mempool/mainnet/backend/node_modules/mysql2/promise.js:93
    const localErr = new Error();
                     ^

Error: Access denied; you need (at least one of) the SUPER privilege(s) for this operation
    at PromisePoolConnection.query (/mempool/mainnet/backend/node_modules/mysql2/promise.js:93:22)
    at checkDbConnection (/mempool/mainnet/backend/dist/database.js:23:20)
    at async Server.startServer (/mempool/mainnet/backend/dist/index.js:79:13) {
  code: 'ER_SPECIFIC_ACCESS_DENIED_ERROR',
  errno: 1227,
  sql: "SET GLOBAL time_zone = '+0:00';",
  sqlState: '42000',
  sqlMessage: 'Access denied; you need (at least one of) the SUPER privilege(s) for this operation'
}
```

backend/src/api/database-migration.ts
```diff
@@ -168,6 +168,19 @@ class DatabaseMigration {
...
+        await this.$executeQuery(connection, 'ALTER TABLE `hashrates` DROP FOREIGN KEY `hashrates_ibfk_1`');
+        await this.$executeQuery(connection, 'ALTER TABLE `hashrates` MODIFY `pool_id` SMALLINT UNSIGNED NOT NULL DEFAULT "0"');
+      }
+
+      if (databaseSchemaVersion < 15 && isBitcoin === true) {
+        logger.warn(`'hashrates' table has been truncated. Re-indexing from scratch.`);
+        await this.$executeQuery(connection, 'TRUNCATE hashrates;'); // Need to re-index
+        await this.$executeQuery(connection, 'ALTER TABLE `hashrates` MODIFY `avg_hashrate` DOUBLE UNSIGNED NOT NULL DEFAULT "0"');
+      }
+
```
wiz, line 183: seems like we should remove this

wiz (changes requested):
few things:
* performance is very slow using core RPC calls instead of esplora, would prefer to have @shesek fix the bug in blockstream/electrs instead of doing this RPC call way for each block
* also noticed that `coinbase_raw` is NULL because we don't get this from esplora and we need to change this PR to use the ascii scriptsig instead

wiz (approved):
Tested ACK @ v2.4.0-dev [[edddf259](https://github.com/mempool/mempool/commit/edddf259)]

Full sync of all blocks took about 8 hours on ninja, needs to be improved before we can ship v2.4

PR #1358, opened 2022-03-13
Increase default INDEXING_BLOCKS_AMOUNT by 10x to 11000 blocks

wiz:
Now that @nymkappa optimized some of the block and hashrate indexing code, this default value takes about 5 minutes to run on our production server, so it should be less than an hour on Raspberry Pi hardware


wiz:
@knorrium can you remind me what other places do we need to add this in, like README etc. ?

PR #1372, opened 2022-03-15
Index blocks using RPC batching

nymkappa:


nymkappa:
Closed because Bitcoin Core cannot handle the multi thread RPC loads. At least we tried... :)

PR #1373, opened 2022-03-15
fix and improve block time predictions

antonilol:
dce775e but then on the backed after refractor #1352

explanation:
https://github.com/mempool/mempool/pull/1322#issue-1163270849 (first part)
https://github.com/mempool/mempool/pull/1322#issuecomment-1062803334

### DIFF
--- a/backend/src/api/difficulty-adjustment.ts
+++ b/backend/src/api/difficulty-adjustment.ts
@@ -22,7 +22,7 @@ class DifficultyAdjustmentApi {
     let difficultyChange = 0;
     if (remainingBlocks < 1870) {
       if (blocksInEpoch > 0) {
-        difficultyChange = (600 / (diff / blocksInEpoch ) - 1) * 100;
+        difficultyChange = (600 / (diff / blocksInEpoch) - 1) * 100;
       }
       if (difficultyChange > 300) {
         difficultyChange = 300;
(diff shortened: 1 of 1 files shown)


backend/src/api/difficulty-adjustment.ts
```diff
@@ -32,22 +32,18 @@ class DifficultyAdjustmentApi {
...
-    let timeAvgMins = 10;
-
-    if (timeAvgDiff > 0) {
-      timeAvgMins -= Math.abs(timeAvgDiff);
-    } else {
-      timeAvgMins += Math.abs(timeAvgDiff);
-    }
-
-    // Testnet difficulty is set to 1 after 20 minutes of no blockSize,
+    // Testnet difficulty is set to 1 after 20 minutes of no blocks,
```
antonilol, line 37: accidental `tab` was here

backend/src/api/difficulty-adjustment.ts
```diff
@@ -32,22 +32,18 @@ class DifficultyAdjustmentApi {
...
     // therefore the time between blocks will always be below 20 minutes (1200s).
     let timeOffset = 0;
-    if (config.MEMPOOL.NETWORK === 'testnet' && now - latestBlock.timestamp + timeAvgMins * 60 > 1200) {
-      timeOffset = -Math.min(now - latestBlock.timestamp, 1200) * 1000;
-      timeAvgMins = 20;
```
antonilol, line 50: just realized `timeAvgMins` will be set to 20 even if it was lower before

softsimon (approved):
tested ACK

PR #1376, opened 2022-03-15
Fix pagination on /mining/blocks

nymkappa:
### Before (can't reach genesis block)

image

### After

image


wiz:
Cool but why does it say Unknown miner? I think we all know who mined the Genesis block :)

nymkappa:
Let's fork and maintain our own pools.json registry so we can show Satoshi's and other know actors in the miners ranking and so on.

PR #1378, opened 2022-03-16
Update nginx.conf

cryptodread:
updated line 147,148 so the conf file works with nginx version: nginx/1.18.0 (Ubuntu)

### DIFF
--- a/nginx.conf
+++ b/nginx.conf
@@ -144,7 +144,8 @@ http {
 	}
 
 	server {
-		listen 127.0.0.1:80;
+		listen 80;
+		server_name 127.0.0.1;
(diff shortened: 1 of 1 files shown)


antonilol:
> I'd prefer to simply improve the documentation that explains how to open this up if the user wants to - everyone has their own specific use cases and binding to all interfaces by default might not be the best idea for security / privacy reasons.

https://github.com/mempool/mempool/pull/1005#issuecomment-991874311

it hasn't been added to the docs though

wiz:
Yeah, I guess you guys are right and enough people have requested it that we could change the default configuration to listen on all hosts - will review this PR later

wiz:
I will make a separate PR to address this issue

PR #1400, opened 2022-03-17
Don't navigate to root when sub network url is wrong.

### LINKED ISSUE #1399: Navigating to a non existent url takes you to the root without switching network
This affects routes to /testnet for example.

softsimon:
When going to for example [mempool.ninja/testnet/foo](url) it should redirect to `/testnet` not `/`

fixes #1399


wiz (approved):
Tested ACK @ v2.4.0-dev [[d1ca9198](https://github.com/mempool/mempool/commit/d1ca9198)]

PR #1401, opened 2022-03-17
Fix some mobile UI issues on mining dashboard

### LINKED ISSUE #1396: Small UX issue on mining dashboard mobile
Bottom text is too close to the bottom

Screen Shot 2022-03-17 at 16 43 03

nymkappa:
Fixes #1396 

image image

image image

image image


nymkappa:
Okay it actually does not adapt well to all mobile display. Will work on making this more responsive.

nymkappa:
Okay I think I made it more stable on mobile. Can you try with you phones and let me know if there are things that are still very broken? (_here I'm talking about "must fix", not pixel perfect stuff 😄, we need actual html/css devs for that because I suck._)

softsimon (approved):
tested ACK @ 2b79d6c9

PR #1417, opened 2022-03-20
Address highlighting tests

knorrium:
This PR adds new tests for the address highlighting feature and a new directive to support `data-cy` test attributes.


softsimon (approved):
utACK

Issue #1405, opened 2022-03-18
Next halving estimate is broken


knorrium:
Off by a few years

![AC37CA65-C459-4184-ACC1-745B55480AF9](https://user-images.githubusercontent.com/100320/159022010-3095b24f-aa33-4787-8dbb-3950b7f12062.jpeg)

PR #1422, opened 2022-03-21
Use 10 minutes avg block time for halving calculation

nymkappa:
Use hard coded 10 minutes average block time for calculating the next halving time, since over long period of time the average should be around 10 minutes.

Fixes https://github.com/mempool/mempool/issues/1405


wiz (approved):
Tested ACK @ v2.4.0-dev [[3e8a34f3](https://github.com/mempool/mempool/commit/3e8a34f3)]

PR #1424, opened 2022-03-21
Set weekly hashrates timestamp to midnight

nymkappa:
Because of early timestamp truncating I was doing in the weekly hashrates indexing loop, all timestamp were off by a bit, instead of being set to exactly midnight (for timestamps, javascript uses milliseconds, mariadb uses seconds).

We now divide javascript timestamps by `/ 1000` only when we need to make a db query, so we avoid rounding error during the indexing loop.

### How to test

* Note the shifted timestamp on weekly hashrates
```mysql
MariaDB [mempool]> select hashrate_timestamp from hashrates where type='weekly' limit 1;
+-------------------------+
| MIN(hashrate_timestamp) |
+-------------------------+
| 2022-03-13 23:49:55     |
+-------------------------+
```
* `TRUNCATE hashrates;`
* Restart node backend and wait for hashrates to re-index
* Verify if weekly hashrates are now on midnight 
```mysql
MariaDB [mempool]> select hashrate_timestamp from hashrates where type='weekly' limit 1;
+-------------------------+
| MIN(hashrate_timestamp) |
+-------------------------+
| 2022-03-14 00:00:00     |
+-------------------------+
```

### DIFF
--- a/backend/src/api/mining.ts
+++ b/backend/src/api/mining.ts
@@ -92,11 +92,11 @@ class Mining {
 
       const indexedTimestamp = await HashratesRepository.$getWeeklyHashrateTimestamps();
       const hashrates: any[] = [];
-      const genesisTimestamp = 1231006505; // bitcoin-cli getblock 000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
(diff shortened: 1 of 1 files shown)


wiz:
@nymkappa please rebase onto master and fix merge conflict

wiz (approved):
Tested ACK @ 7ace0cfb

Issue #1387, opened 2022-03-17
Mining dashboard reward stats


softsimon:
Change reward stats to use past 144 blocks.

PR #1427, opened 2022-03-22
More dynamic mining reward 

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1387

* Move reward stats to component 
* Add `/api/v1/mining/reward-stats/{blockCount}` (default to 144 blocks)
  * Added `/api/v1/mining/reward-stats/144` in in the production cache warmer
* Updated labels for a more verbose stat description 
  
image

### DIFF
--- a/backend/src/api/mining.ts
+++ b/backend/src/api/mining.ts
@@ -1,4 +1,4 @@
-import { PoolInfo, PoolStats } from '../mempool.interfaces';
+import { PoolInfo, PoolStats, RewardStats } from '../mempool.interfaces';
 import BlocksRepository from '../repositories/BlocksRepository';
 import PoolsRepository from '../repositories/PoolsRepository';
 import HashratesRepository from '../repositories/HashratesRepository';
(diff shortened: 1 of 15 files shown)


nymkappa:
@wiz Do you think we should have `blockCount` as a dynamic parameter? I don't think we can cache all possible value right that would be a waste of resources.
Also I'm not sure if we should let users query any `blockCount` as it may be heavy on the mysql server.

softsimon:
Looks good, but we might want the data to update if you leave the page open for a long time?

nymkappa:
> Looks good, but we might want the data to update if you leave the page open for a long time?

Oh yeah lol how did I forgot about this. I'll update. Good point!

nymkappa:
> Looks good, but we might want the data to update if you leave the page open for a long time?

Oh yeah lol how did I forgot about this. Good point!
Fixed in https://github.com/mempool/mempool/pull/1427/commits/0bffb312fd37f0b1ac1719e18abd3ee72bb23dc0

softsimon (approved):
utACK

softsimon:
I think the bottom labels are redundant. Stating what is already clear enough.

I would suggest the bottom part should be the dollar equivalent below a line just like on the main dashboard.

Screen Shot 2022-03-23 at 01 47 42

nymkappa:
> I think the bottom labels are redundant. Stating what is already clear enough.
> 
> I would suggest the bottom part should be the dollar equivalent below a line just like on the main dashboard.
> 
> Screen Shot 2022-03-23 at 01 47 42

Fixed in https://github.com/mempool/mempool/pull/1427/commits/dcd50802e456b60a71756fc17f9582c45066765e

https://user-images.githubusercontent.com/9780671/159620024-9b4235b2-33fe-47c2-9173-0a2b20c75f17.mp4

softsimon (approved):
tested ACK @ dcd50802

Issue #1403, opened 2022-03-18
Blocks page: responsive layout gets broken after jumping to the last page


knorrium:
Can’t see all the pagination links and the nav bar menu grows after going to the last page

![4AD803A7-CC1D-4234-82A7-C63BCDB1C8DD](https://user-images.githubusercontent.com/100320/158919607-76f67424-cfa6-4a17-83df-51b2a221dff2.png)

PR #1428, opened 2022-03-22
Fix blocks list mobile layout

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1403 (TM)


knorrium (approved):
Tested ACK @ a384328f

Issue #1276, opened 2022-02-22
Missing rounding in incoming tx chart tooltip


nymkappa:
![image](https://user-images.githubusercontent.com/9780671/155100955-a353405e-3770-44c6-bc9b-01b29da681e3.png)

PR #1429, opened 2022-03-22
Format Transaction vBytes per second (vB/s) tooltip value

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1276

It seems that production database is storing `statistics.vbytes_per_second` into a `double` instead of an `unsigned integer`. In the database migration, it shows that it's supposed to be an `unsigned integer` (see https://github.com/mempool/mempool/blob/master/backend/src/api/database-migration.ts#L379). I guess this is just a legacy table.

To fix the decimal point issue you can either:
* Merge this PR (format in the front end), which also format the number using the user's locale
* Manually change production db `statistics.vbytes_per_second` db field type to `unsigned integer`, which should truncate all values to the nearest integer (loss of precision, non reversible)

### Before

image

### After

image


softsimon (approved):
utACK

PR #1431, opened 2022-03-22
Don't show coinbase tooltip on mining dashboard

nymkappa:
@wiz mentioned this last week, unsure where we stand today on this. Feel free to close if you want to keep the tooltip enabled in the mining dashboard, this PR disables it when we're in widget mode.

image


softsimon (approved):
utACK

PR #1435, opened 2022-03-23
Last hashrate indexing check needs to be in milliseconds

softsimon:


wiz (approved):
Tested ACK, thank you

PR #1436, opened 2022-03-23
Append missing addresses to bisq transactions

softsimon:
The new Bisq JSON data format is missing the addresses so we just use the address data from the actual Bitcoin transaction.

Before
Screen Shot 2022-03-24 at 01 19 40

After
Screen Shot 2022-03-24 at 01 19 45

### DIFF
--- a/frontend/src/app/bisq/bisq-transaction/bisq-transaction.component.html
+++ b/frontend/src/app/bisq/bisq-transaction/bisq-transaction.component.html
@@ -67,14 +67,11 @@
                 </tr>
                 <tr>
                   <td i18n="transaction.fee-per-vbyte|Transaction fee">Fee per vByte</td>
-                  <td *ngIf="!isLoadingTx; else loadingTxFee">
+                  <td>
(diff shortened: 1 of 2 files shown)


wiz (changes requested):
I deployed your PR on 201.fra and changed my /etc/hosts to that IP for bisq.markets but seems still broken - in other words, it works for the TX page but not other places like the blocks page view https://bisq.markets/block/000000000000000000097d1c944dbe937d574235d6f7d9a1f84f89deffb5378f

knorrium (changes requested):
Tested NACK @ 99268b1b

@softsimon testing on 201.fra and I can't load this address https://bisq.markets/address/Bbc1qmefe8wrxy4pg9nvq0c7k5kgrjppd49kcywsrlu - it says invalid Bisq address (also via the API)

Also the block view is missing the full address, i.e: https://bisq.markets/block/000000000000000000052382ef683117553af74f390d5078716194d03fadcb33

softsimon:
Ok this is not the correct solution.

Issue #1448, opened 2022-03-24
Echarts console errors on the new mining dashboard


knorrium:
Load the mining dashboard and echarts will complain about missing attributes:

![Screenshot from 2022-03-24 12-06-04](https://user-images.githubusercontent.com/100320/159992696-dd8acd63-2311-4c4e-b59a-24b373c9198a.png)

PR #1449, opened 2022-03-25
Remove unnecessary echart init option

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1448

### Testing

* Check if warning are gones from pages with echart
* Make sure charts still fill properly their space on pages


softsimon (approved):
tested ACK

Issue #1447, opened 2022-03-24
url in mining pool pie chart must be relative to network


antonilol:
### Description

ye basically the title

#### Version

mempool.ninja v2.4.0-dev [77df0c52]

### Steps to reproduce

go to https://mempool.ninja/signet/mining and click the yellow circle
it sends you to the mainnet pool

### Expected behaviour

it sends you to the signet pool (it not a pool but anyway)

PR #1450, opened 2022-03-25
Use relative pipe for pie chart click event

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1447 (but I can't manage to test `signet` and `testnet` locally 😄)


softsimon (approved):
tested ACK @ 7e7dd121

by running with `npm run start:local-staging`

PR #1446, opened 2022-03-24
Added slug into `pools` table

nymkappa:
Related to https://github.com/mempool/mempool/issues/1440

Database schema is upgraded to 17. No data/cache truncation is required.

This PR add slugs for mining pools in the database and the `pools.json` parser. It is not used yet in the user facing application yet.

### Testing

* Before testing https://github.com/mempool/mining-pools/pull/2/ must be merged
* Restart your node backend
* Check:
  * db `state.schema_version` = 17
  * db `pools.slug` field exsits
  * slugs are filled properly for all pools, even for `Unknown`

PR #1451, opened 2022-03-25
If pool slug does not exist, generate one on the fly, avoid crash

nymkappa:
Follow up on https://github.com/mempool/mempool/pull/1446. @softsimon reported a crash at start
```
Mar 24 19:12:15 [41735] DEBUG: <bisq> Identify unique mining pools
Mar 24 19:12:15 [41735] DEBUG: <bisq> Found 133 unique mining pools
/Users/simon/repos/mempool/backend/dist/index.js:94
                throw new Error(e instanceof Error ? e.message : 'Error');
                      ^
Error: Cannot read properties of undefined (reading 'Luxor')
    at Server.startServer (/Users/simon/repos/mempool/backend/dist/index.js:94:23)
```

This PR add error handler if we cannot find pool slugs in the `pools.json` file, in order to avoid node backend crash. It generates a slug dynamically.

### Testing (missing some slugs)
* Set `"EXTERNAL_ASSETS": []` in your backend `mempool-config.json`
* In `/backend/pools.json`, delete **some** entry from the `slugs` object
* Run the node backend:
  * it should print some message like `No slug found for POOLNAME, generating...` with `POOLNAME` the pool name you've deleted
* Check that db `pools.slug` is properly filled up
```mysql
MariaDB [mempool_mainnet]> select name, slug from pools where name="POOLNAME";
+-------------+------------+
| name        | slug       |
+-------------+------------+
| POOLNAME    | poolname   |
+-------------+------------+
```

### Testing (no slugs at all)

* Set `"EXTERNAL_ASSETS": []` in your backend `mempool-config.json`
* In `/backend/pools.json`, delete the `slugs` object if you have one
* Run the node backend:
  * look for the `WARN: pools.json does not seem to contain the 'slugs' object`
  * it should also print lot of message `No slug found for POOLNAME, generating...`
* Check that db `pools.slug` is properly filled up
```mysql
MariaDB [mempool_mainnet]> select name, slug from pools;
+---------------------------+-------------------------+
| name                      | slug                    |
+---------------------------+-------------------------+
| BlockFills                | blockfills              |
| ULTIMUSPOOL               | ultimuspool             |
| Terra Pool                | terrapool               |
| Luxor                     | luxor                   |
| 1THash                    | 1thash                  |
| BTC.com                   | btccom                  |
| Bitfarms                  | bitfarms                |
| ...                       | ...                     |
```

### DIFF
--- a/backend/src/api/pools-parser.ts
+++ b/backend/src/api/pools-parser.ts
@@ -12,6 +12,8 @@ interface Pool {
 }
 
 class PoolsParser {
+  slugWarnFlag = false;
+
   /**
    * Parse the pools.json file, consolidate the data and dump it into the database
    */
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK

Issue #1452, opened 2022-03-25
nginx fails to start with systemd after reboot


orenfromberg:
### Description

nginx would not start on boot and gave this error:
```
admin@n83bsaje632:~$ sudo systemctl status nginx

● nginx.service - A high performance web server and a reverse proxy server
     Loaded: loaded (/lib/systemd/system/nginx.service; enabled; vendor preset: enabled)
     Active: failed (Result: exit-code) since Thu 2022-03-24 23:22:58 EDT; 15min ago
       Docs: man:nginx(8)

Mar 24 23:22:49 n83bsaje632 systemd[1]: Starting A high performance web server and a reverse proxy server...
Mar 24 23:22:57 n83bsaje632 nginx[766]: nginx: [emerg] host not found in upstream "mempool.space" in /etc/nginx/snippets/nginx-mempool.conf:49
Mar 24 23:22:58 n83bsaje632 nginx[766]: nginx: configuration file /etc/nginx/nginx.conf test failed
Mar 24 23:22:58 n83bsaje632 systemd[1]: nginx.service: Control process exited, code=exited, status=1/FAILURE
Mar 24 23:22:58 n83bsaje632 systemd[1]: nginx.service: Failed with result 'exit-code'.
Mar 24 23:22:58 n83bsaje632 systemd[1]: Failed to start A high performance web server and a reverse proxy server.
```

#### Version

v2.3.1

### Steps to reproduce

run `sudo reboot now` and upon startup check `sudo systemctl status nginx`

### Expected behaviour

I expect nginx to be running

### Actual behaviour

nginx is not running

### Screenshots

n/a

#### Device or machine

Linux n83bsaje632 5.13.0-37-generic #42~20.04.1-Ubuntu SMP Tue Mar 15 15:44:28 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
 Ubuntu 20.04.4 LTS

#### Additional info

n/a

antonilol:
i also have this issue
i think it tries to resolve the domain name mempool.space before it has an internet connection
i worked around this with a delay
```ini
ExecStartPre=/usr/bin/sleep 30
```

VajraOfIndra:
We also have this issue, see e.g. https://github.com/raspibolt/raspibolt/issues/919

wiz:
Fixed by #1666

PR #1455, opened 2022-03-25
Update nginx-mempool.conf: do not exit on startup if cannot reach mempool.space

omfgd00d:
Fixes https://github.com/mempool/mempool/issues/1452

From [this page](https://sandro-keil.de/blog/let-nginx-start-if-upstream-host-is-unavailable-or-down/):

>If you use proxy_pass or fastcgi_pass definitions in your nginx server config, then nginx checks the hostname during the startup phase. If one of these servers is not available, nginx will not start.

The workaround to this problem is to use an nginx variable for the upstream host. This will prevent the nginx server from exiting if it cannot reach the host on startup.

### DIFF
--- a/contributors/omfgd00d.txt
+++ b/contributors/omfgd00d.txt
@@ -1,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of March 25, 2022.
+
+Signed: omfgd00d
--- a/nginx-mempool.conf
+++ b/nginx-mempool.conf
@@ -46,22 +46,28 @@
 
 	# mainnet API
 	location /api/v1/donations {
-		proxy_pass https://mempool.space;
+		set $upstream mempool.space;
+		proxy_pass https://$upstream;
 	}
 	location /api/v1/donations/images {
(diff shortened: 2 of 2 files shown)


RooSoft:
I have found another fix that works well on my raspibolt.

Added this line

```ini
ExecStartPre=/bin/bash -c 'until host mempool.space; do sleep 1; done;'
```

in the `[Service]` section of the nginx systemd configuration file: `/lib/systemd/system/nginx.service`

The service definition file already had a `ExecStartPre` statement, but it's ok to have more than one.

wiz:
We solve this in a different better way using the resolver statement. Take a look at the production nginx.conf we just need to copy that one line over

omfgd00d:
@wiz if we add the resolver line then does that imply that nginx will only start if the DNS for mempool.space is configured? Wouldn't it be more robust if nginx didn't depend on being able to resolve any hostname?

wiz:
The bot works, it's just that for your first contribution the file needs to be merged into our master branch for it to see it. Also, thanks for your PR but we've just removed all of the mempool.space proxy lines from nginx.conf so we don't need this variable workaround anymore 😅

Issue #1406, opened 2022-03-18
Improve responsiveness on the miner details page


knorrium:
The addresses and tags lists are scrolling on mobile

![45D14DB8-2A58-4A72-BE25-1AAF1DF78CFF](https://user-images.githubusercontent.com/100320/159022773-1558ae31-08d2-41b7-a483-c4683b0f8087.png)

nymkappa:
Scolling is by design (_though it's not aligned well with other fields so I need to fix that_). It also scrolls on desktop when the list is long.

Do you have a suggestion on how to display those? For example, check `AntPool`, they have lot of addresses.
![image](https://user-images.githubusercontent.com/9780671/159217607-81d6fd9e-1f84-4316-83c8-4fbad1adf66e.png)

This also applies to regexes. Maybe a dropdown could be used instead?

nymkappa:
@knorrium I cannot reproduce the vertical align issue on iphone/safari, did you take the screenshot while pull the addresses to the right or is that the default view after loading the page?

nymkappa:
> Add scroll space, button/modal, or collapsible to the Addresses section. (if you check https://mempool.ninja/mining/pool/41, you can see that there are a few dozen addresses that need to be scrolled through, not sure if it is designed to be like that but it may be beneficial to see all the addresses at once via a button/modal or collapsible.

PR #1457, opened 2022-03-25
Pool addresses collapse - Cleanup mobile layout

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1406

Frontend only 👍🏻 

* Replace scrollable addresses with collapse (maybe we want the same thing for the tag, TBD)
* On mobile, show the label and the data on two different lines for all data and show title in bold (see screenshots)

#### Desktop

https://user-images.githubusercontent.com/9780671/160088917-0756afea-1095-4e2a-8cd9-e0b8f21969e8.mp4

#### Mobile

https://user-images.githubusercontent.com/9780671/160088241-c100488c-5c8f-4bb6-8166-187a6642db11.mp4

### Before

image
image

### After

image
image


softsimon (changes requested):
* We have a button called "Load all (18)" on the transaction page. So maybe this new button should also display "Show all (18)"
* The new Show button causes a slightly jumpy transition between skeleton and view on desktop due to the button height
* The mobile skeleton loader does not match the end result well. The skeleton suggests the data should be to the right, but it appears below

Sorry for OCD

nymkappa:
> 

Fixed in https://github.com/mempool/mempool/pull/1457/commits/5d9e8d0177d53ba2f77fc193630f4a9e614fa98d

softsimon (approved):
tested ACK

PR #1464, opened 2022-03-27
Detect more lightning scripts

antonilol:
- expired htlc
- anchor
- swept anchor

Fix indentation what i messed up in 7565aa7

Add explanation

and fix some tslint things like `==` -> `===` and radix to 10 instead of undefined

maybe move the regexes to one object for better readability because lines get to long

https://github.com/mempool/mempool/blob/master/frontend/tslint.json#L48-L51

review please on:
- [x] https://github.com/mempool/mempool/pull/1464#discussion_r836346572
- [ ] maybe move the regexes to one object for better readability because lines get to long

### DIFF
--- a/frontend/src/app/components/address-labels/address-labels.component.ts
+++ b/frontend/src/app/components/address-labels/address-labels.component.ts
@@ -46,19 +46,36 @@ export class AddressLabelsComponent implements OnInit {
         return;
       }
 
+      const topElement = this.vin.witness[this.vin.witness.length - 2];
       // https://github.com/lightning/bolts/blob/master/03-transactions.md#commitment-transaction-outputs
       if (/^OP_IF OP_PUSHBYTES_33 \w{66} OP_ELSE OP_PUSHBYTES_(1 \w{2}|2 \w{4}) OP_CSV OP_DROP OP_PUSHBYTES_33 \w{66} OP_ENDIF OP_CHECKSIG$/.test(this.vin.inner_witnessscript_asm)) {
-          if (this.vin.witness[this.vin.witness.length - 2] == '01') {
-              this.lightning = 'Revoked Force Close';
-          } else {
-              this.lightning = 'Force Close';
-          }
+        if (topElement === '01') {
+          // top element is '01' to get in the revocation path
+          this.lightning = 'Revoked Force Close';
+        } else {
+          // top element is '', this is a delayed to_local output
+          this.lightning = 'Force Close';
+        }
       // https://github.com/lightning/bolts/blob/master/03-transactions.md#offered-htlc-outputs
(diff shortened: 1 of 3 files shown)


antonilol:
idea: add a label also to outputs
handleVout can check if the output is spent and run handleVin for the spending tx's input

softsimon:
> idea: add a label also to outputs handleVout can check if the output is spent and run handleVin for the spending tx's input

Would be cool but don't think it's viable as of now since we would need to fetch the spent TX just for that.

antonilol:
> > idea: add a label also to outputs handleVout can check if the output is spent and run handleVin for the spending tx's input
> 
> Would be cool but don't think it's viable as of now since we would need to fetch the spent TX just for that.

esplora already does this, right? only the witness data needs to be sent to the client as the witnessscript is given at spend time (or p2sh redeemscript)

frontend/src/app/components/address-labels/address-labels.component.ts
```diff
@@ -46,19 +46,36 @@ export class AddressLabelsComponent implements OnInit {
...
+          // top element is '', this is a delayed to_local output
+          this.lightning = 'Force Close';
+        }
       // https://github.com/lightning/bolts/blob/master/03-transactions.md#offered-htlc-outputs
       } else if (/^OP_DUP OP_HASH160 OP_PUSHBYTES_20 \w{40} OP_EQUAL OP_IF OP_CHECKSIG OP_ELSE OP_PUSHBYTES_33 \w{66} OP_SWAP OP_SIZE OP_PUSHBYTES_1 20 OP_EQUAL OP_NOTIF OP_DROP OP_PUSHNUM_2 OP_SWAP OP_PUSHBYTES_33 \w{66} OP_PUSHNUM_2 OP_CHECKMULTISIG OP_ELSE OP_HASH160 OP_PUSHBYTES_20 \w{40} OP_EQUALVERIFY OP_CHECKSIG OP_ENDIF (OP_PUSHNUM_1 OP_CHECKSEQUENCEVERIFY OP_DROP |)OP_ENDIF$/.test(this.vin.inner_witnessscript_asm)) {
-        if (this.vin.witness[this.vin.witness.length - 2].length == 66) {
-            this.lightning = 'Revoked HTLC';
+        if (topElement.length === 66) {
+          // top element is a public key
+          this.lightning = 'Revoked HTLC';
```
antonilol, line 63: this can only show the word Lightning plus a message, so this gives Lightning Revoked HTLC because of
```html
<span
  *ngIf="liquid"
  class="badge badge-pill badge-warning"
  i18n="address-labels.upper-layer-peg-out"
>Liquid {{ liquid }}</span>
```

i actually want Revoked Lightning HTLC here, so i suggest not to have 3 seperate label that show up for multisig, liquid and lightning, but 1 `AddressLabelsComponent.label` that sets a label and is localized in the script, not the html. this also adds more flexibilty for future changes.

softsimon: I guess we can remove those i18n strings and do it in the script like you said. The words Lightning and Liquid in combination with the other technical terms is not viable to make translationable right now.. technical terms is often not possible to translate anyway.

antonilol: > technical terms is often not possible to translate anyway.

yes that's true

antonilol: will do it later today, i got an assignment to finish now

antonilol:
493b44d doesnt work yet because there is a bug because a `PUSHNUM_1` (`OP_1`) in bitcoin script is a `OP_PUSHBYTES_1 1` in  the `Vout.scriptpubkey_asm` example: 85a1335bf9a35d0ba81fc166df272a231791ab8397420051fdf0adc228a5eeff:0 (i dont see it on mempool.space so probably something to do with the backend)

angular live dev server (bitcoind + electrs):

image


mempool.space:

image

softsimon (approved):
tested ACK @ 105b67e5

PR #1467, opened 2022-03-29
Use slug instead of id in mining blocks list component

nymkappa:


wiz (approved):
Tested ACK @ v2.4.0-dev [[cc27b963](https://github.com/mempool/mempool/commit/cc27b963)]

PR #1468, opened 2022-03-29
Use mining pool slug in block component

nymkappa:
If the mining dashboard is enabled in the frontend `MINING_DASHBOARD: true`, use the mining pool slug from `pools.json` and point to `/mining/pool/{slug}` instead of opening external pool webpage.


softsimon (approved):
tested ACK

PR #1470, opened 2022-03-29
Fix query to insert unknown mining pool

nymkappa:
Fix regression introduced in [bb0fd78f2857e0568cfafa7fe3f9d292f4b05e68](https://github.com/mempool/mempool/commit/bb0fd78f2857e0568cfafa7fe3f9d292f4b05e68#diff-e5f65c2e3cf3c60bab8efba5ca79e7be5b42ebcc586a6332b2f0456e5821a793R167).

The query which is supposed to insert the `Unknown` mining pool in the database was wrong and therefore block indexing was stuck at the most recent block mined by an `Unknown` entity, since the pool was not in the database.

Poor testing was the cause of this bug. My bad.

### DIFF
--- a/backend/src/api/pools-parser.ts
+++ b/backend/src/api/pools-parser.ts
@@ -135,10 +135,11 @@ class PoolsParser {
     logger.debug(`Update pools table now`);
 
     // Add new mining pools into the database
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK @ 7ab950

Issue #1442, opened 2022-03-24
Allow timespan selection for pool hashrate chart


nymkappa:
![image](https://user-images.githubusercontent.com/9780671/159874622-80bbd22a-8a40-40da-905d-7ac8a9ae5ee6.png)

PR #1472, opened 2022-03-29
Add data zoom on pool hashrate chart

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1442

The original issue was to add timespan switcher in the chart, but it only makes sense for mining pools which are still active today. Therefore I re-added the datazoom element below the chart so users can select their own timespan, whether the pool is still active.

I also slightly changed the colors so it is slightly less visible than other charts which don't have data below them.

### Desktop

image

### Mobile

image


softsimon (approved):
tested ACK

PR #1473, opened 2022-03-29
Add faq

hunicus:
These changes add a new FAQ tab to the docs section.
- only shows for Bitcoin networks (i.e. mainnet, testnet, and signet on the Mempool site...not Liquid or Bisq)
- FAQ tab is now default docs tab whenever it's available

As an experiment, I added an iframe of the blocks visualization from the main dashboard (see `/docs/faq#what-is-a-mempool-explorer`). It's heavy but I think it looks cool...if people don't like it, I can remove it.

### DIFF
--- a/frontend/src/app/components/docs/api-docs-data.ts
+++ b/frontend/src/app/components/docs/api-docs-data.ts
@@ -4433,7 +4433,7 @@ export const faqData = [
     showConditions: bitcoinNetworks,
     fragment: "what-is-a-mempool-explorer",
     title: "What is a mempool explorer?",
-    answer: "<p>A mempool explorer is a tool that enables you to view real-time and historical information about a node's mempool, visualize its transactions, and search and view those transactions.</p><p>The mempool.space website invented the concept of visualizing a Bitcoin node's mempool as <b>projected blocks</b>. These blocks are the inspiration for our half-filled block logo.</p><p>Here's a snapshot of this visualization (which you can find on <a href='/'>the main dashboard</a>) as of block 729,556 in March 2022. Projected blocks are on the left of the dotted white line, and confirmed blocks are on the right.</p><iframe id='mempool-blocks' style='width: 100%;height: 275px;border: none' src='./resources/mempool-blocks.html' scrolling='no'></iframe>"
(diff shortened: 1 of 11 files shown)


hunicus:
Thanks @softsimon for the hack to get live mempool blocks in FAQ!

The latest commits implement live blocks in FAQ, remove the iframes from git history (was like ~0.5mb combined, total deadweight), and resolve merge conflicts.

softsimon (changes requested):
I noticed that when you go to /docs/faq and reload the page the mempool and blocks doesn't load because the websocket doesn't activate.

Solve this by simply adding this to the constructor function in the docs.component

`private websocket: WebsocketService,`

Just importing it and not using it .. will connect the websocket and get the mempool/blocks.

frontend/src/app/components/docs/api-docs-data.ts
```diff
@@ -4411,3 +4411,177 @@ export const restApiDocsData = [
...
+    title: "What is a mempool?",
+    answer: "<p>A mempool (short for \"memory pool\") holds the queue of pending and unconfirmed transactions for a cryptocurrency network node. There is no one global mempool: every node on the network maintains its own mempool, so different nodes may hold different transactions in their mempools.</p>"
+  },
+  {
+    type: "endpoint",
+    category: "basics",
+    showConditions: bitcoinNetworks,
+    fragment: "what-is-a-mempool-explorer",
+    title: "What is a mempool explorer?",
+    answer: "<p>A mempool explorer is a tool that enables you to view real-time and historical information about a node's mempool, visualize its transactions, and search and view those transactions.</p><p>The mempool.space website invented the concept of visualizing a Bitcoin node's mempool as <b>projected blocks</b>. These blocks are the inspiration for our half-filled block logo.</p><p>Projected blocks are on the left of the dotted white line, and confirmed blocks are on the right.</p>"
```
katesalazar, line 4436: This is a hellish large line, isn't it easier to maintain,
to find some way to limit text width in source code?

hunicus: Not sure that makes sense here...this is a data file. Adding a bunch of multi-line concatenation would be more annoying, I think.

hunicus:
> I noticed that when you go to /docs/faq and reload the page the mempool and blocks doesn't load because the websocket doesn't activate.

I wasn't able to reproduce this locally but added the suggested changes in 01f474628a19f9c24c237b5380e4e49383db49da.

frontend/src/app/components/docs/api-docs-data.ts
softsimon: The issue would be valid if it was source code. This is a special text data file.

softsimon (approved):
tested ACK @ 01f47462

PR #1474, opened 2022-03-29
Use slugs in cache warmer

nymkappa:
Execution directory must be in the `pools.json` folder, or we need to set the full path


wiz (approved):
Tested ACK

Issue #1477, opened 2022-03-29
Mining dashboard: mining style blocks should be displayed on signet and testnet


knorrium:
Currently the new block component is only used on mainnet
![Screenshot from 2022-03-29 11-01-00](https://user-images.githubusercontent.com/100320/160675911-4b7af6ec-81e5-4833-be76-91cde3528bd8.png)

PR #1484, opened 2022-03-30
Show block reward on blockchain blocks for all Bitcoin networks

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1477


softsimon (changes requested):
There is a similar check in the `mempool-blocks.component.ts` for the mempool blocks mining info.

nymkappa:
> There is a similar check in the `mempool-blocks.component.ts` for the mempool blocks mining info.

Thanks. Fixed in https://github.com/mempool/mempool/commit/1f20a56ae7ab58805a8f96985f3cae763282b8a9

softsimon (approved):
tested ACK @ 1f20a56a

Issue #1471, opened 2022-03-29
Pool page executes 5 extra requests


softsimon:
Executing 6 identical requests where all fail except 1 should not be needed.

<img width="1298" alt="Screen Shot 2022-03-29 at 13 24 10" src="https://user-images.githubusercontent.com/8561090/160579782-9b4a3f56-bae0-497c-89f7-03b86871b807.png">

PR #1489, opened 2022-03-30
Fix spam call to `/api/v1/mining/pool/{slug}`

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1471
Also removed duplicated calls to `/api/v1/mining/pool/blocks`


softsimon (approved):
tested ACK @ 4b2698ee

PR #1491, opened 2022-03-30
Apply smooth scrolling to docs only

hunicus:
Went up the tree and the smooth-scrolling property only seemed to work when applied to the `html` element.


nymkappa (approved):
Tested ACK @315b7593bf93508855b0343f06e157837a121c91

softsimon (approved):
utACK

PR #1495, opened 2022-04-01
Set zlevel for all graphs series

nymkappa:
Fixes `z / z2 / zlevel of displayable is invalid, which may cause unexpected errors`


softsimon (approved):
tested ACK

PR #1496, opened 2022-04-01
Mining stats does not depends on the websocket blocks number anymore

nymkappa:
The mining stats component assumes the websocket will send `this.stateService.env.MEMPOOL_BLOCKS_AMOUNT` when we load the mining dashboard page.

`MEMPOOL_BLOCKS_AMOUNT` has nothing to do with how many past blocks are sent through the websocket, but it was still working fine because it is usually set to `8`, which matches the backend setting `config.MEMPOOL.INITIAL_BLOCKS_AMOUNT`.

Also, as a side note, it seems to me that the frontend actually has no idea how many blocks to expect from the websocket when loading the page. Therefore I got rid of any assumption about this.

The same issue exists in other places in the mining dashboard code, and I will make separate PR to fix them.

### DIFF
--- a/backend/src/repositories/BlocksRepository.ts
+++ b/backend/src/repositories/BlocksRepository.ts
@@ -386,10 +386,14 @@ class BlocksRepository {
       connection = await DB.getConnection();
 
       // We need to use a subquery
-      const query = `SELECT SUM(reward) as totalReward, SUM(fees) as totalFee, SUM(tx_count) as totalTx
-        FROM (SELECT reward, fees, tx_count FROM blocks ORDER by height DESC LIMIT ${blockCount}) as sub`;
+      const query = `
+        SELECT MIN(height) as startBlock, MAX(height) as endBlock, SUM(reward) as totalReward, SUM(fees) as totalFee, SUM(tx_count) as totalTx
+        FROM
+          (SELECT height, reward, fees, tx_count FROM blocks
+          ORDER by height DESC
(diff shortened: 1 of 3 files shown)


wiz:
> Also, as a side note, it seems to me that the frontend actually has no idea how many blocks to expect from the websocket when loading the page. Therefore I got rid of any assumption about this.

IIRC this is a frontend configuration variable

nymkappa:
> > Also, as a side note, it seems to me that the frontend actually has no idea how many blocks to expect from the websocket when loading the page. Therefore I got rid of any assumption about this.
> 
> IIRC this is a frontend configuration variable

AFAIK there is not such thing for the initial connection.
When the websocket is initialized, an `init` message is sent to the node backend, which will send `config.MEMPOOL.INITIAL_BLOCKS_AMOUNT` most recent blocks to the websocket.

The frontend does not know (nor have a matching setting) for this `INITIAL_BLOCKS_AMOUNT` setting.

The problem is that new blocks are received through the same subscriptions in our components, there is not differentiation between the initial blocks (upon connection), and actually new blocks, so I had to do some skipping trick, which used faulty assumptions.

The PR solves the issue by using block height to know if we need to refresh stats, no configuration variable are involved anymore 👍🏻

softsimon (approved):
tested ACK @ 105b67e5

This seems to fix the issue of reward stats not updating on new blocks, it does now.

PR #1502, opened 2022-04-03
fix wrong display of other sighash types + p2sh redeemScript when <=4 bytes + consistency with esplora backend + tapscript fixes

### LINKED ISSUE #1260: local hosted mempool wrong script opcodes
### Steps to reproduce

go on your local hosted mempool which uses bitcoind and romanz/electrs to the testnet txid under screenshot and look at the details of the nonstandard output (:3) 

### Actual behaviour

2 bugs:
the arrow is red (gettxout returns nothing since it is *unspendable*, however this is interpreted as *unspent*
the local hosted mempool shows `OP_PUSHBYTES_4` asm for `0x24` opcode (`0x24` is a 36 byte push)

### Screenshots

image

left: mempool.space
right: local hosted mempool
txid: 941fe65310e49347ba7643f5cbc92d8f33f40d26e86b8bc22aabe70ed513b9db (testnet)

#### Device or machine

arch linux
chromium

#### Additional info

arch linux
chromium

antonilol:
tx: f06e7c91f741c0320a6aed477aa4b702c4b592b3ffa6089f51f90d9cab9e99e1

more info on sighash types: https://en.bitcoin.it/wiki/OP_CHECKSIG#How_it_works

before:
image

after:
image

other bug:
tx: 7d5e417287b32a166127042cd8fc2e2bf6205cc718c333e6f37778c358fb5307

before:
image

after:
image

just found out this fixes #1260


antonilol:
a small change was needed to get this working: `convertScriptSigAsm` now expects a script in **hex**, not asm anymore. i changed this were it is used

antonilol:
there were also some inconsistencies between output script and redeemScript/witnessScript because the output script came in ASM from bitcoin core, and the redeemScript/witnessScript was converted to ASM with bitcoinjs-lib, which output different things for the same input. Now every script is passed in as hex and converted to ASM with bitcoinjs-lib

frontend/src/app/components/transactions-list/transactions-list.component.html
```diff
@@ -111,7 +111,10 @@
                         <td style="text-align: left;" [innerHTML]="vin.inner_redeemscript_asm | asmStyler"></td>
                       </tr>
                       <tr *ngIf="vin.inner_witnessscript_asm">
-                        <td i18n="transactions-list.p2wsh-witness-script">P2WSH witness script</td>
+                        <td *ngIf="vin.prevout && vin.prevout.scriptpubkey_type == 'v1_p2tr'; else p2wsh" i18n="transactions-list.p2tr-tapscript">P2TR tapscript</td>
```
antonilol, line 114: > `i18n="transactions-list.p2tr-tapscript"`

i dont know how this localization exactly works, but i think this should do it

antonilol: or does it has to be added as a translation key somewhere?

antonilol:
53d68a3 adds `OP_CHECKSIGADD` (new tapscript opcode) to `convertScriptSigAsm`. this will only be visible now with bitcoind
it is already added to rust-bitcoin, but its version has to be bumped before it will be visible on mempool.space
PR to master (merged): https://github.com/rust-bitcoin/rust-bitcoin/pull/644/files#diff-1145b219d5b2128f8d2cc7b8b47c087b5cc1bc51e5f505443f66a73f0dd81a0cR421
PR to bump version: https://github.com/rust-bitcoin/rust-bitcoin/pull/785

antonilol:
taproot script spend with OP_CHECKSIGADD i made: 7a58768481f63f310119a5a8b02bd94613cddd3f380a108c5134f1be6a774d0b

hmmm the script is not there on mempool.space (dont know if this also will come with rust-bitcoin v0.28)

image

bitcoind backend:

image

frontend/src/app/components/transactions-list/transactions-list.component.html
softsimon: It is fully automatic. Just add a unique key for a unique string and it will be easier to recognize later in the translation-tool

Issue #1443, opened 2022-03-24
Additional general info in the pool detail page


nymkappa:
- [x] Block count for the past 24h
- [x] Current hashrate (average for the past 24h) - Estimated (block count) vs Reported (if API is provided)
- [x] Pool luck percentage (if the pool provides real time hashrate API - Should this be enabled on self hosted setup @wiz @softsimon)

PR #1503, opened 2022-04-04
Updated pool summary page to display more info on hashrate and blocks

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1443 (still have some todo after merging this PR around the reported hashrate).

Added some more info on the pool page:
* Hashrate
  * Estimated for the past 24h
  * _Reported (todo)_
  * _Luck (todo at the same time as Reported)_
* Mined blocks
  * 24h + network share
  * 1w + network share
  * All  + network share

Screen Shot 2022-04-05 at 12 36 25 AM
Screen Shot 2022-04-05 at 12 38 18 AM

### DIFF
--- a/backend/src/api/mining.ts
+++ b/backend/src/api/mining.ts
@@ -45,8 +45,8 @@ class Mining {
     const blockCount: number = await BlocksRepository.$blockCount(null, interval);
     poolsStatistics['blockCount'] = blockCount;
 
-    const blockHeightTip = await bitcoinClient.getBlockCount();
-    const lastBlockHashrate = await bitcoinClient.getNetworkHashPs(144, blockHeightTip);
+    const totalBlock24h: number = await BlocksRepository.$blockCount(null, '24h');
+    const lastBlockHashrate = await bitcoinClient.getNetworkHashPs(totalBlock24h);
     poolsStatistics['lastEstimatedHashrate'] = lastBlockHashrate;
 
     return poolsStatistics;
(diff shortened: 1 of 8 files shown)


backend/src/api/mining.ts
```diff
@@ -62,12 +62,30 @@ class Mining {
...
-    const emptyBlocksCount = await BlocksRepository.$countEmptyBlocks(pool.id);
+    const totalBlock: number = await BlocksRepository.$blockCount(null, null);
+
+    const blockCount24h: number = await BlocksRepository.$blockCount(pool.id, '24h');
+    const totalBlock24h: number = await BlocksRepository.$blockCount(null, '24h');
+
+    const blockCount1w: number = await BlocksRepository.$blockCount(pool.id, '1w');
+    const totalBlock1w: number = await BlocksRepository.$blockCount(null, '1w');
+
+    const currentEstimatedkHashrate = await bitcoinClient.getNetworkHashPs(totalBlock24h);
```
softsimon, line 73: I am concerned about adding direct calls to Bitcoin RPC from a public API call as Bitcoin RPC only can handle as much. If it will be handled with some Nginx cache I guess it will be fine?

frontend/src/app/components/pool/pool.component.html
```diff
@@ -10,48 +11,46 @@ <h1 class="m-0 pt-1 pt-md-0">{{ poolStats.pool.name }}</h1>
...
-                  </a>
-                  <button *ngIf="poolStats.pool.addresses.length >= 2" style="transform: translateY(-3px);"
-                    type="button" class="btn btn-outline-info btn-sm float-right" (click)="collapse.toggle()"
+              <tr *ngIf="!isMobile()" class="taller-row">
+                <td class="label addresses" i18n="mining.addresses">Addresses</td>
+                <td *ngIf="poolStats.pool.addresses.length else nodata" style="padding-top: 25px">
+                  <button *ngIf="poolStats.pool.addresses.length >= 2" type="button"
+                    class="btn btn-outline-info btn-sm float-right small-button" (click)="collapse.toggle()"
                     [attr.aria-expanded]="!gfg" aria-controls="collapseExample">
                     <span i18n="show-all">Show all</span> ({{ poolStats.pool.addresses.length }})
```
softsimon, line 42: The new desktop position of Show all looks a bit weirdly placed. Maybe move it to below the addresses on both desktop and mobile ... Just like it's for "load more" below addresses on the transaction page?

backend/src/api/mining.ts
nymkappa: Yes, `/api/v1/mining/pool/{slug}` is cache warmed. HTTP header `Expires` is set to 1 minute in the node backend right now.

softsimon (approved):
tested ACK

Issue #1153, opened 2022-01-17
small RTL issue


EZ0010:
### Description
A small RTL issue with mempool graph, the time and date in the title tag aren't correctly aligned.
#### Version

v2.3.0 [b9067ed9]

### Actual behaviour
![image](https://user-images.githubusercontent.com/75021359/149742223-cbc83b24-7846-42b5-9237-4281dbe2d032.png)

### Expected behaviour
I added  `.title { direction: rtl;}` and it fixed the issue.

![image](https://user-images.githubusercontent.com/75021359/149742108-36e08fe9-cc03-4eb6-af9a-44219ac47670.png)

#### Additional info
The other "Transaction vBytes per second (vB/s)" graph doesn't have the issue with RTL but it has another issue (on both RTL and LTR layout)
![image](https://user-images.githubusercontent.com/75021359/149743300-f81cbd0f-0d78-4d31-95f7-b070a4deabdb.png)

wiz:
@TechMiX did you want to help us with this one

TechMiX:
Sure. I will take care of this along with some other minor issues on other pages. Thanks for reporting, @EZ0010.

EZ0010:
Hi again. here are other RTL issues.
![8081DED3-1509-4763-8E58-0AD1469C8D65](https://user-images.githubusercontent.com/75021359/161530810-114aa210-8081-46f7-99c7-989a63ec723c.jpeg)
![FF65D291-86DB-4652-A357-5C9663C54EEC](https://user-images.githubusercontent.com/75021359/161530815-ff6e9f13-6fc0-4384-983d-e7671f299abd.jpeg)

TechMiX:
This is fixed in #1506, but not deployed yet.

PR #1506, opened 2022-04-05
fix RTL layout issues

TechMiX:
fixes various RTL layout issues including the one mentioned in #1153


softsimon (approved):
tested ACK @ 9a389cc9

PR #1509, opened 2022-04-06
WebSocket uid

### LINKED ISSUE #1334: Refactor the loading indicator websocket push
It should support multiple loading stuff by unique IDs.
In the specific view we can subscribe and filter for updates that matches the ID.

Replacing the current general loading indicator that doesn't say what it is about.

We should use this for the block, hash rate syncing, and current block and address page loading.

softsimon:
fixes #1334

The websocket connection will provide the client with the unique user id that will be passed in the header of HTTP requests in order to make per-user loading websocket indicators possible.

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -1,5 +1,6 @@
 import { CpfpInfo, TransactionExtended, TransactionStripped } from '../mempool.interfaces';
 import config from '../config';
+import * as crypto from 'crypto';
 export class Common {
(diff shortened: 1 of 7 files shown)


nymkappa:
Hey @softsimon your branch was created of master branch. This PR should only contains this commit https://github.com/mempool/mempool/pull/1509/commits/ce34acdae6b89dcc34911a6d0dab5caadc52d36c

wiz:
Closing due to stalled

Issue #1395, opened 2022-03-17
Handle pools.json changes


softsimon:
Store the hash of the `pools.json` in our `state` table

Detect changes and handle current data (delete Unknown pools / truncate hashrates)

PR #1511, opened 2022-04-07
Use github api to fetch and update the pools database, once a week

nymkappa:
This PR replace the parsing of on-disk `pools.json` file with github api.

A new class `PoolsUpdater` has been added. It executes once a week, check if the sha of https://github.com/mempool/mining-pools/blob/master/pools.json has changed from the latest one on record, and update the `pools` db table if changes are detected.

This is the first step into fixing https://github.com/mempool/mempool/issues/1395

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -135,6 +135,12 @@ class Blocks {
       } else {
         pool = await poolsRepository.$getUnknownPool();
       }
+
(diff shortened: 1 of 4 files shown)


softsimon (approved):
tested ACK

ayanamidev:
@nymkappa @softsimon I don't think it is good to fetch static assets using the native 'https' module while we already use axios which is more convenient, promise native and have an ability to use proxies like tor, would you mind if I change this to axios?

Issue #1507, opened 2022-04-05
Historical reward and fees graphs


softsimon:
With the new database data on fees, we are able to add the following graphs:

* Historical block median fee graph
* Historical block reward graph

Optional:

* Historical block reward graph (that later can be merged with a historical $USD reward graph)

softsimon:
Fixed in https://github.com/mempool/mempool/pull/1513 and https://github.com/mempool/mempool/pull/1512

PR #1512, opened 2022-04-08
Add block fees graph

nymkappa:
Context https://github.com/mempool/mempool/issues/1507

This PR add two things:
* I've merged all mining charts into a dropdown
* I've added a new chart to show the historical blocks fees, data is averaged (the same way we average mempool stats for longer timespan)

Screen Shot 2022-04-09 at 1 05 36 AM

Screen Shot 2022-04-09 at 1 05 21 AM

Screen Shot 2022-04-09 at 1 19 45 AM

_Note: I originally wanted to have the block reward (or subsidy) + block total fees in the same chart, but the UX will be very tricky to get right if we do that. I can provide more info about this if needed. This explains the odd branch name_


softsimon (changes requested):
It seems like the default timespan selection is something that is not among the choices (1Y)

Also for Hashrate & difficulty and Pools dominance which seems to default for "All" but it's not selected for me.

Screen Shot 2022-04-09 at 23 10 41

nymkappa:
> It seems like the default timespan selection is something that is not among the choices (1Y)
> 
> Also for Hashrate & difficulty and Pools dominance which seems to default for "All" but it's not selected for me.
> 
> Screen Shot 2022-04-09 at 23 10 41

I've updated the timespan selection overall now. All mining charts share a variable in localstorage `miningWindowPreference`. Timespan will be pre-selected properly.
Also, if the localstorage preference is lower than what the graph supports (for example, hashrate graph support only from 1w+), it will not change the user preference, but will still preselect a correct value.

softsimon (approved):
tested ACK

PR #1514, opened 2022-04-09
Fix Lightning HTLC detection with options_anchors

antonilol:
change `OP_CHECKSEQUENCEVERIFY` to `OP_CSV` in regex


softsimon (approved):
utACK

PR #1517, opened 2022-04-12
Refactor database pool use

nymkappa:
This PR is an attempt to fix `Can't add new command when connection is in closed state` errors. This PR remove all the manual db pool connection handling and let the library manages this. This approach was inspired by this thread https://github.com/sidorares/node-mysql2/issues/939.

As this is quite a critical change, it would be wise to let this PR run for a couple of day on some .fra servers @wiz before we can ACK it.

### DIFF
--- a/backend/src/database.ts
+++ b/backend/src/database.ts
@@ -1,11 +1,14 @@
 import config from './config';
-import { createPool, PoolConnection } from 'mysql2/promise';
+import { createPool, Pool, PoolConnection } from 'mysql2/promise';
 import logger from './logger';
 import { PoolOptions } from 'mysql2/typings/mysql';
 
 export class DB {
-  static poolConfig = ():PoolOptions => {
-    let poolConfig:PoolOptions = {
+  static connectionsReady: number[] = [];
+  static pool: Pool | null = null;
+
+  static poolConfig = (): PoolOptions => {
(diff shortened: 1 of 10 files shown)


backend/src/database.ts
```diff
@@ -1,49 +1,52 @@
...
 import { PoolOptions } from 'mysql2/typings/mysql';
 
 export class DB {
-  static poolConfig = ():PoolOptions => {
-    let poolConfig:PoolOptions = {
+  static connectionsReady: number[] = [];
+  static pool: Pool | null = null;
+
+  static poolConfig = (): PoolOptions => {
+    const poolConfig: PoolOptions = {
```
softsimon, line 11: I believe we can move the `const poolConfig` above the class instead of calling the `getPool()` to generate the config object every database query.

softsimon (approved):
tested ACK locally

But need to test in production as well.

nymkappa:
Rebased on master and fixed conflict with the db refactoring

wiz (approved):
Tested ACK @ v2.4.0-dev [[6919393e](https://github.com/mempool/mempool/commit/6919393e)]

Seems fine after running for a few hours with no issues, so going to merge and upgrade a few more servers for further testing.

PR #1521, opened 2022-04-14
Add block fee rates percentiles chart

nymkappa:
Screen Shot 2022-04-15 at 12 17 31 AM
Screen Shot 2022-04-15 at 12 17 18 AM
Screen Shot 2022-04-15 at 12 17 08 AM


nymkappa:
@softsimon Feel free to take a first look.
- On the `All` timespan, I capped the y-axis value to 5000 other the chart is almost all flat (dunno what happened in 2021 but there were some insane fees values there).
- If you select `24h` or `3D`, you can click on the graph bars, and it will open the block page. It's not available on higher timeframe because it represents an average of multiple blocks.
- I will see if I can save which percentiles are selected by the user so it's not lost when we switch timespan.

wiz:
Looks very cool. I think you can change the y-axis label from sats/vB to s/vB to save space. Please also add new API endpoints to cache warmer script.

wiz (approved):
Tested ACK @ v2.4.0-dev [[ed495cc0](https://github.com/mempool/mempool/commit/ed495cc0)]

There are a few minor issues which I will report as separate bugs, but overall looks great and works well.

PR #1548, opened 2022-04-15
Use block count instead of oldest block for timespan selection

nymkappa:
Today if you don't index all blocks, but browse the oldest blocks in the `/mining/blocks` page, then all mining charts will be pretty much broken. The reason is that we used the oldest indexed block timestamp as a proxy to know how much data we have.
I've replaced this by the total number of indexed block instead, which is more much more accurate.

Now you'll be able to index only 10k blocks, and browser Satoshi's blocks, without breaking your mining charts 😃


wiz:
Any idea why the global hashrate is so far off with this PR deployed?

### Expected

Screen Shot 2022-04-24 at 12 28 19

### Actual

Screen Shot 2022-04-24 at 12 28 24

wiz (approved):
Partially Tested ACK @ v2.4.0-dev [[0803fed0](https://github.com/mempool/mempool/commit/0803fed0)]

Issue #1547, opened 2022-04-15
Invalid reorg detection when `Fast forwarding`


nymkappa:
Nothing critical, it will not result in invalid or missing indexed data, but it just add unnecessary overhead and `warn` spam.

```
Apr 15 17:40:18 [46849] INFO: Downloading external asset pools.json over clearnet...
Apr 15 17:40:18 [46849] INFO: Connected to Electrum Server at 127.0.0.1:50001 (["electrs-esplora 0.4.1","1.4"])
Apr 15 17:40:19 [46849] INFO: External asset pools.json saved to ./pools.json
Apr 15 17:40:19 [46849] INFO: Restoring mempool and blocks data from disk cache
Apr 15 17:40:19 [46849] INFO: Database connection established.
Apr 15 17:40:19 [46849] INFO: Starting statistics service
Apr 15 17:40:19 [46849] INFO: Starting currency rates service
Apr 15 17:40:19 [46849] INFO: Currency rates service will be queried over clearnet using https://price.bisq.wiz.biz/getAllMarketPrices
Apr 15 17:40:19 [46849] NOTICE: Mempool Server is running on port 8999
Apr 15 17:40:19 [46849] INFO: 95 blocks since tip. Fast forwarding to the 8 recent blocks
Apr 15 17:40:19 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:40:19 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:40:19 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:40:19 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:40:19 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:40:19 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:40:19 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:40:19 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:40:22 [46849] WARN: Chain divergence detected at block 731948, re-indexing most recent data
Apr 15 17:41:00 [46849] NOTICE: The mempool is now in sync!
Apr 15 17:41:00 [46849] INFO: Delete 10 most recent indexed blocks from the database
[...]

PR #1554, opened 2022-04-18
Validate block hash chain after indexing and for new blocks

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1547 and also remove the automatic deletion of 10 blocks upon each node backend restart.

Two new ideas introduced in this PR:
* After the initial block indexing, valid the whole chain of block hash, and re-index blocks from where a mismatch is detected
* When a new block is detected, if the new block previoushash does not match the last indexed block hash, re-index the past 10 blocks, the last daily network hashrate and the last weekly pools hashrates (_The code still assumes that we don't have a reorg deeper than 10 blocks_)


wiz (approved):
Partially Tested ACK @ v2.4.0-dev [[8d279edf](https://github.com/mempool/mempool/commit/8d279edf)]

Not sure how to fully test a re-org scenario, but it seems to have detected a previous re-org on testnet after restarting the first time and has re-indexed a point since then, so it seems working well :)

Issue #1393, opened 2022-03-17
Refactor away pools.json from the frontend


softsimon:
Now that the backend and the new blocks api contains the pool identification, we can remove the pools.json download from `sync-assets.js` and the miner component.

nymkappa:
This requires a refactor the `/block` API. It would make sense to put this to 2.5 now that we're trying to polish 2.4
Ref: https://github.com/mempool/mempool/pull/1348

PR #1556, opened 2022-04-20
Update block API

nymkappa:
This sensitive PR must be tested thoroughly on all networks because it changes the `/block` API code.

This PR updates the `/block` API on Bitcoin network to always return our `block.extras` data structure, regardless if indexing is enabled or not. This allows us to:
* Tag miner without having to fetch the `pools.json` file (Fixes https://github.com/mempool/mempool/issues/1393, some code cleanup it required as we may not need the miner component anymore)
* Show the correct value for block reward and subsidy (Fixes https://github.com/mempool/mempool/pull/1156)

If the block is cached in memory, we return it directly.
If indexing is enabled, we return block from the database when available. If not available, we index it on the fly and save it.
If indexing is not available, we also index the block extended on the fly, but we don't save it.

Small change compared to previous behavior regarding the miner tag in the block component: We do not show the tooltip anymore saying `Identified by coinbase tag/address`. While it's possible to re-add it by adding a new field in the block.extras data structure, I do not think it is appropriate for this PR and would require block re-indexing.


wiz (changes requested):
Tested NACK @ v2.4.0-dev [[38648a86](https://github.com/mempool/mempool/commit/38648a86)]

Please see https://mempool.ninja/block/0000000000000000004b27f9ee7ba33d6f048f684aaeb0eea4befd80f1701126 which crashes with the following error in JS console:
```
ERROR TypeError: Cannot read properties of undefined (reading 'pool')
ERROR TypeError: Cannot read properties of undefined (reading 'totalFees')
```

wiz:
Here are some more errors for you to fix :)
```
Apr 24 07:06:33 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 10 sec.
24h
Apr 24 07:06:33 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 10 sec.
24h
Apr 24 07:06:43 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 20 sec.
24h
Apr 24 07:06:43 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 20 sec.
24h
Apr 24 07:06:44 node206.fmt  mempool[28609]: WARN <bisq> runMainLoop error: Table 'mempool_bisq.pools' doesn't exist. Retrying in 10 sec.
24h
Apr 24 07:06:45 node206.fmt  mempool[25986]: NOTICE The mempool is now in sync!
24h
Apr 24 07:06:56 node206.fmt  mempool[28609]: WARN <bisq> runMainLoop error: Table 'mempool_bisq.pools' doesn't exist. Retrying in 20 sec.
24h
Apr 24 07:07:03 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 40 sec.
24h
Apr 24 07:07:03 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 40 sec.
24h
Apr 24 07:07:17 node206.fmt  mempool[28609]: WARN <bisq> runMainLoop error: Table 'mempool_bisq.pools' doesn't exist. Retrying in 40 sec.
24h
Apr 24 07:07:43 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 60 sec.
24h
Apr 24 07:07:43 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 60 sec.
24h
Apr 24 07:07:58 node206.fmt  mempool[28609]: WARN <bisq> runMainLoop error: Table 'mempool_bisq.pools' doesn't exist. Retrying in 60 sec.
24h
Apr 24 07:08:43 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 60 sec.
24h
Apr 24 07:08:43 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 60 sec.
```

nymkappa:
> 

Disk cache issue most likely. Make sure to clear the `backend/cache` folder content 👍🏻

nymkappa:
> Here are some more errors for you to fix :)
> 
> ```
> Apr 24 07:06:33 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 10 sec.
> 24h
> Apr 24 07:06:33 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 10 sec.
> 24h
> Apr 24 07:06:43 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 20 sec.
> 24h
> Apr 24 07:06:43 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 20 sec.
> 24h
> Apr 24 07:06:44 node206.fmt  mempool[28609]: WARN <bisq> runMainLoop error: Table 'mempool_bisq.pools' doesn't exist. Retrying in 10 sec.
> 24h
> Apr 24 07:06:45 node206.fmt  mempool[25986]: NOTICE The mempool is now in sync!
> 24h
> Apr 24 07:06:56 node206.fmt  mempool[28609]: WARN <bisq> runMainLoop error: Table 'mempool_bisq.pools' doesn't exist. Retrying in 20 sec.
> 24h
> Apr 24 07:07:03 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 40 sec.
> 24h
> Apr 24 07:07:03 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 40 sec.
> 24h
> Apr 24 07:07:17 node206.fmt  mempool[28609]: WARN <bisq> runMainLoop error: Table 'mempool_bisq.pools' doesn't exist. Retrying in 40 sec.
> 24h
> Apr 24 07:07:43 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 60 sec.
> 24h
> Apr 24 07:07:43 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 60 sec.
> 24h
> Apr 24 07:07:58 node206.fmt  mempool[28609]: WARN <bisq> runMainLoop error: Table 'mempool_bisq.pools' doesn't exist. Retrying in 60 sec.
> 24h
> Apr 24 07:08:43 node206.fmt  mempool[27043]: WARN <liquid> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 60 sec.
> 24h
> Apr 24 07:08:43 node206.fmt  mempool[30434]: WARN <liquidtestnet> runMainLoop error: Unknown column 'slug' in 'field list'. Retrying in 60 sec.
> ```

Fixed. Also fixed the frontend so we show the pool tag only on Bitcoin networks. Please deploy the PR again and let me know if there are any other issues.
I've tested it locally using mainnet and liquidtestnet.

nymkappa:
Rebased on top of master

wiz:
> Execute the following mysql query: delete from blocks where height = 501726;

That SQL query fixed the page loading, but there is still an error in the JS console:
```
ERROR TypeError: Cannot read properties of undefined (reading 'pool')
```

Here is the other info you requested:
```
root@localhost [mempool]> select height, pool_id, coinbase_raw from blocks where height in (501725, 501726, 501727, 700000, 733453);
+--------+---------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| height | pool_id | coinbase_raw                                                                                                                                                                                             |
+--------+---------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 501725 |      11 | 03dda70741d691e2d183f1ef41d691e2d08ff7f92f4254432e544f502ffabe6d6d18ae6d666053eab96f9675c9b02a7dc925fb814fdd39811bbcf31c2a4c6a9a3a80000000000000005e00c78f5261010000000000                               |
| 501727 |       6 | 03dfa707043c8f475a622f4254432e434f4d2ffabe6d6d2f8f373e994f2719a9b22e76ad772bb5ad663b7afea311014de5593aaa9f6f4f01000000000000002f007daba95f000000000000                                                   |
| 700000 |     137 | 0360ae0a2cfabe6d6d078b52bc69cf0982356d3eb439efe5489aee2c74da629dcc1a8b50a309cc0800100000000212030400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005008a1c0000 |
| 733453 |      73 | 030d310b082f5669614254432f2cfabe6d6d73c38bb174a77684602f59926edacd85e281a13e0a88b69777d725ce3243e83c100000000000000010b2f8e8035faacecba0b9fef1d3eefe0100000000                                           |
+--------+---------+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
4 rows in set (0.000 sec)
```

nymkappa:
Fixed your reported issues:
* In staging/prod, esplora was queried directly from the frontend which is why you did not have the `.extras` structure in the API response. We now always query the node backend (which will use esplora when needed/available)
* We now always expose the `/block/:hash` api in the node backend

wiz:
> esplora was queried directly from the frontend which is why you did not have the .extras structure in the API response. We now always query the node backend (which will use esplora when needed/available)

Wait a minute, isn't this a major change in the architecture of our site? It sounds like this will cause the frontend to be slower, so for starters I would like to see the hard data from performance testing to see how much slower the site is with this change.

Personally I think we should reject any change which makes the site slower, since we want the site to be faster, not slower. @softsimon what do you think?

wiz:
That being said, perhaps we could warm-cache all of the Bitcoin blocks into nginx RAM so it's not a performance hit. We need to see some performance numbers first, as a base line, and then we can try adding the nginx warm-cache and see how it changes.

nymkappa:
> > esplora was queried directly from the frontend which is why you did not have the .extras structure in the API response. We now always query the node backend (which will use esplora when needed/available)
> 
> Wait a minute, isn't this a major change in the architecture of our site? It sounds like this will cause the frontend to be slower, so for starters I would like to see the hard data from performance testing to see how much slower the site is with this change.
> 
> Personally I think we should reject any change which makes the site slower, since we want the site to be faster, not slower. @softsimon what do you think?

We already discussed this. There are two approach we can follow.

1. Keep querying esplora from the frontend and add an extra API call to the node backend to fetch our extra data. 2. Have one unified API in the node backend to send all data at once.

The only significantly slower response time would happen for blocks which are not yet indexed (not a concern for prod), and possibly for older blocks for which we have to query mysql db (out of the memory cache).

Yet since blocks data never changes (aside in case of reorg or reindex of blocks tagged with unknown), the API endpoint can be cached for long period of time.

What would be your approach in measuring performance here? We want to measure a full page load time, not just API response time.

wiz:
Okay cool then my only requested change is to simply add all the block API endpoints (for all 700K+ blocks) to the warm cache script, but instead of every 10 seconds like the other APIs, set the cache time to 1 month or so and have the cache warmer query them at script startup and thereafter every 2 weeks or so.

softsimon (commented):
I guess 
    `this.getMiningPools$ = this.httpClient.get(apiBaseUrl + '/resources/pools.json').pipe(shareReplay(1));`

Can be removed from `assets.service.ts`

wiz (approved):
Tested ACK @ v2.4.0-dev [[094257a9](https://github.com/mempool/mempool/commit/094257a9)]

Let's merge this for now and improve performance using warm-caching later on

PR #1558, opened 2022-04-21
Wait for external assets file writers to complete

nymkappa:
This PR fixes the external asset download which was running asynchronously. Therefore the node backend was sometimes trying to read assets file while it was not fully written on disk yet.

This fixes liquid icons parsing failure issues.

_Also cleanup unrelated things I've noticed on the way 👍🏻._

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -77,7 +77,7 @@ export class Common {
     };
   }
 
-  static sleep(ms: number): Promise<void> {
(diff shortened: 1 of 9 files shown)


softsimon:
Append $ on all methods that are async / returning promise.

nymkappa:
> Append $ on all methods that are async / returning promise.

Force pushed

wiz (approved):
Tested ACK @ v2.4.0-dev [c51216f9]

Issue #1559, opened 2022-04-21
Websocket is constantly spamming loading indicator to everyone


softsimon:
This shows up even though I am just sitting at the / dashboard.

<img width="268" alt="Screen Shot 2022-04-21 at 16 16 36" src="https://user-images.githubusercontent.com/8561090/164456199-23c6db20-4af2-4493-a12c-4532f6d6b40d.png">

nymkappa:
Related to https://github.com/mempool/mempool/issues/1334
The /block-extras has loading indicator websocket push, even tho I did not implement any loading indicator (and it's near instant anyway) so I can just remove it.

PR #1563, opened 2022-04-22
Remove loading indicator for blocks-extras API

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1559


softsimon (approved):
utACK

Issue #1497, opened 2022-04-01
Block 729893 shows wrong miner (?)


nymkappa:
This block https://mempool.space/block/00000000000000000002b81ccc3766c2083413ed982098957acb12f3fce0b819 shows `Unknown` mining pool.

The coinbase asciiScriptSig `%#binance/818�dkMtmmY䡸aLEDD�������j���:b/��` should match Binance tag `/Binance/`

_I'm trying to confirm on Twitter if that's their block or not_

nymkappa:
Missing coinbase tag
https://github.com/mempool/mining-pools/pull/4
Also related https://github.com/mempool/mempool/pull/1564

PR #1564, opened 2022-04-22
Ignore case on coinbase tag regexes

nymkappa:
Ignore regex case when matching coinbase tags
Related to https://github.com/mempool/mempool/issues/1497


softsimon (approved):
utACK

wiz (approved):
Tested ACK @ v2.4.0-dev [[0860e16b](https://github.com/mempool/mempool/commit/0860e16b)]

Fully re-indexed from db schema version 2 and the "unknown" mined blocks from `binance` are now properly identified:

Screen Shot 2022-04-26 at 7 37 53

Issue #1562, opened 2022-04-22
`blocks-extras` API returns incorrect structure during first start


nymkappa:
I just setup a new mempool instance in my VM, and I was not able to open the new blocks list page because the `/api/v1/blocks-extras` API does not return a `pool` object for all blocks.

Need to investigate why and fix before tagging v2.4.

Tested @ e8d68726201069f579361d12cada1be1d1ec5cdf.

PR #1565, opened 2022-04-22
Update pools table before fetching the first blocks

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1562

To test this you need to truncate your `blocks` table and **clear your disk cache** to emulate a clean mempool install. Then load `/mining/blocks` page as soon as the node backend is restarted and confirm that blocks are showing, which is not the case today without this PR.


nymkappa:
Failed the rebase so it closed the PR automatically, re-opening :D

wiz (approved):
Tested ACK @ v2.4.0-dev [[5281c486](https://github.com/mempool/mempool/commit/5281c486)]

PR #1568, opened 2022-04-24
Moving Docs and Faq to separate lazy loaded module

softsimon:
This moves the **~200 kB** (uncompressed) Docs and Faq module to a separate lazy loaded module to improve the main bundle size.


wiz (approved):
Partially tested ACK @ v2.4.0-dev [2fc45380]

Seems to work fine but I want to compare the before and after loading resources and transfer size a bit more before merging to verify it's working as intended

wiz:
@softsimon looks good, can you do this for other pages as well? also please rebase

hunicus (approved):
All looks to be functional. Noticing a ~300kb docs resource being loaded upon navigating to the docs section from elsewhere.

ACK from my end.

wiz:
@hunicus since you moved the docker docs back to its original location, will need to update the FAQ link as well now - cc @knorrium for lulz

Issue #1567, opened 2022-04-24
loadBisqDumpFile(): Cannot create a string longer than 0x1fffffe8 characters


wiz:
Seems like we should be using a stream buffer instead of a string somewhere:
```
Apr 24 06:00:14 node206.fmt  mempool[17249]: INFO <bisq> loadBisqDumpFile() error.Cannot create a string longer than 0x1fffffe8 characters
```

Also please fix the log priority and message format, i.e. `ERR <bisq> loadBisqDumpFile(): Cannot create a string longer than 0x1fffffe8 characters`

wiz:
@nymkappa please look into this asap, since it's a critical error that totally breaks bisq: https://bisq.ninja/transactions

PR #1569, opened 2022-04-25
Read bisq dump blocks line by line and parse manually

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1567

This PR is an attempt to fix bisq blocks parsing. Previously we were loading the whole file into a `string`. But nodejs has a max string memory size of 512mb so the file does not fit in there.

I now read the file line by line, and use tokens to identify when a block start and ends. So it's 100% dependent on the file formatting, including spacing.

Block start token: `XXXX{` (X = space, does not render in github lol)
Block end token: `XXXX},`

This is not ideal, but unless the file format changes, it should(TM) solve the issue. I did not want to add a new 3rd party library (JSON streaming) because they usually have many feature we don't care about here, and the parsing is trivial, yet not very robust as it heavily depends on the syntax.

Note: In terms of performance, it seems the JSON parsing was multi threaded before. It is now single threaded. On my machine I get the following:
```
Apr 25 11:45:33 [89986] DEBUG: Processing Bisq data dump...
Apr 25 11:45:35 [89986] DEBUG: Bisq dump processed in 2335 ms
````

### DIFF
--- a/backend/src/api/bisq/bisq.ts
+++ b/backend/src/api/bisq/bisq.ts
@@ -157,8 +157,7 @@ class Bisq {
 
   private async loadBisqDumpFile(): Promise<void> {
     try {
-      const data = await this.loadData();
-      await this.loadBisqBlocksDump(data);
+      await this.loadData();
       this.buildIndex();
       this.calculateStats();
     } catch (e) {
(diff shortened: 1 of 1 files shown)


wiz (approved):
Tested ACK @ v2.4.0-dev [[addfa455](https://github.com/mempool/mempool/commit/addfa455)]

Thanks for fixing this so quickly.

Issue #1575, opened 2022-04-26
mempool backend crashes due to OOM while processing bisq market offers


wiz:
```
Apr 26 05:07:29 [50282] DEBUG: <bisq> Updating Bisq Market Offers Data

<--- Last few GCs --->

[50282:0x804eb3500] 21346878 ms: Mark-sweep 3783.7 (4130.4) -> 3764.4 (4126.5) MB, 10206.5 / 0.0 ms  (average mu = 0.102, current mu = 0.011) allocation failure scavenge might not succeed
[50282:0x804eb3500] 21356105 ms: Mark-sweep 3783.3 (4129.7) -> 3764.3 (4126.3) MB, 9168.3 / 0.0 ms  (average mu = 0.058, current mu = 0.006) allocation failure scavenge might not succeed


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x2fb2c00 node::Abort() [/mempool/.nvm/versions/node/v16.10.0/bin/node]
 2: 0x2fb2d75 node::OnFatalError(char const*, char const*) [/mempool/.nvm/versions/node/v16.10.0/bin/node]
 3: 0x3146d43 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/mempool/.nvm/versions/node/v16.10.0/bin/node]
 4: 0x3146ce9 v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/mempool/.nvm/versions/node/v16.10.0/bin/node]
 5: 0x32ecdc5 v8::internal::Heap::FatalProcessOutOfMemory(char const*) [/mempool/.nvm/versions/node/v16.10.0/bin/node]
 6: 0x32f0829 v8::internal::Heap::RecomputeLimits(v8::internal::GarbageCollector) [/mempool/.nvm/versions/node/v16.10.0/bin/node]
 7: 0x32ed6e2 v8::internal::Heap::PerformGarbageCollection(v8::internal::GarbageCollector, v8::GCCallbackFlags) [/mempool/.nvm/versions/node/v16.10.0/bin/node]
 8: 0x32ead5d v8::internal::Heap::CollectGarbage(v8::internal::AllocationSpace, v8::internal::GarbageCollectionReason, v8::GCCallbackFlags) [/mempool/.nvm/versions/node/v16.10.0/bin/node]
[...]

PR #1579, opened 2022-04-27
Clean local block cache before loading dump

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1575


wiz (approved):
Tested ACK @ v2.4.0-dev [[0aac4109](https://github.com/mempool/mempool/commit/0aac4109)]

Issue #1577, opened 2022-04-26
The reward stats skeleton loaders has incorrect titles


softsimon:
The 3 Miners Reward component titles on the mining dashboard are wrong in the skeleton loader.

The Next Halving also states "Current Period" during loading in the Difficulty adjustment component.

PR #1580, opened 2022-04-27
Fix reward stats skeleton titles

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1577


wiz (approved):
Tested ACK @ v2.4.0-dev [[452e9277](https://github.com/mempool/mempool/commit/452e9277)]

PR #1584, opened 2022-04-28
Upgrading npm packages and

softsimon:
Bumping angular package versions and other packages that dependabot noticed.


knorrium (approved):
utACK

wiz (approved):
Tested ACK @ v2.4.0-dev [[5336ba81](https://github.com/mempool/mempool/commit/5336ba81)]

Issue #1581, opened 2022-04-27
Deprecate the MEMPOOL_BACKEND configuration field


OliverOffing:
### Description

Raspiblitz automatic installation of mempool creates a configuration file that assumes Electrum Server is installed. This means that if you haven't installed Electrum Server, mempool will fail to query the backend. This issue can obviously be fixed in either project (mempool or raspiblitz) but I'm thinking that if we implement it here other projects might benefit from a simplified installation process.

### Problem to be solved

The configuration file currently doesn't allow for both bitcoind and electrum server to work as backends at the same time.
```yml
  api:
    environment:
      MEMPOOL_BACKEND: "none"
```

### Proposed solution

- Ditch the `MEMPOOL_BACKEND` configuration field
- Check `RPC_*` and `ELECTRUM_*` fields for available backends
- Attempt to connect to any and all configured backends
- Use whichever connects

#### Additional info

Original issue: https://github.com/rootzoll/raspiblitz/issues/3077

PR #1586, opened 2022-04-29
Clarify configuration options

OliverOffing:
I've recently [misunderstood how the backend configuration works](https://github.com/mempool/mempool/issues/1581#issuecomment-1113466544) and so decided to suggest a few changes in the documentation to help clarify this point.

Overall, I think we should move away from saying things like "`electrum` as the backend", given that you still need to have `bitcoind` alongside.

Similarly, I think the valid options for `MEMPOOL_BACKEND` should change:

- `none` => `bitcoind`
- `electrum` => `bitcoind_electrum`

This might help clarify that `electrum` is an _optional addon_, and not a self-standing backend by itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@ Mempool can be self-hosted on a wide variety of your own hardware, ranging from 
 
 The `docker` directory contains the Dockerfiles used to build and release the official images and a `docker-compose.yml` file that is intended for end users to run a Mempool instance with minimal effort.
 
-## bitcoind only configuration
+## bitcoind-only configuration
+
+> If you need the ability to do address lookups, please use the [bitcoind+electrum configuration](#bitcoindelectrum-configuration) instead.
 
 To run an instance with the default settings, use the following command:
 
@@ -51,14 +53,16 @@ You can check if the instance is running by visiting http://localhost - the grap
 
 ## bitcoind+electrum configuration
 
-In order to run with a `electrum` compatible server as the backend, in addition to the settings required for running with `bitcoind` above, you will need to make the following changes to the `docker-compose.yml` file:
+> This configuration gives you the ability to do address lookups.
+
+In order to add on an `electrum` compatible server, in addition to the settings required for running with `bitcoind` above, you will need to make the following changes to the `docker-compose.yml` file:
 
 - Under the `api` service, change the value of the `MEMPOOL_BACKEND` key from `none` to `electrum`:
 
 ```
   api:
     environment:
-      MEMPOOL_BACKEND: "none"
+      MEMPOOL_BACKEND: "electrum"
 ```
 
 - Under the `api` service, set the `ELECTRUM_HOST` and `ELECTRUM_PORT` keys to your Docker host IP address and set `ELECTRUM_TLS_ENABLED` to `false`:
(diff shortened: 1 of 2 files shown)


nymkappa:
I agree. But IMO it would be better to rename the option instead of the value. I would rename `backend` to something like `index_engine` and keep the existing values.

wiz:
Thanks for the feedback. I think @hunicus is currently rewriting our documentation so perhaps he could incorporate your suggestions into his rewrite

README.md
```diff
@@ -51,14 +53,16 @@ You can check if the instance is running by visiting http://localhost - the grap
...
+
+In order to add on an `electrum` compatible server, in addition to the settings required for running with `bitcoind` above, you will need to make the following changes to the `docker-compose.yml` file:
 
 - Under the `api` service, change the value of the `MEMPOOL_BACKEND` key from `none` to `electrum`:
 
 ```
   api:
     environment:
-      MEMPOOL_BACKEND: "none"
+      MEMPOOL_BACKEND: "electrum"
```
OliverOffing, line 65: This was a typo in the original documentation.

README.md
```diff
@@ -71,6 +75,8 @@ In order to run with a `electrum` compatible server as the backend, in addition
       ELECTRUM_TLS_ENABLED: "false"
 ```
 
+## Configuration overrides
```
OliverOffing, line 78: Adding a new heading here to make it visually clear that this is a separate section.

contributors/OliverOffing.txt
```diff
@@ -0,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of May 1, 2022.
```
OliverOffing, line 1: What else do I need to do here? @cla-bot is still mad at me :(

OliverOffing (commented):
> I agree. But IMO it would be better to rename the option instead of the value. I would rename backend to something like index_engine and keep the existing values.

I think that's a great idea @nymkappa.

wiz:
Hey @OliverOffing thanks for the PR - since @hunicus just rewrote the entire README file, moved the Docker documentation into its own file, and clarified that the electrs backend is optional for address lookups I'm going to close this for now. Hope that helps new users avoid having the same confusion you had :)

OliverOffing:
Hi @wiz, I'm glad to learn the documentation is being looked after.

I'd still have some feedback if you're interested:

- The point about electrum being optional hasn't been made more clear at all—I'd use something to the effect of my copy suggestions if I were you.
- Also, the point about electrum being optional is now hidden inside the docker documentation—I've run into this problem while _not_ on docker—and should be front and foremost as it applies to basically all installation methods.
- We should have a ticket to track [@nymkappa's suggestion of renaming `backend` to `index_engine`](https://github.com/mempool/mempool/pull/1586#issuecomment-1113948505).

Overall, scarcely any of the traps that I'd fallen into have been addressed and I'm pretty sure others will fall into the same traps again but might not open an issue or a PR—so you'll never hear from these experiences. I'd urge you to reconsider this issue. Thanks.

Katnabox:
I hope to learn from this

Issue #1566, opened 2022-04-24
mining dashboard global hashrate is inconsistent across instances


wiz:
### Node A

<img width="177" alt="Screen Shot 2022-04-24 at 12 28 19" src="https://user-images.githubusercontent.com/232186/164955158-09243e75-35ec-402e-b0eb-77b54ac0fcb8.png">

### Node B

<img width="161" alt="Screen Shot 2022-04-24 at 12 28 24" src="https://user-images.githubusercontent.com/232186/164955156-8260f789-681e-4613-909c-97c22c1a4438.png">

PR #1590, opened 2022-04-30
Run hashrate indexing after midnight

nymkappa:
Related to https://github.com/mempool/mempool/issues/1566

This PR makes hashrate indexing more consistent between instances, by running indexing around midnight UTC for everyone.

Note that if a block is found at midnight, then the network hashrate may still be slightly different between instances due to block propagation delays. This is fine as it won't impact much the estimated hashrate, values between instances will be very close anyway.

### Testing

After deploying the PR you must wait for the current UTC day to end
* After midnight UTC, check if `/graphs/mining/hashrate-difficulty` has a new entry after midnight (daily network hashrate indexing), not that the new value may take some time to show due to caching, you can clear your cache in your network tab if needed
* At the beginning of the next week (Monday, a bit after midnight UTC), check if `/graphs/mining/pools-dominance` has new entry. The previous remarks about caching also applies

### DIFF
--- a/backend/src/api/mining.ts
+++ b/backend/src/api/mining.ts
@@ -139,10 +139,13 @@ class Mining {
 
     try {
       this.weeklyHashrateIndexingStarted = true;
+      const lastestRunDate = await HashratesRepository.$getLatestRun('last_weekly_hashrates_indexing');
 
-      // We only run this once a week
-      const latestTimestamp = await HashratesRepository.$getLatestRunTimestamp('last_weekly_hashrates_indexing') * 1000;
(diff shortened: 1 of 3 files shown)


wiz (approved):
Tested ACK @ v2.4.0-dev [[b56f110f](https://github.com/mempool/mempool/commit/b56f110f)]

I deployed and restarted over the weekend, and waited until Monday, May 9th at 1am UTC and the expected results were displayed:
Screen Shot 2022-05-09 at 10 43 34
Screen Shot 2022-05-09 at 10 43 19

PR #1592, opened 2022-04-30
Show current network hashrate and difficulty using core RPC

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1566

This PR shows the result of `getnetworkhashps` and `getdifficulty` bitcoin core rpc result in the mining dashboard, instead of using the latest daily network hashrate snapshot value we've indexed. So it's more real time now (based on the last 120 blocks and current tip, by default in core). Also, we don't need to wait for indexing to complete before showing those values anymore.

As a side note, I had to disable the auto refresh (every 60 seconds) of the data. So after this PR a page reload will be required for displaying the chart once indexing is completed. I think this is okay because we will push indexing progress on the websocket (https://github.com/mempool/mempool/pull/1596), so we can use this to refresh the component only when indexing is done in a future PR.

Also updated the hashrate unit as mentioned long time ago.

Screen Shot 2022-05-02 at 5 42 11 PM

### DIFF
--- a/backend/src/routes.ts
+++ b/backend/src/routes.ts
@@ -630,6 +630,7 @@ class Routes {
       res.json({
         hashrates: hashrates,
         difficulty: difficulty,
+        currentHashrate: await bitcoinClient.getNetworkHashPs()
       });
     } catch (e) {
       res.status(500).send(e instanceof Error ? e.message : e);
--- a/backend/src/routes.ts
+++ b/backend/src/routes.ts
@@ -630,7 +630,8 @@ class Routes {
       res.json({
         hashrates: hashrates,
         difficulty: difficulty,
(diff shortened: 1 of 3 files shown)


wiz (changes requested):
Tested NACK @ v2.4.0-dev [[ab28c034](https://github.com/mempool/mempool/commit/ab28c034)]

### Bitcoin Core version v0.21.1 @ node206.fmt
Screen Shot 2022-05-06 at 13 48 39

### Bitcoin Core version v22.0 @ node206.fra
Screen Shot 2022-05-06 at 14 02 56

wiz (approved):
Acktually it was just nginx cache. Tested ACK @ v2.4.0-dev [[ab28c034](https://github.com/mempool/mempool/commit/ab28c034)]

PR #1599, opened 2022-05-03
Revise main readme and docker install docs

hunicus:
These changes overhaul the main README, which ended up including revisions to the Docker install notes since they currently live in the main README.

The goal was to make to make these docs better organized, more readable, and easier to follow.

Production install notes will be revised separately.

Notable changes:
- Add Start9 to one-click installation list
- Move Docker installation docs to `./docker/README`
- Correct Docker settings override example ([proposed changes](https://github.com/mempool/mempool/commit/b6c80c7dc395664752e63722da39d2c3d4fa9cca#diff-da8fcbe728a9172b578e5d754f8e2df214c658c4321f610e63dd68bea828ab49R26-R29) vs [existing example](https://github.com/mempool/mempool#bitcoind-only-configuration))
- Remove HA deployment link

PR #1600, opened 2022-05-03
Add screenshot for 2.4.0 to readme

hunicus:
Not sure if too early, but I did this as part of #1599 and realized it should be a separate PR.


softsimon (approved):
utACK

PR #1609, opened 2022-05-07
Add back overwritten sentence to main readme

hunicus:
It was [here](https://github.com/mempool/mempool/pull/1599/commits/a02e44313017d144e2ed569af0864b8e740ec334#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R5) but then got removed [here](https://github.com/mempool/mempool/pull/1600/files).


wiz:
Yeah that was intentional but ok if you want to add it back we can

PR #1625, opened 2022-05-10
Update the recommended node version to v16.5.0 (LTS)

knorrium:


wiz:
Since this is Docker related I suppose we have to tag a dev build, or perhaps it's time for `v2.4.0-alpha1`

wiz (approved):
Tested on FreeBSD baremetal but not Linux/Docker ACK

PR #1627, opened 2022-05-10
Correcting and harmonizing i18n strings

softsimon:
* Harmonized many i18n strings across the project and concerning the mining dashboard
* Changed "Load all" to "View all"
* Corrected capitalization for titles and text on the mining dashboard and graph pages


softsimon:
Strings that I merged duplicates:

* "Timestamp" The same string is now used on Transaction page and Block page.
* "Size" Now same on Block pages and Transaction pages referring to the TX size or Block size
* "Weight" is not the same string on the Transaction and Block page

Restored the following strings as the context is different
* "Confirmed" Bisq block column and Transaction pages

PR #1632, opened 2022-05-11
Reset outspends cache when switching to new tx page.

### LINKED ISSUE #1164: [frontend] caching issue with outspends
reliably reproducible, see the video

[simplescreenrecorder-2022-01-19_15.34.44.mkv.zip](https://github.com/mempool/mempool/files/7897926/simplescreenrecorder-2022-01-19_15.34.44.mkv.zip)
(github doesn't let me upload videos directly but does allow zips)

### LINKED ISSUE #1613: Outspends is broken when navigating block transaction pages
When switching to a new page among the the block transactions, the new outspends requests are not triggered and the wrong red/green arrows are displayed.

I believe I broke this recently so I should be the one to fix it.

softsimon:
When loading more transactions on the address page I wanted to cache the previous outspends in a cleaner way that introduced a bug on the Block transactions page. 

This fix clears the outspends cache when switching page on paginated transactions views.

fixes #1613
fixes #1164


wiz (approved):
Tested ACK @ v2.4.0-dev [[a9012234](https://github.com/mempool/mempool/commit/a9012234)]

PR #1633, opened 2022-05-12
Fix the socks5 connection after updating the socks lib

knorrium:
The `host` field was renamed to `hostname` in a recent release


wiz (approved):
utACK since it's just a variable name change

Issue #1619, opened 2022-05-09
Indexing gaps when the node is offline for too long


knorrium:
Due to the fast forwarding logic, blocks will be skipped by the indexer if the node is offline:

https://github.com/mempool/mempool/blob/953f9405fceb4889edda191dc28bd235bb967a99/backend/src/api/blocks.ts#L293

Visiting the /mining/blocks page will index them on demand but we should detect the gaps and force a reindex.

PR #1635, opened 2022-05-12
Make sure to re-index skipped block when backend is offline for too long

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1619

I've simplified a lot the indexing logic, removing all useless state flags and moving the high level code into an `Indexer` class which make sure we don't have multiple indexing running at the same time as well as ensuring the indexing order (blocks, network hashrate then pool hashrate).

#### Mainnet

I don't think we have a way to emulate the backend being offline (no internet connection) for a long time during run time without changing the code.

I was able to reproduce the bug manually but obviously a real test would be better. Here is a possible way to test the fix.

* Stop node backend
* Delete the `backend/cache` folder content
* Set `INITIAL_BLOCKS_AMOUNT` to 1 in mempool backend config
* Start node backend
* Disconnect your internet connection
* Wait for 2 new blocks to be mined (monitor on mempool.space)
* Reconnect your internet connection
* Wait for 1 new block to be mined
* In the debug log you should notice `...blocks since tip. Fast forwarding to the 1 recent blocks`
* You should also notice `Indexing blocks from ...` (we run the block indexing loop again)
* Confirm that all recent blocks are properly indexed by querying the database. You can also compare with prod.

#### Regtest

* CTRL + Z your node backend
* Generate 100 blocks
* `fg` your node backend and confirm all skipped blocks are properly indexed


nymkappa:
I've tested it by disconnecting my internet connection all nights and reconnect it in the morning. It took a bit of time on my slow internet  but eventually core and the node backend were able to catch up. Fast forward was triggered at some point and the re-indexing happened automatically once core was fully synced. Seems good on my end.

knorrium:
Tested ACK 449d6b17

Killed bitcoind for hours and the backend was able to reindex the missing blocks after restarting bitcoind only.

knorrium (approved):

PR #1637, opened 2022-05-12
Tweaking the websocket retry timers

### LINKED ISSUE #1560: Lower the websocket retry timeout
When the websocket connection disconnects we wait 10 seconds before trying to reconnect again.

This could be lowered to something like 1 or 2 seconds(?).

Or it should start at 0 and increase on each retry.

softsimon:
fixes #1560

1. When going offline the timeout before retrying to reconnect reduced from 10 seconds to 1 second. 
2. The delay between Ping check reduced from 30s to 10s
3. The wait for Pong response increased from 4s to 5s

This is an attempt to improve the stability of the websocket connection.
This also improves the experience when switching networks like VPN as the site will reconnect back faster.


wiz (approved):
Tested ACK @ v2.4.0-dev [[8022cec3](https://github.com/mempool/mempool/commit/8022cec3)]

I have some concerns about reducing the reconnect retry time since it will result in 10x more queries to the servers, but let's try it and see how it goes :)

Issue #1626, opened 2022-05-10
Bisq backend should wait for JSON files to appear instead of exiting immediately


wiz:
### Description

Currently, the Bisq daemon dumps JSON files to a certain location and the mempool backend imports this data.
However, if these JSON files are not present, the backend immediately exits:

```
> mempool-backend@2.4.0-dev start-production
> node --max-old-space-size=4096 dist/index.js

May 10 07:45:37 [98804] DEBUG: <bisq> Starting Mempool Server... (da3c758)
May 10 07:45:37 [98804] INFO: <bisq> Downloading external asset pools.json over clearnet...
May 10 07:45:37 [98804] INFO: <bisq> External asset pools.json saved to ./pools.json
May 10 07:45:37 [98804] INFO: <bisq> Starting currency rates service
May 10 07:45:37 [98804] INFO: <bisq> Currency rates service will be queried over clearnet using https://price.bisq.wiz.biz/getAllMarketPrices
May 10 07:45:37 [98804] DEBUG: <bisq> Querying currency rates service...
May 10 07:45:37 [98804] WARN: <bisq> /bisq/statsnode-data/btc_mainnet/db/json/all/blocks.json doesn't exist. Make sure Bisq is running and the config is correct before starting the server.
```

### Problem to be solved

Instead of giving up and exiting, the backend should wait for the files to be created and valid so it can proceed.

### Proposed solution

Improve the existing file-watcher code that detects when the JSON files are modified so that it is also used at startup.

PR #1642, opened 2022-05-13
If bisq data folder is not ready, retry every 3 minutes instead of exit

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1626

I choose to fix it the simplest way possible, as I did not want to get into existing watchers and potentially break something else.


wiz:
Thanks for working on this. I am preparing a new server to test this and should review in the next 1-2 days.

wiz (approved):
Tested ACK @ v2.4.0-dev [[de8a51fe](https://github.com/mempool/mempool/commit/de8a51fe)]

Excellent, now the bisq backend starts even if bisq is off in the corner chewing paint and patiently waits for bisq to dump the JSON data so it can import it :)

PR #1643, opened 2022-05-13
Lazy load tx inputs in Bitcoin Core mode

### LINKED ISSUE #465: Bitcoind: Lazy load inputs to speed up loading

softsimon:
fixes #465

## The issue
When opening block 736192 (0000000000000000000405c3eb4af4ca6361275baa347c35c8afb6bef09a9d4b) when running Bitcoin Core.. The transactions list is timing out due to the amount of transactions with 100 inputs each.\
Screen Shot 2022-05-13 at 16 01 16

And in other occasions with less inputs the transactions list can be extremely slow when running Bitcoin core, for Raspberry Pi users. 

## What this PR does
Limiting the fetching of inputs to 12 (which is the limit we use before displaying the "Load all (90)" button. Suddenly we can load the page without timeout. And we can flip through pages without super slowdowns. This affects the Block transactions view and Address page view, and it won't effect the individual transaction view.

The only downside is that we can't calculate the fee (missing from down left) without knowing all the inputs.

Screen Shot 2022-05-13 at 15 58 51

After clicking Load all the inputs are lazily loaded:

Screen Shot 2022-05-13 at 15 55 19

Scrolling to the bottom we can now see the fee.

Screen Shot 2022-05-13 at 15 55 49

## How to test
Make sure the features mentioned above should work on the Block transactions page and the Address page containing transactions with many inputs. Going to an individual transaction should still work. This should not be present when using Esplora.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
+++ b/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
@@ -2,7 +2,7 @@ import { IEsploraApi } from './esplora-api.interface';
 
 export interface AbstractBitcoinApi {
   $getRawMempool(): Promise<IEsploraApi.Transaction['txid'][]>;
-  $getRawTransaction(txId: string, skipConversion?: boolean, addPrevout?: boolean): Promise<IEsploraApi.Transaction>;
+  $getRawTransaction(txId: string, skipConversion?: boolean, addPrevout?: boolean, lazyPrevouts?: boolean): Promise<IEsploraApi.Transaction>;
   $getBlockHeightTip(): Promise<number>;
   $getTxIdsForBlock(hash: string): Promise<string[]>;
   $getBlockHash(height: number): Promise<string>;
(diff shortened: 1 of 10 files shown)


antonilol:
maybe explicitly state that the fee is unknown other than just leaving it out

the fee can also be cached or calculated when the block is received and median fee etc is calculated

softsimon:
Will wait before https://github.com/mempool/mempool/pull/1665 is merged.

softsimon:
I have added a explaining text string instead of emptiness.

Screen Shot 2022-05-20 at 21 30 26

softsimon:
A good example

Go to block 0000000000000000000865d5262df0b0ba520b2e869d5dd71b76b49a2a778693
Navigate to last page (29)
This takes so long time that the nginx will timeout error.

Now it only takes 3 seconds at most to load these extreme cases.

wiz (approved):
Tested ACK @ v2.4.0-dev [[dabfa053](https://github.com/mempool/mempool/commit/dabfa053)]

Screen Shot 2022-05-21 at 3 00 10

PR #1649, opened 2022-05-15
pools-updater: Support secure Tor connection to sync data with Github

ayanamidev:
Using Axios instead of native https module from node.js.

Also, I have added a retry in case of rate limit applied due to the use of proxy or maybe some minor connection problem.

### DIFF
--- a/backend/src/tasks/pools-updater.ts
+++ b/backend/src/tasks/pools-updater.ts
@@ -1,8 +1,10 @@
-const https = require('https');
+import axios from 'axios';
 import poolsParser from '../api/pools-parser';
 import config from '../config';
 import DB from '../database';
 import logger from '../logger';
+import { SocksProxyAgent } from 'socks-proxy-agent';
+import * as https from 'https';
 
 /**
  * Maintain the most recent version of pools.json
@@ -28,6 +30,13 @@ class PoolsUpdater {
 
     this.lastRun = now;
 
+    logger.info('Updating latest mining pools from Github');
+    if (config.SOCKS5PROXY.ENABLED) {
+      logger.info('List of public pools will be queried over the Tor network');
+    } else {
+      logger.info('List of public pools will be queried over clearnet');
+    }
+
     try {
       const dbSha = await this.getShaFromDb();
       const githubSha = await this.fetchPoolsSha(); // Fetch pools.json sha from github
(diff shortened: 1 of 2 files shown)


wiz:
@knorrium since you last worked on fetching resource files over Tor, did you want to review this PR

backend/src/tasks/pools-updater.ts
```diff
@@ -105,35 +111,45 @@ class PoolsUpdater {
...
-    });
+      if (config.SOCKS5PROXY.USERNAME && config.SOCKS5PROXY.PASSWORD) {
+        socksOptions.username = config.SOCKS5PROXY.USERNAME;
+        socksOptions.password = config.SOCKS5PROXY.PASSWORD;
+      }
+
+      axiosOptions.httpsAgent = new SocksProxyAgent(socksOptions);
+    }
+
+    while(retry < 5) {
```
ayanamidev, line 139: Should we expose this value to `config.json`?

backend/src/tasks/pools-updater.ts
```diff
@@ -105,35 +111,45 @@ class PoolsUpdater {
...
+        const data = await axios.get(path, axiosOptions);
+        if (data.statusText !== 'OK' || !data.data) {
+          throw new Error(`Could not fetch data from Github, Error: ${data.status}`);
+        }
+        return data.data;
+      } catch (e) {
+        logger.err('Could not connect to Github. Reason: '  + (e instanceof Error ? e.message : e));
+        retry++;
+      }
+      await setDelay();
```
ayanamidev, line 150: Maybe adding this value to config file as well, I think 1 second spacing for retry could not be enough with unstable connections like Tor network.

ayanamidev:
I have added a retry loop so that the worker could retry in case of connection failure or failure with rate limits, could anyone take a look?

backend/src/tasks/pools-updater.ts
wiz: Yeah that would be great

backend/src/tasks/pools-updater.ts
wiz: Probably 10 seconds would be better since it's not urgent to get the updated data?

wiz (approved):
Partially Tested ACK @ v2.4.0-dev [[d1671c4f](https://github.com/mempool/mempool/commit/d1671c4f)]

Only tested over clearnet, so @knorrium will have to test over Tor when he starts the release testing on Umbrel target devices

Issue #1516, opened 2022-04-11
Make copying or selecting addresses easier,


Transisto:
Few problems;

The address is trimmed while there is plenty of room to display it. https://github.com/mempool/mempool/issues/1465

The link area to select goes about 1 inch further to the right than the text is. (see clicking hand)

Solution, put a copy to clipboard button next to each addresses. (might be simpler if the addresses weren't needlessly trimmed)

![image](https://user-images.githubusercontent.com/388716/162766909-d329d0db-bbea-4ef7-855f-2923a0028dd9.png)

Issue #1465, opened 2022-03-28
Address trimming should be dynamic based on windows size.


Transisto:
No matter how much free space there is, about 7 character will be replaced by 3 dots "..." making it hard to CTRL-F an address.

![image](https://user-images.githubusercontent.com/388716/160448513-3df1de95-6a43-4030-b7e0-6658920181d8.png)

nymkappa:
I believe you are requesting two different improvements here:
* Improve truncation behavior so we don't have unnecessary blank spaces
* Provide a way to quickly highlight an address in a transaction. I know we have something for `txid` using url parameters, for example check out https://mempool.space/tx/a30a92627045a077029d0ca4c087062310618ba0b76782fabe4335915d0adaf1:1, maybe we can implement something similar for addresses. Maybe @softsimon has more insight on this.

PRs are welcome.

In the mean time, if you need to repeat this "CTRL F" operation often, try to use the API directly. For example using https://mempool.space/api/tx/a30a92627045a077029d0ca4c087062310618ba0b76782fabe4335915d0adaf1 you can "CTRL F" without any issue. If you use Firefox, the JSON will be automatically formatted, otherwise I recommend using a JSON parser extension.

Transisto:
I just mean that full addresses should be displayed when there's room for
it.

Some people might want to just select and copy it?

softsimon:
I have one suggestion that will solve this in most cases.

Add a monospace font just to display addresses on the page. We then don't have to shorten them to remove risk of overflow.

nymkappa:
> I have one suggestion that will solve this in most cases.
> 
> Add a monospace font just to display addresses on the page. We then don't have to shorten them to remove risk of overflow. Segwit multisig addresses will still be too long but most won't.

Taproot addresses are 62 char long. Even with a monospace font it won't fit.

<img width="1213" alt="Screen Shot 2022-05-16 at 12 01 50 PM" src="https://user-images.githubusercontent.com/9780671/168568958-1f25f771-1a23-416c-8e10-72575ac7fb10.png">

PR #1653, opened 2022-05-16
Allow CTRL F an address on the transaction list component

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1465
Now you can CTRL-F an address in the transaction list component.

I've only applied this change for Bitcoin, but I can apply it everywhere if it's needed.

I've set the font to `monospace` (will use the OS default monospace font) so please let me know if it does not work well on your setup.

I believe this can be further improved to also fix https://github.com/mempool/mempool/issues/1516 later

Screen Shot 2022-05-16 at 4 56 30 PM


wiz (approved):
Tested ACK @ v2.4.0-dev [[9c73e5be](https://github.com/mempool/mempool/commit/9c73e5be)]

Thanks for fixing the display and the CTRL+F functionality on desktop, but the tablet and mobile layouts could use some love too. Maybe make another PR to make the fonts on those consistent with the new monospace font used on desktop?

Issue #1646, opened 2022-05-13
Add new API documentation for the new mining dashboard endpoints


softsimon:
### New endpoints

```
mining/pools/:interval
mining/pools/all
mining/pool/:slug/hashrate
mining/pool/:slug/blocks
mining/pool/:slug/blocks/:height
mining/pool/:slug
mining/pool/:slug/:interval
mining/hashrate/pools
mining/hashrate/pools/:interval
mining/hashrate
mining/hashrate/:interval
mining/reward-stats/:blockCount
mining/blocks/fees/:interval
mining/blocks/rewards/:interval
mining/blocks/fee-rates/:interval
mining/blocks/sizes-weights/:interval
```

# Available intervals:
```

24h
3d
1w
1m
3m
6m
1y
2y
3y

```

hunicus:
@softsimon @nymkappa below are some items that came up as I've been working through this. 

I'm not super familiar with the reasoning that went into making the API as it is, so I'm not saying the following things are necessarily _problems_, but they seem worth pointing out in case they're not intended behavior.

- It seems that `mining/pool/:slug/:interval` returns the same data no matter which interval you specify.
- For `mining/hashrate/pools/[interval]`, 24h and 3d intervals don't seem to work.
  - Also it seems like results are ordered by hashrate, for the most part, but if this is the case, some items are slightly out of order.
- For `mining/pool/:slug/hashrate`, it seems 18 data points are returned representing 9 weeks, with 2 data points per week. What time period is `avgHashrate` measured over?
- Right now I can get a maximum of 75 days / 10 weeks of hashrate and difficulty data from `mining/hashrate/[:interval]`. Will that remain the maximum trailing time period of data available going forward?
- For `mining/blocks/fees/:interval` and `mining/blocks/rewards/:interval`, metrics are labeled as averages but aren't averages (`avg_height`, `avg_rewards`, `avg_fees`). 
  - Also intervals seem to work up to 3m, but with 6m and after they seem to get wonky.
- For `mining/blocks/fee-rates/:interval`, metrics are labeled as averages but aren't averages (`avg_height`, `avg_fee_*`, etc)...the fee figures seem like percentiles.
  - Also intervals seem to work up to 3m, but with 6m and after they seem to get wonky.
- `mining/blocks/sizes-weights/:interval` seems to have the same issue with intervals as the other blocks endpoints, as described above.
[...]

PR #1654, opened 2022-05-16
Add mining api docs

hunicus:
~This is a draft because of pending items listed [here](https://github.com/mempool/mempool/issues/1646#issuecomment-1128083068).~

~Also I still need to hide the JS tabs for these new mining endpoints since they won't be available in `mempool.js` for now.~

Ready for review.


hunicus:
@nymkappa this is ready for review, if you could take a look.

Notes:

- I added a note on the `INDEXING_BLOCKS_AMOUNT` option for the hashrate endpoint, but I'm not sure if it's relevant for any of the other endpoints.
- I kept the original `/api/blocks` example the same (i.e., no `extras` field) and made a new entry for `/api/blocks-extras`.

nymkappa:
* [GET /api/v1/mining/hashrate/pools/[:timePeriod]](https://mempool.ninja/api/v1/mining/hashrate/pools/1m)
While it's not throwing an error, it does not make much sense to query this API with :timePeriod less than `1m` (see https://mempool.space/graphs/mining/pools-dominance). You can hide smaller :timePeriod from the doc.

* [GET /api/v1/mining/hashrate/[:timePeriod]](https://mempool.ninja/api/v1/mining/hashrate/3d)
Similar as above.
Additionally, `currentHashrate` and `currentDifficulty` are real time data queried directly to our Bitcoin Core node. It's the most up to date data, using latest blocks, while arrays returned by this endpoint are historical data, so not real time.

I've noticed some things I need to fix as well which will require updating the doc later (we can make another PR once it's done)
* https://github.com/mempool/mempool/issues/1678
* https://github.com/mempool/mempool/issues/1679

Aside the two first comments, it's good for me.

Issue #1655, opened 2022-05-17
DEBUG: This slug does not match any known pool


wiz:
Please clarify this syslog message, I have no idea what it means and it will confuse users

PR #1657, opened 2022-05-17
Add escaped slug into logs when slug does not match any pool

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1655

The reason we did not include more details in this log was because we're unsure if it's safe to include user input in there. I used mysql2 `escape` function to sanitize the string, I don't know if that's safe enough.

@softsimon


wiz:
I don't understand why we would want to log this at all. What's wrong with silently returning 404?

nymkappa:
> I don't understand why we would want to log this at all. What's wrong with silently returning 404?

As discussed, I've removed the logging of this error. It will still throw a 404 with the invalid mining pool slug used:
> nymkappa@nymkappas-MacBook-Pro ~ % curl -s localhost:4200/api/v1/mining/pool/nymkappa               
This mining pool does not exist 'nymkappa'%

wiz (approved):
Tested ACK @ v2.4.0-dev [[6cd3c312](https://github.com/mempool/mempool/commit/6cd3c312)]

Issue #1605, opened 2022-05-06
Use passive listeners to improve scrolling performance


wiz:
Can you look into this to improve the scrolling performance?
https://web.dev/uses-passive-event-listeners/

nymkappa:
The documentation mentioned
> Add a passive flag to every event listener that Lighthouse identified.

Do you have such a list of identified event listeners? I only found an occurrence of `addEventListener` on the documentation component.

wiz:
Thanks, that might be the only one.

This bug report is simply a copy/paste from the lighthouse report which mostly contains easy to fix things that improve the site, so I guess it's a new one since I ran the test last time.

PR #1658, opened 2022-05-17
Use passive listeners to improve scrolling performance #1605

nymkappa:


wiz (approved):
Tested ACK @ v2.4.0-dev [[25d5b31f](https://github.com/mempool/mempool/commit/25d5b31f)]

Seems nicer?

PR #1659, opened 2022-05-17
Cleanup mining API endpoints

nymkappa:
Fixes some feedback https://github.com/mempool/mempool/issues/1646#issuecomment-1128599061

* We don't allow calling endpoints with empty `:interval` anymore when it's required. Before, we were always using `all` as default value. We now only fallback to `all` if the passed `:interval` is not valid. Maybe this can be further restricted later.
* Removed unused `/mining/pool/:slug/:interval` endpoint (`:interval` was never used)

cc @hunicus


wiz (approved):
Tested ACK @ v2.4.0-dev [[3fd2d74f](https://github.com/mempool/mempool/commit/3fd2d74f)]

@nymkappa Do we need to update the nginx warm cache script?

Issue #1402, opened 2022-03-17
After a difficulty adjustment estimated block times are not 10 minutes


wiz:
Obviously after the difficulty adjusts the estimated block time is exactly 10 minutes, but a wildly off estimate is being used/displayed:

<img width="1663" alt="Screen Shot 2022-03-17 at 23 03 01" src="https://user-images.githubusercontent.com/232186/158902056-d639b12c-f0de-4c3b-a613-f6c32916a374.png">

softsimon:
We should set it to default 10 min as long as our estimate is a dash -, which is the first 146 blocks after an adjustment

PR #1661, opened 2022-05-17
Define avg block time to 10 min until we have at least 146 block in the current epoch

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1402


wiz (approved):
Tested ACK @ v2.4.0-dev [[dddf83a2](https://github.com/mempool/mempool/commit/dddf83a2)]

### Before
Screen Shot 2022-05-18 at 20 07 54

### After
Screen Shot 2022-05-18 at 20 08 00

PR #1664, opened 2022-05-17
Adding some missing i18n strings

softsimon:
* Went through and added i18n (mostly reused) strings where I found it was missing in the tooltips and legends of the new graphs.

Screen Shot 2022-05-18 at 02 39 06
 
* Removed duplicate "Pools Dominance" header due to my bad rebase merge


wiz (approved):
Partially Tested ACK @ v2.4.0-dev [[b7ed45cb](https://github.com/mempool/mempool/commit/b7ed45cb)]

Amazing to see how many strings we missed. We should also add i18n tags on the new FAQ page and other docs when @hunicus finishes them.

PR #1665, opened 2022-05-18
Fix for transaction inputs being fetched recursively

softsimon:
This bug is affecting bitcoin core backend.

When calculating fees and getting the address by fetching the previous-transaction (prevouts), it recursively did the same for the prevout TXs without stopping. 

This PR adds a stop so it only goes 1 layer deep which is the only thing required.

The performance difference of loading the block transactions pages now is extremely noticeable.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api.ts
+++ b/backend/src/api/bitcoin/bitcoin-api.ts
@@ -14,14 +14,31 @@ class BitcoinApi implements AbstractBitcoinApi {
     this.bitcoindClient = bitcoinClient;
   }
 
-  $getRawTransaction(txId: string, skipConversion = false, addPrevout = false, blockHash?: string): Promise<IEsploraApi.Transaction> {
+  static convertBlock(block: IBitcoinApi.Block): IEsploraApi.Block {
+    return {
+      id: block.hash,
+      height: block.height,
+      version: block.version,
+      timestamp: block.time,
+      bits: parseInt(block.bits, 16),
+      nonce: block.nonce,
+      difficulty: block.difficulty,
+      merkle_root: block.merkleroot,
+      tx_count: block.nTx,
+      size: block.size,
+      weight: block.weight,
+      previousblockhash: block.previousblockhash,
+    };
+  }
+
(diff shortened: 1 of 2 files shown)


backend/src/api/bitcoin/bitcoin-api.ts
```diff
@@ -174,35 +191,20 @@ class BitcoinApi implements AbstractBitcoinApi {
...
-    } else {
-      esploraTransaction = await this.$appendMempoolFeeData(esploraTransaction);
-      if (addPrevout) {
-        esploraTransaction = await this.$calculateFeeFromInputs(esploraTransaction, addPrevout);
+    if (addPrevout) {
+      if (transaction.confirmations) {
+        esploraTransaction = await this.$calculateFeeFromInputs(esploraTransaction);
+      } else {
+        esploraTransaction = await this.$appendMempoolFeeData(esploraTransaction);
+        if (addPrevout) {
```
nymkappa, line 199: This second `if (addPrevout)` is unnecessary because it's already nested in the exact same test

nymkappa (approved):
I don't know that part of the code very well, but loading blocks is much faster when `config.MEMPOOL.BACKEND` is set to `none` with this PR.

So I'll tested ACK @ caff758ef99ee8a01792d9357c0c6c9f25c73f7b

wiz (approved):
Tested ACK @ v2.4.0-dev [[f91d9239](https://github.com/mempool/mempool/commit/f91d9239)]

### Before
```
% time curl -s -i http://127.0.0.1:8999/api/v1/tx/aeaadc58c6f3e7ee08c2a2a46f0174ea712d18ac06478d898af6798dae3becb1
HTTP/1.1 200 OK
...
curl -s -i   0.00s user 0.01s system 0% cpu 11.501 total
```
### After

```
% time curl -s -i http://127.0.0.1:8999/api/v1/tx/aeaadc58c6f3e7ee08c2a2a46f0174ea712d18ac06478d898af6798dae3becb1         
HTTP/1.1 200 OK
...
curl -s -i   0.00s user 0.01s system 0% cpu 1.437 total
```

PR #1666, opened 2022-05-18
Remove unused config from nginx

ayanamidev:
Since the express router does the proxy for backend api endpoint https://github.com/mempool/mempool/blob/master/backend/src/index.ts#L231,

I think it would be good to remove the proxy_pass directive to avoid confusion for users installing mempool instances.


wiz (approved):
utACK, will need @knorrium to test later on that all About page functionality still works being proxied thru the backend instead of nginx

PR #1667, opened 2022-05-18
Improving loading indicator UX

softsimon:
**Problem**
When running Bitcoin Core and loading block transaction pages and address page transactions, the loading indicator that pops up causes the browser window to jump or flicker.

**Solution**
Moved the progress bar from the bottom to the txid skeleton placeholder while it's loading to not cause any UI-jumping.

**New UX:**
Screen Shot 2022-05-18 at 08 15 32

**How to test**
Switch between transaction pages on Block page and load more transactions on Address page in Bitcoin Core mode. Loading indicators are not visible with Esplora.


nymkappa (approved):
The UI is a bit more stable when switching pages on /blocks/XXX or when scrolling down on /address/XXX

Tested ACK @ cac8c717ad6bf3306a04b3efedd9cdf8a6d3cd9b

PR #1675, opened 2022-05-19
Fix for reward stats race condition trigger

softsimon:
Reward stats was sometimes triggered multiple times.

This fix should make sure that reward stats is fetched only first time, and then on new blocks.

**EDIT** I am not able to reproduce the race condition bug... So the original code "should" work fine. I can't explain the reason for the blocks$ to get triggered before the $getRewardStats$ has returned any result.


softsimon:
Closing as this is not a race condition bug.

PR #1682, opened 2022-05-19
Moving graphs and mining dashboard to a lazy loaded module

softsimon:
This is an attempt to separate the EChart and Graphs component to a separate lazy loaded module.

So if the user directly navigates to
`/tx/`
`/block/`
`/address/`
and other pages that doesn't contain a graph, that module won't be loaded and we should increase the initial loading time a lot.

Upon loading
`/`
`/mining`
`/graphs`
`/tv`

The lazy loaded bundle containing the graphs will be loaded.

The uncompressed main bundle is down from **2.3 MB** to **1 M**B.

To test this we need to make sure that nothing broke. That mainnet, testnet, Liquid and Bisq still works. Also make sure we didn't introduce any other annoyance or flicker. For example opening a block and clicking on a Mempool Block will now cause a small delay because the graph-bundle is required to display a mempool block.


wiz:
> For example opening a block and clicking on a Mempool Block will now cause a small delay because the graph-bundle is required to display a mempool block.

That's because you need to pre-load the lazy-loaded bundles asynchronously in the background so they're already pre-cached in RAM by the time you need them to render the next page.

wiz (approved):
Tested ACK @ v2.4.0-dev [[062a864a](https://github.com/mempool/mempool/commit/062a864a)]

Excellent refactoring, huge performance gains for FCP 🚀 

### Before (Lighthouse Report)
Screen Shot 2022-05-20 at 20 04 58

### After
Screen Shot 2022-05-20 at 20 06 34

Issue #1679, opened 2022-05-19
Use camel case for /mining/blocks/XXX APIs


nymkappa:
`avg_height` => `avgHeight`
`avg_fees` => `avgFees`
`avg_rewards` => `avgRewards`
`avg_fee_XXX` => `avgFee_XXX`
`avg_size` => `avgSize`

This would match `avgHashrate`

PR #1685, opened 2022-05-20
Update case in some mining API endpoint response

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1679

Documentation has already been updated, but need to be double checked.


hunicus (approved):
utACK

Seems straightforward.

wiz (approved):
Tested ACK @ v2.4.0-dev [[f42da0e3](https://github.com/mempool/mempool/commit/f42da0e3)]

Seems fine from quick testing, but definitely needs full release testing

PR #1686, opened 2022-05-20
Set /mining/blocks/xxx APIs expiration to 60 seconds instead of 5 min

nymkappa:
I'm not sure why I've set those API endpoint expiration to 5 minutes since they depend on blocks. I've set them to 60 seconds now.

```
/mining/blocks/fees/:interval
/mining/blocks/rewards/:interval
/mining/blocks/fee-rates/:interval
/mining/blocks/sizes-weights/:interval
```


wiz (approved):
utACK

PR #1687, opened 2022-05-20
Replace all oldestIndexedBlockTimestamp with X-total-count header

nymkappa:
Depends on https://github.com/mempool/mempool/pull/1685 which must be merged first to avoid conflict.

Fixes https://github.com/mempool/mempool/issues/1678

Documentation has already been updated, but need to be double checked.


hunicus (approved):
Could it be worthwhile to mention `X-Total-Count` in docs (not familiar with reasoning for the move to HTTP header). 

Otherwise ACK for the docs part.

nymkappa:
> Could it be worthwhile to mention `X-Total-Count` in docs? Otherwise ACK for the docs part.

I think it's worth mentioning it for people who use their own instance API, as they may get incomplete data if they don't index all blocks.

wiz:
plz2rebase

wiz (commented):
Tested @ v2.4.0-dev [[df9c9e33](https://github.com/mempool/mempool/commit/df9c9e33)]

Works fine, but what is the purpose of this PR ? The linked issue has no description, so I don't understand why we are replacing a JSON field with a HTTP header

PR #1691, opened 2022-05-20
Handle Error with basic retry while syncing external assets ( Price Data )

ayanamidev:
- [x] Removed unused External Assets value
- [x] Make static URL dynamic
- [x] Added config options for syncing pool data
- [x] Added retry interval & max retry


backend/mempool-config.sample.json
```diff
@@ -15,9 +15,9 @@
     "INDEXING_BLOCKS_AMOUNT": 11000,
     "PRICE_FEED_UPDATE_INTERVAL": 600,
     "USE_SECOND_NODE_FOR_MINFEE": false,
-    "EXTERNAL_ASSETS": [
-      "https://raw.githubusercontent.com/mempool/mining-pools/master/pools.json"
-    ],
+    "EXTERNAL_ASSETS": [],
```
ayanamidev, line 18: Since this resource is already being fetched & served with https://github.com/mempool/mempool/blob/master/backend/src/tasks/pools-updater.ts, fetching same resource with https://github.com/mempool/mempool/blob/master/backend/src/sync-assets.ts would be useless

backend/mempool-config.sample.json
```diff
@@ -66,13 +66,22 @@
...
+    "CLEARNET_URL": "https://price.bisq.wiz.biz/getAllMarketPrices",
+    "BISQ_URL": "https://bisq.markets/api",
+    "BISQ_ONION": "http://bisqmktse2cabavbr2xjq7xw3h6g5ottemo5rolfcwt6aly6tp5fdryd.onion/api"
+  },
+  "EXTERNAL_DATA_SERVER": {
+    "MEMPOOL_API": "https://mempool.space/api/v1",
+    "MEMPOOL_ONION": "http://mempoolhqx4isw62xs7abwphsq7ldayuidyx2v2oethdhhj6mlo2r6ad.onion/api/v1",
+    "LIQUID_API": "https://liquid.network/api/v1",
+    "LIQUID_ONION": "http://liquidmom47f6s3m53ebfxn47p76a6tlnxib3wp6deux7wuzotdr6cyd.onion/api/v1"
   }
```
ayanamidev, line 86: Exposed static URLs used to fetch resources to `mempool-config.json` ( So it could be updated by users if they have an alternative & Would be better to manage external resources )

backend/src/tasks/pools-updater.ts
```diff
@@ -136,18 +136,18 @@ class PoolsUpdater {
       axiosOptions.httpsAgent = new SocksProxyAgent(socksOptions);
     }
 
-    while(retry < 5) {
+    while(retry < config.MEMPOOL.EXTERNAL_MAX_RETRY) {
       try {
-        const data = await axios.get(path, axiosOptions);
-        if (data.statusText !== 'OK' || !data.data) {
+        const data: AxiosResponse = await axios.get(path, axiosOptions);
+        if (data.statusText === 'error' || !data.data) {
```
ayanamidev, line 142: Changed this one since not every servers respond with `statusText === "OK"`

```js
{
  status: 200,
  statusText: '',
  headers: {
    'content-type': 'application/json',
    'transfer-encoding': 'chunked',
    date: 'Fri, 20 May 2022 13:54:11 GMT',
    connection: 'close'
  },
  config: {
    transitional: {
      silentJSONParsing: true,
      forcedJSONParsing: true,
      clarifyTimeoutError: false
    },
    adapter: [Function: httpAdapter],
    transformRequest: [ [Function: transformRequest] ],
    transformResponse: [ [Function: transformResponse] ],
    timeout: 30000,
    xsrfCookieName: 'XSRF-TOKEN',
    xsrfHeaderName: 'X-XSRF-TOKEN',
    maxContentLength: -1,
    maxBodyLength: -1,
    env: { FormData: [Function] },
    validateStatus: [Function: validateStatus],
    headers: {
      Accept: 'application/json, text/plain, */*',
      'User-Agent': 'mempool/v2.4.0-dev'
    },
    httpAgent: SocksProxyAgent {
      _events: [Object: null prototype] {},
      _eventsCount: 0,
      _maxListeners: undefined,
      timeout: null,
      maxFreeSockets: 1,
      maxSockets: 1,
      maxTotalSockets: Infinity,
      sockets: {},
      freeSockets: {},
      requests: {},
      options: {},
      shouldLookup: false,
      proxy: [Object],
      tlsConnectionOptions: {},
      promisifiedCallback: [Function: callback],
      [Symbol(kCapture)]: false
    },
    method: 'get',
    url: 'http://wizpriceje6q5tdrxkyiazsgu7irquiqjy2dptezqhrtu7l2qelqktid.onion/getAllMarketPrices',
    data: undefined
  },
}
```

backend/src/tasks/pools-updater.ts
```diff
@@ -136,18 +136,18 @@ class PoolsUpdater {
...
           throw new Error(`Could not fetch data from Github, Error: ${data.status}`);
         }
         return data.data;
       } catch (e) {
         logger.err('Could not connect to Github. Reason: '  + (e instanceof Error ? e.message : e));
         retry++;
       }
-      await setDelay();
+      await setDelay(config.MEMPOOL.EXTERNAL_RETRY_INTERVAL);
```
ayanamidev, line 150: Exposed this value to configuration as per discussed from https://github.com/mempool/mempool/pull/1649#discussion_r873354446

ayanamidev:
This PR is an attempt to add missing configuration value & handle errors with retrys for unstable networks like tor.

knorrium (changes requested):
@ayanamidev thanks for these changes.

When adding new variables to the config file, we also need to update a few other things related to the Docker image:

1) The README sections that explains the overrides: https://github.com/mempool/mempool/blob/b4beb29f31170a6847b288c01a0275ce4546beeb/docker/README.md?plain=1#L73
2) The template backend config file that is used by the startup script: https://github.com/mempool/mempool/blob/b4beb29f31170a6847b288c01a0275ce4546beeb/docker/backend/mempool-config.json#L17
3) The startup script itself, that replaces the values in the template with defaults or the overrides specified in the docker-compose.yml file: https://github.com/mempool/mempool/blob/b4beb29f31170a6847b288c01a0275ce4546beeb/docker/backend/start.sh#L74
4) The docker-compose-yml file depending on what's being added. In your case you don't need to add any overrides at this time.

backend/src/config.ts
```diff
@@ -145,14 +158,23 @@ const defaults: IConfig = {
...
     'PORT': 9050,
     'USERNAME': '',
     'PASSWORD': ''
   },
   "PRICE_DATA_SERVER": {
     'TOR_URL': 'http://wizpriceje6q5tdrxkyiazsgu7irquiqjy2dptezqhrtu7l2qelqktid.onion/getAllMarketPrices',
-    'CLEARNET_URL': 'https://price.bisq.wiz.biz/getAllMarketPrices'
+    'CLEARNET_URL': 'https://price.bisq.wiz.biz/getAllMarketPrices',
+    'BISQ_URL': 'https://bisq.markets/api',
+    'BISQ_ONION': 'http://bisqmktse2cabavbr2xjq7xw3h6g5ottemo5rolfcwt6aly6tp5fdryd.onion/api'
```
knorrium, line 171: @ayanamidev shouldn't this be under `EXTERNAL_DATA_SERVER` as well?

backend/src/config.ts
```diff
@@ -145,14 +158,23 @@ const defaults: IConfig = {
...
+  },
+  "EXTERNAL_DATA_SERVER": {
+    'MEMPOOL_API': 'https://mempool.space/api/v1',
+    'MEMPOOL_ONION': 'http://mempoolhqx4isw62xs7abwphsq7ldayuidyx2v2oethdhhj6mlo2r6ad.onion/api/v1',
+    'LIQUID_API': 'https://liquid.network/api/v1',
+    'LIQUID_ONION': 'http://liquidmom47f6s3m53ebfxn47p76a6tlnxib3wp6deux7wuzotdr6cyd.onion/api/v1'
```
knorrium, line 177: @ayanamidev I think these should be called `MEMPOOL_CLEARNET`, `LIQUID_CLEARNET` and `BISQ_CLEARNET`, containing just the base URL, then `/api/v1` gets appended at the time you make the request.

backend/src/config.ts
ayanamidev: @knorrium You are right, it would make more sense to put them as `EXTERNAL_DATA_SERVER`

backend/src/config.ts
ayanamidev: @knorrium In case that the configuration should be changed to connect localhost instance without the help of the nginx reverse proxy.

https://mempool.space/api/donations

Also, accessing /api/donations endpoint will result in `endpoint does not exist "/donations"` error.

https://mempool.space/api/v1/donations

While this one works well

backend/src/api/fiat-conversion.ts
```diff
@@ -33,48 +35,75 @@ class FiatConversion {
...
-        }
+    let retry = 0;
 
-        const agent = new SocksProxyAgent(socksOptions);
-        fiatConversionUrl = config.PRICE_DATA_SERVER.TOR_URL;
-        logger.debug('Querying currency rates service...');
-        response = await axios.get(fiatConversionUrl, { httpAgent: agent, headers: headers, timeout: 30000 });
-      } else {
-        fiatConversionUrl = config.PRICE_DATA_SERVER.CLEARNET_URL;
+    while(retry < config.MEMPOOL.EXTERNAL_MAX_RETRY) {
```
knorrium, line 58: @ayanamidev I missed this on the previous PR but the retries are blocking the backend from performing other tasks. Notice the logs after getting 403s below

```
May 23 21:37:58 [99503] NOTICE: Mempool Server is running on port 8999
May 23 21:37:58 [99503] INFO: Updating latest mining pools from Github
May 23 21:37:58 [99503] INFO: List of public pools will be queried over the Tor network
May 23 21:37:58 [99503] ERR: Cannot fetch pools.json sha from db. Reason: connect ENOENT /var/run/mysql/mysql.sock
May 23 21:37:59 [99503] ERR: Could not connect to Github. Reason: Request failed with status code 403
May 23 21:37:59 [99503] DEBUG: USD Conversion Rate: 29306.538
May 23 21:39:00 [99503] ERR: Could not connect to Github. Reason: Request failed with status code 403
May 23 21:40:01 [99503] ERR: Could not connect to Github. Reason: Request failed with status code 403
May 23 21:41:03 [99503] ERR: Could not connect to Github. Reason: Request failed with status code 403
May 23 21:42:04 [99503] ERR: Could not connect to Github. Reason: Request failed with status code 403
May 23 21:43:05 [99503] DEBUG: Pools.json sha | Current: undefined | Github: cc8d3e01dea2787525ed1f182a0f9a685fe3a850
May 23 21:43:05 [99503] WARN: Pools.json is outdated, fetch latest from github
May 23 21:43:07 [99503] DEBUG: Parse coinbase_tags
May 23 21:43:07 [99503] DEBUG: Parse payout_addresses
May 23 21:43:07 [99503] DEBUG: Identify unique mining pools
May 23 21:43:07 [99503] DEBUG: Found 136 unique mining pools
May 23 21:43:07 [99503] ERR: Cannot get existing pools from the database, skipping pools.json import
May 23 21:43:07 [99503] ERR: Cannot save github pools.json sha into the db. Reason: connect ENOENT /var/run/mysql/mysql.sock
May 23 21:43:07 [99503] NOTICE: PoolsUpdater completed
May 23 21:43:07 [99503] DEBUG: Initial difficulty adjustment data set.
May 23 21:43:07 [99503] DEBUG: New block found (#737663)!
May 23 21:43:07 [99503] DEBUG: Indexing tx 1 of 830 in block #737663
May 23 21:43:07 [99503] DEBUG: 0 of 830 found in mempool. 1 fetched through backend service.
```

or when we can't connect to the proxy:

```
May 23 22:00:04 [102230] INFO: Updating latest mining pools from Github
May 23 22:00:04 [102230] INFO: List of public pools will be queried over the Tor network
May 23 22:00:04 [102230] ERR: Cannot fetch pools.json sha from db. Reason: connect ENOENT /var/run/mysql/mysql.sock
May 23 22:00:04 [102230] ERR: Could not connect to Github. Reason: connect ECONNREFUSED 127.0.0.1:9051
retry: 1
May 23 22:01:04 [102230] DEBUG: Querying currency rates service...
May 23 22:01:04 [102230] ERR: Error updating fiat conversion rates: connect ECONNREFUSED 127.0.0.1:9051
May 23 22:01:04 [102230] ERR: Could not connect to Github. Reason: connect ECONNREFUSED 127.0.0.1:9051
retry: 2
May 23 22:02:04 [102230] DEBUG: Querying currency rates service...
May 23 22:02:04 [102230] ERR: Error updating fiat conversion rates: connect ECONNREFUSED 127.0.0.1:9051
May 23 22:02:04 [102230] ERR: Could not connect to Github. Reason: connect ECONNREFUSED 127.0.0.1:9051
retry: 3

``` 

```

File: backend/mempool-config.sample.json [hunk 8]
```diff
@@ -15,9 +15,10 @@
     "INDEXING_BLOCKS_AMOUNT": 11000,
     "PRICE_FEED_UPDATE_INTERVAL": 600,
     "USE_SECOND_NODE_FOR_MINFEE": false,
-    "EXTERNAL_ASSETS": [
-      "https://raw.githubusercontent.com/mempool/mining-pools/master/pools.json"
-    ],
+    "EXTERNAL_ASSETS": [],
+    "EXTERNAL_MAX_RETRY": 10,
+    "EXTERNAL_RETRY_INTERVAL": 60,
```
knorrium, line 20: @ayanamidev this should be in milliseconds to match `POLL_RATE_MS` set just a few lines above

backend/package.json
```diff
@@ -41,8 +41,9 @@
   },
   "devDependencies": {
     "@types/compression": "^1.7.2",
-    "@types/ws": "~8.5.3",
     "@types/express": "^4.17.13",
+    "@types/ws": "~8.5.3",
+    "cross-env": "^7.0.3",
```
knorrium, line 46: @ayanamidev why is this dependency being added? It's no longer maintained and doesn't seem to be needed or in scope for this change

knorrium (changes requested):

backend/src/api/fiat-conversion.ts
ayanamidev: This seems odd, maybe the code isn't asynchronous and would run with await keyword instead which is not very efficient as well

backend/mempool-config.sample.json
ayanamidev: No problem

backend/package.json
ayanamidev: @knorrium Seems like adding `node` would be sufficient to enable `npm run build` for Windows. Replacing this one.

backend/src/api/fiat-conversion.ts
ayanamidev: @knorrium I believe you should ask @nymkappa about the issue since the reason for blocking tasks is mainly the pool updater has 'await' keyword to wait until Promise resolves https://github.com/mempool/mempool/blame/master/backend/src/index.ts#L155.

And since the indexer function https://github.com/mempool/mempool/blame/master/backend/src/index.ts#L158 depends on https://github.com/mempool/mempool/blame/master/backend/src/index.ts#L155, removing the await keyword would likely break other updater's behavior.

So my recommendation is to go as is but if people want, they would be able to reduce the retry limit per job.

backend/mempool-config.sample.json
ayanamidev: @knorrium Checking this again looks like there is a similar config called `PRICE_FEED_UPDATE_INTERVAL` which is a config defined with seconds not ms. If the config key doesn't include the term `MS`, wouldn't it be good to leave this value as is?

backend/src/api/fiat-conversion.ts
ayanamidev: Have addressed issues with Tor connections https://github.com/mempool/mempool/pull/1691/commits/a1fbe582b4724fdb43b6208ac8d6dc852d8a3893, maybe try this commit.

backend/package.json
```diff
@@ -20,8 +20,8 @@
   ],
   "main": "index.ts",
   "scripts": {
-    "ng": "./node_modules/@angular/cli/bin/ng",
-    "tsc": "./node_modules/typescript/bin/tsc",
+    "ng": "node ./node_modules/@angular/cli/bin/ng",
+    "tsc": "node ./node_modules/typescript/bin/tsc",
```
wiz, line 24: This looks committed by mistake - obviously not related to price data :)

wiz:
Sorry, what is the goal of this PR exactly? The linked issue doesn't seem to be related to the majority of the PR, so I'm very confused.

backend/package.json
ayanamidev: It isn't, maybe I should separate the PR to avoid confusion, but this is what enables building for Windows environment

ayanamidev:
@wiz 

1. Automatic retry after immediate failure ( With timeout of course )

ayanamidev:
@wiz

1. Adding automatic retry for fetching external assets after immediate failure ( With time spacing of pre defined seconds )

2. Expose hardcoded URL value to config so that we could update the external server anytime ( If the static URL is blocked by some circumstances, people could use an alternative URL or mirror of it )

3. Customizable 'User-Agent' header value to not leak privacy that we are running an mempool instance

ayanamidev:
Rebased conflicts

softsimon (approved):
Can't speak for the docker file.

But other than some simple TSLint warnings the code looks good to me.

utACK @ 56dc3376

ayanamidev:
@softsimon Shouldn't the TSLint considered as deprecated? https://www.npmjs.com/package/tslint

I use ESLint for TypeScript btw https://github.com/ayanamitech/axios-auto/blob/main/.eslintrc.json

knorrium (approved):
Tested ACK @ 56dc3376

Will retest again on Umbrel under the next RC tag

PR #1697, opened 2022-05-20
Fix websockets tab

hunicus:
Caused by show/hide property for JS examples introduced in e8ef4a39ec247e075fb4cb98ee38d0dedb3f141a.


wiz (approved):
Tested ACK @ v2.4.0-dev [[85ce3ba9](https://github.com/mempool/mempool/commit/85ce3ba9)]

Issue #1704, opened 2022-05-21
Improve latest blocks widget responsiveness


knorrium:
The TXs column is missing on smaller screens:

![Screenshot from 2022-05-21 09-30-08](https://user-images.githubusercontent.com/100320/169660865-c2af0f8d-21b1-480f-8a53-09e39b11c274.png)

PR #1708, opened 2022-05-23
Fix missing tx column in main dashboard on mobile

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1704

### DIFF
--- a/frontend/src/app/dashboard/dashboard.component.scss
+++ b/frontend/src/app/dashboard/dashboard.component.scss
@@ -198,9 +198,7 @@
     display: none;
     text-align: right;
     width: 20%;
-    @media (min-width: 376px) {
-      display: table-cell;
-    }
+    display: table-cell;
   }
   .table-cell-size {
     display: none;


wiz:
Tested but I don't see any change in behavior from prod.
Maybe @knorrium can verify this fix does anything?

Screen Shot 2022-05-25 at 17 26 00
Screen Shot 2022-05-25 at 17 25 20

nymkappa:
@wiz This must have break with a recent change because I don't see it in prod as well. But on `master` branch you can see the last column disappear if you set your screen width to 375 or lower.

wiz:
Sorry, disregard previous comment, need to re-test

wiz (approved):
Tested ACK @ v2.4.0-dev [[c16c7ee0](https://github.com/mempool/mempool/commit/c16c7ee0)]

### Before
Screen Shot 2022-05-25 at 18 16 44

### After
Screen Shot 2022-05-25 at 18 16 51

PR #1714, opened 2022-05-24
Move manual install notes to separate docs

hunicus:
These changes move manual install docs off the main README to separate docs, emphasize 1-click installs, and clarify some recently-asked questions (e.g., electrs being optional).

[See result here](https://github.com/hunicus/mempool/tree/install-manual).


wiz:
Very nice job on these docs, the new architecture of putting everything into the sub-folders feels much cleaner, and now we can put more app screenshots and app store screenshots in the top-level README. I need to read through them more so will assign this to myself for now.

wiz:
> Node.js 16 and npm 7 are recommended.

We currently require nodeJS 16.15, this is not a recommendation, but a requirement

PR #1717, opened 2022-05-24
NgBootstrap library import optimization

softsimon:
Using the recommended import method from here https://ng-bootstrap.github.io/#/getting-started to only import the NgbBoostrap modules we are needed instead of the whole library.

I didn't see any immediate size differences in the bundle..


wiz (approved):
Tested ACK @ v2.4.0-dev [[63ab2d7c](https://github.com/mempool/mempool/commit/63ab2d7c)]

Seems fine

Issue #1715, opened 2022-05-24
Pools parser and pools updater tasks do not check if the database is enabled before attempting to connect


knorrium:
Disabling the database causes issues with the pools tasks and probably other mining tasks too. I don't think the mining dashboard can work properly without a database but we should handle this case both on the backend and frontend.

```
  "DATABASE": {
    "ENABLED": false,
    "HOST": "127.0.0.1",
    "PORT": 3306,
    "SOCKET": "/var/run/mysql/mysql.sock",
    "DATABASE": "mempool",
    "USERNAME": "mempool",
    "PASSWORD": "mempool"
  },
``` 
```
May 23 21:21:30 [98097] ERR: Cannot fetch pools.json sha from db. Reason: connect ENOENT /var/run/mysql/mysql.sock
...
May 23 21:21:33 [98097] ERR: Cannot get existing pools from the database, skipping pools.json import
May 23 21:21:33 [98097] ERR: Cannot save github pools.json sha into the db. Reason: connect ENOENT /var/run/mysql/mysql.sock
```

https://github.com/mempool/mempool/blob/b4beb29f31170a6847b288c01a0275ce4546beeb/backend/src/tasks/pools-updater.ts#L67-L91

https://github.com/mempool/mempool/blob/b4beb29f31170a6847b288c01a0275ce4546beeb/backend/src/api/pools-parser.ts#L60-L67

PR #1719, opened 2022-05-24
Don't try to run pools parser if db is not enabled

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1715


backend/src/index.ts
```diff
@@ -25,9 +25,6 @@ import databaseMigration from './api/database-migration';
 import syncAssets from './sync-assets';
 import icons from './api/liquid/icons';
 import { Common } from './api/common';
-import mining from './api/mining';
```
knorrium, line 28: @nymkappa did you mean to remove these? It's now showing up as a merge conflict

nymkappa: Rebased

wiz (approved):
Partially Tested ACK @ v2.4.0-dev [[288a96ed](https://github.com/mempool/mempool/commit/288a96ed)]

PR #1735, opened 2022-05-26
Remove keybase link from about page

hunicus:
## Before

Screenshot from 2022-05-26 22-21-06

## After

Screenshot from 2022-05-26 22-22-26


wiz (approved):
utACK

PR #1736, opened 2022-05-26
Adjust stuck tx faqs for angry people

hunicus:
Every time fees spike, we see a spike in angry people who think Mempool is holding their bitcoin hostage. This PR's goal is to make the stuck tx FAQs more informative for such people so they can be linked (instead of having to explain the same thing over and over again on Twitter, Telegram, etc).

These updates:
- better address peoples' frustration
- make it clear that there is nothing Mempool can do about such situations since Mempool merely provides data about the Bitcoin network
- suggest that people consult the tool that helped them make the stuck transaction (wallet, exchange company, etc) if they want the transaction to confirm quicker but don't know what to do

Three FAQs are updated. This is how they look:

Screenshot from 2022-05-26 17-19-22


wiz (approved):
utACK

PR #1739, opened 2022-05-27
Use block hash instead of block height in urls

nymkappa:
Use `block.id` (hash) instead of `block.height` in router links


softsimon (changes requested):
When navigating to another page in the block list, the id is not a block hash so the link to the block is broken.

softsimon (approved):
tested ACK @ 99902e7

PR #1738, opened 2022-05-27
Add `economyFee` field in /api/fees/recommended API

nymkappa:
Redo of https://github.com/mempool/mempool/pull/1734
Context https://github.com/mempool/mempool/issues/1726

Added `economyFee` field in `/api/fees/recommended` API which is always `2 x minimumFee`.

Issue #1726, opened 2022-05-25
Improve the fee suggestion widget, add new "economy" fee rate


wiz:
### Problem
<img width="630" alt="Screen Shot 2022-05-25 at 19 23 08" src="https://user-images.githubusercontent.com/232186/170240976-d2bfa0c5-0628-474f-a815-c3b3aafbed4d.png">

In this type of mempool situation, people complain "low" is too high, so we need to add an "economy" rate which is somewhere in between "low" and "minimum", and we also need to move the minimum fee from the 2nd widget up to the top widget, so that there is a sliding scale with 5 fee rates, indicating to the user that it is more of a *suggestion* on a *sliding scale*, that should be combined with your own logic, rather than something that should be taken as-is

See the Sparrow Wallet UI for inspiration, animated GIF below:

https://user-images.githubusercontent.com/232186/170242339-557fc055-2f5c-42e5-a4e5-45c9a0f7b600.mp4

Source: https://twitter.com/SparrowWallet/status/1370011461238554625

nymkappa:
### Design

Idea/general direction

![fee_concept](https://user-images.githubusercontent.com/9780671/170481978-b226bb26-b148-4de4-aed2-f33b685e6419.png)

### Specs

* 5 fee "tiers"
  * Minimum
  * Economy (2x minimum)
  * Low (same as today's "Low priority")
  * Medium (same as today's "Medium priority")
  * Priority (same as today's "High priority")
* When the mempool is not full, minimum is 1 sat/vB (defined by core RPC), and the 5 tiers colors are visible
* When the mempool is full and the node start purging, show the purge rate, and start shifting color from right to left (therefore the green color will be less visibile)

#### Empty mempool

<img width="732" alt="Screen Shot 2022-05-26 at 2 05 51 PM" src="https://user-images.githubusercontent.com/9780671/170484689-ac65d0cf-660a-4316-8aeb-57d851bb6988.png">

### Mempool full (purging)

_* Obviously fees value are wrong on the screenshot_

<img width="729" alt="Screen Shot 2022-05-26 at 2 06 53 PM" src="https://user-images.githubusercontent.com/9780671/170484811-49b06181-ea7e-4619-8446-bcad195cdc0e.png">

wiz:
Don't forget to update the `frontend/resources/screenshots/v2.4.0-dashboard.png`

PR #1741, opened 2022-05-27
Remove fee calculation from the frontend

nymkappa:
Context https://github.com/mempool/mempool/issues/1726

PR https://github.com/mempool/mempool/pull/1738 must be merged first.

Until now, the frontend was reading mempool blocks from the websocket and computing fees on its own (duplicated backend code).

With this PR, the backend takes care of computing recommended fees and push them through the websocket after each mempool updates (every 2 seconds). The frontend simply listen to updates and display values without any further processing.

### Testing

* Deploy the PR on staging
* Compare prod and staging servers fees and see if they match. Monitor for a couple of minutes to see if fees are approximately the same of both instances. Of course there may be a few sats difference according to tx propagation time.


frontend/src/app/interfaces/websocket.interface.ts
```diff
@@ -65,3 +66,11 @@ export interface IBackendInfo {
   gitCommit: string;
   version: string;
 }
+
+export interface Recommendedfees {
+  fastestFee: number;
+  halfHourFee: number;
+  hourFee: number;
+  minimumFee: number;
+  economyFee: number;
```
wiz, line 75: I think the order should be preserved so @hunicus doesn't nitpick, but other than that looks good :)

PR #1743, opened 2022-05-27
Add Tor support for fetching static assets for frontend

ayanamidev:


frontend/package.json
```diff
@@ -20,9 +20,9 @@
   ],
   "main": "index.ts",
   "scripts": {
-    "ng": "./node_modules/@angular/cli/bin/ng",
-    "tsc": "./node_modules/typescript/bin/tsc",
-    "i18n-extract-from-source": "./node_modules/@angular/cli/bin/ng extract-i18n --out-file ./src/locale/messages.xlf",
+    "ng": "node ./node_modules/@angular/cli/bin/ng",
+    "tsc": "node ./node_modules/typescript/bin/tsc",
+    "i18n-extract-from-source": "node ./node_modules/@angular/cli/bin/ng extract-i18n --out-file ./src/locale/messages.xlf",
```
wiz, line 25: How is this related to adding Tor support to fetch static assets? Was this committed by mistake?

wiz (changes requested):
NACK 

It appears you are adding Tor support into the frontend build process, which is not the correct place to implement this.... The frontend is only built by us once, and then these assets are shipped together along with the docker images downstream to the Umbrel and other fullnode distros. What we need to do is move the fetching of all these static assets into the backend, run as a periodic service like the fiat rates, etc.

ayanamidev:
@wiz Yes this sounds more reasonable, will be submitting the PR for fetching static assets from backend

Issue #1744, opened 2022-05-28
cant read property of undefined


antonilol:
#### Version

mempool.space v2.4.0-dev [774cd98a]
not on bitcoind mempool

### Steps to reproduce

https://mempool.space/testnet/tx/86d5de9fd4e284d285e547216576ca94cdf6fddfe8a7afc9b9dc67fa5afec2db

### Expected behaviour

nothing special

### Actual behaviour

parts of the ui gone and errors in devtools

### Screenshots

![image](https://user-images.githubusercontent.com/49868160/170827704-1ed5618f-1136-463e-b269-4fb6fe821555.png)

#### Device or machine

chromium

#### Additional info

frontend/src/app/components/transactions-list/transactions-list.component.html lines 72 and 73
vin.prevout.scriptpubkey_address is undefined when the script cant be represented by an address
vin.prevout.scriptpubkey_address.length will throw an error when accessed

PR #1745, opened 2022-05-28
stricter multisig check + detect bare multisig when spent

antonilol:
the following script is not a standard bare multisig (wont relay on mainnet) but gets the label because the `detectMultisig` function checks from the end but does not check if there are opcodes and pushes left

```
OP_PUSHBYTES_33 02e8a6b194fc65a21e2400f85fece058b9e1d3c3ad5ef2a4e1f0b5f4bd2d781e97
OP_CHECKSIGVERIFY
OP_PUSHNUM_1
OP_PUSHBYTES_33 031400754ca5d3c257c95ee209874107c3602c90c1b603fab0d9bd78b9e71cf423
OP_PUSHBYTES_33 0389cd2b809c4477725ab452a5238432ed7db6351f4035bf0ffc8da490d98092ba
OP_PUSHBYTES_33 0326d3997d41a2345e49d69ea44852b36ecab419d3c20f9db42f98862eb5ee3970
OP_PUSHNUM_3
OP_CHECKMULTISIG
```

this is fixed + the label is now also added to the input when the a multisig is spent

came across this at the same time i found #1744 but these two are not linked


softsimon (approved):
tested ACK at @ b930b9bf

PR #1751, opened 2022-05-31
Handle nonstandard inputs

### LINKED ISSUE #1744: cant read property of undefined

softsimon:
fixes #1744

**Problem**
Blank row when there is no input address, (and js crash).

**This fix**
Same as we already handle outputs. When the address is non existent instead print out the type which in the reported test case is NONSTANDARD from Bitcoin Core... Note that on Esplora the type is UNKNOWN.

Testnet TX
86d5de9fd4e284d285e547216576ca94cdf6fddfe8a7afc9b9dc67fa5afec2db

Screen Shot 2022-05-31 at 04 12 32

### DIFF
--- a/frontend/src/app/components/transactions-list/transactions-list.component.html
+++ b/frontend/src/app/components/transactions-list/transactions-list.component.html
@@ -66,13 +66,16 @@
                         </ng-template>
                       </ng-template>
                       <ng-template #defaultAddress>
-                        <a [routerLink]="['/address/' | relativeUrl, vin.prevout.scriptpubkey_address]" title="{{ vin.prevout.scriptpubkey_address }}">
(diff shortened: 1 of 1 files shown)


wiz (approved):
Tested ACK @ v2.4.0-dev [[85b17927](https://github.com/mempool/mempool/commit/85b17927)]


Note: displays UNKNOWN instead of NONSTANDARD

Screen Shot 2022-06-01 at 2 41 39

PR #1764, opened 2022-05-31
Only attempt to save fx rate if database is enabled

nymkappa:
Fix bisq instances stuck because they're trying to save the BTC fx rate even though the database is disabled.


wiz (approved):
Tested ACK @ v2.4.0-dev [[c7e9b47a](https://github.com/mempool/mempool/commit/c7e9b47a)]

Issue #1762, opened 2022-05-31
chain divergence detected even when no divergence exists?


wiz:
It's most frequently seen on Testnet:
```
May 31 15:05:23 node201.fra  mempool[83481]: WARN <testnet> Chain divergence detected at block 2252853, re-indexing most recent data
May 31 15:05:24 node201.fra  mempool[83481]: WARN <testnet> Chain divergence detected at block 2252854, re-indexing most recent data
May 31 15:05:24 node201.fra  mempool[83481]: WARN <testnet> Chain divergence detected at block 2252855, re-indexing most recent data
May 31 15:05:24 node201.fra  mempool[83481]: WARN <testnet> Chain divergence detected at block 2252856, re-indexing most recent data
May 31 15:05:24 node201.fra  mempool[83481]: WARN <testnet> Chain divergence detected at block 2252857, re-indexing most recent data
May 31 15:05:24 node201.fra  mempool[83481]: WARN <testnet> Chain divergence detected at block 2252858, re-indexing most recent data
May 31 15:05:25 node201.fra  mempool[83481]: WARN <testnet> Chain divergence detected at block 2252859, re-indexing most recent data
```
But it's also seen on Mainnet:
```
May 31 14:25:12 node201.fra  mempool[60848]: WARN Chain divergence detected at block 738736, re-indexing most recent data
```

PR #1769, opened 2022-06-01
Re-add hash field in the mysql block query

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1762


wiz (approved):
Tested ACK @ v2.4.0-dev [[d860344b](https://github.com/mempool/mempool/commit/d860344b)]

PR #1774, opened 2022-06-01
Feature: Projected block visualization

mononaut:
### Summary

This PR adds a Bitfeed-style visualization to the projected block page.

https://user-images.githubusercontent.com/83316221/171306706-2719e7a8-05d7-4e06-95fb-ccd2a6fe5411.mp4

Demo deployment running at [mempool.monospace.live](https://mempool.monospace.live/mempool-block/0)

The area of each transaction is proportional to its size in vbytes, scaled so that the full visualization area represents the maximum block capacity.

Transactions are sorted right-to-left by fee rate, so that the visualization looks like a zoomed-in view of the corresponding mini projected block at the top.

As the contents of the projected block change, the visualization animates updates to the layout. These layout updates try to affect as few transactions as possible, while keeping everything roughly sorted over time.

Entry and exit transitions are also animated.

Hovering over or clicking transactions in the visualization shows a summary of that transaction on the left side.

The visualization is quietly disabled if the browser doesn't support WebGL.

#### Backend

Extended to support a new 'track-mempool-block' subscription.

Sends a full set of stripped transactions for that block when a client starts the subscription, then continues to send deltas (lists of txids removed and stripped transactions added) every time the mempool changes.

I assume the websocket transport is reliable: sending deltas instead of full blocks cuts the data requirements down significantly, but won't work if updates are dropped or delivered out of order.

The MempoolBlocks class now calculates and caches deltas for each projected block on every mempool update to provide this data.

#### Frontend

The visualization itself is contained within a new component `mempool-block-overview`, used in `mempool-block`.

Everything is rendered on an html canvas using WebGL with some custom shaders. Animations are interpolated entirely within the graphics shaders, so CPU load stays reasonable even with a large number of independently moving squares.

`mempool-block` is extended to display the transaction details on mouse-over, and to detect WebGL support to decide whether to show the visualization.

Added support in `state.service` and `websocket.service` for starting and updating the mempool block subscription, and processing responses.

### DIFF
--- a/backend/src/api/mempool-blocks.ts
+++ b/backend/src/api/mempool-blocks.ts
@@ -112,6 +112,7 @@ class MempoolBlocks {
       medianFee: Common.percentile(transactions.map((tx) => tx.effectiveFeePerVsize), config.MEMPOOL.RECOMMENDED_FEE_PERCENTILE),
       feeRange: Common.getFeesInRange(transactions, rangeLength),
       transactionIds: transactions.map((tx) => tx.txid),
+      transactions: transactions.map((tx) => Common.stripTransaction(tx)),
     };
   }
 }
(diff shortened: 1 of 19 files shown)


ayanamidev:
@mononaut Great feature, thank you!

wiz (approved):
Tested ACK @ v2.4.0-dev [[82471073](https://github.com/mempool/mempool/commit/82471073)]

I think this is the best PR that we've ever received :wink:
Simon's requested changes can be done in a separate PR later.
Thanks so much for this great new feature!

PR #1786, opened 2022-06-02
Adjust HU translation, one strange-looking word

catenocrypt:


catenocrypt:
See also #1787

wiz:
sorry, but translation work is performed on transifex, and imported into the repo that way, so we cannot accept your pull request. please see https://www.transifex.com/mempool/mempool/dashboard/

PR #1797, opened 2022-06-03
Skip blocks disk cache loading after db migration

nymkappa:
Fixes the situation where blocks in disk cache are not matching the same data structure as the codebase, creating bunch of random errors a bit everywhere in the app.

Fixes https://github.com/mempool/mempool/issues/1791

### DIFF
--- a/backend/src/api/disk-cache.ts
+++ b/backend/src/api/disk-cache.ts
@@ -9,6 +9,8 @@ import { TransactionExtended } from '../mempool.interfaces';
 import { Common } from './common';
 
 class DiskCache {
+  private cacheSchemaVersion = 1;
+
   private static FILE_NAME = config.MEMPOOL.CACHE_DIR + '/cache.json';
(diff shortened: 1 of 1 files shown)


backend/src/api/disk-cache.ts
```diff
@@ -57,6 +59,13 @@ class DiskCache {
     }
   }
 
+  wipeCache() {
```
knorrium, line 62: @nymkappa why don’t you scan/filter the cache dir for files instead?

wiz (approved):
Tested ACK @ v2.4.0-dev [[256dbc8c](https://github.com/mempool/mempool/commit/256dbc8c)]

```
Jun  3 14:40:09 node201.fra  mempool[37757]: INFO Restoring mempool and blocks data from disk cache
Jun  3 14:40:09 node201.fra  mempool[37757]: NOTICE Disk cache contains an outdated schema version. Clearing it and skipping the cache loading.
```

PR #1800, opened 2022-06-03
Re-added missing cache version

nymkappa:


wiz (approved):
Tested ACK @ v2.4.0-dev [[3bc55d80](https://github.com/mempool/mempool/commit/3bc55d80)]

PR #1801, opened 2022-06-03
Improve disk cache logging

nymkappa:


wiz (approved):
Tested ACK @ v2.4.0-dev [[b60c2a93](https://github.com/mempool/mempool/commit/b60c2a93)]

```
Jun  3 16:20:56 node201.fra  mempool[54192]: WARN <testnet> Failed to parse mempoool and blocks cache. Skipping. Reason: Cannot read properties of undefined (reading 'timestamp')
```

PR #1803, opened 2022-06-03
Update fee faq

hunicus:
Looks like this:

Screenshot from 2022-06-04 09-49-33


wiz:
I like it but I don't think we can make block targets like "the next few blocks" - maybe more vague phrases like "as soon as possible" for high priority, etc.

hunicus:
It's tricky because without _some_ kind of specificity, the explanation doesn't really explain anything...but then the specificity itself is vague and unpredictable.

I suppose we'll have the same issue with #1804...one difference is there I used "attempt" to convey that the "several blocks" thing is not to be relied on.

hunicus:
It's tricky because without _some_ kind of specificity, the explanation doesn't really explain anything...but then the specificity itself is vague and unpredictable.

I suppose we'll have the same issue with #1804...one difference is there I used "attempt" to convey that the "several blocks" thing is not to be relied on.

PR #1804, opened 2022-06-03
Update fee estimate tooltips

hunicus:
Since this PR replaces the note on fiat fee, I added it to the FAQ with https://github.com/mempool/mempool/pull/1803/commits/67b4a4d3b8ec7e6ddf102827b239f6c05cd5ff52.

All changes in `messages.xlf` are result of running `i18n-extract-from-source`.


wiz:
I think we should add new tooltips on the "high priority" titles, instead of replacing the existing fiat fee calculation tooltip text, especially since it's already translated into all languages

PR #1811, opened 2022-06-05
Reload mempool visualization on reconnect

### LINKED ISSUE #1799: Mempool visualizer: Reload block on reconnect
If the backend is restarted, the visualization gets out of sync and out of whack.

Subscribe to the online/offline observable event and reload the block transactions on reconnect.

softsimon:
fixes #1799

### DIFF
--- a/frontend/src/app/components/mempool-block-overview/mempool-block-overview.component.ts
+++ b/frontend/src/app/components/mempool-block-overview/mempool-block-overview.component.ts
@@ -1,8 +1,9 @@
 import { Component, ElementRef, ViewChild, HostListener, Input, Output, EventEmitter, OnInit,
   OnDestroy, OnChanges, ChangeDetectionStrategy, NgZone, AfterViewInit } from '@angular/core';
 import { StateService } from 'src/app/services/state.service';
-import { MempoolBlockWithTransactions, MempoolBlockDelta, TransactionStripped } from 'src/app/interfaces/websocket.interface';
-import { Subscription, BehaviorSubject } from 'rxjs';
+import { MempoolBlockDelta, TransactionStripped } from 'src/app/interfaces/websocket.interface';
+import { Subscription, BehaviorSubject, merge, of } from 'rxjs';
+import { switchMap, filter } from 'rxjs/operators';
 import { WebsocketService } from 'src/app/services/websocket.service';
 import { FastVertexArray } from './fast-vertex-array';
 import BlockScene from './block-scene';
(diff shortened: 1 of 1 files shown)


wiz (approved):
Tested ACK @ v2.4.0-dev [[967a2a44](https://github.com/mempool/mempool/commit/967a2a44)]

Seems to correct the previous issue we saw where the mempool would look like this
Screen Shot 2022-06-03 at 23 40 26

mononaut:
This didn't totally fix #1799 for me.

I actually couldn't reproduce the issue at all until I deleted the backend cache files, but then it persisted even after applying this PR, so I think there must be an underlying bug.

mononaut:
We already attempt to re-subscribe to `track-mempool-block` events when the websocket reconnects here:
https://github.com/mempool/mempool/blob/967a2a446165e028aa3af84f6705dc954bad2ba2/frontend/src/app/services/websocket.service.ts#L106-L110

which should request a full copy of the current block and re-sync the visualization.

I think the root cause is that after a restart, it's possible for the client to reconnect and send that `track-mempool-block` message *before* the backend has finished loading the mempool. Especially in a dev environment, where client-server communication is fast, but loading the mempool could be slow without blocks already cached.

If that mempool block isn't loaded yet, the backend doesn't respond at all:
https://github.com/mempool/mempool/blob/967a2a446165e028aa3af84f6705dc954bad2ba2/backend/src/api/websocket-handler.ts#L114-L128
Which gets us out of sync. Then the next message is a 'delta' update, which could include a bunch of transactions that the client has already seen, messing up visualization layout.

I haven't tested to verify yet, but I think the fix might be to respond to `track-mempool-block` with an empty list of transactions if data isn't available yet. Something like:

```diff
 if (parsedMessage && parsedMessage['track-mempool-block'] !== undefined) { 
   if (Number.isInteger(parsedMessage['track-mempool-block']) && parsedMessage['track-mempool-block'] >= 0) { 
     const index = parsedMessage['track-mempool-block']; 
     client['track-mempool-block'] = index; 
     const mBlocksWithTransactions = mempoolBlocks.getMempoolBlocksWithTransactions(); 
-    if (mBlocksWithTransactions[index]) { 
-      response['projected-block-transactions'] = { 
-        index: index, 
-        blockTransactions: mBlocksWithTransactions[index].transactions 
-      }; 
-    }
+    response['projected-block-transactions'] = {
+      index: index,
+      blockTransactions: mBlocksWithTransactions[index]?.transactions || []
+    };
   } else { 
     client['track-mempool-block'] = null; 
   } 
 } 
```

softsimon:
> I haven't tested to verify yet, but I think the fix might be to respond to `track-mempool-block` with an empty list of transactions if data isn't available yet. Something like:

I pushed this change. It seems to work.. Thanks for the review!

wiz (approved):
Tested ACK @ v2.4.0-dev [[5360f6dd](https://github.com/mempool/mempool/commit/5360f6dd)]

Can no longer reproduce the issue

Issue #175, opened 2020-11-16
Get hourly or daily average historical Bitcoin to fiat prices in a database to use when displaying fiat price at the time of transaction


wiz:
### Official mempool.space backend
- [x] Implement with 5-6 above exchanges
- [x] Fetch 10-12 currencies from the above exchanges, insert into SQL
- [x] Manually find and insert historical data on the internet to get full history from 2010
~- [ ] Implement new API to expose above historical data over prod hostnames / Tor onion~
- [x] Display historical price for transactions (at the time of the transaction)
 
### Unofficial RPI users backend
- [x] When starting up, check to see what data is missing
~- [ ] Fetch missing historical data from centralized mempool.space oracle over Tor~
~- [ ] Hit mempool.space every hour to get updated fiat rates for the past hour?~

PR #1816, opened 2022-06-06
If price is not yet available, delay saving block rate

nymkappa:
When you start the node backend, bisq rates may not yet be available. Instead of skipping it (current behavior), we now simply wait for bisq rates to be fetched before saving the exchange rate for a block.

We also wait for Core to be synced before saving exchange rate for new blocks, otherwise we may save today's exchange rate for blocks which are much older than today.

#### Testing

* Stop the node backend
* Clear your disk cache, so that `MEMPOOL.INITIAL_BLOCKS_AMOUNT` blocks are fetched at start
* If you already have exchange rate for the `MEMPOOL.INITIAL_BLOCKS_AMOUNT` most recent blocks in your `rates` table, delete their entries
* Start the node backend
* Check the `rates` table and confirm that the exchange rate has properly been inserted for the `MEMPOOL.INITIAL_BLOCKS_AMOUNT` most recent blocks

Without this PR, following the same step would most likely lead to missing exchange rates for the most recent blocks.
Also note that this does not solve the problem of backward filling the table for blocks older than `MEMPOOL.INITIAL_BLOCKS_AMOUNT`.

### DIFF
--- a/backend/src/api/blocks-rates.ts
+++ b/backend/src/api/blocks-rates.ts
@@ -1,0 +1,42 @@
+import logger from "../logger";
+import RatesRepository from "../repositories/RatesRepository";
+import fiatConversion from "./fiat-conversion";
+
+class BlocksRates {
+  private pendingBlocksRates: number[] = [];
+  private saveInProgress = false;
+
+  public async savePendingBlockRates() {
+    if (this.saveInProgress === false && fiatConversion.ratesInitialized === true) {
+      this.saveInProgress = true;
+
+      const pending = [...this.pendingBlocksRates];
+      this.pendingBlocksRates = [];
+
+      for (let height of pending) {
+        try {
+          await RatesRepository.$saveRate(height, fiatConversion.getConversionRates());
+        } catch (e) {
+          this.pendingBlocksRates.push(height);
+          logger.debug(`Cannot save pending block rates, trying again later`);
+        }
+      }
+    }
+
+    this.saveInProgress = false;
+  }
+
+  public async saveRateForBlock(height: number) {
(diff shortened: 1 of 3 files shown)


backend/src/api/blocks-rates.ts
```diff
@@ -0,0 +1,42 @@
...
+          logger.debug(`Cannot save pending block rates, trying again later`);
+        }
+      }
+    }
+
+    this.saveInProgress = false;
+  }
+
+  public async saveRateForBlock(height: number) {
+    if (fiatConversion.ratesInitialized === true) {
```
nymkappa, line 30: If we did not fetch bisq price yet, we simply push the block height into a buffer and try to save the exchange rate later again.

backend/src/api/blocks.ts
```diff
@@ -343,8 +343,11 @@ class Blocks {
           await blocksRepository.$saveBlockInDatabase(blockExtended);
         }
       }
-      if (fiatConversion.ratesInitialized === true && config.DATABASE.ENABLED === true) {
-        await RatesRepository.$saveRate(blockExtended.height, fiatConversion.getConversionRates());
+      if (config.DATABASE.ENABLED === true) {
+        const blockchainInfo = await bitcoinClient.getBlockchainInfo();
+        if (blockchainInfo.blocks === blockchainInfo.headers) {
```
nymkappa, line 348: We wait for Core to be synced before attempting to save block exchange rate, to avoid saving current exchange rate for old blocks (it could happen in the node backend is offline for a long time)

nymkappa:
See https://github.com/mempool/mempool/issues/175

Issue #1520, opened 2022-04-14
Fetch and store historical USD/BTC rates per block


softsimon:
Fetch and continuously store the USD value of bitcoin in the `blocks` table and integrate the historical USD price into the historical block reward graph.

nymkappa:
I can take care of this one. Tho, I think it would be better to have a separate table just for this, let's not touch the `blocks` table for now as it seems to be stable.

If that's okay with you I'll start working on it.

PR #1817, opened 2022-06-06
Add USD serie in block fee/reward charts

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1520

This PR adds a new `USD` series to the block fees and rewards charts. It needs some polish.

Screen Shot 2022-07-09 at 6 27 17 PM
Screen Shot 2022-07-09 at 6 27 48 PM

### DIFF
--- a/backend/src/api/mining/mining.ts
+++ b/backend/src/api/mining/mining.ts
@@ -31,7 +31,7 @@ class Mining {
    */
   public async $getHistoricalBlockFees(interval: string | null = null): Promise<any> {
     return await BlocksRepository.$getHistoricalBlockFees(
-      this.getTimeRange(interval),
+      this.getTimeRangeForAmounts(interval),
       Common.getSqlInterval(interval)
     );
   }
@@ -41,7 +41,7 @@ class Mining {
    */
   public async $getHistoricalBlockRewards(interval: string | null = null): Promise<any> {
     return await BlocksRepository.$getHistoricalBlockRewards(
-      this.getTimeRange(interval),
+      this.getTimeRangeForAmounts(interval),
       Common.getSqlInterval(interval)
     );
   }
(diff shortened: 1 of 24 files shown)


nymkappa:
Need to index block prices using the new `prices` table in the first place. Then this PR can be easily updated and reviewed.

wiz:
please rebase and also change the colors back lol

softsimon:
> ... and also change the colors back lol

I too prefer this old design. And maybe the USD graph could be in green which is the same color as fiat price / USD on the site.

Screen Shot 2022-07-10 at 21 15 21

nymkappa:
Screen Shot 2022-07-11 at 9 21 24 AM
Screen Shot 2022-07-11 at 9 21 03 AM

backend/src/api/mining/mining.ts
```diff
@@ -31,7 +33,7 @@ class Mining {
    */
   public async $getHistoricalBlockFees(interval: string | null = null): Promise<any> {
     return await BlocksRepository.$getHistoricalBlockFees(
-      this.getTimeRange(interval),
+      this.getTimeRange(interval, 5),
```
knorrium, line 36: @nymkappa what are you trying to do with this line?

I'm a bit confused given the current implementation of `getTimeRange`:

https://github.com/mempool/mempool/blob/a97675c5388e14edddfe87930112e75e7bd7f5c8/backend/src/api/mining/mining.ts#L531-L544

backend/src/repositories/BlocksRepository.ts
```diff
@@ -288,6 +288,7 @@ class BlocksRepository {
       LIMIT 10`;
 
     try {
+      console.log(query, params);
```
knorrium, line 291: Use debug.log for this and add a TODO to remove it later

backend/src/api/mining/mining.ts
nymkappa: It reduces the resolution x5 for the block fees charts to be more visually pleasing (less noisy)

This is what the chart look like without that scaling.

Screen Shot 2022-07-12 at 8 46 47 AM

VS now

Screen Shot 2022-07-12 at 8 47 35 AM

backend/src/repositories/BlocksRepository.ts
nymkappa: Nah I just forgot it :), deleting

wiz (approved):
Tested ACK @ v2.5.0-dev [908635b3]

image

PR #1820, opened 2022-06-06
Update feerate tooltips

hunicus:
Revision of #1804.

Updates:
- leaves fiat fee estimates in place and creates new tooltips for describing feerates
- uses more literal terms to describe priority instead of speed or block confirmation targets
- leaves i18n xlf files alone

Screenshot from 2022-06-06 14-38-01

### DIFF
--- a/frontend/src/app/components/fees-box/fees-box.component.html
+++ b/frontend/src/app/components/fees-box/fees-box.component.html
@@ -12,24 +12,24 @@
   </div>
   <div class="fee-estimation-container">
     <div class="item">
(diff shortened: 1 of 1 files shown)


wiz (changes requested):
I think we should put the tooltips on the "high priority" text to define what high priority means - not on the actual suggested fee rate itself

hunicus:
Moved tooltips to feerate labels. 

Should tooltips be shown on top now? They currently show on bottom, covering the feerates.

Screenshot from 2022-06-07 13-14-06

hunicus:
I just moved tooltip placement to top as I don't see any downside.

Issue #1821, opened 2022-06-07
Backend still tries to use the mysql socket when the database is disabled


knorrium:
```
Jun  6 22:30:44 [496921] DEBUG: New block found (#739661)!
Jun  6 22:30:44 [496921] DEBUG: Indexing tx 1 of 1128 in block #739661
Jun  6 22:30:44 [496921] DEBUG: 0 of 1128 found in mempool. 1 fetched through backend service.
Jun  6 22:30:44 [496921] DEBUG: runMainLoop error: connect ENOENT /var/run/mysql/mysql.sock. Retrying in 5 sec.
Jun  6 22:30:44 [496921] DEBUG: {"message":"connect ENOENT /var/run/mysql/mysql.sock","code":"ENOENT","errno":-2}
Jun  6 22:30:49 [496921] DEBUG: New block found (#739661)!
Jun  6 22:30:49 [496921] DEBUG: Indexing tx 1 of 1128 in block #739661
Jun  6 22:30:49 [496921] DEBUG: 0 of 1128 found in mempool. 1 fetched through backend service.
Jun  6 22:30:49 [496921] WARN: runMainLoop error: connect ENOENT /var/run/mysql/mysql.sock. Retrying in 10 sec.
Jun  6 22:30:49 [496921] DEBUG: {"message":"connect ENOENT /var/run/mysql/mysql.sock","code":"ENOENT","errno":-2}
``` 

```
  "DATABASE": {
    "ENABLED": false,
    "HOST": "127.0.0.1",
    "PORT": 3306,
    "SOCKET": "/var/run/mysql/mysql.sock",
    "DATABASE": "mempool",
    "USERNAME": "mempool",
    "PASSWORD": "mempool"
  },

```

PR #1822, opened 2022-06-07
Add 'db-less' mining pool tagging support

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1821

This PR fixes a regression introduced when we refactored away the mining pool tagging logic from the frontend. Today we rely on the backend to tag mining pools, but the problem is that it also assume the database is available.

With this PR, RAM memory is used as a temporary storage for mining pools when the database is disabled.

### Testing

* Clear your disk cache, to make sure we process at least a few blocks for testing this new code
* Set `DATABASE.ENABLED: false` in `mempool-config.json`
* Set `MINING_DASHBOARD: false` in `mempool-frontend-config.json`
* Start the node backend
* Browse blocks and confirm that they have proper mining pool tag, even the older one

Also, there should be no changes at all from `master` with `DATABASE.ENABLED: true` and `MINING_DASHBOARD: true`


wiz (approved):
utACK - need to test it as part of our full release testing cc @knorrium

PR #1830, opened 2022-06-07
Change "projected block" to "mempool block" in feerate faq

hunicus:
Late, but I just realized it. Fixes same issue as #1827, but in the feerate FAQ.


wiz (approved):
utACK

PR #1841, opened 2022-06-08
Add note on rate limits to api docs

hunicus:
Only shows when `OFFICIAL_MEMPOOL_SPACE` is `true`.

Looks like this:

Screenshot from 2022-06-07 23-31-43


softsimon (approved):
utACK

wiz (approved):
Tested ACK @ v2.4.0-dev [[e1611d1e](https://github.com/mempool/mempool/commit/e1611d1e)]

Will merge after v2.4.0 release since this modifies i18n strings

PR #1842, opened 2022-06-08
Bob the builder

knorrium:
fixes #1201 

- Add mempool.js as a submodule so we can build it with the mempool frontend
- Add new script to generate `liquid.js`, `mempool.js` or `bisq.js` based on the `BASE_MODULE`


erikarvstedt (changes requested):
`frontend/package.json` still includes `@mempool/mempool.js`

knorrium:
Closing this stale PR

PR #1851, opened 2022-06-08
Stop docs links from reloading app

### LINKED ISSUE #1783: Avoid triggering page reload when clicking anchor links loaded from doc data file
Angular syntax in the [docs data file](https://github.com/mempool/mempool/blob/master/frontend/src/app/docs/api-docs/api-docs-data.ts) isn't evaluated when inserted into a page. 

This prevents us from linking properly since we cannot use built-in Angular directives like `routerLink` and `fragment`. So we use plain HTML, but doing that triggers a page reload and so the UX sucks.

It would be great to find a workaround that provides a better experience, better performance, etc.

hunicus:
These changes rework some docs routing logic to fix #1783.

Only links _within_ content blocks are effected. In the example below, the "ways to accelerate the process" link is being clicked.

## Anchor Link, Before

https://user-images.githubusercontent.com/93150691/172710794-d9887ba9-fc73-4f76-809f-044251a25ec4.mp4

## Anchor Link, After

https://user-images.githubusercontent.com/93150691/172710812-b8c83205-8ccd-490d-b495-3358b012152e.mp4

### DIFF
--- a/frontend/src/app/docs/api-docs/api-docs-data.ts
+++ b/frontend/src/app/docs/api-docs/api-docs-data.ts
@@ -6052,7 +6052,7 @@ export const faqData = [
     showConditions: bitcoinNetworks,
     fragment: "why-is-transaction-stuck-in-mempool",
     title: "Why isn't my transaction confirming?",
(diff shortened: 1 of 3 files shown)


softsimon (approved):

wiz:
@hunicus I think simon is going to re-implement this

softsimon:
I belive I have a better solution to this now.

wiz:
Replaced by #1874

PR #1864, opened 2022-06-10
Make all mining charts tooltips uniform

nymkappa:
Currently, mining charts do not follow the same layout rules and formatting. This PR cleans this up my making them uniform as much as possible.

Also changed from `short` months to `long` months in tooltip titles to avoid weird formatting in some language like french, showing for example `2 juil., 18:23`


softsimon (approved):
tested ACK @ da3272df

wiz (approved):
Tested ACK @ v2.4.1-dev [[02b34c98](https://github.com/mempool/mempool/commit/02b34c98)]

Issue #1862, opened 2022-06-10
"Latest blocks" should link to list of blocks


Xekyo:
### Description

I have found myself multiple times lately trying to click "Latest Blocks" expecting that it would take me to the page that lists the blocks. After that I scroll down to find "View more" and click that instead. Could perhaps the title of that tile also link to the block page?

This is the view I usually have when opening the site, so I sometimes forget that I need to scroll down to find the link:
![image](https://user-images.githubusercontent.com/4060799/173103481-23e343b4-5734-455e-8339-83ccd51e5850.png)
![image](https://user-images.githubusercontent.com/4060799/173104161-47540d1e-2406-4e83-bb93-0f7ef07a876e.png)

softsimon:
It's a fine idea but I don't think it's intuitive to use the header titles as links. It would cause inconsistency. We currently use the "View more" links to reveal the full views, but we have tried to come up with a better way as we don't think it's perfect.

So for those reasons I will NACK

PR #1865, opened 2022-06-10
Make `Latest blocks` and `Adjustments` widget title clickable

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1862

I made some widget title clickable, but disable all `<a>` css effect so that the overall design stays the same. Let me know if you wanna to add more links somehwere else.


wiz:
IMO we can do better than simply making the titles clickable, personally I would like to see some icon added

softsimon:
It's a fine idea but I don't think it's intuitive and good UX to use "some" header titles as links. It would cause inconsistency. We currently use the "View more" links to reveal the full views, but we have tried to come up with a better way as we don't think it's perfect.

So for those reasons I will NACK

wiz:
Can you try the icon idea?

PR #1866, opened 2022-06-10
Fix latest blocks rewards column not large enough to fit high reward

nymkappa:
Fixes the following

#### Now

Screen Shot 2022-06-10 at 11 54 20 PM

#### With the PR

Screen Shot 2022-06-10 at 11 55 05 PM


wiz (approved):
Tested ACK @ v2.4.1-dev [[035b29e7](https://github.com/mempool/mempool/commit/035b29e7)]

PR #1870, opened 2022-06-11
Fix anchor link scrolling on load (chrome)

hunicus:
Tested on desktop and mobile on Firefox and Chrome.


wiz (approved):
Tested ACK @ v2.4.1-dev [[abaaef22](https://github.com/mempool/mempool/commit/abaaef22)]

Tested using https://mempool.ninja/docs/faq#why-is-transaction-stuck-in-mempool

PR #1873, opened 2022-06-12
Fix canvas resolution on high-DPI screens

mononaut:
Block visualizations were rendered according to the canvas size measured in CSS pixels.

On high-DPI/retina screens, this can look blurry, because the canvas is scaled up to display at a higher device resolution.

This PR properly accounts for `window.devicePixelRatio` to render at full resolution on these displays.

|before|after|
|-|-|
|before|after|

(screenshots from testing on a high-DPI phone)


wiz (approved):
Tested ACK @ v2.4.1-dev [[1f6b59f2](https://github.com/mempool/mempool/commit/1f6b59f2)]

wiz:
Thanks a lot for fixing this, really appreciate it!

PR #1874, opened 2022-06-12
Move FAQ HTML to angular templates

### LINKED ISSUE #1783: Avoid triggering page reload when clicking anchor links loaded from doc data file

softsimon:
fixes #1783

Alternative solution to https://github.com/mempool/mempool/pull/1851

Moving the HTML data from the javascript file to the template in order to be able to execute angular code like [routerLink] and other components like the Mempool blocks without using any hacky javascript tricks.

Only downside is that now, everything is not in the same place, for new Faq items we need the items in the component ts and html template. But I think it's required in order to being able to make the FAQ html more interactive without hacks

### DIFF
--- a/frontend/src/app/docs/api-docs/api-docs-data.ts
+++ b/frontend/src/app/docs/api-docs/api-docs-data.ts
@@ -5981,7 +5981,6 @@ export const faqData = [
     showConditions: bitcoinNetworks,
     fragment: "what-is-a-mempool",
     title: "What is a mempool?",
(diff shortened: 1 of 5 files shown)


wiz (approved):
Tested ACK @ v2.4.1-dev [[465053f3](https://github.com/mempool/mempool/commit/465053f3)]

Issue #1788, opened 2022-06-02
Write setup guide for regtest development


nymkappa:
What I use https://gist.github.com/nymkappa/ed398c119169d941e89bff40591b5927
Something I found helpful to get started https://gist.github.com/System-Glitch/cb4e87bf1ae3fec9925725bb3ebe223a

@hunicus How do you suggest we integrate this into mempool's repo? Maybe in the backend readme?

hunicus:
Ahh sorry I didn't get a notification about this for some reason. Yeah I think another section in the [development tips area of the backend setup docs](https://github.com/mempool/mempool/tree/master/backend#development-tips) would be good.

PR #1878, opened 2022-06-13
Added regtest notes to backend README

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1788

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -199,7 +199,7 @@ nodemon src/index.ts --ignore cache/ --ignore pools.json
    ./src/bitcoin-cli sendtoaddress # will print the help
    ```
 
-* Mini script to generate txs with random fx rate (between 1 to 100 sat/vB). It's slow so don't expect to use this to test mempool spam, except if you let it run for a long time, or maybe with multiple regtest node connected to each other.
+* Mini script to generate transactions with random TX fee-rate (between 1 to 100 sat/vB). It's slow so don't expect to use this to test mempool spam, except if you let it run for a long time, or maybe with multiple regtest node connected to each other.
    ```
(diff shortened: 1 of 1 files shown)


backend/README.md
```diff
@@ -159,3 +159,58 @@ nodemon src/index.ts --ignore cache/ --ignore pools.json
...
+   ```
+   ./src/bitcoin-cli -named -regtest -rpcport=8332 sendtoaddress address=$(./src/bitcoin-cli -regtest -rpcport=8332 getnewaddress) amount=0.1 fee_rate=5
+   ```
+
+* See more example of `sendtoaddress`
+   ```
+   ./src/bitcoin-cli sendtoaddress # will print the help
+   ```
+
+* Mini script to generate txs with random fx rate (between 1 to 100 sat/vB). It's slow so don't expect to use this to test mempool spam, except if you let it run for a long time, or maybe with multiple regtest node connected to each other.
```
hunicus, line 202: Haven't seen fx rate refer to feerate before...might be better to just say feerate?

hunicus (commented):
This looks good to me overall.

Most of the changes I would suggest are minor wording and formatting stuff...I can suggest those in a separate PR.

backend/README.md
wiz: yeah FX rate is BTC/USD rate, must be a typo

wiz, line 202:
```suggestion
* Mini script to generate transactions with random TX fee-rate (between 1 to 100 sat/vB). It's slow so don't expect to use this to test mempool spam, except if you let it run for a long time, or maybe with multiple regtest node connected to each other.
```

PR #1881, opened 2022-06-14
Fix anchor links and typos in faq

hunicus:
Commits 1 and 2 fix some issues resulting from #1874. FAQ content is loaded after `scrollIntoView()` fires, sometimes resulting in partially-scrolled pages when the browser directly requests the FAQ page with an anchor link (example: go directly to https://mempool.space/docs/faq#host-my-own-instance-linux-server in desktop Chrome).

Commit 3 makes 2 minor updates.


wiz (approved):
Tested ACK @ v2.4.1-dev [[b82abc28](https://github.com/mempool/mempool/commit/b82abc28)]

Issue #1868, opened 2022-06-11
Add hashtages with timespan into all mining charts


nymkappa:
For example you can share 24h mempool chart with https://mempool.space/graphs/mempool#24h
You cannot do that with mining charts, so I need to implement it.

PR #1887, opened 2022-06-15
Add #{timespan} in mining charts urls for easier sharing

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1868. Adds support for #{timespan} in all mining charts, similar to how mempool chart can be shared using a timespan.

For example: https://mempool.space/graphs/mining/block-fee-rates#3y

#### Testing

* For each mining charts, click on all timespan one by one and make sure the url is correct
* Copy one of this url, and open a new tab with it (avoid the `1y` timespan because it's the default one)
* Also try to paste the previous url in inconito tab to see if it also works fine


wiz (approved):
Mostly Tested ACK @ v2.4.1-dev [[95323ac4](https://github.com/mempool/mempool/commit/95323ac4)]

PR #1891, opened 2022-06-15
Use Bitcoin Core's ZMQ publisher to instantly discover new blocks

antonilol:
This PR adds zmq support to get updates _pushed_ from bitcoind instead of _pulling_ them every `POLL_RATE_MS` (2 seconds)

some readme on how to set this up on the bitcoin core side can be helpful but haven't made it (yet)

previously, mempool would poll bitcoind every 2 seconds, i put a zmq subscriber in between that skips the rest of the ongoing 2 seconds when a new block is discovered by bitcoind. blocks end up up to 2 seconds faster on the website

i decided to use `Promise.any` for the delay + zmq, to return which one of the 2 triggers, either the first after 2 seconds or the second after a block has been discovered. when zmq is disabled in mempool (i made a config thingy for that) `this.newBlockTrigger` will stay `undefined` forever and do nothing. es2021 is required for `Promise.any` (tsconfig.json)

to test:

add 
```
zmqpubhashblock=tcp://127.0.0.1:<some port>
```
to bitcoin.conf

and
```json
  "CORE_ZMQ": {
    "ENABLED": true,
    "HOST": "tcp://127.0.0.1:<port>"
  },
```
to mempool-config.json on regtest

start everything (database nginx backend etc)

run 
```
bitcoin-cli -regtest -generate
```
to mine a block and it will instantly appear in mempool

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -18,7 +18,8 @@
         "node-worker-threads-pool": "^1.5.1",
         "socks-proxy-agent": "^6.2.0",
         "typescript": "~4.7.2",
-        "ws": "~8.7.0"
+        "ws": "~8.7.0",
+        "zeromq": "^6.0.0-beta.6"
       },
       "devDependencies": {
         "@types/compression": "^1.7.2",
@@ -1037,6 +1038,16 @@
         "node": ">= 0.6"
       }
     },
+    "node_modules/node-gyp-build": {
+      "version": "4.4.0",
+      "resolved": "https://registry.npmjs.org/node-gyp-build/-/node-gyp-build-4.4.0.tgz",
+      "integrity": "sha512-amJnQCcgtRVw9SvoebO3BKGESClrfXGCUTX9hSn1OuGQTQBOZmVd0Z0OlecpuRksKvbsUqALE8jls/ErClAPuQ==",
+      "bin": {
+        "node-gyp-build": "bin.js",
+        "node-gyp-build-optional": "optional.js",
+        "node-gyp-build-test": "build-test.js"
+      }
+    },
     "node_modules/node-worker-threads-pool": {
       "version": "1.5.1",
       "resolved": "https://registry.npmjs.org/node-worker-threads-pool/-/node-worker-threads-pool-1.5.1.tgz",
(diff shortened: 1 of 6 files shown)


softsimon:
I have not tested yet, but I believe this could cause some synchronization issues. For example, a new block comes in and confirming your transaction, but the mempool is still the same unaffected and won't be updated until the next 2 sec loop.

antonilol:
How i implemented it now the whole update loop is ran when a block is discovered, it waits 2 seconds (usual) or until a block arrives (Promise.any)

nymkappa:
If we can use ZMQ for both getting notified of new blocks and new transactions, and it's as reliable as polling (meaning we cannot miss any data), then should not we just migrate it all to ZMQ approach?

I don't think having this only used for new blocks is a good improvement due to @softsimon 's concerns about possible synchronization/racing issues.

wiz:
The problem is that esplora backend won't have the transaction yet. IMO the current poll implementation is fine and a 1 second time delay is not the end of the world...

wiz:
On production mempool.space servers we use poll time of 1000ms and it works well

backend/package-lock.json
```diff
@@ -18,7 +18,8 @@
         "node-worker-threads-pool": "^1.5.1",
         "socks-proxy-agent": "^6.2.0",
         "typescript": "~4.7.2",
-        "ws": "~8.7.0"
+        "ws": "~8.7.0",
+        "zeromq": "^6.0.0-beta.6"
```
wiz, line 22: NACK for adding a new third party npm module dependency

antonilol: is there something wrong specifically with the zeromq package? like the others are also 3rd party packages

antonilol:
> The problem is that esplora backend won't have the transaction yet.

does esplora have some sort of bitcoind's zmq equivalent? that could be used for esplora then and bitcoind's zmq for home node runners

nymkappa:
> is there something wrong specifically with the zeromq package? like the others are also 3rd party packages

For the backend part, we avoid adding new 3rd party dependencies because it is a security hole. Those 3rd party will run on our production servers, which run critical Bitcoin infrastructure use by hundred thousands people. There were multiple occurrences of npm packages repo getting hacked to include malicious code, and we cannot take that risk for mempool prod servers.

We prefer not having an optional feature rather than having an extra 3rd party dep on the backend.

For this reason, plus the fact that I don't see the need to get new blocks displayed 2 seconds faster than usual, I will also NACK.

PR #1892, opened 2022-06-16
Mined block visualization

mononaut:
## Summary

This PR improves the block visualization feature, and expands it to mined blocks:

- Refactors the canvas & webgl related code into a separate, reusable component.
- Integrates that component into the `/block` page.
- Draft implementation of a new backend API method `/block/:hash/summary` to supply the data.
- Adds a tooltip-style transaction info dialog on touch/hover.
- Optimizes webgl rendering.

The mined block visualization is laid out vertically, to match the orientation of the mini blocks at the top of the page, but uses the same mempool fee colors.

half-full-hover

Demo deployment running at [mempool.monospace.live](https://mempool.monospace.live)

### Transaction tooltip/dialog

On touch devices, tapping a square opens the dialog. Users can then tap the txid hyperlink in the dialog to navigate to that transaction.

Otherwise, hovering over a square opens the dialog, and clicking navigates to the transaction page.

For mined blocks, would it be better to scroll to the tx entry in the paginated transaction list on this page, instead of navigating away to the transaction page?

### Backend

This PR only adds a minimal implementation of the `/block/:hash/summary` API method to supply data for the mined blocks visualization.

It hits the bitcoinClient's `getBlock` method with `verbosity=2` to get a full list of transaction data (including fees), then returns a list of stripped transactions to the client.

That works ok on my machine, using a core backend on Bitcoin mainnet. I'm not sure about other backends, or compatibility with e.g. Liquid.

The data is preloaded and cached in the backend for recent blocks, but I haven't implemented any indexing yet.


wiz (approved):
Tested ACK @ v2.4.1-dev [[288bddca](https://github.com/mempool/mempool/commit/288bddca)]

Works well, and the new tooltips are a great improvement over the previous UI design.
However, since other networks like Testnet, Signet, and Liquid are mostly empty blocks I think we should make the mined blocks visualization a mainnet only feature.

softsimon:
Tooltip with click to transaction looks great.
Need to get used to the new block page design, maybe we need tweaking, there is a lot of empty space down left.

The only problem now is the use of `bitcoinClient.getBlock` every time an old block is loaded. The issue is that Bitcoin Core RPC can't handle all the parallel requests in a live environment. We need to add this transaction data to the block data indexer, this is @nymkappa specialty.

wiz:
Good point. @nymkappa How about we merge this for now, and you can look into indexing / caching the bitcoind RPC data in MySQL as part of the full block indexing?

wiz:
> Yeah, perhaps it looks cleaner if the size of the visualization is limited to the height of the table on the left?

> Need to get used to the new block page design, maybe we need tweaking, there is a lot of empty space down left.

Perhaps a bit off-topic for this PR but personally I feel we could add one more row for the block subsidy reward (separate from the transaction fees) so that newbies can more clearly see that the subsidy was 6.25 and the fees were 0.05 so the total was 6.30.

nymkappa:
Regarding the design; I know this makes sense from a technical POV because it matches fees colors gradient, but it feels a bit off for me to see mined block in blue/purple gradient at the top and red/green gradient at the bottom.

mononaut:
For this feature we only need to keep stripped transaction data (txid, vsize, value and fee), rather than the whole verbose block RPC response. I think bitcoin mainnet is up to about 750m total transactions, so that still sets an absolute minimum lower bound of ~42gb for a full index.

Agree we shouldn't try to index or cache that much on RPIs.

Nginx caching makes sense to me. That's how I handle similar block data on Bitfeed, and it works well there. Expensive parallel requests might still be a problem until the cache warms up, depending how much traffic you get to old block pages.

nymkappa:
> For this feature we only need to keep stripped transaction data (txid, vsize, value and fee), rather than the whole verbose block RPC response. I think bitcoin mainnet is up to about 750m total transactions, so that still sets an absolute minimum lower bound of ~42gb for a full index.
> 
> Agree we shouldn't try to index or cache that much on RPIs.
> 
> Nginx caching makes sense to me. That's how I handle similar block data on Bitfeed, and it works well there. Expensive parallel requests might still be a problem until the cache warms up, depending how much traffic you get to old block pages.

I think we can just warm cache for all blocks on production, @wiz should confirm but I don't think that's an issue, we already though about doing something similar for the `/block` APIs.

wiz:
The problem is that the time required to warm cache all the Bitcoin blocks into nginx cache would take hours, and the script would have to be quite complex, so I think it would be cleaner to store that large amount of data that never changes in mysql as part of kappa's much more intelligently implemented block indexing process. Then the mysql data can be easily copied to other nodes as well.

nymkappa:
> The problem is that the time required to warm cache all the Bitcoin blocks into nginx cache would take hours, and the script would have to be quite complex, so I think it would be cleaner to store that large amount of data that never changes in mysql as part of kappa's much more intelligently implemented block indexing process. Then the mysql data can be easily copied to other nodes as well.

Noted. I'll add an additional indexing that will be disabled by default (new config option) to store what we need for the visualization.
It will index the same amount of block as defined in `INDEXING_BLOCKS_AMOUNT`, and "on the fly" indexing will also be available.

PR #1895, opened 2022-06-18
Add moving average to the hashrate & difficulty chart

nymkappa:
Something I worked on couple of days ago after seeing some comment on Twitter about big accounts "promoting new ATH in network hashrate" while it was most likely just luck (specifically, Luxor got very luck at that time).

I added a moving average on the hashrate & difficulty chart after to reduce the noise due to the luck factor. I also removed the 1 month resolution because there is not enough data to have a decent looking chart.

### Dashboard widget

Screen Shot 2022-06-18 at 10 23 35 AM

### Chart (3m)

For 3m and 6m timespan, I use a 7 days moving average

Screen Shot 2022-06-18 at 10 24 02 AM

### Dashboard widget (3y)

For 1y+ timespan, I use a 30 days moving average

Screen Shot 2022-06-18 at 10 24 30 AM


softsimon (approved):
utACK

wiz (approved):
Tested ACK @ v2.4.1-dev [[625dba94](https://github.com/mempool/mempool/commit/625dba94)]

wiz (changes requested):
The feature works well, but I think we need some MySQL indexing. Currently this URL takes about 35 seconds to return: https://mempool.ninja/testnet/api/v1/mining/hashrate/all

PR #1896, opened 2022-06-18
Index block summaries in db

nymkappa:
Index data needed by https://github.com/mempool/mempool/pull/1892

The  logic is simple. I get all the block hash we have indexed in our database, and I index their visualization data. It is done as the last indexing task (after block and hashrate indexing).

Also, visualizing a block that has not yet been indexed will simply index its summary on the fly for later reuse.

Logs are identical to blocks indexing. You will notice that it is much slower to use the verbose mode of `getBlock` RPC call:
`Jun 18 16:39:26 [30312] DEBUG: Indexing block summary for #738086 | ~5.17 blocks/sec | total: 3212/741299 (0.43%) | elapsed: 280 seconds | left: ~142856 seconds`

There are possible optimizations to explore, but I think this will be a production only index anyway, which can easily be copy/pasted between database once the initial sync is done on one server, so I'm not going to explore them for now.

It should take less than 24h to index everything as it will become much faster when hitting old/empty blocks.


mononaut (approved):
Tested ACK @ [`72a603a`](https://github.com/mempool/mempool/commit/72a603ac370c7b23121abb581041a8a6fde3c134)

wiz (changes requested):
Could you please add HTTP cache headers? As for the `Expires:` time, I think we can safely cache this data forever since the contents will never change, the only concern is that if we modify the API structure in the future we would need some kind of cache buster. Maybe set it to 1 month?

wiz:
I've begun running an index but it will take a few days to complete. After that I can copy the SQL data to other servers and upgrade prod.

wiz:
@nymkappa How does this code handle re-orgs?

nymkappa:
> @nymkappa How does this code handle re-orgs?

Good call, it does not. I'll add this as well as the caching.

wiz:
Need to set the cache lower for the 2 most recently mined blocks

nymkappa:
Just FYI, this is the size on my db for all blocks

Table | Size (MB)
-- | --
blocks_summaries | 99327

wiz (changes requested):
NACK - the index completed fine but then it seems backend got stuck when it tried to index TX in a newly mined block?

Screen Shot 2022-06-21 at 23 35 26

wiz:
Seems related to the daily DB backup locking the tables:
```
| 5234 | mempool               | localhost:36249 | mempool               | Query   |  490 | Waiting for table metadata lock | UPDATE state SET number = 0 WHERE name = 'last_hashrates_indexing'                  
```

nymkappa:
For tracking purpose, it has been decided to not lock db during backups. This should fix the issue you mentioned but let me know if there is something else.

```
ok kappa i modified the backup script so i think i can merge your PR but i want to test it another day or so
```

wiz:
I modified the daily backup script to not lock the DB tables and also run faster, so going to re-test this

PR #1903, opened 2022-06-22
Start adding backend unit tests

knorrium:
coverage


backend/package.json
```diff
@@ -41,8 +41,12 @@
   },
   "devDependencies": {
     "@types/compression": "^1.7.2",
-    "@types/ws": "~8.5.3",
     "@types/express": "^4.17.13",
+    "@types/jest": "^28.1.3",
+    "@types/ws": "~8.5.3",
+    "jest": "^28.1.1",
+    "ts-jest": "^28.0.5",
+    "ts-node": "^10.8.1",
```
wiz, line 49: Am I correct in assuming that these new npm deps are only installed with --dev-deps or some command line option?

PR #1904, opened 2022-06-22
Remove all block animation and remove debounce time

nymkappa:


nymkappa:
Todo before it's a reviewable PR

- [ ] Fix the loading animation
- [ ] Try to remove the very small `duration` / `delay` instead of setting them to 1ms
- [ ] Preload the previous block summary in order to cache it in the browser

Issue #1906, opened 2022-06-22
Disable all animations on mined block visualization


nymkappa:
### Today 

Currently, there is a long transition between mined blocks. It makes the user experience slower in comparison to the mempool project general feeling.

https://user-images.githubusercontent.com/9780671/175103934-b2daed10-21dd-4667-8dec-4cdf57a2e52e.mov

### The goal

We should disable all animation/transition for mined blocks. The best situation would be where the current block rendering only clears when the new one is ready to be displayed.

https://user-images.githubusercontent.com/9780671/175103860-167a5c87-16d7-4856-8aa6-a483c641eafb.mov

### Notes and context

Context https://github.com/mempool/mempool/pull/1904
Choosen approach https://github.com/mempool/mempool/pull/1904#issuecomment-1163292873

nymkappa:
Fixed in https://github.com/mempool/mempool/pull/1907

PR #1907, opened 2022-06-22
Disable mined block animations

mononaut:
This PR addresses issue #1906, disabling entry/exit transition animations for mined blocks.

The current block only clears when the new block is ready to display, but we cover with a loading spinner if that takes long enough to be confusing.

https://user-images.githubusercontent.com/83316221/175119832-836572fb-9d64-4736-bce6-4fa3124eb451.mp4

### DIFF
--- a/frontend/src/app/components/block/block.component.ts
+++ b/frontend/src/app/components/block/block.component.ts
@@ -2,9 +2,9 @@ import { Component, OnInit, OnDestroy, ViewChild, ElementRef } from '@angular/co
 import { Location } from '@angular/common';
 import { ActivatedRoute, ParamMap, Router } from '@angular/router';
 import { ElectrsApiService } from '../../services/electrs-api.service';
-import { switchMap, tap, debounceTime, catchError, map, shareReplay, startWith, pairwise } from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


frontend/src/app/components/block/block.component.ts
```diff
@@ -170,13 +170,9 @@ export class BlockComponent implements OnInit, OnDestroy {
...
         this.isLoadingOverview = true;
-        this.isAwaitingOverview = true;
-        this.overviewError = true;
-        if (this.blockGraph) {
-          this.blockGraph.exit(direction);
-        }
+        this.overviewError = null;
       }),
-      debounceTime(300),
+      throttleTime(300, asyncScheduler, { leading: true, trailing: true }),
```
mononaut, line 175: I'm inclined to keep the debounce behavior, because otherwise holding down the the left/right arrow keys to quickly scroll through blocks generates a huge number of requests and lots of busy-work.

We can avoid the initial 300ms delay by switching to `throttleTime` with this config, which resolves the first change immediately, then debounces subsequent changes.

wiz: This works great, it's the best of both worlds :)

wiz (approved):
Tested ACK @ v2.4.1-dev [[f2780e65](https://github.com/mempool/mempool/commit/f2780e65)]

PR #1908, opened 2022-06-22
Batch outspends requests

### LINKED ISSUE #1902: Batch outspends requests

softsimon:
fixes #1902

Created a new NodeJS backend API endpoint named `/api/v1/outspends` that takes a list of TXIDs to have it's current outspends returned batched in the same order. There is a limit of 50 per request to prevent spam and overload.

This should work both in bitcoind and esplora mode.


wiz (approved):
Tested ACK @ v2.4.1-dev [[960513c3](https://github.com/mempool/mempool/commit/960513c3)]

I am concerned about hitting the max URL length limit which is quite close, currently you are around 1950 chars when using a Tor v3 onion hostname and the max is 2000 chars or so - maybe consider changing `txId` to `tx` to save 50 bytes heh

PR #1911, opened 2022-06-23
Fix double http query when switching timespan for the first time

nymkappa:
Screen Shot 2022-06-23 at 3 31 04 PM


wiz (approved):
Tested ACK @ v2.4.1-dev [[4c8eaac1](https://github.com/mempool/mempool/commit/4c8eaac1)]

PR #1912, opened 2022-06-23
Don't preload /block APIs for genesis block

nymkappa:
Genesis block does not have `previousblockhash` so don't try to preload it.


wiz (approved):
Tested ACK @ v2.4.1-dev [[98b9f007](https://github.com/mempool/mempool/commit/98b9f007)]

PR #1919, opened 2022-06-24
Reduce maximum size of mined block visualization

mononaut:
Shrinks the mined block visualization to eliminate dead space below the block info table on larger screens.

after-layout


wiz:
Hey @mononaut I discussed with @softsimon and I think we'd prefer to keep the current size and instead maybe put more information into the empty space. Any idea what to add to those rows?

PR #1922, opened 2022-06-24
Fix block visualization color for txs with 0 < feerate < 1 sat/vb

mononaut:
Fixes a bug where transactions in the visualization with fee rates more than 0 but less than 1 sat/vb were colored incorrectly (pink, instead of green).

This never happens in the mempool visualization, but mined blocks do occasionally have transactions in this fee range.

Most obviously reproducible with block 421770, where a large transaction pays 0.6 sats/vb.


wiz (approved):
Tested ACK @ v2.4.1-dev [[f5f53c93](https://github.com/mempool/mempool/commit/f5f53c93)]

before vs after
Screen Shot 2022-06-27 at 0 30 23

PR #1925, opened 2022-06-25
Index difficulty adjustments

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1924

This PR adds:
* a new `difficulty_adjustments` index in the database for Bitcoin networks.
* a new API endpoint (warm cached) at `/api/v1/mining/difficulty-adjustments/:interval` -> Returns a JSON array to avoid field name spam in the response
* fix the rounding issue between the adjustment widget showing `-2.35 %` and the difficulty adjustment table showing `-2.4%`

Note: This PR does not fix the resolution issue with the hashrate chart (create very fat API response on testnet). A following PR will fix that.

### Self testing

* ✅ Bitcoin (mainnet / testnet)
* ✅ Liquid testnet
* ⁉️ Bisq

Issue #1924, opened 2022-06-25
Index difficulty adjustment to get rid of that monster query


nymkappa:
https://github.com/mempool/mempool/blob/master/backend/src/repositories/BlocksRepository.ts#L391

Indexing this will allow us to handle testnet in a more efficient way, and in the long term will provide better perf also for mainnet.

PR #1928, opened 2022-06-25
Nymkappa/feature/hashrate resolution

nymkappa:
**Depends on https://github.com/mempool/mempool/pull/1925 which must be merged first before this one.**
Linked to https://github.com/mempool/mempool/issues/1924

This PR adds resolution scaling for the hashrate and difficulty chart. As @wiz noticed, the test API response for `/api/v1/mining/hashrate/all` on testnet is more than 13MB. The reason is that testnet difficulty is very unstable and changes all the time, resulting in hundreds thousands (or more?) difficulty adjustments.

So here is the new resolutions:
* On mainnet we use (quite arbitrary numbers, can be tweaked)
  * 1 day resolution up to `1y` timespan (included)
  * 2 days resolution for the `2y` timespan
  * 3 days resolution for the `3y` timespan
  * 7 days resolution the the `all` timespan
* On testnet we basically double all the resolutions above. I could not yet test with full indexed db (currently re-indexing)

Now of course the chart looks a bit less accurate, but trends are still well visible, and up to `1y` timespan, we still have an accurate chart since we keep the 1d resolution (aside testnet but that's pretty much irrelevant here).

The PR also tweaked the following:
* The hashrate moving average has been set to 15 data points average regardless of the timespan. Again the goal is to show trends. By definition moving averages are very arbitraty and a lagging number
* When you click on `View more` in the hashrate rate from the mining dashboard, we open `/graphs/mining/hashrate-difficulty#1y` to show the same chart as the widget
* The legend toogles are now saved (you can turn off the noisy `Hashrate` serie for example and switch between timespan, preferences are kept)


wiz:
please rebase onto current master

wiz (changes requested):
NACK - charts are not only too low resolution but also contain incorrect data

PR #1929, opened 2022-06-25
Minimum transaction fee is 0

nymkappa:
A bug introduced in https://github.com/mempool/mempool/commit/ecbd18087b267ba61b74477c973281b740edaf30 was setting the minimum transaction fee rate to 0.1 sats/vb on Liquid and 1 sats/vb on Bitcoin.

However, those two networks allows lower fee rates (all the way down to `0`), so the minimum should be effectively `0` instead of `0.1` and `1`.

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -114,7 +114,7 @@ export class Common {
       totalFees += tx.bestDescendant.fee;
     }
 
-    tx.effectiveFeePerVsize = Math.max(Common.isLiquid() ? 0.1 : 1, totalFees / (totalWeight / 4));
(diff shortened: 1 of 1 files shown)


wiz (approved):
Tested ACK @ v2.4.1-dev [[a8de738e](https://github.com/mempool/mempool/commit/a8de738e)]

Works well to fix the issue we observed on liquid (left side).
Behavior is now correctly estimating 1 minute ETA (right side).
However, we need to be careful to test this didn't break anything else cc @knorrium 
Screen Shot 2022-06-27 at 0 03 54

Issue #1913, opened 2022-06-23
Handle gracefuly query to pool dominance when there is no indexed block


nymkappa:
If `topPoolsId` is empty https://github.com/mempool/mempool/blob/master/backend/src/repositories/HashratesRepository.ts#L77 then the query https://github.com/mempool/mempool/blob/master/backend/src/repositories/HashratesRepository.ts#L92 will have an invalid syntax.

nymkappa:
```
Jun 26 10:36:44 [66132] ERR: Cannot fetch weekly pools hashrate history. Reason: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ') ORDER by hashrate_timestamp, FIELD(pool_id, )' at line 4
```

PR #1935, opened 2022-06-26
 Handle pool dominance query error with 0 indexed blocks

nymkappa:
Mini PR - Fixes https://github.com/mempool/mempool/issues/1913


wiz (approved):
utACK

Issue #1848, opened 2022-06-08
Umbrel v0.5 electrs errors are not handled properly, no pretty error page


wiz:
<img width="1185" alt="Screen Shot 2022-06-09 at 0 32 02" src="https://user-images.githubusercontent.com/232186/172669964-5c71da35-b5b0-40fe-901f-fd4fafa5acd7.png">

wiz:
@lukechilds how did you want to handle this?

lukechilds:
The issue is just the Umbrel proxy swallowing the 504 status code and returning a 502, right? If so we'll get that fixed in the proxy so it returns the correct status code.

Are you detecting the error based on status code alone or also testing the response message from Umbrel?

wiz:
@nymkappa luke told me they fixed umbrel so it won't "swallow" the 504 anymore, but apparently our code is checking for other HTTP codes like 429 so can you modify it to check for 504 as well? should be a one-liner PR

PR #1936, opened 2022-06-26
Handle electrs error 504

nymkappa:
Mini PR - Fixes https://github.com/mempool/mempool/issues/1848

With this PR, we check for the following error codes: 413, 405, 504


wiz (approved):
utACK, however I think this could potentially reproduce on production if our backend times out for some reason

Issue #1927, opened 2022-06-25
Use columns in prices table instead of JSON object


nymkappa:
Since we did not integrated this table anywhere yet, let's refactor currencies into their own column in order to save db space.

PR #1940, opened 2022-06-26
Replace json `prices.avg_prices` with table columns - update prices logs

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1927

This PR also fixes the logging for http queries retry mechanism (_price oracle only_):
* `warn` when the query fails then `err` when `config.MEMPOOL.EXTERNAL_MAX_RETRY` attempts have been depleted


nymkappa:
Will conflict with https://github.com/mempool/mempool/pull/1925 because the database migration

wiz (approved):
Tested ACK @ e8175a90f4a6d856a006f7e40a2df5a15bc5867f

PR #1942, opened 2022-06-26
Remove fade animation when switching docs tabs

hunicus:
There's no fade animation when going to/from the FAQ tab, so for consistency, this change removes the fade animation from the other tabs too.


knorrium (approved):
Tested ACK @ 5b9d6a31

wiz (approved):
utACK

PR #1943, opened 2022-06-26
Handle occasional blank side nav on desktop docs

hunicus:
Sometimes (not sure exactly why) the desktop view's side navigation menu doesn't appear on load, but it appears as soon as the user scrolls.

This change forces the browser to show the side nav on load instead of waiting for a scroll event.

Example of the issue:

https://user-images.githubusercontent.com/93150691/175832022-f9aa66a9-3c1e-4009-adec-e1b4d95af95d.mp4


nymkappa:
I integrated this change in a different way in another PR that supposedly fixes the js crash when loading the FAQ (which makes all E2E tests fails for a while now)

https://github.com/mempool/mempool/pull/1945

PR #1944, opened 2022-06-26
Add 'usually' to feerate tooltips

hunicus:
To handle times when tooltips appear misleading. 

Would like to complement these tooltips with more details in a FAQ in a future PR.


knorrium (approved):

wiz (approved):
utACK

PR #1945, opened 2022-06-26
Fix js crash when loading faq - fix crash when switching tabs with an active fragment

nymkappa:
Replace https://github.com/mempool/mempool/pull/1943

Hopefully that should also fix E2E tests? @knorrium


knorrium (approved):
Tested ACK @ b97ea010

I don't see any functional differences in the docs page and all the tests pass now

hunicus (approved):
ACK

In testing this PR, I realized there is an issue with opening accordion panels on mobile from same-page links. Looks like a lingering issue from #1874, not these changes, so I will fix separately.

PR #1947, opened 2022-06-29
Add faq on vb, wu, and sat/vb

hunicus:


frontend/src/app/docs/api-docs/api-docs.component.html
```diff
@@ -134,6 +134,19 @@ <h3 *ngIf="( item.type === 'category' ) && ( item.showConditions.indexOf(network
   Mining pools are groups of miners that combine their computational power in order to increase the probability of finding new blocks.
 </ng-template>
 
+<ng-template type="what-are-vb-wu">
+  <p>Virtual bytes (vB) and weight units (WU) are used to measure the size of transactions and blocks on the Bitcoin network.</p>
+  <p>A Bitcoin transaction's size in the blockchain is <i>not</i> determined how much bitcoin it transfers—instead, a transaction's size is determined by technical factors such how many inputs and outputs it has, how many signatures it has, and the format it uses (legacy, SegWit, etc). Since space in the Bitcoin blockchain is limited, bigger transactions pay more in mining fees than smaller transactions.</p>
```
antonilol, line 139: > bigger transactions pay more in mining fees than smaller transactions

i would say "bigger transactions **need to** pay more mining fees than smaller transactions **for the same priority**

hunicus: You're totally right, but this FAQ is intended for beginners, and I think mentioning priority here may confuse some people. 

In my opinion, in this case, given the likely audience for this part of the FAQ, it's worth being a little less than 100% precise in order to convey a more approachable answer.

antonilol: > I think mentioning priority here may confuse some people.

i dont think so, you explain it in the next section ("The priority of a pending Bitcoin transaction is determined by its feerate.")

> In my opinion, in this case, given the likely audience for this part of the FAQ, it's worth being a little less than 100% precise in order to convey a more approachable answer.

yeah, agree with that, just choose for yourself whats best for the faq

wiz (approved):
utACK

PR #1948, opened 2022-06-29
Tell git to ignore .swp files

hunicus:
Makes it easier to use tab completion to add a file to staging if it's open in vim.


knorrium (approved):

wiz:
Why not put it in the top-level .gitignore?

wiz (approved):

PR #1949, opened 2022-06-29
Save block predictions results in db and show results in a chart

nymkappa:
This PR adds a new table in the db, which saves our block predictions vs actual mined blocks, which we call "Block audit". This is Bitcoin only.

Screen Shot 2022-06-29 at 8 08 00 PM

We currently tracks the following:
* `missing_txs` - an array of missing tx ids
* `added_txs` - an array of added tx ids
* `match_rate` - In %, how accurate we were in our block prediction

The `match_rate` has been **updated** as follow:
```javascript
// Round to two decimals
matchRate = Math.round((Math.max(0, matches.length - missing.length - added.length) / txIds.length * 100) * 100) / 100;
```

Since this PR does not provide yet the new UI for visualizing each block audit, I've added  a new chart which allow to see our historical block prediction accuracy, and makes testing this PR simpler. You can see it at `/graphs/mining/block-predictions`.
Note that for the first 24h or so, since we don't have much data to display, the bar chart will look funny with super thick bars.

Screen Shot 2022-06-29 at 9 54 18 AM


nymkappa:
@softsimon I need your review on this part mostly (`backend/src/api/websocket-handler.ts`). 

https://github.com/mempool/mempool/pull/1949/files#diff-eb8b32a0032f3de01af050a97b4e03d2d0b7dd19c59d04e193c5fb6470dba1c2

There is some odd results with this code. For block https://mempool.space/block/0000000000000000000166db4d0cb187bb680a3e2c31f274a643110358c74125, I got ~60 missing txs, which were included in the next block.

Is it possible that all those txs were too recent? Or is there a racing code between the time a block is mined and the time `const _mempoolBlocks = mempoolBlocks.getMempoolBlocksWithTransactions();` is updated?

knorrium:
@nymkappa can you solve the merge conflicts?

nymkappa:
**Please review and merge https://github.com/mempool/mempool/pull/1940 first to avoid database migration conflict 😃**

wiz (approved):
Tested ACK - seems ok but needs more testing

wiz:
reminder to @softsimon to please review the above question

PR #1951, opened 2022-06-29
fixes for 0 sat/vB minimum fee

antonilol:
2 fixes for mempool when running bitcoind with `minrelaytxfee=0.00000000` (i do for testnet)

1. No Priority color
2. always says Purging

1
there was no color for 0 sat/vB, so `feeLevelIndex` in frontend/src/app/components/fees-box/fees-box.component.ts:31 was `-1` which set it to the last color of the list (meant for the highest fee rate)

2
it before always said `Purging` because minrelaytxfee was not equal to a hardcoded value, but bitcoind has a way to see if the mempool if overflowing and pruning:

```json
$ bitcoin-cli -testnet getmempoolinfo
{
  "loaded": true,
  "size": 11,
  "bytes": 2913,
  "usage": 16976,
  "total_fee": 0.00049031,
  "maxmempool": 300000000,
  "mempoolminfee": 0.00000000,   <-- minimum fee needed to enter the mempool
  "minrelaytxfee": 0.00000000,   <-- minrelaytxfee from the config
  "unbroadcastcount": 0
}
```

if those 2 values are not equal bitcoind is pruning

i didnt test with liquid but i think this should be fine because elementsd has this same rpc (its a fork of bitcoind)

another color for 0 sat/vB is also possible of course to let those transaction stick out (coinbase txs, txs from miners etc) but i just copied the color from 1 sat/vB

before:
image

after:
image

### DIFF
--- a/frontend/src/app/components/fees-box/fees-box.component.ts
+++ b/frontend/src/app/components/fees-box/fees-box.component.ts
@@ -33,11 +33,11 @@ export class FeesBoxComponent implements OnInit {
         tap((fees) => {
           let feeLevelIndex = feeLevels.slice().reverse().findIndex((feeLvl) => fees.minimumFee >= feeLvl);
           feeLevelIndex = feeLevelIndex >= 0 ? feeLevels.length - feeLevelIndex : feeLevelIndex;
-          const startColor = '#' + (mempoolFeeColors[feeLevelIndex - 1] || mempoolFeeColors[mempoolFeeColors.length - 1]);
+          const startColor = '#' + (mempoolFeeColors[feeLevelIndex - 1] || mempoolFeeColors[0]);
 
           feeLevelIndex = feeLevels.slice().reverse().findIndex((feeLvl) => fees.fastestFee >= feeLvl);
           feeLevelIndex = feeLevelIndex >= 0 ? feeLevels.length - feeLevelIndex : feeLevelIndex;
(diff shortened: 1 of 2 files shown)


antonilol:
i updated (rebased) this branch, @softsimon @wiz can you review it?

wiz (dismissed):

wiz (changes requested):
Tested NACK @ v2.5.0-dev [[ffe9f4a1](https://github.com/mempool/mempool/commit/ffe9f4a1)]

Looks like you need to fix a lot more things :)

Screen Shot 2022-11-21 at 17 08 27

antonilol:
first i changed it in the wrong place, other things dependent on that data (the graph) broke, now i fixed in the right place, the startColor was set to the last color (highest fee rate) when it was outside the range (0 is outside the range, the colors start at 1 sat/vB)
i think it is actually better to check if it falls below or above the range when no color is found

antonilol:
oh hmm i re-requested review and now @softsimon's review request got removed, still feel free to review though :)

antonilol:
rebased

antonilol:
can this be reviewed/merged?

softsimon:
Replaced with https://github.com/mempool/mempool/pull/3996

antonilol:
#3996 only fixed it always displaying 'Purging', 'No Priority' (0 sat/vB) is still red

image

PR #1952, opened 2022-06-29
set error correction level to low for qr codes

### LINKED ISSUE #1807: qr code optimization
### Description

long addresses, such as p2wsh, I guess will happen with p2tr too, appear on 57x57 grid on mempool ,while bitinfocharts is capable of packing them in the 33x33 grid

### Problem to be solved

I think 57x57 qr codes require a little bit more "zoom" for scanning, than 33x33

### Proposed solution

I guess all codes are being generated with a single configuration when analyzing the string could lead to a different optimal qr configuration for the string resulting in a smaller code

### Additional info

[BIP 173][0] (["Base58 needs a lot of space in QR codes..."][1]) (["For presentation, lowercase is usually preferable, but inside QR codes uppercase SHOULD be used, as those permit the use of alphanumeric mode, which is 45% more compact than the normal byte mode."][2])

(linked from the previous: https://www.thonky.com/qr-code-tutorial/alphanumeric-mode-encoding)

(linked from the previous: https://www.thonky.com/qr-code-tutorial/alphanumeric-table)

[0]: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki
[1]: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#motivation
[2]: https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#specification

antonilol:
fixes #1807

i first thought that alphanumeric mode wasnt used/supported but the npm package supports it and segwit addresses are converted to uppercase (btw regtest addresses weren't, fixed that too)

error correction was set to high, which is causing the large qrcodes, in my opinion it is not needed as low is good enough and bech32 has error correction

qrcodes before were very pixely on my screen (because of scaling some qrcode pixels were larger than other)
with smaller qrcodes this will be smoother and less pixely

@katesalazar does this fix your issue?


katesalazar:
I still havent been able to run mempool but I believe u

katesalazar:
what about lowering qr error correction but only for base32

keeping base58 with high qr error correction

funds should be safu

antonilol:
> what about lowering qr error correction but only for base32 keeping base58 with high qr error correction funds should be safu

oh i said bech32 has error correction, but base58 can still _detect_ error (with sha256)
(both have error _detection_, only bech32 has _correction_)

qr code error detection is useful when printed on a piece of paper, where some ink can get vague, but in the mempool case it is a screen where the code gets displayed on, for possible dead pixels on screens low error correction is fine

katesalazar:
I printed an address page and it looked kinda horrible
but was usable enough

somewhere someone will use it as a paper wallet,
surely they deserve high error correction

antonilol:
> I printed an address page and it looked kinda horrible but was usable enough

lol

> somewhere someone will use it as a paper wallet, surely they deserve high error correction

dont think so, there are website for that that give nice printable things that also have a private key

but a button (or clicking the qr) could be added to scale up the qr (for better scanning) and/or increase error correction level

katesalazar:
users will commit mistake
resistance is futile

I think it's a good idea allowing different error correction triggered by user, yeah

wiz (approved):
Tested ACK @ v2.4.1-dev [[98db8b1b](https://github.com/mempool/mempool/commit/98db8b1b)]

Issue #1941, opened 2022-06-26
Potentiel high CPU usage in dashboards


nymkappa:
I've noticed my laptop running hot recently without running any heavy task and it seems to me that dashboards are using a fair amount of CPU.

I would say it's probably because of the "breathing" color effect on the mempool blocks (which afaik don't have any frame rate limitation?), but it needs to be investigated if you can reproduce.

Chrome is using less CPU than Firefox, but both shows a significant difference between empty tab and mempool app. See screenshots below.

### Firefox (new tab vs mempool)
<img width="1728" alt="Screen Shot 2022-06-26 at 4 39 25 PM" src="https://user-images.githubusercontent.com/9780671/175819923-9670e587-99e6-4363-9b03-21d476a3e7d2.png">
<img width="1728" alt="Screen Shot 2022-06-26 at 4 39 50 PM" src="https://user-images.githubusercontent.com/9780671/175819926-6a0b7808-4228-4bbf-a8d0-7ab074403edb.png">

### Chrome (new tab vs mempool)
<img width="1728" alt="Screen Shot 2022-06-26 at 4 41 05 PM" src="https://user-images.githubusercontent.com/9780671/175819930-2f708121-3634-4900-8f6d-9ffc934b0a26.png">
<img width="1728" alt="Screen Shot 2022-06-26 at 4 41 31 PM" src="https://user-images.githubusercontent.com/9780671/175819932-fe92cb16-e744-4238-ba8e-46310d5b1bfd.png">

### Firefox, other mempool pages
<img width="1728" alt="Screen Shot 2022-06-26 at 4 43 46 PM" src="https://user-images.githubusercontent.com/9780671/175819935-a7cbf1df-3ef9-44ed-858e-259acf7e7b98.png">
<img width="1728" alt="Screen Shot 2022-06-26 at 4 43 59 PM" src="https://user-images.githubusercontent.com/9780671/175819936-efd2aeed-451f-4c98-80f6-40d84718b760.png">
<img width="1728" alt="Screen Shot 2022-06-26 at 4 44 10 PM" src="https://user-images.githubusercontent.com/9780671/175819939-156dea20-c981-48ab-adec-fdd086446b97.png">

eugeneprateek:
Not happening on my windows machine. mempool is as good as empty tab.

wiz:
Just a comment but the FPS of the mempool block breathing effect is most noticeable on devices like Raspberry Pies that don't have a proper GPU

[...]

PR #1953, opened 2022-06-29
limit pulsing blocks animation frame rate to 30FPS

mononaut:
This PR applies the CSS `steps()` timing function to effectively limit the maximum frame rate of the 'breathing' blocks effect, and reduce the high CPU and GPU usage reported in #1941.

`steps()` is [about as well supported as CSS animations in general](https://caniuse.com/?search=steps), so compatibility should be fine.

30 FPS seemed like a good compromise between smoothness and performance, but we can adjust up or down as needed.

note that this method requires *linear* interpolation, instead of *ease-out*, which changes the overall effect slightly.

### DIFF
--- a/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.scss
+++ b/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.scss
@@ -114,7 +114,11 @@
 }
 
 .flashing {
-  animation: opacityPulse 2s ease-out;
+  /* force compositing */
+  will-change: opacity;
+  transform: translateZ(0);
+  /* effective max frame rate = (#keyframes - 1) x steps / duration */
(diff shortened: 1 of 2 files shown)


wiz:
@nymkappa does this solve the issue you reported?

nymkappa (approved):
The animation is going easier on the CPU and GPU now yes. I think it would be a nice improvement for low spec and old hardware.

I can't see any difference in the animation honestly, you have to know it changed to slightly notice it.

Tested ACK @ f9dfbf94efa6e7b690ae939da7f8cfd814d581ce

Tested using Pro Motion (high screen framerate).

Prod:
Screen Shot 2022-07-07 at 10 53 37 AM

This PR:
Screen Shot 2022-07-07 at 10 54 10 AM

PR #1954, opened 2022-07-02
Block error status code missing fix

softsimon:
Before
Screen Shot 2022-07-02 at 18 47 47

After:
Screen Shot 2022-07-02 at 18 48 01


wiz (approved):
utACK

PR #1968, opened 2022-07-05
Fix sed command for Linux systems in the prod install script

Emzy:
GNU seed needs the file extension directly behind the "-i" option.


wiz (approved):
utACK

PR #1971, opened 2022-07-05
Install packages, Bitcoin and Liquid core services in prod install script

Emzy:


wiz (approved):
utACK
did you use our special keyboard for this?

PR #1985, opened 2022-07-06
Add crontab creation for linux and bitcoin-minfee.service in prod install

Emzy:


production/linux/bitcoin-minfee.service
```diff
@@ -0,0 +1,22 @@
+[Unit]
+Description=Bitcoind-minfee
+After=network.target
+
+[Service]
+ExecStart=/usr/local/bin/bitcoind -daemon -printtoconsole -pid=/minfee/bitcoind-minfee.pid
+ExecStop=/usr/local/bin/bitcoin-cli stop
```
wiz, line 7: I would prefer to kill -TERM ${pid} with a long timeout (90 or 120 seconds?) until kill -9 occurs

PR #1995, opened 2022-07-06
Minor CI improvements

erikarvstedt:
#### Commit "CI: Use `npm ci` instead of `npm install"
See also: [`npm ci` docs](https://docs.npmjs.com/cli/v8/commands/npm-ci)


knorrium (approved):
Tested ACK

PR #2000, opened 2022-07-07
Add editorconfig and prettier integration with eslint

knorrium:
This PR adds some initial formatting rules and will be applied if the editorconfig plugin is installed. After the other big PR is merged we can format all the code and avoid merge conflicts.


wiz (approved):
utACK

wiz (approved):

Issue #1939, opened 2022-06-26
Automatic block re-indexing upon pools.json update


nymkappa:
We can re-index unknown blocks automatically using https://github.com/mempool/mining-pools/pull/7#issuecomment-1166475654

If we update an existing mining pool, we must identify which one and delete all its related blocks from the database as well for re-indexing.

All hashrates must be re-indexed regardless of the situation.

PR #2002, opened 2022-07-07
 Automatic block re-indexing upon pools.json update

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1939

This PR _finally_ adds automatic blocks/hashrates re-indexing when minning pools.json file is updated.

* All `unknown` tagged blocks from height > 130635 are always deleted for re-indexing
* All blocks related to any mining pool for which regexes or addresses has been changed are deleted for re-indexing

### Testing

We have two mining pools.json PR opened, so you can use those to test. For example:

* https://github.com/mempool/mining-pools/pull/8 is a mining pool update (we add new regexes). Right now, some blocks are wrongly tagged to `unknown`

Screen Shot 2022-07-07 at 1 52 54 PM

* So to test it, you need to manually change the url in `pools-updater.ts` as we don't have config option for that yet

```
poolsUrl: string = 'https://raw.githubusercontent.com/mempool/mining-pools/nymkappa/feature/emcdpool/pools.json';
treeUrl: string = 'https://api.github.com/repos/mempool/mining-pools/git/trees/nymkappa/feature/emcdpool';
```

* Restart the node backend and you'll notice this in the logs `NOTICE: Deleting blocks from ECMD Pool mining pool for future re-indexing`
* Once the re-indexing is complete, the blocks won't show in the `unknown` mining pool anymore but in the EMCD Pool
Screen Shot 2022-07-07 at 1 50 13 PM

* This process can be re-verified with https://github.com/mempool/mining-pools/pull/7 as well


wiz (approved):
Tested ACK @ v2.4.1-dev [[7ba7440b](https://github.com/mempool/mempool/commit/7ba7440b)]

Tested using new EMCD pool coinbase tags PR to mining-pools repo. Worked well on this PR branch to automatically index the past and future blocks, and tested using v2.4.0 which worked well to automatically update the data to only index future blocks.

PR #2006, opened 2022-07-07
Merging the 3 "block predictions" i18n string into 1

softsimon:


wiz (approved):

wiz (approved):
Tested ACK v2.4.1-dev [[a55eb653](https://github.com/mempool/mempool/commit/a55eb653)]

PR #2011, opened 2022-07-07
Fix npm install commands in Dockerfiles and ops scripts

wiz:


knorrium (approved):
utACK - CI passes and we'll tag a new RC for Docker soon to test the rest of the PR

PR #2033, opened 2022-07-08
Fix hashrate indexing, log difficulty adjustment progress

nymkappa:
* Removed hardcoded genesis timestamps for hasharate indexing
* Fix duplicated genesis hashrate attempt (forgot to divide a timestamp by 1000 before checking if it was already indexed), leading to the following error `Jul  8 09:24:53 node201.fmt  mempool[82115]: ERR <testnet> Indexer failed, trying again in 10 seconds. Reason: Duplicate entry '2009-01-03 18:15:05-0' for key 'hashrate_timestamp_pool_id'`
* Add log during difficulty adjustment indexing, it takes quite a while for testnet so it deserves basic progress log

### Testing

* Nuke the db and re-index everything from scratch (db schema v2) to make sure the full indexing cycle is working properly
* Check hashrate & difficulty charts, weekly mining pool chart and any block related chart, look for the first entry in the `All` timespan to see if it ~matches the network genesis time

### DIFF
--- a/backend/src/api/mining.ts
+++ b/backend/src/api/mining.ts
@@ -189,10 +189,12 @@ class Mining {
     }
 
     try {
+      const genesisBlock = await bitcoinClient.getBlock(await bitcoinClient.getBlockHash(0));
+      const genesisTimestamp = genesisBlock.time * 1000;
+
       const indexedTimestamp = await HashratesRepository.$getWeeklyHashrateTimestamps();
       const hashrates: any[] = [];
-      const genesisTimestamp = 1231006505000; // bitcoin-cli getblock 000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f
-
+ 
       const lastMonday = new Date(now.setDate(now.getDate() - (now.getDay() + 6) % 7));
       const lastMondayMidnight = this.getDateMidnight(lastMonday);
       let toTimestamp = lastMondayMidnight.getTime();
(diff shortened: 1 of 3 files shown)


wiz (changes requested):
NACK - did a full index and got same errors
```
Jul  8 17:55:40 node201.fra  mempool[81525]: ERR <testnet> Cannot save indexed hashrate into db. Reason: Duplicate entry '2009-01-03 18:15:05-0' for key 'hashrate_timestamp_pool_id'
```

nymkappa:
> NACK - did a full index and got same errors
> 
> ```
> Jul  8 17:55:40 node201.fra  mempool[81525]: ERR <testnet> Cannot save indexed hashrate into db. Reason: Duplicate entry '2009-01-03 18:15:05-0' for key 'hashrate_timestamp_pool_id'
> ```

The bug was because I was using `AVG()` when selecting existing indexed hashrates timestamp.
The indexer was then thinking that some timestamp where missing in the index therefore attempting to insert them into the db. Which lead to the error seen.
I'm not using the raw db timestamps instead of `AVG()` during the indexing.

wiz (approved):
Tested ACK @ v2.4.1-dev [[07cb4a49](https://github.com/mempool/mempool/commit/07cb4a49)]

Seems to be working well, and the mining dashboard is now usable while the bitfeed visualization is indexing silently in the background. All of Bitcoin mainnet took about 6 hours on our server.

PR #2039, opened 2022-07-08
Use bitcoin core to fetch blocks

nymkappa:
This PR replaces querying blocks through esplora vs bitcoin core when it is requested by users from the frontend.

In very rare occasion, it was possible for the user to query a block right before we index it. Since we were using esplora, the block was indexed on the fly in the db using an rounded difficulty (esplora bug). This created a follow up bug on the difficulty adjustments table.

### Testing

* Deploy the PR, explore some blocks (some new, and some old to make sure they're not in memory cache). Confirm everything still works fine
* Delete 100 most recent blocks
```mysql
delete from blocks order by height desc limit 100;
```
* Explore a block you've just deleted, confirm everything works fine
* Open the block list page `/blocks`, confirm everything works fine, browse 6~7 pages
* Check if blocks are properly re-indexed on the fly
```mysql
select * from blocks order by height desc limit 100;
```
* **Clear your disk cache** and test with `config.DATABASE.ENABLED: false` and `MINING_DASHBOARD: false` (can still browse blocks, and see the blocks list)

Note: Bisq and liquid are not be impacted because it uses a different code branch (the legacy /blocks backend code)

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -168,7 +168,7 @@ class Blocks {
       blockExtended.extras.avgFeeRate = stats.avgfeerate;
     }
 
-    if (['mainnet', 'testnet', 'signet', 'regtest'].includes(config.MEMPOOL.NETWORK)) {
+    if (['mainnet', 'testnet', 'signet'].includes(config.MEMPOOL.NETWORK)) {
       let pool: PoolTag;
       if (blockExtended.extras?.coinbaseTx !== undefined) {
         pool = await this.$findBlockMiner(blockExtended.extras?.coinbaseTx);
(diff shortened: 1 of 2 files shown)


wiz (changes requested):
Tested NACK @ v2.4.1-dev [[8c995356](https://github.com/mempool/mempool/commit/8c995356)]

When browsing a yet-unindexed block, I get the following error
Screen Shot 2022-07-08 at 15 49 32

wiz (approved):
Tested ACK @ v2.4.1-dev [[e1fe6078](https://github.com/mempool/mempool/commit/e1fe6078)]

wiz (approved):
Tested ACK @ v2.4.1-dev [[e1fe6078](https://github.com/mempool/mempool/commit/e1fe6078)]

PR #2041, opened 2022-07-08
Install nginx config also for freebsd on prod install

Emzy:

### DIFF
--- a/production/install
+++ b/production/install
@@ -183,6 +183,9 @@ case $OS in
         TOR_PKG=tor
         TOR_USER=_tor
         TOR_GROUP=_tor
+        NGINX_USER=www
+        NGINX_ETC_FOLDER=/usr/local/etc/nginx
+        NGINX_CONFIGURATION=/usr/local/etc/nginx/nginx.conf
         CERTBOT_PKG=py39-certbot
     ;;
 
@@ -197,6 +200,7 @@ case $OS in
         TOR_USER=debian-tor
         TOR_GROUP=debian-tor
         CERTBOT_PKG=python3-certbot-nginx
+        NGINX_USER=www-data
         NGINX_ETC_FOLDER=/etc/nginx
         NGINX_CONFIGURATION=/etc/nginx/nginx.conf
     ;;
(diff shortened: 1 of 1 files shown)


production/install
```diff
@@ -1536,26 +1540,17 @@ NGINX_BISQ_ONION=$(cat "${TOR_RESOURCES}/bisq/hostname")
...
-        osSudo "${ROOT_USER}" sed -i.orig "s!__NGINX_MEMPOOL_ONION__!${NGINX_MEMPOOL_ONION%.onion}!" "${NGINX_CONFIGURATION}"
-        osSudo "${ROOT_USER}" sed -i.orig "s!__NGINX_BISQ_ONION__!${NGINX_BISQ_ONION%.onion}!" "${NGINX_CONFIGURATION}"
-        osSudo "${ROOT_USER}" sed -i.orig "s!__NGINX_LIQUID_ONION__!${NGINX_LIQUID_ONIONi%.onion}!" "${NGINX_CONFIGURATION}"
-        echo "[*] Restarting Nginx"
-        osSudo "${ROOT_USER}" service nginx restart
-    ;;
-esac
+osSudo "${ROOT_USER}" install -c -o "${ROOT_USER}" -g "${ROOT_GROUP}" -m 644 "${MEMPOOL_HOME}/${MEMPOOL_REPO_NAME}/production/nginx/nginx.conf" "${NGINX_CONFIGURATION}"
+mkdir -p /var/cache/nginx/services /var/cache/nginx/api
+chown ${NGINX_USER}: /var/cache/nginx/services /var/cache/nginx/api
```
wiz, line 1545: Just make /var/cache/nginx and nginx will manage everything under that

production/install
```diff
@@ -1536,26 +1540,17 @@ NGINX_BISQ_ONION=$(cat "${TOR_RESOURCES}/bisq/hostname")
...
+ln -s /mempool/mempool /etc/nginx/mempool
```
wiz, line 1546: This is only needed on linux because on FreeBSD we do something like this:
```
nginx_enable="YES"
nginx_profiles="mempool"
nginx_mempool_flags="-p /mempool"
nginx_mempool_configfile="/mempool/mempool/production/nginx/nginx.conf"
```

wiz (changes requested):

Issue #2050, opened 2022-07-09
[Difficulty adjustments] - Use raw db value instead of avg for the diff table


nymkappa:
<img width="561" alt="Screen Shot 2022-07-09 at 12 10 00 PM" src="https://user-images.githubusercontent.com/9780671/178101286-0450496e-d2d8-4133-afae-053bfec28143.png">

PR #2051, opened 2022-07-09
Fix diff adj table using raw db value

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2050

Also fixes some indexing issues when INDEXING_BLOCK_AMOUNT is not set to -1


wiz (approved):
Tested ACK @ v2.4.1-dev [[30e8b134](https://github.com/mempool/mempool/commit/30e8b134)]

### Before

Screen Shot 2022-07-10 at 13 13 17

### After

Screen Shot 2022-07-10 at 13 13 22

Issue #2048, opened 2022-07-09
[Prices index] - Don't insert price if there is no match USD entry


nymkappa:
There are 4 entries where the EUR price timestamp does not have any matching USD price timestamp, let's ignore them during the indexing.

<img width="570" alt="Screen Shot 2022-07-09 at 10 29 54 AM" src="https://user-images.githubusercontent.com/9780671/178098203-b392a29d-bc5e-414a-b2e5-c4ed8338ee5a.png">

PR #2054, opened 2022-07-09
Ignore prices without USD exchange rate

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2048

Will be useful for an incoming PR linking blocks and prices.


wiz (approved):
Tested ACK @ v2.4.1-dev [[7e389c88](https://github.com/mempool/mempool/commit/7e389c88)]

PR #2056, opened 2022-07-09
Disable automatic block re-indexing by default

nymkappa:
Make automatic block re-indexing upon mining pools.json change configurable

Untested 😁


wiz (approved):
Partially Tested ACK @ v2.4.1-dev [[d697c0c4](https://github.com/mempool/mempool/commit/d697c0c4)]

Issue #2060, opened 2022-07-10
Liquid Block API has changed, breaking block page


softsimon:
Current master:

https://liquid.place/liquid/api/v1/block/ce66634b1ffb86c2afce7c3b17398ab260b94bbb79d6d6b8e46d7d8126521c7d

v2.4.1-dev [[001bd1d4](https://github.com/mempool/mempool/commit/001bd1d4)] block API response:

https://liquid.network/liquid/api/v1/block/ce66634b1ffb86c2afce7c3b17398ab260b94bbb79d6d6b8e46d7d8126521c7d

Bonus question: why is the block API returning all the TXIDs for the block?

PR #2065, opened 2022-07-11
Liquid always uses esplora (regression of #2039)

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2060
Regression introduced in https://github.com/mempool/mempool/issues/2039

* Tested locally with liquidtestnet (liquid node only)
```
    "NETWORK": "liquidtestnet",
    "BACKEND": "none",
```

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -527,13 +527,12 @@ class Blocks {
       }
     }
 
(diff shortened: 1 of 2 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[0093eab2](https://github.com/mempool/mempool/commit/0093eab2)]

PR #2068, opened 2022-07-11
Block audit page

nymkappa:
Screen Shot 2022-07-11 at 10 38 07 AM

This draft PR adds the first piece of the block audit page, which offers users to visually compare our expected block (currently using getsimontemplate), to the actual mined block.

This leverage @mononaut's visualization work and extend it to highlight the block diff.

Currently I only show the diff using static colors, but as discussed in the dev chat there are many possibilities, some of them already demonstrated here https://mempool.monospace.live/block-audit/000000000000000000049a5f64153078219b447cebb838fdcb486dd403b9763d#mined

This feature requires us to save our block template in database. As this is based on the mempool state, it's not possible to retrieve historical data, so we also need some elegant way to inform the user that "this feature is not available for this block".

Feedbacks appreciated.

Note: The draft status is mostly related to the UX. The code is pretty much ready but we need to polish. Also, a second PR later on will complement this one in order to show the transaction list below the rendering.

### DIFF
--- a/backend/src/repositories/BlocksSummariesRepository.ts
+++ b/backend/src/repositories/BlocksSummariesRepository.ts
@@ -17,14 +17,24 @@ class BlocksSummariesRepository {
     return undefined;
   }
 
-  public async $saveSummary(height: number, summary: BlockSummary) {
+  public async $saveSummary(height: number, mined: BlockSummary | null = null, template: BlockSummary | null = null) {
+    const blockId = mined?.id ?? template?.id;
     try {
-      await DB.query(`INSERT INTO blocks_summaries VALUE (?, ?, ?)`, [height, summary.id, JSON.stringify(summary.transactions)]);
+      const [dbSummary]: any[] = await DB.query(`SELECT * FROM blocks_summaries WHERE id = "${blockId}"`);
+      if (dbSummary.length === 0) { // First insertion
+        await DB.query(`INSERT INTO blocks_summaries VALUE (?, ?, ?, ?)`, [
+          height, blockId, JSON.stringify(mined?.transactions ?? []), JSON.stringify(template?.transactions ?? [])
+        ]);
+      } else if (mined !== null) { // Update mined block summary
+        await DB.query(`UPDATE blocks_summaries SET transactions = ? WHERE id = "${mined.id}"`, [JSON.stringify(mined?.transactions)]);
+      } else if (template !== null) { // Update template block summary
+        await DB.query(`UPDATE blocks_summaries SET template = ? WHERE id = "${template.id}"`, [JSON.stringify(template?.transactions)]);
+      }
     } catch (e: any) {
       if (e.errno === 1062) { // ER_DUP_ENTRY - This scenario is possible upon node backend restart
-        logger.debug(`Cannot save block summary for ${summary.id} because it has already been indexed, ignoring`);
+        logger.debug(`Cannot save block summary for ${blockId} because it has already been indexed, ignoring`);
       } else {
-        logger.debug(`Cannot save block summary for ${summary.id}. Reason: ${e instanceof Error ? e.message : e}`);
+        logger.debug(`Cannot save block summary for ${blockId}. Reason: ${e instanceof Error ? e.message : e}`);
         throw e;
       }
     }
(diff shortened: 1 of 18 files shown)


mononaut (commented):
Looks good! The visualization implementation is exactly how I'd do it, and doesn't break any of the other pages as far as I can tell :)

mononaut:
I noticed the toggle to switch between mined block and template view on mobile isn't implemented yet.

When you get around to wiring that up, I'd suggest re-using a single `block-overview-graph` component on mobile, and switch between the two sets of transactions using the `destroy`/`setup` methods.

Swapping out component instances at the template level seems cleaner at first, but I think you might run into some awkward component lifecycle issues.

backend/src/repositories/BlocksSummariesRepository.ts
```diff
@@ -17,14 +17,24 @@ class BlocksSummariesRepository {
     return undefined;
   }
 
-  public async $saveSummary(height: number, summary: BlockSummary) {
+  public async $saveSummary(height: number, mined: BlockSummary | null = null, template: BlockSummary | null = null) {
```
knorrium, line 20: @nymkappa consider simplifying this a bit with a small refactor so you don't have to pass around null values which can be confusing:

```
async function $saveSummary(params : {height: number, mined?: BlockSummary, template?: BlockSummary}) {
     const { height, mined, template } = params;
...
if (mined) {
}

if (template) {
}
```

Then call it with the parameters you will actually use:
```
await BlocksSummariesRepository.$saveSummary({height: 123, mined: minedBlockSummary});
await BlocksSummariesRepository.$saveSummary({height: 123, template: templateBlockSummary})
```

frontend/src/app/components/block-overview-graph/tx-view.ts
```diff
@@ -140,6 +142,16 @@ export default class TxView implements TransactionStripped {
   }
 
   getColor(): Color {
+    // Block audit
+    if (this.status === 'found') {
+      // return hexToColor('1a4987');
```
knorrium, line 147: Is this intentional?

frontend/src/app/components/block-audit/block-audit.component.html
```diff
@@ -0,0 +1,111 @@
...
+        <div class="col-sm" *ngIf="blockAudit">
+          <table class="table table-borderless table-striped">
+            <tbody>
+              <tr>
+                <td class="td-width" i18n="shared.transaction-count">Transactions</td>
+                <td>{{ blockAudit.tx_count }}</td>
+              </tr>
+              <tr>
+                <td i18n="block.match-rate">Match rate</td>
+                <td>{{ blockAudit.matchRate }}%</td>
```
knorrium, line 65: @nymkappa don't you have to format the match rate with `| number: '1.2-2'`?

frontend/src/app/components/block-overview-graph/tx-view.ts
nymkappa: Yes, we're still in draft 👍🏻 . Once we settle on the color/animation style this will be cleaned up.

frontend/src/app/components/block-audit/block-audit.component.html
nymkappa: It has only 2 digits in db, but that could be good to add this anyway in case javascript decides to send us a million digits because meh floats

wiz (approved):
Tested ACK @ v2.5.0-dev [[1be7c953](https://github.com/mempool/mempool/commit/1be7c953)]

Screen Shot 2022-07-14 at 16 50 41

wiz:
@nymkappa any reason not to merge this?

nymkappa:
> @nymkappa any reason not to merge this?

I've set in draft because I wanted to discuss a bit about the feature itself but it seems we have consensus. The page is not feature complete but since it's not integrated in the app yet (no link to this page) we can merge it and start saving our GST templates already.

* @mononaut will make a future PR to change the way we highlight the txs diff.
* I will work on displaying the transaction list below (another PR).
* Also, we discussed the possibility of showing the [mempool blocks | blockchain blocks] at the top, which would be another PR as well.

PR #2069, opened 2022-07-11
[Indexer] Set log level accordingly - Remove indexing ETAs

nymkappa:
Only show `... indexed X things` messages in NOTICE. If `X is 0`, use DEBUG instead.
Removed indexing ETAs because they are not accurate at all right now.


wiz (approved):
Tested ACK @ v2.5.0-dev [[475bb119](https://github.com/mempool/mempool/commit/475bb119)]

PR #2070, opened 2022-07-11
frontend: Always reference `resources` relative to root

erikarvstedt:
In some cases, the frontend references `resources` via relative URLs.
Example: `./resources/bitcoin-logo.png`, which resolves to `/<lang>/resources/bitcoin-logo.png` for non-default langs.
Convert all of these instances to root-relative URLs (`/resources/bitcoin-logo.png`).

With this fix, the language-invariant `resource` dir no longer has to be copied to each lang dir, which shrinks the frontend size by 55% (279M -> 124M).
Also, the nginx location config can be simplified.

#### Correctness

I've manually checked that relative `resources` access no longer happens by logging browser requests.

Also, searching the frontend for instances of string `resource` that are not preceded by `"/`, `'/` or  `` `/ ``  shows no offending results.
```bash
grep -irP "$(cat <<'EOF'
(?<!["'`]/)resources
EOF
)" frontend
```

Still, these checks might be incomprehensive, so the frontend requires further real-world testing before switching off copying of `resources`.

This patch is deployed at https://mempool.nixbitcoin.org

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -4,7 +4,7 @@ import logger from '../logger';
 import { Common } from './common';
 
 class DatabaseMigration {
-  private static currentVersion = 27;
+  private static currentVersion = 28;
   private queryTimeout = 120000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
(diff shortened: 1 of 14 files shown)


wiz:
@erikarvstedt in the future can you please sign all commits? thanks!

PR #2071, opened 2022-07-11
Add maxmind geoip-db update utility to prod installer

wiz:
@Emzy is this the correct debian linux package?


Emzy (approved):
Package name is right.

Issue #1888, opened 2022-06-15
dist folder almost 4x increase in size (274M to 1014M) because of large file with a lot of copies


antonilol:
### Description

while trying to clean up some files i found out my local mempool git repo is really big an has grown recently.
it is 2.4G now where the majority comes from frontend (2.3G) which has dist of 1 gig and node modules of 767M

i have not seen this before, so took a look at the mempool instance running on my bitcoin node, which was updated around a week ago (e99a6843547b2e69f4f398cdc34651f986716e5d), and its dist folder is way smaller, 274M

when searching for the cause i found the resources folder of every locale was the same (4.4M on my node and 30M on my pc) x30
the difference in size comes from two files, `assets-testnet.json` and `assets-testnet.minimal.json` (26M). when i divide this difference over the size difference of the whole dist folder it tells me there are about 31 resources folders, each in one folder for e specific language.

i actually dont need a single copy of this file because i dont run liquid testnet mempool. there is already this `mempool-frontend-config.json` where you can specify networks, maybe some script can not include resources not needed by the current setup (for example this: no `assets.*.json` when not running liquid or no signet logo when not running signet etc)

i think this can be solved with symlinks between the resources folder (requires fs support, fat32 doesnt have this) or some nginx config

#### Version

c5bcf76353f2fd220e7ac9ba69cfa52e337353a9

#### Additional info

must have happened somewhere between e99a6843547b2e69f4f398cdc34651f986716e5d (6 jun) and c5bcf76353f2fd220e7ac9ba69cfa52e337353a9 (15 jun)

antonilol:
this issue is almost solved by #2078, but i still have the liquid asset jsons in my dist folder

```
$ ls dist/mempool/browser/resources/ -lah
-rw-r--r--  1 antoni antoni 3.7K Sep  9 15:29 assets.json
-rw-r--r--  1 antoni antoni  704 Sep  9 15:29 assets.minimal.json
-rw-r--r--  1 antoni antoni  23M Sep  9 15:29 assets-testnet.json
-rw-r--r--  1 antoni antoni 3.0M Sep  9 15:29 assets-testnet.minimal.json
```

for liquid these files could be minified and the `*.minimal.json` could be removed because they contain information also it the non-`minimal`
for non liquid builds they could be removed but i don't know if this is easy because they use the same frontend sources

PR #2078, opened 2022-07-11
frontend: Don't copy `resources`, shrink static size

erikarvstedt:
This PR can be merged as soon as it's certain that the frontend no longer accesses `resources` in language dirs (like `/fr/resources/bitcoin-logo.png`).
This should be the case since #2070.

This PR is currently deployed at https://mempool.nixbitcoin.org

#### Copy of commit msg
Since 355e89ce5, the frontend references resources via root-relative URLs.
This means that `resources` dirs in the language dirs are no longer accessed and can be removed.

As of fd35c8f4a, this shrinks the frontend size by 55% (279M -> 124M).

Also, the nginx location configs now can be simplified.

### DIFF
--- a/frontend/angular.json
+++ b/frontend/angular.json
@@ -170,6 +170,10 @@
           },
           "configurations": {
             "production": {
+              "assets": [
+                "src/favicon.ico",
+                "src/robots.txt"
+              ],
               "fileReplacements": [
                 {
                   "replace": "src/environments/environment.ts",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -34,7 +34,7 @@
     "start:local-staging": "npm run generate-config && npm run sync-assets-dev && npm run ng -- serve -c local-staging",
     "start:mixed": "npm run generate-config && npm run sync-assets-dev && npm run ng -- serve -c mixed",
     "build": "npm run generate-config && npm run ng -- build --configuration production --localize && npm run sync-assets && npm run build-mempool.js",
-    "sync-assets": "node sync-assets.js && rsync -av ./dist/mempool/browser/en-US/resources ./dist/mempool/browser/resources",
+    "sync-assets": "rsync -av ./src/resources ./dist/mempool/browser && node sync-assets.js",
     "sync-assets-dev": "node sync-assets.js dev",
     "generate-config": "node generate-config.js",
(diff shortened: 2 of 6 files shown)


antonilol:
I think this fixes a large part of #1888 (the copying, not the liquid part)

Capi5879 (approved):

Capi5879 (approved):

erikarvstedt:
@wiz, are you still interested in this? I'd rebase it then.

wiz:
Yes, very much interested, please rebase and I'll take a look at it soon.

wiz:
Thanks, I'll review and test this today

erikarvstedt:
`angular serve` (and thus the Cypress tests) are curently broken because `src/resources` are not served.
This is caused by the removal of `"src/resources"` from `architect/build/assets` in `angular.json`.

How can we configure angular so that `ng build` doesn't copy `src/resources` to `dist`, but serves them with `ng serve`?

To locally reproduce the current bug, run:
```bash
npm ci
npm run start:local-prod
curl -f http://localhost:4200/resources/mempool-tube.png
# => 404
```

wiz:
I think we can just modify the proxy.conf.json to serve them from the en-US folder? Talk to you on matrix about it

wiz:
When using the local development server we could just copy them to dist/ like we did before, the important thing is to not do that for production builds / docker builds, right?

wiz (approved):
Tested ACK for prod build @ 997bff116df1953b51a4a0a4158e42651233e73c
but NACK for dev build due to missing `src/resources/` in `dist/` folder - so please fix that before we can merge

### Master
```
% du -sh *
315M	bisq/
324M	liquid/
315M	mainnet/
```
### Using this PR
```
% du -sh *
 74M	bisq/
 79M	liquid/
 74M	mainnet/
```

erikarvstedt:
`ng serve` is now fixed and the tests pass.

@wiz, I've added a fix to your `production/mempool.crontab`, can you check if this is correct?

wiz:
Thanks for the great PR, @softsimon @nymkappa @knorrium please keep an eye out for any side-effects to the dev env or cypress tests related to this

erikarvstedt:
Note: This PR only affects the production config, the dev env (and the cyprus tests) are unchanged. The resources are served from `src/resources`, just like before.

Issue #2084, opened 2022-07-12
Hashrate charts js error


nymkappa:
<img width="1540" alt="Screen Shot 2022-07-12 at 8 38 15 AM" src="https://user-images.githubusercontent.com/9780671/178424893-7592423a-6f3b-4a29-a1af-febeb98725a9.png">
<img width="1728" alt="Screen Shot 2022-07-12 at 8 38 05 AM" src="https://user-images.githubusercontent.com/9780671/178424896-240ab677-edc6-4d44-a763-1fd465457b44.png">

PR #2085, opened 2022-07-12
[Hashrate chart] Fix javascript error if difficulty array is empty

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2084


wiz (approved):
utACK since not sure how to reproduce/test the issue

PR #2086, opened 2022-07-12
Missing blocks prices

nymkappa:
This PR adds the link between newly mined blocks and their respective prices. Up to today, we were only indexing previous blocks to their prices, and were simply catching up once every hour when the price indexer runs.

This PR also makes sure that we don't attempt to link blocks to their prices unless the price updater has completed inserting the missing historical prices, in order to not link a block against an outdated price.

### Testing
* Unlink last blocks prices
```mysql
delete from blocks_prices order by height desc limit 10;
```
* Unlink some random blocks prices, for example
```mysql
delete from blocks_prices where height in (123456, 234567, 345678, 456789, 567890)
```
* Clear your disk cache (so we fetch the most recent 8 blocks)
* Tail nodejs backend logs with DEBUG level enabled
* Start the node backend, notice the following message
```
Jul 16 08:34:17 [77751] DEBUG: New block found (#745179)!
[...]
Jul 16 08:34:17 [77751] INFO: Cannot save block price for 745179 because the price updater hasnt completed yet. Trying again in 10 seconds.
[...]
Jul 16 08:36:17 [78315] INFO: Fetching daily price history from exchanges and saving missing ones into the database, this may take a while
[...]
Jul 16 08:36:33 [78315] DEBUG: Blocks prices indexer is waiting for the price updater to complete
[...]
Jul 16 08:40:37 [79409] DEBUG: Blocks prices indexer will run now
[...]
Jul 16 08:40:37 [90445] DEBUG: Linking 15 blocks to their closest price
```
* Confirm there is not gap in block prices using (should not return any row)
```mysql
SELECT
 CONCAT(z.expected, IF(z.got-1>z.expected, CONCAT(' thru ',z.got-1), '')) AS missing
FROM (
 SELECT
  @rownum:=@rownum+1 AS expected,
  IF(@rownum=height, 0, @rownum:=height) AS got
 FROM
  (SELECT @rownum:=0) AS a
  JOIN blocks_prices
  ORDER BY height
 ) AS z
WHERE z.got!=0;
```
* Testing will continue if merged into master, as we will in the future use `blocks_prices` to show reward USD value, tx value etc, so we need to monitor if some prices are missing


wiz:
@knorrium can you review and test this plz

nymkappa:
Screen Shot 2022-08-08 at 10 16 25 AM

wiz (approved):
Tested ACK @ 33cf698

```
Aug 10 13:32:59 node201.fra  mempool[43081]: DEBUG Linking 15 blocks to their closest price
```


```
root@localhost [mempool]> SELECT
    ->  CONCAT(z.expected, IF(z.got-1>z.expected, CONCAT(' thru ',z.got-1), '')) AS missing
    -> FROM (
    ->  SELECT
    ->   @rownum:=@rownum+1 AS expected,
    ->   IF(@rownum=height, 0, @rownum:=height) AS got
    ->  FROM
    ->   (SELECT @rownum:=0) AS a
    ->   JOIN blocks_prices
    ->   ORDER BY height
    ->  ) AS z
    -> WHERE z.got!=0;
Empty set (0.508 sec)
```

PR #2088, opened 2022-07-12
[LN] Add `as_organization` in nodes table

nymkappa:
Screen Shot 2022-07-12 at 12 12 55 PM

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -4,7 +4,7 @@ import logger from '../logger';
 import { Common } from './common';
 
 class DatabaseMigration {
-  private static currentVersion = 29;
+  private static currentVersion = 30;
   private queryTimeout = 120000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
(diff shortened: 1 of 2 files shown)


softsimon:
I added `as_number` in the nodes table so that we can add additional information for the specific AS in the `geo_names` to not repeat name strings. Maybe add a new "type" in the geo_names table instead?

nymkappa:
> I added `as_number` in the nodes table so that we can add additional information for the specific AS in the `geo_names` to not repeat name strings. Maybe add a new "type" called `organization` in the geo_names table instead?

Isn't this table used for translatable fields?

softsimon (approved):
utACK

wiz (approved):
Tested ACK @ v2.5.0-dev [[3d2ff7ef](https://github.com/mempool/mempool/commit/3d2ff7ef)]
```
root@localhost [mempool_mainnet_lightning]> select * from geo_names where id = 397444;
+--------+-----------------+----------+
| id     | type            | names    |
+--------+-----------------+----------+
| 397444 | as_organization | "AS-OKI" |
+--------+-----------------+----------+
1 row in set (0.000 sec)
```

PR #2097, opened 2022-07-12
Add nodes per AS chart

nymkappa:
This PR add a new chart for the lightning explorer, which show which AS are used the most by clearnet LN nodes.

Screen Shot 2022-07-16 at 11 31 15 AM

Screen Shot 2022-07-16 at 11 31 24 AM

@softsimon Could you please double check if [that query](https://github.com/mempool/mempool/pull/2097/files#diff-f84fbb84918b2fec7a767289be371d57ae68dc0fb1bf4e42ba01b750f368aea8R99) makes sense, to get the "running capacity" per AS?

### DIFF
--- a/backend/src/api/explorer/nodes.api.ts
+++ b/backend/src/api/explorer/nodes.api.ts
@@ -93,6 +93,35 @@ class NodesApi {
       throw e;
     }
   }
+
+  public async $getNodesAsShare() {
+    try {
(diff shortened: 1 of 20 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[01ec0751](https://github.com/mempool/mempool/commit/01ec0751)]

PR #2098, opened 2022-07-12
Add nodes per country table page

nymkappa:
This PR adds a very simple page to browser a list of nodes per country (no pagination). You can access it via the following url `/lightning/nodes/country/:country`.

For example `/lightning/nodes/country/france` or `/lightning/nodes/country/sweden`

Screen Shot 2022-07-16 at 3 57 43 PM
Screen Shot 2022-07-16 at 3 58 48 PM

Screen Shot 2022-07-16 at 3 56 48 PM
Screen Shot 2022-07-16 at 3 59 04 PM

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -4,7 +4,7 @@ import logger from '../logger';
 import { Common } from './common';
 
 class DatabaseMigration {
-  private static currentVersion = 32;
+  private static currentVersion = 33;
   private queryTimeout = 120000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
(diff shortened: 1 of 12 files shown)


wiz (changes requested):
Need a matching /graphs page pie chart for "Lightning nodes per country" with a table where clicking on a pie slice country goes to this page. Also the URL should use the 2 letter country code like US or DE or JP

nymkappa:
> Need a matching /graphs page pie chart for "Lightning nodes per country" with a table where clicking on a pie slice country goes to this page. Also the URL should use the 2 letter country code like US or DE or JP

Fixed in https://github.com/mempool/mempool/pull/2098/commits/93e93d44f4b607178b4c80fffc738c073c7a510c
You can now access the list using country codes, for example `/lightning/nodes/country/us`

softsimon (approved):
code looks good

utACK

wiz (approved):
Tested ACK @ v2.5.0-dev [[93e93d44](https://github.com/mempool/mempool/commit/93e93d44)]

PR #2099, opened 2022-07-13
Re-add missing migration that was dropped during merge conflict

nymkappa:
This migration code was dropped in 40634a0eb85be423b0a519840d4f01036f195160


wiz (approved):
utACK

PR #2112, opened 2022-07-16
Ignore Kraken historical price without USD

nymkappa:
Mini PR that fixes the incorrect log we see at each nodejs backend restart. Those 4 prices entries only contain `EUR` prices, therefore we never insert them into DB. So every time we restart the nodejs backend, the price updater thinks they are missing, and we print this log line.

`Jul 16 09:27:53 [92458] NOTICE: Inserted 4 Kraken EUR, USD, GBP, JPY, CAD, CHF and AUD weekly price history into db`


wiz (approved):
utACK

Issue #2116, opened 2022-07-16
Create LN nodes per country graph


nymkappa:
Need a matching /graphs page pie chart for "Lightning nodes per country" with a table where clicking on a pie slice country goes to this page. Also the URL should use the 2 letter country code like US or DE or JP

_Originally posted by @wiz in https://github.com/mempool/mempool/pull/2098#pullrequestreview-1041029397_

PR #2118, opened 2022-07-17
LN nodes per country pie chart

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2116

Depends on https://github.com/mempool/mempool/pull/2098. You can either test everything country related with this PR and close the other other, or test them separately.

Note: When you switch between the pie chart and the node per country list pages, it may be very slow/buggy for the top countries. I believe this is due to the lack of pagination in the node per country list component, which I will address in a separate PR.

Screen Shot 2022-07-17 at 11 02 45 AM
Screen Shot 2022-07-17 at 11 03 20 AM

### DIFF
--- a/frontend/src/app/lightning/nodes-per-country-chart/nodes-per-country-chart.component.html
+++ b/frontend/src/app/lightning/nodes-per-country-chart/nodes-per-country-chart.component.html
@@ -35,7 +35,11 @@
         <tr *ngFor="let country of countries">
           <td class="text-left rank">{{ country.rank }}</td>
           <td class="text-left text-truncate name">
-            <a [routerLink]="['/lightning/nodes/country' | relativeUrl, country.iso]">{{ country.name.en }}</a>
+            <div class="d-flex">
+              <span style="font-size: 20px">{{ country.flag }}</span>
+              &nbsp;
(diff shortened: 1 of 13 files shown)


frontend/src/app/lightning/nodes-per-country-chart/nodes-per-country-chart.component.html
```diff
@@ -0,0 +1,54 @@
...
+    </div>
+
+    <div class="text-center loadingGraphs" *ngIf="isLoading">
+      <div class="spinner-border text-light"></div>
+    </div>
+
+    <table class="table table-borderless text-center m-auto" style="max-width: 900px">
+      <thead>
+        <tr>
+          <th class="text-left rank" *ngIf="!isMobile()" i18n="mining.rank">Rank</th>
```
softsimon, line 27: Function calls from template can cause performance problems so consider setting a static variable instead

nymkappa: Yeah actually I already hide them with css but I forgot to remove those

softsimon (approved):
utACK

wiz (approved):
Tested ACK @ v2.5.0-dev [[ccdeb108](https://github.com/mempool/mempool/commit/ccdeb108)]

Later we should make the flags clickable in the list of nodes by country, and also add flag emojis to the node's country on the node page

PR #2123, opened 2022-07-17
Fix error 500 for `Isle of Man` nodes list

nymkappa:
Fixes(TM) and error 500 when querying https://ln.fra.mempool.space/api/v1/lightning/nodes/country/im


wiz (approved):
Tested ACK @ v2.5.0-dev [[3633d36b](https://github.com/mempool/mempool/commit/3633d36b)]

PR #2129, opened 2022-07-18
Create lightning nodes world map to show where most nodes are

nymkappa:
This draft PR goal is to define if introducing a new charting library is acceptable in terms of bundle size, as well as experimenting with it.

The problem with echart is that it has limited map feature available, some of them requiring some external plugins, which only one of them AFAIK is still maintained. 

Therefore I decided to explore https://www.amcharts.com/demos/#maps as it seems to offer cool advanced map charting. It also contains a lot of predefined map with different level of details.

I've implemented a very basic "lightning node heat world map" using it. As you can notice, the source code is quite small compared to our echart components https://github.com/mempool/mempool/pull/2129/files#diff-e9c9e7fd183609a374822af7a46378b89b88378d7f9ac3776dfbcf77283f0c19

Screen Shot 2022-07-18 at 12 57 56 PM

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -9,6 +9,9 @@
       "version": "2.5.0-dev",
       "license": "GNU Affero General Public License v3.0",
       "dependencies": {
+        "@amcharts/amcharts5": "^5.2.10",
+        "@amcharts/amcharts5-fonts": "^5.0.1",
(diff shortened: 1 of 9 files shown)


nymkappa:
`default-src_app_graphs_graphs_module_ts.js` is currently 5.41 MB on master. With this PR it increases to 6.90 MB.

Issue #2107, opened 2022-07-14
Add Taproot to transaction features menu


Bosch-0:
Hey guys,

To help with spreading awareness, and hopefully adoption, of bech32m (P2TR), Mempool could add a Taproot option in the features section when viewing a transaction. When hovering over it you could highlight the fee savings and potential privacy benefits this transaction could have had if using this address type. 

Highlighted the section below:

![image](https://user-images.githubusercontent.com/55287964/178893908-8ec23889-ad9a-4bb4-8389-cd19d0b2dbbd.png)

softsimon:
We do have a Taproot feature tag showing up when the transaction is a taproot-spend with a basic tooltip.

https://mempool.space/tx/d3bdc0d73c31ba6fb28c88caf66a9ebcb842fc03c0e206954fe5a57253dd339d

<img width="414" alt="Screen Shot 2022-07-14 at 12 57 20" src="https://user-images.githubusercontent.com/8561090/178967461-5875dd21-c66e-42c9-aaab-eed6414d977b.png">

So the question is what we should write in the tooltip. The Segwit tag I believe already contain the segwit savings.

Bosch-0:
Oh cool, missed that. Some suggestions below:

**Multisig transactions**

Could include a grey pill called 'Privacy' that when hovered over suggests that this multisig output could be more private if using PT2R. This wouldn't be shown on P2TR multisig wallets obviously as we wouldn't know that they exist <3 

For the Taproot pill tooltip, the copy could state "This transaction uses Taproot. Future outputs will will be cheaper by x%" Hard to give an exact number, could just drop the "by x%" part and mention how overall future outputs will be cheaper. 

![image](https://user-images.githubusercontent.com/55287964/179143373-93f61085-3933-4cb5-bfcd-d449286d655b.png)

**P2TR transactions**

The tooltip could say "This transaction uses Taproot. Future outputs will will be cheaper by x%." Like I mentioned above, could drop the by x% and make it more general as its hard to know exact figures. 

![image](https://user-images.githubusercontent.com/55287964/179144006-e1734a53-1b97-430c-a3af-34aefbdb1934.png)

**Transactions with no P2TR outputs**

For transactions that don't use Taproot at all you could show a red strikethroughed Taproot like the SegWit pill below. The tooltip could say "This transaction could make future outputs cheaper by using Taproot."

[...]

PR #2141, opened 2022-07-20
Always show taproot button

softsimon:
refs #2107

This addresses one of the 3 suggestions described in https://github.com/mempool/mempool/issues/2107.

The taproot tag is now always visible to inform if the transaction uses taproot or not.

Next step is to show the actual fee savings.

Screen Shot 2022-07-20 at 19 28 56


wiz (approved):
utACK

PR #2144, opened 2022-07-21
Add worldmap.json, downloaded from https://geojson-maps.ash.ms/

nymkappa:
The website mentioned in the footer

```
These maps are sourced from [Natural Earth](http://naturalearthdata.com/) and are in the public domain, so you are free to use them in any way without attribution.
```


wiz:
I don't see any advantage to using this and none was given in the PR description, so let's wait on this PR until after we implement the zoomed-in maps for individual node and channel pages - then we can compare before and after on all the maps views and I have a feeling we will want to keep the current higher resolution one for zoomed-in maps.

PR #2147, opened 2022-07-21
Create world map of clearnet LN nodes and channels

nymkappa:
After playing around for a couple of hours with settings, I think the result is not to bad. There are some issue I'd like to fix in the future but it will need more hours of testing/playing around, or maybe writing custom code in order to fix them. Maybe mononaut can take over and do some magic in there.

You can move the map around, play around with the 3d view, zoom in/out, and click on the nodes to open their page.

Screen Shot 2022-07-21 at 10 43 22 PM

### DIFF
--- a/backend/src/api/explorer/channels.api.ts
+++ b/backend/src/api/explorer/channels.api.ts
@@ -32,7 +32,7 @@ class ChannelsApi {
         row.node2_public_key, row.node2_alias, row.node2_longitude, row.node2_latitude,
         row.capacity]);
     } catch (e) {
-      logger.err('$getAllChannels error: ' + (e instanceof Error ? e.message : e));
+      logger.err('$getAllChannelsGeo error: ' + (e instanceof Error ? e.message : e));
       throw e;
     }
   }
--- a/backend/src/api/explorer/channels.api.ts
+++ b/backend/src/api/explorer/channels.api.ts
@@ -13,6 +13,30 @@ class ChannelsApi {
     }
   }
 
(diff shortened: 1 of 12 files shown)


softsimon:
The titles on hover needs a higher z-index as I can see lines on top

backend/src/api/explorer/channels.api.ts
```diff
@@ -13,6 +13,30 @@ class ChannelsApi {
...
+      WHERE nodes_1.latitude IS NOT NULL AND nodes_1.longitude IS NOT NULL
+        AND nodes_2.latitude IS NOT NULL AND nodes_2.longitude IS NOT NULL
+      `;
+      const [rows]: any = await DB.query(query);
+      return rows.map((row) => [
+        row.node1_public_key, row.node1_alias, row.node1_longitude, row.node1_latitude,
+        row.node2_public_key, row.node2_alias, row.node2_longitude, row.node2_latitude,
+        row.capacity]);
+    } catch (e) {
+      logger.err('$getAllChannels error: ' + (e instanceof Error ? e.message : e));
```
softsimon, line 35:
```suggestion
      logger.err('$getAllChannelsGeo error: ' + (e instanceof Error ? e.message : e));
```

wiz (approved):
Tested ACK @ v2.5.0-dev [[96983394](https://github.com/mempool/mempool/commit/96983394)]

backend/src/api/explorer/channels.api.ts
Sniper1h: - [ ]

PR #2151, opened 2022-07-22
Navbar logos fix

softsimon:
* Logos are clickable
* Doesn't break on mobile
* Center connection-badge


wiz (approved):
Tested ACK @ v2.5.0-dev [[773e43e0](https://github.com/mempool/mempool/commit/773e43e0)]

PR #2155, opened 2022-07-23
Make flags clickable

nymkappa:
Flags are now clickable in `Lightning nodes per country` chart


wiz (approved):
Tested ACK @ v2.5.0-dev [[8de89a9f](https://github.com/mempool/mempool/commit/8de89a9f)]

PR #2171, opened 2022-07-24
Remove duplicated nodes from the world map

nymkappa:
Right now we display two nodes for each channels, even if the node has already been registered with a previous channel. This PR makes sure we only draw one point for each node instead of many. The rendering should be almost identical from master, just slightly less bright.


wiz (approved):
Tested ACK @ v2.5.0-dev [[441a5fa2](https://github.com/mempool/mempool/commit/441a5fa2)]

Issue #2180, opened 2022-07-25
feature request: add `/block/:hash/raw` api route for node backend


oleonardolima:
### Description
Currently the `/block/:hash/raw` route is only available in production environment (mempool.space), the route is not available in [bitcoin.routes.ts](https://github.com/mempool/mempool/blob/master/backend/src/api/bitcoin/bitcoin.routes.ts) and only for the `/api/` prefix (esplora backend) as all the `/api` calls are redirected to esplora backend.

When using the mempool/backend with Bitcoin Core RPC, as the route is not available there is no endpoint to access the raw format of a block, it would be helpful to have such a feature.

### Problem to be solved
This solves the problem of doing different calls to get the header and then get all the transactions of a block when using mempool with Bitcoin Core RPC backend, as it reuses the existing RPC call.

### Proposed solution
Add the route to be available when using Bitcoin Core RPC as backend, and ignore it when using esplora or in production environment.

#### Additional info
As discussed in the matrix chat I've addressed this issue in a PR. #2181

wiz:
Fixed by #2181

PR #2181, opened 2022-07-25
feature: add /block/:hash/raw api route

oleonardolima:
### Description
As decribed in #2180 this PR adds the /block/:hash/raw API route, it's now available when using mempool backend, with Bitcoin Core RPC.

In the production environment and using `esplora` as backend, it's still being ignored and must be redirected by the nginx configuration.

It also updates the doc message indicating that the type must differ by each backend used:
- esplora: block in binary representation
- bitcoin core rpc: block in hex


docker/backend/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:16.16.0-buster-slim AS builder
+FROM node:17.9.1-buster-slim AS builder
```
softsimon, line 1: Is the change to the Dockerfile intended?

I don't see the relation to the PR so should be reverted.

softsimon (changes requested):
Our goal for the Mempool API is to be identical to the Esplora API (when possible) so it doesn't matter which backend you run https://github.com/Blockstream/esplora/blob/master/API.md

So you need to change it so our bitcoin core based api returns the data in binary.

oleonardolima:
> Our goal for the Mempool API is to be identical to the Esplora API (when possible) so it doesn't matter which backend you run https://github.com/Blockstream/esplora/blob/master/API.md
> 
> So you need to change it so our bitcoin core based api returns the data in binary.

Sure, I'll work on it

oleonardolima:
> Our goal for the Mempool API is to be identical to the Esplora API (when possible) so it doesn't matter which backend you run https://github.com/Blockstream/esplora/blob/master/API.md
> 
> So you need to change it so our bitcoin core based api returns the data in binary.

@

> > Our goal for the Mempool API is to be identical to the Esplora API (when possible) so it doesn't matter which backend you run https://github.com/Blockstream/esplora/blob/master/API.md
> > So you need to change it so our bitcoin core based api returns the data in binary.
> 
> Sure, I'll work on it

@softsimon Sorry for the delay, I've updated it now 😁

oleonardolima:
I've tested it locally with curl, postman, and with the following snippet:
``` rust
use std::ops::Deref;

use bitcoin::{consensus::deserialize, Block};

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    env_logger::init();

    let hash = "1a72d2a7d9c7f0894b8e83344ec5d73e0c58588a7bee95f8552feb39423bca9e";
    let url = format!("http://localhost:8999/api/v1/block/{}/raw", hash);

    let res = reqwest::get(url).await?;
    log::debug!("{:#?}", res);

    let res_bytes = res.bytes().await?;
    log::debug!("{:#?}", res_bytes);
    let block: Block = deserialize(res_bytes.deref()).unwrap();
    // log::debug!("{:#?}", block);
    print!("{:#?}", block);

    Ok(())
}

```

backend/src/api/bitcoin/bitcoin.routes.ts
```diff
@@ -470,6 +471,16 @@ class BitcoinRoutes {
     }
   }
 
+  private async getRawBlock(req: Request, res: Response) {
+    try {
+      const result = await bitcoinApi.$getRawBlock(req.params.hash);
+      res.setHeader('content-type', 'text/plain');
```
softsimon, line 477:
```suggestion
      res.setHeader('content-type', 'application/octet-stream');
```

softsimon (approved):
tested ACK @ v2.5.0-dev [[a9248a5f](https://github.com/mempool/mempool/commit/a9248a5f)]

wiz (approved):
Tested ACK @ d66cc8a213f9e4f78197133ebc7db95f13617b4f

Output matches esplora exactly :+1:

PR #2185, opened 2022-07-26
[Node page] Add link to node list per country/isp in node page

nymkappa:
You can now access the node list per country/isp directly from the node page.

image


wiz (approved):
Tested ACK @ v2.5.0-dev [[898fef19](https://github.com/mempool/mempool/commit/898fef19)]

PR #2187, opened 2022-07-26
[LN Dashboard] - Show 3y charts instead of 1y

nymkappa:
It just makes sense. The LN network does not evolve fast enough to get interesting charts on a 1y timeframe right now.

### 1y

Very boring

Screen Shot 2022-07-26 at 3 18 16 PM

### 3y

Now we talkin

Screen Shot 2022-07-26 at 3 18 05 PM


wiz (approved):
utACK

PR #2190, opened 2022-07-26
[LN stats] Order lightning_stats by added timestamp instead of id

nymkappa:
Fix a future bug where the `lightning_stats` table rows may not be returned in the correct order and break time based charts and other calculation.


wiz (approved):
Tested ACK @ v2.5.0-dev [[a49cb2d6](https://github.com/mempool/mempool/commit/a49cb2d6)]

PR #2192, opened 2022-07-26
Fix for mempool logo not being centered vertically

softsimon:


wiz (approved):
Partially Tested ACK @ v2.5.0-dev [[9bf04b0a](https://github.com/mempool/mempool/commit/9bf04b0a)]

Only tested with official logo but looks fine

PR #2194, opened 2022-07-26
Fix zmq ports in prod bitcoin.conf

wiz:
Currently testnet bitcoind fails to start because mainnet bitcoind is using the wrong port for zmq


Emzy:
utACK

PR #2201, opened 2022-07-27
Add capacity/nodes, include/exclude Tor from ISP chart

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2189
I think we should also apply this for the country pie chart.

https://user-images.githubusercontent.com/9780671/181238189-cad7fc93-46d2-44ff-a158-dbf010d8cd73.mov


wiz (approved):
Tested ACK @ v2.5.0-dev [[512739ae](https://github.com/mempool/mempool/commit/512739ae)]

Issue #2199, opened 2022-07-27
Make sure channels stats are only computed on active channels (status = 1)


nymkappa:
This is incorrect for example as we should exclude inactive channels https://github.com/mempool/mempool/blob/master/backend/src/api/explorer/nodes.api.ts#L12

PR #2202, opened 2022-07-27
Ony consider channel stats = 1 for stats calculation

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2199

When we run any kind of calculation using channels (capacity, channel count, and other things), we should only consider `active` channels (channel.status = 1).


wiz (approved):
Tested ACK @ v2.5.0-dev [[d6e9500b](https://github.com/mempool/mempool/commit/d6e9500b)]

PR #2203, opened 2022-07-27
Don't set all channels to inactive when the updater runs

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2197


backend/src/tasks/lightning/node-sync.service.ts
```diff
@@ -106,7 +108,22 @@ class NodeSyncService {
 
     try {
       // @ts-ignore
-      const [channels]: [ILightningApi.Channel[]] = await DB.query(`SELECT channels.id FROM channels WHERE channels.status = 1 AND ((SELECT COUNT(*) FROM nodes WHERE nodes.public_key = channels.node1_public_key) = 0 OR (SELECT COUNT(*) FROM nodes WHERE nodes.public_key = channels.node2_public_key) = 0)`);
+      const [channels]: [ILightningApi.Channel[]] = await DB.query(`
```
nymkappa, line 111: Note: This is just a formatting update. The query is still the same but we can at least read it now

wiz: I approve of this subtle trolling

wiz (approved):
Partially Tested ACK @ v2.5.0-dev [[48496abb](https://github.com/mempool/mempool/commit/48496abb)]

Not sure how to test this so I just truncated the lightning_stats table and it looks good

PR #2204, opened 2022-07-27
Silence LN db migration is CONFIG.LIGHTNING.ENABLED = false

nymkappa:
When we run the db migration on servers which does not have the lightning backend enabled, don't print out LN migration related messages


wiz (changes requested):
NACK - not sure why you want to silence all the DB migration logs, those are very important to know what's going on - you only need to remove the "Will re-generate historical data from scratch." part since that won't happen if LN is disabled 😅

wiz (approved):
Tested ACK @ v2.5.0-dev [[f610699e](https://github.com/mempool/mempool/commit/f610699e)]

PR #2205, opened 2022-07-27
Better block unfurls

mononaut:
**_builds on PR #2173_**

This PR improves the link unfurling approach, restyles the block preview images for better legibility, and sets link preview titles.

block-unfurl-700007

Open Graph/Twitter meta tags are now set by frontend code, which is server-side rendered for crawler bots by the unfurler daemon using Puppeteer.

Hopefully this makes the procedure for adding new previews easier:
  1. Add a data attribute ```data: { ogImage: true }``` to your routing entry.
  2. Create a matching routing entry at `/preview/<page>...` with a new "preview" version of the page component.
  3. Style that component to look good at 1024x512 resolution.
  
Routes that haven't opted in via the ```data: { ogImage: true }``` attribute keep their existing static preview image:

unfurl-default


wiz (approved):
Tested ACK @ v2.5.0-dev [[9c44bf17](https://github.com/mempool/mempool/commit/9c44bf17)]

Screen Shot 2022-07-28 at 3 47 25

PR #2207, opened 2022-07-28
Reduce mobile menus padding

nymkappa:


wiz (approved):
Tested ACK @ v2.5.0-dev [[c3780ada](https://github.com/mempool/mempool/commit/c3780ada)]

PR #2208, opened 2022-07-28
Update mouse UX on LN map in dashboard

nymkappa:
This PR removes the LN world map interactions from the LN dashboard.

The scroll is a bit broken/weird specially when using a mouse, but I'm gonna improve it later. At least now you can scroll the page properly without triggering the map zoom feature.

You can click on the map and it will open the graph map page.


wiz (approved):
Tested ACK @ v2.5.0-dev [[d7fe9276](https://github.com/mempool/mempool/commit/d7fe9276)]

I can still drag the world map around, but this does change the click to go to the full map view, which is kinda better than what we have now ...

PR #2209, opened 2022-07-28
Remove useless view more links that does not link to anywhere

nymkappa:
Until we have something to link to, I remove those two href.

Screen Shot 2022-07-28 at 7 51 43 AM


wiz (approved):
utACK

PR #2214, opened 2022-07-28
use @grpc/grpc-js directly instead of the wrapper lib

antonilol:
This pr replaces the npm package `lightning` with 2 of its dependencies `@grpc/grpc-js` and `@grpc/proto-loader`

interfaces have been rewritten to match LND's grpc output

and...
- i made a utility script that downloads `lightning.proto` from the lnd github repo and generates typescript types from it
that script can be given a specific lnd version or let it auto-detect it from `lnd --version`. usage: `npm run configure-lnd [version]` or `node configure-lnd [version]`
- adds `lightning` as label in package.json, thought it would fit there :)
- noticed admin.macaroon is used by default. this is not needed and only imposes extra security risks. i replaced it with the readonly.macaroon which still allows mempool to read the needed things from lnd

### DIFF
--- a/backend/configure-lnd.js
+++ b/backend/configure-lnd.js
@@ -1,0 +1,82 @@
+const { execSync } = require('child_process');
+const https = require('https');
+const fs = require('fs');
+
+const LND_PROTO_FILE_LOCATION = 'src/api/lightning/lnd/proto/lightning.proto';
+const LND_PROTO_TYPES_DIR = 'src/api/lightning/lnd/types/';
+var lndVersion = process.argv[2];
+
+function downloadProto(commit, old=false) {
+  return new Promise(resolve => {
+    var data = '';
+    const req = https.request(`https://raw.githubusercontent.com/lightningnetwork/lnd/${commit}/lnrpc/${old ? 'rpc' : 'lightning'}.proto`, async res => {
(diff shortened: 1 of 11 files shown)


softsimon:
Downloading the `lightning.proto`-file from github feels like a bit too much added complexity.

antonilol:
> Downloading the `lightning.proto`-file from github feels like a bit too much added complexity.

yeah that's true, i will remove it from the gitignore

the configure script is then only needed for development (compiling without it will generate a fallback type `any`) or if someone wants to be 100% sure for running it with another version of lnd (using a proto file from a different version should generally work but can give random unpredictable errors/bugs sometimes)

wiz:
Removing the lightning npm dep and connecting directly to lightning GRPC is a good idea, but this PR seems to want to do much more than that and feels quite messy - for example, downloading proto from github as part of the build process feels very wrong. Can you just make a new PR that only has the GRPC integration in it, and no other random things mixed in?

wiz:
Why use the GRPC API at all when lnd has a REST API we can use instead?

antonilol:
> Why use the GRPC API at all when lnd has a REST API we can use instead?

good point, made a new pr to use the rest api

PR #2221, opened 2022-07-29
use lnd rest api

antonilol:
#2214 but with rest api now, (lnd-api.ts looks way cleaner now lol)

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -16,7 +16,8 @@
     "mempool",
     "blockchain",
     "explorer",
-    "liquid"
+    "liquid",
+    "lightning"
   ],
   "main": "index.ts",
   "scripts": {
@@ -34,10 +35,8 @@
     "@types/node": "^16.11.41",
     "axios": "~0.27.2",
     "bitcoinjs-lib": "6.0.1",
-    "bolt07": "^1.8.1",
     "crypto-js": "^4.0.0",
     "express": "^4.18.0",
-    "lightning": "^5.16.3",
     "maxmind": "^4.3.6",
     "mysql2": "2.3.3",
     "node-worker-threads-pool": "^1.5.1",
--- a/backend/src/tasks/lightning/node-sync.service.ts
+++ b/backend/src/tasks/lightning/node-sync.service.ts
@@ -1,4 +1,3 @@
-import { chanNumber } from 'bolt07';
 import DB from '../../database';
 import logger from '../../logger';
 import channelsApi from '../../api/explorer/channels.api';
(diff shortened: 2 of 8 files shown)


softsimon (approved):
tested ACK @ v2.5.0-dev [[887fb13f](https://github.com/mempool/mempool/commit/887fb13f)]

backend/package.json
```diff
@@ -16,7 +16,8 @@
     "mempool",
     "blockchain",
     "explorer",
-    "liquid"
+    "liquid",
+    "lightning"
```
nymkappa, line 20: We don't need this `lightning` dep anymore right?

backend/src/tasks/lightning/node-sync.service.ts
```diff
@@ -418,8 +419,18 @@ class NodeSyncService {
...
+  private toShortId(id: string): string {
+    const n = BigInt(id);
+    return [
+      n >> 40n, // nth block
+      (n >> 16n) & 0xffffffn, // nth tx of the block
+      n & 0xffffn // nth output of the tx
+    ].join('x');
+  }
+
+  private utcDateToMysql(date?: number): string {
```
nymkappa, line 432: I still don't understand what's that function for. `lnd` returns timestamp that we can directly manipulate and insert into the db without ever the need to convert it to string. @softsimon 

We can use `FROM_UNIXTIME(timestamp)` to insert in db.

nymkappa (changes requested):

backend/package.json
antonilol: This one is an npm keyword/search word, the deps are a few more lines down

nymkappa: Oh my bad!

backend/src/tasks/lightning/node-sync.service.ts
nymkappa: Actually nevermind I'll clean this up when I make the clightning PR

nymkappa:
I got an issue running the PR, I'm not sure if it's related to the change or not, maybe the previous library was handling this because I never ran into it before?

```
Aug  1 10:28:48 [86969] ERR: <lightning> $scanForClosedChannels() error: connect ETIMEDOUT 127.0.0.1:8332
```

And now I have 2 channels close, and all others channels status is set to 0.

Screen Shot 2022-08-01 at 10 34 17 AM

nymkappa (approved):

wiz (approved):
Tested ACK @ v2.5.0-dev [[2a3111af](https://github.com/mempool/mempool/commit/2a3111af)]

Thanks for fixing this and removing the extra npm deps :+1:

PR #2222, opened 2022-07-30
Fix tor config for FreeBSD on prod installer

Emzy:


production/install
```diff
@@ -183,7 +183,7 @@ case $OS in
         ROOT_GROUP=wheel
         ROOT_HOME=/root
         TOR_HOME=/var/db/tor
-        TOR_CONFIGURATION=/usr/local/etc/tor/torrc
+        TOR_CONFIGURATION=/var/db/tor/torrc
```
wiz, line 186:
```suggestion
        TOR_CONFIGURATION=/usr/local/etc/tor/torrc
```

wiz (approved):

PR #2226, opened 2022-08-01
Fix missing pub key, capacity and channel count for node lists

nymkappa:
This PR fixes the following issues in the nodes for X isp and nodes for X country lists components:
* Missing nodes in the list because there is not associated node stats
* Missing or nullified fields (public key, capacity and channels) for the same reason


wiz (approved):
Tested ACK @ v2.5.0-dev [[3fd8af0c](https://github.com/mempool/mempool/commit/3fd8af0c)]

PR #2227, opened 2022-08-01
Re-enabled channels world map click event

nymkappa:
Right now you cannot click on nodes in /graphs/lightning/nodes-channels-map


wiz (approved):
Tested ACK @ v2.5.0-dev [[fc463a95](https://github.com/mempool/mempool/commit/fc463a95)]

Issue #2229, opened 2022-08-01
Show shorten public key if no alias exists


nymkappa:
<img width="1134" alt="Screen Shot 2022-08-01 at 1 30 39 PM" src="https://user-images.githubusercontent.com/9780671/182138934-c55a89f9-a50e-44b4-bcbf-cc12bdbce706.png">

PR #2233, opened 2022-08-01
Set default values when pubkey, capacity and channels are missing from top nodes

nymkappa:
The top nodes component may show empty data if node has an empty alias, display the 20 first char of the public key. Also send zeroes value instead of missing fields.

Fixes https://github.com/mempool/mempool/issues/2229

May be present in other components, will submit other PR when I see them.


wiz (approved):
Tested ACK @ v2.5.0-dev [[03dfca3b](https://github.com/mempool/mempool/commit/03dfca3b)]

PR #2234, opened 2022-08-01
Fix UX interaction with channels map

nymkappa:
On the dashboard, can only click to the map to open `/charts`.
On the graph page, can only zoom and pan.


wiz (approved):
Tested ACK @ v2.5.0-dev [[73c9ec20](https://github.com/mempool/mempool/commit/73c9ec20)]

Issue #2216, opened 2022-07-28
Integrate Core Lightning


nymkappa:

nymkappa:
Need https://github.com/mempool/mempool/pull/2221 to be merged in the first place before I can open a PR

PR #2231, opened 2022-08-01
Import LN historical statistics (network wide + per node)

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2124

This PR adds the historical LN stats import feature.

Database migration has been updated to add a new lightning_stats field named `clearnet_tor_nodes`, allowing to track properly LN nodes presence on different networks.

Screen Shot 2022-08-02 at 3 01 25 PM

Screen Shot 2022-08-02 at 3 03 21 PM

### Testing

1. Generate network topology files
2. Update your mempool nodejs backend config

You need to indicate where your topology output has been written. You can do so by setting a full path of the output folder at `MEMPOOL.LIGHTNING.TOPOLOGY_FOLDER`.

#### Run the importer

Simply run the nodejs backend and you should notice the following messages in the logs
```
Aug  1 17:49:37 [7949] INFO: <lightning> Importing historical lightning stats
Aug  1 17:49:40 [7949] DEBUG: <lightning> Processing /Users/nymkappa/Documents/topologies/topology_1632009600
Aug  1 17:49:40 [7949] DEBUG: <lightning> Sun, 19 Sep 2021 00:00:00 GMT (1632009600): Found 13198 nodes and 102464 channels
Aug  1 17:49:40 [7949] DEBUG: <lightning> Caching funding txs for Sun, 19 Sep 2021 00:00:00 GMT (1632009600)
Aug  1 17:49:40 [7949] DEBUG: <lightning> Imported 53447 funding tx amount from the disk cache
Aug  1 17:50:12 [7949] DEBUG: <lightning> Indexing channels funding tx 25272 of 102464 (24.66%) | elapsed: 32 seconds
```

This process is very long. I had not done yet a full import from scratch so I can't give any number yet. We start with the most recent `topology_XXXXXX` timestamp, and back in time. Feel free to suggest which log message you are interested in seeing on INFO level.

Frontend is not yet aware of the progress, but to limit the scope of that PR I'll submit a future PR to indicate the import progress.

PR #2240, opened 2022-08-02
Add clightning support to the lightning backend

nymkappa:
**Depends on https://github.com/mempool/mempool/pull/2231**

* Set `MEMPOOL.LIGHTNING.BACKEND: 'cln'`
* Add the following config entry in the backend config to indicate where is your lightningd socket stored. For example:
```
  "CLIGHTNING": {
    "SOCKET": "/Users/nymkappa/.lightning/bitcoin/lightning-rpc"
  },
```
Right now, the code assumes that your node is running for long enough already so that it has a good view of the graph, if that's not the case, the backend will simply inserts whatever is available, until the next run (right now it's every hour) or upon a nodejs backend restart.

While it should(tm) be okay to switch between lnd/cln without cleaning up the db, to properly test the PR, please truncate your `nodes` and `channels` table before running.


wiz (approved):
Tested ACK @ v2.5.0-dev [[215df5ef](https://github.com/mempool/mempool/commit/215df5ef)]

nymkappa:
Fixed https://github.com/mempool/mempool/issues/2216

PR #2247, opened 2022-08-03
Faster unfurls

mononaut:
This PR reworks the unfurler implementation to improve response times (now <250ms on my machine), and also tweaks the preview image layouts & size:

#### Preview images

The preview image rendering process now reuses puppeteer chrome pages instead of spinning up a new page for each request, preloads mempool.space, and avoids reloading the page between screenshots when possible.

Frontend preview pages now explicitly signal screenshot readiness, so the unfurler can take a screenshot as soon as possible instead of relying on heuristics.

#### Meta tags

Switches back to simple templating for serving html meta tags, eliminating all of the puppeteer-related overhead for these requests.

Unlike server-side rendering in puppeteer, this approach doesn't provide automatic localization of open graph link titles and descriptions, but it's much faster and preview images do remain localized:

| block-localization | address-localization |
|-|-|

### Config

This PR adds a few new optional variables to `unfurler/config.json`:

`MEMPOOL.NETWORK`: either "bitcoin", "liquid" or "bisq" - used to customize the open graph titles & descriptions and choose an appropriate fallback image.

`PUPPETEER.MAX_PAGE_AGE`: the maximum time in seconds that a puppeteer instance can reuse a mempool.space session before loading a fresh page.

`PUPPETEER.RENDER_TIMEOUT` maximum time in milliseconds to wait before abandoning an attempt to render an image preview.

### DIFF
--- a/frontend/src/app/components/address/address-preview.component.html
+++ b/frontend/src/app/components/address/address-preview.component.html
@@ -44,7 +44,7 @@
     <div class="w-100 d-block d-md-none"></div>
     <div class="col-md qrcode-col">
       <div class="qr-wrapper">
-        <app-qrcode [data]="address.address" [size]="370"></app-qrcode>
+        <app-qrcode [data]="address.address" [size]="448"></app-qrcode>
       </div>
     </div>
   </div>
--- a/frontend/src/app/components/address/address-preview.component.scss
+++ b/frontend/src/app/components/address/address-preview.component.scss
@@ -1,5 +1,5 @@
 h1 {
-  font-size: 42px;
+  font-size: 52px;
   margin: 0;
 }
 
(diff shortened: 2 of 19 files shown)


wiz (changes requested):
Tested NACK @ 54355b61d6614f4f39f34aaa8b16b2d04d144933

Great PR, works well but with one problem: if you request an invalid URL the unfurler gets into a broken state and stops rendering pages.

### How to reproduce:
1) GET /render/preview/block/600000
2) GET /render/preview/block/asdf
3) GET /render/preview/block/600001

### Actual Result
After step 1, proper rendering is displayed
After step 2, empty page appears
After step 3, empty page appears

wiz:
image

Also seems to display incorrect mismatched data when requesting many queries at the same time

mononaut:
turns out my puppeteer concurrency implementation was broken, so the unfurler workers were trying to concurrently share the _same_ chrome tab :sweat_smile:

just pushed fixes for both that and the invalid url bug, but I'd like to do some more testing before this gets merged.

wiz:
Works quite well, but I was "stress testing" by pasting many URLs at once into telegram and got this error:
```
failed to render page preview for screenshot due to client-side error. probably requested an invalid ID
failed to render page preview for screenshot due to client-side error. probably requested an invalid ID
failed to render page preview for screenshot due to client-side error. probably requested an invalid ID
```

wiz (approved):
Tested ACK @ e60bd52e8b43f53fcf4ba8c56f72c65d8b346f18

While it might not be 100% perfect it feels shippable, and any bugs we find later on can be fixed in future PRs

Issue #2232, opened 2022-08-01
Empty channel list missing label


nymkappa:
`No channels to display` should also be shown on the `Open` tab

nymkappa:
This was actually the pagination not being reset when you browse a different node

Issue #2230, opened 2022-08-01
Show shorten sats amount for small channels or capacity


nymkappa:
<img width="434" alt="Screen Shot 2022-08-01 at 1 32 00 PM" src="https://user-images.githubusercontent.com/9780671/182139123-39a6e2d0-6503-4413-895f-d7ec43fd0fe9.png">

Can be found in multiple places, also need to apply this idea for charts

Issue #2160, opened 2022-07-23
Missing fee rate on channel


knorrium:
![Screenshot from 2022-07-23 10-32-06](https://user-images.githubusercontent.com/100320/180616351-3484646a-ef57-478e-a8d2-8e51fd30a1d2.png)

from node 031371bfd9e935c86bf5fe814d8e6cbc8efe50a46d711f75837a32e35b66494153

antonilol:
sometimes lnd doesn't know the channel policies (yet), mempool will now give `null` values to the frontend for those fields

i think this happens when announcements are still being relayed and not yet received by my node, the 3 channels below all have a low amount of confirmations (< 12)

```
$ lncli describegraph | grep -C6 null

...
            "channel_id": "821754099926368257",
            "chan_point": "d46292df418d0411011e50dc107934c3e84d2d08c202a24efeb721cc9fc9e1c7:1",
            "last_update": 1659288239,
            "node1_pub": "02df315ed19636d66272e2ebbc1f413e460ac5a004810b28ef4714b30f33acf6b0",
            "node2_pub": "03c436af41160a355fc1ed230a64f6a64bcbd2ae50f12171d1318f9782602be601",
            "capacity": "500000",
            "node1_policy": null,
            "node2_policy": {
                "time_lock_delta": 144,
                "min_htlc": "1",
                "fee_base_msat": "2",
                "fee_rate_milli_msat": "2",
                "disabled": true,
--
            "channel_id": "821757398437134337",
            "chan_point": "87e8bad0525e8566192e9d8a619499def8b903f1623d6b278984cf19e7821480:1",
            "last_update": 1659288239,
            "node1_pub": "02df315ed19636d66272e2ebbc1f413e460ac5a004810b28ef4714b30f33acf6b0",
            "node2_pub": "03c436af41160a355fc1ed230a64f6a64bcbd2ae50f12171d1318f9782602be601",
            "capacity": "500000",
            "node1_policy": null,
            "node2_policy": {
                "time_lock_delta": 144,
                "min_htlc": "1",
                "fee_base_msat": "2",
                "fee_rate_milli_msat": "2",
                "disabled": false,
--
            "channel_id": "821757398445522944",
            "chan_point": "ac27c82b2fa98cc3f4ef5afbf14667ceddc8117e8827814c3c7672e1ede14af5:0",
            "last_update": 1659287965,
[...]

PR #2250, opened 2022-08-04
Fix node page and display real time data

nymkappa:
This PR fixes various issues on the node page:
* We don't use `node_stats` anymore, so the data display is what we actually have in db, not the indexed historical version
* Only show stats for active channels at the top
* Fix pagination
* Display zeroed value instead of empty fields
* When you click on a channel node public_key, the channel list is properly updated

Fixes https://github.com/mempool/mempool/issues/2232
Fixes https://github.com/mempool/mempool/issues/2230
Fixes https://github.com/mempool/mempool/issues/2160

### DIFF
--- a/backend/src/api/explorer/channels.api.ts
+++ b/backend/src/api/explorer/channels.api.ts
@@ -1,5 +1,6 @@
 import logger from '../../logger';
 import DB from '../../database';
+import nodesApi from './nodes.api';
 
 class ChannelsApi {
   public async $getAllChannels(): Promise<any[]> {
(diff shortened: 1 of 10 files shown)


nymkappa:
Fixes https://github.com/mempool/mempool/issues/2232
Fixes https://github.com/mempool/mempool/issues/2230
Fixes https://github.com/mempool/mempool/issues/2160

wiz (approved):
Tested ACK @ v2.5.0-dev [[f7cbe30a](https://github.com/mempool/mempool/commit/f7cbe30a)]

We should probably warm cache these new endpoints

Issue #2249, opened 2022-08-04
Properly convert clightning channel short_id to lnd channel id


nymkappa:
Currently when using cln, the `channels.id` field is incorrect. So if you come from or switch to lnd, you will have all channels duplicated.

PR #2252, opened 2022-08-04
Fix CLN channel short_id conversion

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2249

I directly imported source code from 2 repositories. There were originally backend 3rd party deps before https://github.com/mempool/mempool/pull/2221 got merged
* https://github.com/alexbosworth/bolt07/blob/master/ids/encode_chan_id.js

You should now be able to switch between `cln` and `lnd` backend smoothly without having to truncate the channels table.

**NOTE: You need to truncate your `channels` table because it may contains duplicated channels due to this bug, leading to inconsistent results between nodes per isp/country lists and node pages.**

### DIFF
--- a/backend/src/api/lightning/clightning/clightning-convert.ts
+++ b/backend/src/api/lightning/clightning/clightning-convert.ts
@@ -71,8 +71,11 @@ export async function convertAndmergeBidirectionalChannels(clChannels: any[]): P
 }
 
 export function convertChannelId(channelId): string {
-  const s = channelId.split('x').map(part => parseInt(part));
-  return BigInt((s[0] << 40) | (s[1] << 16) | s[2]).toString();
+  if (channelId.indexOf('/') !== -1) {
+    channelId = channelId.slice(0, -2);
+  }
+  const s = channelId.split('x').map(part => BigInt(part));
+  return ((s[0] << 40n) | (s[1] << 16n) | s[2]).toString();
 }
 
 /**
--- a/backend/src/tasks/lightning/network-sync.service.ts
+++ b/backend/src/tasks/lightning/network-sync.service.ts
@@ -430,10 +430,13 @@ class NetworkSyncService {
   }
 
   private toIntegerId(id: string): string {
-    if (config.LIGHTNING.BACKEND === 'lnd') {
+    if (config.LIGHTNING.BACKEND === 'cln') {
+      return convertChannelId(id);
+    }
+    else if (config.LIGHTNING.BACKEND === 'lnd') {
       return id;
     }
-    return convertChannelId(id);
+    return '';
   }
 
   /** Decodes a channel id returned by lnd as uint64 to a short channel id */


antonilol:
in #2221 i added a function that converts to uint64 format from lnd to the "short" (with the `x`s) format, to replace the `bolt07` package

https://github.com/mempool/mempool/blob/master/backend/src/tasks/lightning/network-sync.service.ts#L439-L451

one of the two can be removed to not have duplicated code

nymkappa:
> in #2221 i added a function that converts to uint64 format from lnd to the "short" (with the `x`s) format, to replace the `bolt07` package
> 
> https://github.com/mempool/mempool/blob/master/backend/src/tasks/lightning/network-sync.service.ts#L439-L451
> 
> one of the two can be removed to not have duplicated code

Thanks for pointing that out. I'll also import the https://github.com/alexbosworth/bolt07/blob/master/ids/decode_chan_id.js then and remove your implementation. Does that sounds okay?

antonilol:
> > in #2221 i added a function that converts to uint64 format from lnd to the "short" (with the `x`s) format, to replace the `bolt07` package
> > https://github.com/mempool/mempool/blob/master/backend/src/tasks/lightning/network-sync.service.ts#L439-L451
> > one of the two can be removed to not have duplicated code
> 
> Thanks for pointing that out. I'll also import the https://github.com/alexbosworth/bolt07/blob/master/ids/decode_chan_id.js then and remove your implementation. Does that sounds okay?

yes, i'm fine with that

nymkappa:
@antonilol I've imported the relevant bolt07 part. But that's a lot of new code considering you are doing the same thing with a few lines of code so I'm a bit confused.

Is there any reason we should not use your code instead? Is there some cases were it may break?

antonilol:
> @antonilol I've imported the relevant bolt07 part. But that's a lot of new code considering you are doing the same thing with a few lines of code so I'm a bit confused.
> 
> Is there any reason we should not use your code instead? Is there some cases were it may break?

The only reason for it to not work would be outdated nodejs. It needs node v10+ (released 2018) because i used bigints there.

nymkappa:
> > @antonilol I've imported the relevant bolt07 part. But that's a lot of new code considering you are doing the same thing with a few lines of code so I'm a bit confused.
> > Is there any reason we should not use your code instead? Is there some cases were it may break?
> 
> The only reason for it to not work would be an ancient version of nodejs. It needs node v10+ (released 2018) because i used bigints there.

@knorrium @wiz Do we officially support nodejs version less than 10 ? If not, I'll use @antonilol 's implementation

wiz:
No we do not

backend/src/api/lightning/clightning/clightning-convert.ts
```diff
@@ -71,8 +72,11 @@ export async function convertAndmergeBidirectionalChannels(clChannels: any[]): P
 }
 
 export function convertChannelId(channelId): string {
+  if (channelId.indexOf('/') !== -1) {
+    channelId = channelId.slice(0, -2);
+  }
   const s = channelId.split('x').map(part => parseInt(part));
-  return BigInt((s[0] << 40) | (s[1] << 16) | s[2]).toString();
+  return bolt07_encode_chan_id({block_height: s[0], block_index: s[1], output_index: s[2]});
```
antonilol, line 79: the value inside `BigInt()`, `(s[0] << 40) | (s[1] << 16) | s[2]` is not a bigint, it is a javascript number, which only has 52 bits. precision is lost here and the channel id will have the last few digits replaced by zeros

this can be fixed by converting each part of the id to bigint first:
```ts
const s = channelId.split('x').map(part => BigInt(part));
return ((s[0] << 40n) | (s[1] << 16n) | s[2]).toString();
```

nymkappa: The code you are commenting has been deleted. Do you mean we can use the old code with this simple fix?

antonilol: yes

nymkappa:
Thanks to @antonilol this PR is now back into the mini PR territory and ready for testing @wiz

wiz (approved):
Tested ACK @ v2.5.0-dev [[b725fc8d](https://github.com/mempool/mempool/commit/b725fc8d)]

PR #2254, opened 2022-08-04
Always show channels map in node page - auto zoom on the node

nymkappa:
This PR makes the channels map one of the main component on the node page. When you open a node page, the page will auto focus on the node area as shown in the demo.

The UI of the node page itself has not been polished yet and it is not the purpose of this PR.

https://user-images.githubusercontent.com/9780671/182871143-0dcef361-6b08-40a8-949b-381e80829ca8.mov

### DIFF
--- a/frontend/src/app/lightning/node/node.component.html
+++ b/frontend/src/app/lightning/node/node.component.html
@@ -102,9 +102,7 @@
 
   </div>
 
-  <br>
-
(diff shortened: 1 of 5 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [5d81a13a]

Looks good but please move the map to the top to be same layout as dashboard

PR #2244, opened 2022-08-03
Fix daily LN stats crash

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2243

This PR also wraps the LN backend into a try/catch/retry logic in case there is an unhanded crash in there, to avoid crashing the whole nodejs backend at once.

### Testing

* Run the PR and see if a new `lightning_stats` entry and one new entry per node in `nodes_stats` are properly inserted in the db around midnight without crashing, using both lnd and cln

PR #2255, opened 2022-08-04
Update latest stats every 10 minutes

nymkappa:
**Depends on https://github.com/mempool/mempool/pull/2244**

Refresh the stats (network + nodes) every 10 minutes (by querying the network graph) and update the current day entries. When a new day starts, we create a new entry and update it for the next 24h, etc...

This will allow to display more up to date stats on the dashboard and in node lists while keeping the table size not to massive.


wiz (approved):
Tested ACK @ v2.5.0-dev [[fad73cf3](https://github.com/mempool/mempool/commit/fad73cf3)]

PR #2264, opened 2022-08-06
Right align mempool logo on mobile with dual logos

softsimon:
Preview:

Screen Shot 2022-08-06 at 04 27 34


wiz (approved):
Tested ACK @ v2.5.0-dev [[5612a033](https://github.com/mempool/mempool/commit/5612a033)]

I think it should be slightly more to the right so the edge of the mempool logo aligns with the search box on mobile

Issue #2228, opened 2022-08-01
LN node sync service does not handle backend errors properly during closed channel scan


nymkappa:
I got an issue running the PR, I don't think it's related to the PR but will need to fix.

```
Aug  1 10:28:48 [86969] ERR: <lightning> $scanForClosedChannels() error: connect ETIMEDOUT 127.0.0.1:8332
```

And now I have 2 channels close, and all others channels status is set to 0.

<img width="186" alt="Screen Shot 2022-08-01 at 10 34 17 AM" src="https://user-images.githubusercontent.com/9780671/182107982-a4ad239b-6d11-4b3c-9357-4892e10e26fb.png">

I'll ACK @ 2a3111af6da2c8edf5c6a1849899d9ccc0dc2473

_Originally posted by @nymkappa in https://github.com/mempool/mempool/pull/2221#issuecomment-1200893163_

PR #2265, opened 2022-08-08
Convert short_id to integer id with clightning backend before returning the graph

nymkappa:
We currently convert channel id (between short_channel_id and integer id) at the db insertion step and this is an issue for marking channels as inactive.

Here, when running with LND we receive an array of integer channel id, and therefore it marks all channels as inactive
https://github.com/mempool/mempool/pull/2265/files#diff-afe74c642a8770eaa6f6b83c0e207e308c2288b0b3c37a1b310d86bbf12631b2L392

With this PR, we make sure to convert the CLN short_channel_id to an integer id before returning the graph, and we use integer id across the board ([this is how the interface is defined](https://github.com/mempool/mempool/blob/master/backend/src/api/lightning/lightning-api.interface.ts#L22)) 

Probably fixes https://github.com/mempool/mempool/issues/2228, this bug was most likely the original reason and the timeout has nothing to do with the issue.

### Testing
Run the PR once with `lnd` and once with `cln` and each time make sure channels are not all marked as inactive, for example:

```mysql
# Active channels count
SELECT count(*)
FROM `channels`
WHERE status = 1;

# Inactive channels count
SELECT count(*)
FROM `channels`
WHERE status = 0;
```


wiz (changes requested):
Tested NACK @ c85b6d53c5ba88f261555caf64122833e08fb2cd

```
root@localhost [mempool_mainnet_lightning]> SELECT count(*)
    -> FROM `channels`
    -> WHERE status = 0;
+----------+
| count(*) |
+----------+
|    58115 |
+----------+
```

wiz (approved):
Tested ACK @ v2.5.0-dev [[c85b6d53](https://github.com/mempool/mempool/commit/c85b6d53)]

Had to delete the old bogus channels
```
root@localhost [mempool_mainnet_lightning]> delete from channels where id < 500000000000000000;
Query OK, 58706 rows affected (8.511 sec)
```

PR #2266, opened 2022-08-08
Refactor the LN backend and add more logs

nymkappa:
This PR cleans up a bit the lightning backend, mostly the node and channel updater code.
I've also added logs so we get a progress update every 10 seconds approximately.

We run the node/channels update every `config.MEMPOOL.GRAPH_REFRESH_INTERVAL` now (default to 5 minutes).

Creates and fixes https://github.com/mempool/mempool/issues/2256

PR #2267, opened 2022-08-08
Add `nodes.status` field

nymkappa:
Depends on https://github.com/mempool/mempool/pull/2266

This PR adds a new field in the `nodes` table to monitor the status of a node. Similar to what we do with channels, if a node exists in our db but does not exist in the network graph (`describegraph` or `listnodes`), we mark it as inactive.

Consequently, channels which have no active nodes will also be marked as inactive.

We will indicate this information in the frontend later.

### Testing

Simply by running this PR, you should notice the following message: `Marked XXX nodes as inactive because they are not in the graph`.

Also you can run the following query:
```mysql
# Active channels count
SELECT count(*)
FROM `nodes`
WHERE status = 1;

# Inactive channels count
SELECT count(*)
FROM `nodes`
WHERE status = 0;
```


wiz (approved):
Tested ACK @ v2.5.0-dev [[88e6305b](https://github.com/mempool/mempool/commit/88e6305b)]

Needs documentation of new backend configuration variables in a separate PR later. Also:

> We run the node/channels update every config.MEMPOOL.GRAPH_REFRESH_INTERVAL now (default to 5 minutes).

LIES

PR #2270, opened 2022-08-10
Rewrite channels map component using native echart

nymkappa:
I rewrote the channels world map component without using echarts-gl extension. Here are the changes:

* I've added some "GPS" noise for each nodes, so that nodes with same GPS coordinates are not all on top of each other
* The UX is much better, you have a "google map" like experience with zoom and pan
* Tooltips supports html/css, and we can also add tooltip on channels
* There is no weird 3d glitch and z fightning
* Performances are worst than using the native gl extension, but I believe it can be optimized (we can maybe draw only the nodes and channels seen in the zoomed area for example)
* We can use svg renderer if there is no opengl support (with a performance hit of course), which would also allow for downloading the map with the download button (not implemented).
* We can zoom in a node (in the node page) simply by setting the view center to the node GPS coordinates, it's currently buggy if you see nodes in Australia for example, with this PR it's working great.
* It looks a bit less sexy because there is not post processing (bloom filter). I wonder if we can add some custom rendering pipeline on top of the canvas tho.

#### Dashboard
Screen Shot 2022-08-10 at 11 29 08 AM

#### Map component
Screen Shot 2022-08-10 at 11 43 46 AM

#### Zoomed in Europe
Screen Shot 2022-08-10 at 11 43 59 AM

#### CSS on tooltips
Screen Shot 2022-08-10 at 11 44 41 AM

#### Paris - We show all nodes with some GPS noise
Screen Shot 2022-08-10 at 11 44 15 AM

#### Node page auto zoom perfectly
Screen Shot 2022-08-10 at 11 44 54 AM
Screen Shot 2022-08-10 at 11 47 51 AM

### DIFF
--- a/frontend/src/app/lightning/nodes-channels-map/nodes-channels-map.component.ts
+++ b/frontend/src/app/lightning/nodes-channels-map/nodes-channels-map.component.ts
@@ -20,7 +20,11 @@ export class NodesChannelsMap implements OnInit, OnDestroy {
   @Input() publicKey: string | undefined;
 
   observable$: Observable<any>;
-  center: number[] | undefined = undefined;
+  
+  center: number[] | undefined;
+  zoom: number | undefined;
+  channelWidth = 0.6;
+  channelOpacity = 0.1;
 
   chartInstance = undefined;
   chartOptions: EChartsOption = {};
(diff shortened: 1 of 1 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[d6a42cdf](https://github.com/mempool/mempool/commit/d6a42cdf)]

Fuck yeah 🏆

PR #2274, opened 2022-08-10
Disable country highlight - Update node styling

nymkappa:
Make sure no country is highlighted in the channel map (I could not reproduce)
Update node styling

#### Fixes(tm)
image

#### Show nodes in the front instead of the back

Master

Screen Shot 2022-08-10 at 4 43 53 PM

PR

Screen Shot 2022-08-10 at 4 43 58 PM


softsimon (approved):
tested ACK

PR #2275, opened 2022-08-10
Fix channel rendering issue

nymkappa:
When I ran with CLN I see more than 90,000 channels, and the rendering breaks.
This PR aims to fix that.

Screen Shot 2022-08-10 at 5 02 00 PM


softsimon (approved):
tested ACK

wiz (approved):
Tested ACK @ v2.5.0-dev [[73285ae7](https://github.com/mempool/mempool/commit/73285ae7)]

PR #2277, opened 2022-08-10
Add ISP chart in the dashboard - Fix mobile layout - Start polishing

nymkappa:
This PR is the first of the "polishing" series that will follow. It aims to set a decent starting point by matching the mining dashboard idea.

* Added ISP pie chart in dashboard
* Fix mobile layout overflowing in top nodes by capacity and channels
* Make the 3 first rows of the dashboard matches the mining dashboard dimension (1. map, 2. stats and 3. charts) 

Note: The ISP query is somehow buggy and will need to be fixed, but it is not the aim of this PR

Screen Shot 2022-08-10 at 6 49 31 PM

### DIFF
--- a/backend/src/api/explorer/nodes.api.ts
+++ b/backend/src/api/explorer/nodes.api.ts
@@ -168,7 +168,7 @@ class NodesApi {
     }
   }
 
-  public async $getNodesISP(groupBy: string, showTor: boolean) {
+  public async $getNodesISPRanking(groupBy: string, showTor: boolean) {
     try {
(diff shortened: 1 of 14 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[1d71e26a](https://github.com/mempool/mempool/commit/1d71e26a)]

Things to fix:
* Mobile view displays 3 stats above pie chart, should only show 2 to fit
* Change 3 stats strings "tagged" to "clearnet", i.e. "Clearnet ISPs"

PR #2280, opened 2022-08-10
Remove doc scroll listener after navigating away

hunicus:
Listener was activated but never removed, so it remained active even after navigating to other pages where it wasn't needed.

### DIFF
--- a/frontend/src/app/docs/api-docs/api-docs.component.ts
+++ b/frontend/src/app/docs/api-docs/api-docs.component.ts
@@ -50,9 +50,7 @@ export class ApiDocsComponent implements OnInit, AfterViewInit {
           document.getElementById( this.route.snapshot.fragment ).scrollIntoView();
         }
       }
-      window.addEventListener('scroll', function() {
-        that.desktopDocsNavPosition = ( window.pageYOffset > 182 ) ? "fixed" : "relative";
-      }, { passive: true} );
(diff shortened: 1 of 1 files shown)


softsimon (approved):
I can't see what problem this caused but memory leak fixes are always welcome

tested ACK

v2.5.0-dev [[e0c61f72](https://github.com/mempool/mempool/commit/e0c61f72)]

hunicus:
Just for performance. On its own, not terrible, but don't want them to pile up.

frontend/src/app/docs/api-docs/api-docs.component.ts
```diff
@@ -87,6 +85,14 @@ export class ApiDocsComponent implements OnInit, AfterViewInit {
     });
   }
 
+  ngOnDestroy(): void {
```
Arooba-git, line 88: Hm, we didn't import or implement `OnDestroy` from `angular/core` 🤔 , can we use `ngOnDestroy` without doing that? 🤔

softsimon: We can't, but this changed after merge and is not present in master.

Arooba-git: @softsimon Ah, thanks for the clarification.. didn't use angular for years so just wanted to confirm 👍  :)

PR #2281, opened 2022-08-11
Fix the OOM issue when building the frontend docker image for armv7

knorrium:
Build Docker using tmpfs to work around an issue that was causing the frontend build to crash when building armv7 images.

Update Ubuntu to `ubuntu-latest`

Ran two builds on my fork and had no crashes but the build time increased by about 15-20 minutes #lowtimepreference


knorrium:
@wiz TTP removed and new build worked fine @ v2.5.0-dev [[b082d819](https://github.com/mempool/mempool/commit/b082d819)]

wiz (approved):
utACK

Issue #2276, opened 2022-08-10
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory


nymkappa:
I probably have a leak somewhere in my new LN backend code so I need to investigate, easy to reproduce locally, I just need to let the backend run for a while.

PR #2282, opened 2022-08-11
Fix recursive call in LN network updater

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2276
[This](https://github.com/mempool/mempool/blob/e48eab403ca06f178f8b4ff6c2d620ad0a661e51/backend/src/tasks/lightning/network-sync.service.ts#L111) is obviously not supposed to be here so I must have added it by mistake


wiz (approved):
Tested ACK @ v2.5.0-dev [[4f1ebc25](https://github.com/mempool/mempool/commit/4f1ebc25)]

Usually crashes within 1 hour and didn't crash after 2 hours so it seems working

PR #2285, opened 2022-08-12
Fixing broken Matomo och bisq and liquid

softsimon:


wiz (approved):
Tested ACK @ v2.5.0-dev [[aa51f567](https://github.com/mempool/mempool/commit/aa51f567)]

PR #2286, opened 2022-08-12
Fix LN stats node per network count

nymkappa:
Currently when you run LND, all nodes will be detected as `unannounced` node, this PR fixes that.


wiz (approved):
utACK since I don't have lnd running

PR #2299, opened 2022-08-14
Fix wrong size of translators avatars

softsimon:


wiz (approved):
Tested ACK @ v2.5.0-dev [[23800801](https://github.com/mempool/mempool/commit/23800801)]

PR #2300, opened 2022-08-14
Fix difficulty adjustment

slaninas:
This fixes remaining time and estimated retarged date.
Just moving `now * 1000` from line 50 to 51 where it should be.

I'm not going to create the contribution file nor am I going to sign the commits because I don't use it atm.
If you want this fix, take it. If you don't, just close it.


junderw (approved):
ACK

@wiz I added OP as a co-author and signed the commit. Is that no good?

Do I need to remove them as an author because they didn't agree to the contributors agreement?

junderw:
I'll include my agreement in this PR.

PR #2301, opened 2022-08-15
Add basic transaction link previews

mononaut:
Adds a basic transaction link preview, as a placeholder while we come up with something more aesthetically pleasing.

tx-unfurl
liquid-tx-unfurl


wiz (approved):
Tested ACK @ v2.5.0-dev [[e0ea47b8](https://github.com/mempool/mempool/commit/e0ea47b8)]
Screen Shot 2022-08-16 at 13 47 31

Issue #2245, opened 2022-08-03
Remove XML parser deps from backend


nymkappa:
If the topology script is updated with the json export, we can get rid of the XML parser.

PR #2261, opened 2022-08-05
Only parse .topology file and assume they already are in json format

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2257
Untested yet.

PR #2306, opened 2022-08-16
Refactor LN stats import

nymkappa:
Redo of https://github.com/mempool/mempool/pull/2261

This will avoid the spam on topology file opening everytime we restart as well. Trying to keep thing simple and smooth for now.
Before we deploy to prod I'll add a small piece of code that run after the import which cleans up the corrupted data (hardcoded query).

Fixes https://github.com/mempool/mempool/issues/2290
Fixes https://github.com/mempool/mempool/issues/2245

If we need to regenerate json files for the future historical import with updated backups, we'll use my patched topology script to do so and I will update the code to match the exported json format.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -16,7 +16,6 @@
         "bitcoinjs-lib": "6.0.2",
         "crypto-js": "^4.0.0",
         "express": "^4.18.0",
-        "fast-xml-parser": "^4.0.9",
         "maxmind": "^4.3.6",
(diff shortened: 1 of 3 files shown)


wiz (approved):
Partially Tested ACK @ v2.5.0-dev [[b1e8d0aa](https://github.com/mempool/mempool/commit/b1e8d0aa)]

Not sure exactly how to test this but seems fine

Issue #2309, opened 2022-08-16
Node page channels pagination seems broken


nymkappa:
Cannot jump to last page

PR #2314, opened 2022-08-17
Fix node channel list pagination

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2309
Fixes https://github.com/mempool/mempool/issues/2310

There is a duplicated http request when clicking on a node in the channel list, but I don't want to waste time again with rxjs as we have more importing things to fix, I'll take a look another time.


wiz (approved):
Tested ACK @ v2.5.0-dev [[ea931da3](https://github.com/mempool/mempool/commit/ea931da3)]

PR #2319, opened 2022-08-18
Import json topology

nymkappa:
Note that you need export your topology as JSON format in the first place.
We know read all files which name contains `topology_` so you don't need to change the extension, since we export as `topology_XXXXX` already.


wiz (approved):
Tested ACK @ v2.5.0-dev [[c9a8b91c](https://github.com/mempool/mempool/commit/c9a8b91c)]

Issue #2317, opened 2022-08-18
ISP pie chart wrong %


nymkappa:
Wrong % in tooltip

![image](https://user-images.githubusercontent.com/9780671/185301324-af4fa0a1-dda5-423a-a4ed-628c25b99aca.png)

PR #2320, opened 2022-08-18
Fix wrong tooltip % in isp pie chart

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2317


Carsonmarvel1 (commented):
Return

wiz (approved):
Tested ACK @ v2.5.0-dev [[fe139651](https://github.com/mempool/mempool/commit/fe139651)]

PR #2323, opened 2022-08-18
Increase node size in channel map on node/channel pages

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2293


wiz (approved):
Tested ACK @ v2.5.0-dev [[998dcc6f](https://github.com/mempool/mempool/commit/998dcc6f)]

I was expecting the node size to scale along with its liquidity / routing capacity, but I guess this is okay for now :P

PR #2325, opened 2022-08-18
Fix missing isp ids on isp pie chart

nymkappa:
I broke the ISP chart so when you click on an ISP, you only get one ISP id. For example when you click on `Amazon.com` you land on `/lightning/nodes/isp/16509` instead of `/lightning/nodes/isp/16509,14618`.


wiz (approved):
Tested ACK @ v2.5.0-dev [[9fd4adec](https://github.com/mempool/mempool/commit/9fd4adec)]

PR #2327, opened 2022-08-18
Hide features before feature activated

softsimon:
Now hiding feature flags Taproot, Segwit and RBF on transactions before the feature was activated.

Taproot visible from activation block 709632
Segwit badges visible after activation block 477120
RBF was introduced in Bitcoin Core 0.12 February 23, 2016 so I picked the arbitrary block 399700.

Example:

Before taproot
Screen Shot 2022-08-18 at 22 46 57

Before segwit:
Screen Shot 2022-08-18 at 22 47 11

Before RBF (Downside is empty feature row)

Screen Shot 2022-08-18 at 22 46 36


wiz (approved):
Tested ACK @ v2.5.0-dev [[60fb61f7](https://github.com/mempool/mempool/commit/60fb61f7)]

PR #2329, opened 2022-08-19
Reverted wrong fix in 2300

nymkappa:
Reverted bogus PR https://github.com/mempool/mempool/pull/2300

image

### DIFF
--- a/backend/src/api/difficulty-adjustment.ts
+++ b/backend/src/api/difficulty-adjustment.ts
@@ -47,8 +47,8 @@ class DifficultyAdjustmentApi {
     }
 
     const timeAvg = timeAvgMins * 60 * 1000 ;
-    const remainingTime = remainingBlocks * timeAvg;
-    const estimatedRetargetDate = remainingTime + now * 1000;
+    const remainingTime = (remainingBlocks * timeAvg) + (now * 1000);
+    const estimatedRetargetDate = remainingTime + now;
 
     return {
       progressPercent,


junderw:
Sorry about that. I wasn't aware that the frontend would be affected, I should have said `utACK` instead of `ACK`. I'll be more careful in the future.

It seems as though the backend is returning incorrect info, and the frontend is compensating for the incorrect info to produce the correct info to display.

I should have added a frontend change to coincide with it.

junderw (approved):
utACK. This reverts the only change the PR made, so it should just revert the regression.

wiz (approved):
Tested ACK @ v2.5.0-dev [[298edb64](https://github.com/mempool/mempool/commit/298edb64)]
Screen Shot 2022-08-19 at 17 46 10

slaninas:
> Sorry about that. I wasn't aware that the frontend would be affected, I should have said `utACK` instead of `ACK`. I'll be more careful in the future.
> 
> It seems as though the backend is returning incorrect info, and the frontend is compensating for the incorrect info to produce the correct info to display.
> 
> I should have added a frontend change to coincide with it.

I wasn't aware either this would break something else.

The reason why I created #2300  was that the `/api/v1/difficulty-adjustment` is returning nonsensical values like `"remainingTime": 1661870014570.4146`. With this revert the API is broken again. I wonder why was #2300 reverted instead of properly fixing the frontend.

I was actually put off by your contribution requirements so I'm not going to dig into this to fix it properly.

Anyway, thanks for this project, it's nice otherwise.

PR #2331, opened 2022-08-19
 Use timestamp instead of date in stats tables 

nymkappa:
Depends on https://github.com/mempool/mempool/pull/2319

**DB schema v38**

Currently the `lightning_stats` and `node_stats` tables are using `date` type for `added` field. This creates timezone crap to manage in the code therefore this PR changes the fields to `timestamp` type.

This also fixes the fact that every restart all topology files are processed regardless if they were already imported (it would only works if you're in UTC timezone lol).

### Testing

* Run the PR, your stats tabe will be truncated
* Wait for the stats import to import ~20 files
* Restart the nodejs backend and confirm the import resume where it left and not not re-process all files


wiz (approved):
Tested ACK @ v2.5.0-dev [[e46636f5](https://github.com/mempool/mempool/commit/e46636f5)]

PR #2333, opened 2022-08-19
Fix: Difficulty API (REST) with frontend fixes

junderw:
1. The area with the bug only affected the backend REST API (which is documented, but rarely used outside the frontend, according to @wiz) and the Difficulty page that utilized the REST API. (not websocket)
2. Websocket mining info uses different data, so unaffected.
3. The frontend Difficulty page was using `remainingTime` as if it meant "absolute date in ms of next retarget". (This was the bug)
4. Upon fixing the bug, `estimatedRetargetDate` is the data that the frontend wants to render (since it is an actual absolute Date and not a duration of time between now and the next retarget)

---

I have refactored the difficulty logic to make it a little bit more readable (though not that much).

I have a few questions I will add as comments to the code on this PR.

Related: #2300 and #2329 

TODO: Add unit tests to make the refactor a bit more understandable.


backend/src/api/difficulty-adjustment.ts
```diff
@@ -6,33 +6,37 @@ class DifficultyAdjustmentApi {
...
     const DATime = blocks.getLastDifficultyAdjustmentTime();
     const previousRetarget = blocks.getPreviousDifficultyRetarget();
     const blockHeight = blocks.getCurrentBlockHeight();
     const blocksCache = blocks.getBlocks();
     const latestBlock = blocksCache[blocksCache.length - 1];
 
-    const now = new Date().getTime() / 1000;
-    const diff = now - DATime;
-    const blocksInEpoch = blockHeight % 2016;
-    const progressPercent = (blocksInEpoch >= 0) ? blocksInEpoch / 2016 * 100 : 100;
```
junderw, line 18: Q: All of these blocksInEpoch >= 0 checks are assuming blockHeight is possibly negative. If blockHeight is negative, nextRetargetHeight and remainingBlocks will be messed up. Are these values ok for such a messed up state?

backend/src/api/difficulty-adjustment.ts
```diff
@@ -41,14 +45,14 @@ class DifficultyAdjustmentApi {
       if (timeAvgMins > 20) {
         timeAvgMins = 20;
       }
-      if (now - latestBlock.timestamp + timeAvgMins * 60 > 1200) {
-        timeOffset = -Math.min(now - latestBlock.timestamp, 1200) * 1000;
```
junderw, line 45: Q: What is this for? It is set to the milliseconds value of time since last block timestamp with a hard cap at 20 minutes (in ms) multiplied by -1 And it is returned as-is... What is its usecase? If not testnet, it just is ignored and stays 0.

junderw: I added a test case that barely triggers this, and it makes the data's meaning a bit more clear... however, I am unsure what the use case is, still.

backend/src/api/difficulty-adjustment.ts
```diff
@@ -2,6 +2,83 @@ import config from '../config';
...
+  DATime: number,
+  nowSeconds: number,
+  blockHeight: number,
+  previousRetarget: number,
+  network: string,
+  latestBlockTimestamp: number,
+): DifficultyAdjustment {
+  const ESTIMATE_LAG_BLOCKS = 146; // For first 7.2% of epoch, don't estimate.
+  const EPOCH_BLOCK_LENGTH = 2016; // Bitcoin mainnet
+  const BLOCK_SECONDS_TARGET = 600; // Bitcoin mainnet
```
junderw, line 27: Q: These magic numbers were in the original logic as-is, even though there is also logic that splits based on the config network.

Should we change them based on the network?

backend/src/api/difficulty-adjustment.ts
softsimon: According to @antonilol who made this and wrote the code comment above, this code is made for _testnet_ only since this code is wrapped in this.

So as far as I know it works and shouldn't be deleted

```
/ Testnet difficulty is set to 1 after 20 minutes of no blocks,
// therefore the time between blocks will always be below 20 minutes (1200s).
if (config.MEMPOOL.NETWORK === 'testnet') {
```

antonilol: a negative timeOffset lowers the estimated mine time of a mempool block (the `In ~n minutes`) to ensure the estimation never shows a gap between blocks of more than 20 minutes. on all other networks this value is zero so zero is added to a number = no effect

it is used here: https://github.com/mempool/mempool/blob/master/frontend/src/app/components/mempool-blocks/mempool-blocks.component.html#L29

this rule in the bitcoin source code: https://github.com/bitcoin/bitcoin/blob/master/src/pow.cpp#L23-L27

backend/src/api/difficulty-adjustment.ts
junderw: Thinking about it some more... this is not something that changes often, and we're not adding networks left and right, so it should be fine just using constants like this.

softsimon (approved):
tested ACK @ v2.5.0-dev [[3a4982c5](https://github.com/mempool/mempool/commit/3a4982c5)]

backend/src/api/difficulty-adjustment.ts
junderw: No opinions?

Closing for now, as it shouldn't be a blocker.

wiz (approved):
Tested ACK @ v2.5.0-dev [[738d1f80](https://github.com/mempool/mempool/commit/738d1f80)]

However it seems "previous" is now broken showing 0.00% which is apparently unrelated to this PR??
Screen Shot 2022-08-27 at 9 46 36

Issue #2307, opened 2022-08-16
If channel is already marked as closed in DB, don't update it anymore


nymkappa:
Avoid resetting its status to 1 just because it's still in the network graph, this makes backend restart slow and while we set back all channels to closed status, we will serve outdated information.

PR #2336, opened 2022-08-19
If a channel is closed, stop updating it

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2307

Currently, even if a channel is closed (funding tx has been spent), the channel may still exist in the network graph. Therefore its status will be reset to 1 'active' upon the network lightning graph sync.

This PR makes sure that when we tag a channel as closed, we stop updating it.


wiz (approved):
Partially Tested ACK @ v2.5.0-dev [[7c25fd61](https://github.com/mempool/mempool/commit/7c25fd61)]

Don't have a good TC for this so just going to merge it for now and we'll continue to test it on staging

PR #2337, opened 2022-08-19
Updated mempool debug log

softsimon:
Optimized the Mempool loop log output.

Before:

```
Aug 19 17:56:22 [6183] DEBUG: Updating mempool
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 1
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 2
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 3
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 4
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 5
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 6
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 7
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 8
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 9
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 10
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 11
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 12
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 13
Aug 19 17:56:23 [6183] DEBUG: Fetched transaction 14
Aug 19 17:56:23 [6183] DEBUG: Mempool blocks calculated in 0.007 seconds
Aug 19 17:56:23 [6183] DEBUG: New mempool size: 8198 Change: -251
Aug 19 17:56:23 [6183] DEBUG: Mempool updated in 0.129 seconds
```

After:

```
Aug 19 19:35:44 [7261] DEBUG: Updating mempool...
Aug 19 19:35:44 [7261] DEBUG: Mempool blocks calculated in 0.004 seconds
Aug 19 19:35:44 [7261] DEBUG: Mempool updated in 0.121 seconds. New size: 2740 (+13)
```


wiz:
Thanks for the PR but maybe the correct fix is to disable the mempool loop in the Lightning backend?

softsimon:
> Thanks for the PR but maybe the correct fix is to disable the mempool loop in the Lightning backend?

As long as the mempool loop is working it should debug some output.
If you want to disable the mempool updating completely (not just silent the log) it should be a separate PR

wiz:
Yes, that's what I'm saying, instead of disabling the "Fetching transactions" debug log, isn't what we really want to correctly fix the issue of the mempool looper running in the Lightning backend? This way we still get the debug output on the mempool backends but it won't spam our logs on Lightning AND save CPU cycles and load on bitcoind RPC

softsimon:
> Yes, that's what I'm saying, instead of disabling the "Fetching transactions" debug log, isn't what we really want to correctly fix the issue of the mempool looper running in the Lightning backend? This way we still get the debug output on the mempool backends but it won't spam our logs on Lightning AND save CPU cycles and load on bitcoind RPC

Yes, I agree. But it's a separate issue. Average people are going to run Mainnet and Lightning in the same backend so this PR still makes sense to remove that mempool update spam now when we have so much more going on.

wiz (approved):
Tested ACK @ 8cc252642b93fcd7b903288963f5beaf3fa2aad0

PR #2344, opened 2022-08-20
Fix transaction sorting on frontend

### LINKED ISSUE #424: Electrumx backend: Address transactions sometimes not sorted in expected order
Looking up address `3BJ7yNNYcUzT9BWC7STxc6YPWS7CKNUFYH` the 2 transactions was confirmed in the same block and therefore has the same date, but the spending from the address should be ordered on top of the deposit transaction.

junderw:
Fixes #424 

Background:

1. Electrs (romanz) sends history in ascending block order, and within the same block it sorts them by ascending transaction position within the block. We flip that order, but don't touch the order of tx in the same block height, so the order gets messed up for same-block transactions.
2. Esplora sends in roughly the same order, but within the same block, it is ordered by F for funding, S for spending (this does not look intentional, and merely is a byproduct of not sorting explicitly by position in block. It is possible for some transactions that spend and fund the same address multiple times to be in incorrect order (theoretically)... when we flip the order of transactions, the F/S sorting (due to the database key) just happens to have the correct order when two clear funding and spending transactions are present. (the example in the previous issue)

Possible solutions:

1. Fix Esplora to explicitly sort by position in block (this will require changing the index keys, requiring reindex of the esplora db). Once we do that romanz and esplora will be in line. Then we can do something simple like reversing the array before doing the current sort. It will flip the order from ascending to descending for us before we run sort.
2. (This PR) Add a topological sort to the frontend. Sorting via dependency when the block is the same.

Areas of discussion for this PR:

1. I couldn't find unit tests for frontend, so I made a temp file that runs a simple test to show the logic is sound. Instructions on how to run the file are written in the file at the top. This should be moved into some unit test/removed before merging.
2. Topological sort does not guarantee any deterministic order for items without dependency relationships. ie. if two deposits are done in the same block, and they don't depend on each other, the sort algorithm should leave them alone, but depending on how it shuffles things around while searching there is no guarantee. Is this ok?

Areas of discussion for the Esplora solution:

1. Is reindexing and messing with indexing worth it?
2. Also, the REST API change is still in limbo over at esplora repo. How should we move forward with esplora changes if we decide to go down that path.


wiz:
Could you please rebase this onto current master?

softsimon:
Closing because we're not going this route.

PR #2351, opened 2022-08-22
Remove useless api call in channel page

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2346

#### Testing

* Open the browser network debugger
* After deploying the PR, open any channel page, for example `/lightning/channel/818244458842292224`
* Confirm that there is not call to `/api/v1/lightning/channels-geo` anymore
* Page should load much faster now


wiz (approved):
Tested ACK @ v2.5.0-dev [[c7809628](https://github.com/mempool/mempool/commit/c7809628)]

PR #2354, opened 2022-08-22
Improve channels world map performances

nymkappa:
#### Changes

- Warm cache `/api/v1/lightning/channels-geo?style=widget`  and `/api/v1/lightning/channels-geo?style=graph`
- For the lightning dashboard, since the map is not interactive, don't send all public keys and alias
- Limit the number of channels returned by the API to the top 10,000 biggest channels
- Ignore multiple channels between the same nodes in the same direction
- Added loading spinner animation while the map is loading and rendering all channels. When all channels are done rendering we remove the loading spinner
- API response is cached for 1 minute in the web browser

#### Impact

- Previously the API response for the map in the LN dashboard was 10.27mb uncompressed (transferred 2.14mb). Now it is 986.56kb while testing locally (no compression).
- It is also more visually pleasing now that we don't render all channels

**Before**

https://user-images.githubusercontent.com/9780671/185927978-b7c778aa-4789-4488-a7c0-dbff0275bf40.mov

**After**

https://user-images.githubusercontent.com/9780671/185927985-b440c5cb-8a97-41f2-b4fa-d2e6e2215d88.mov

#### Testing

* After deploying the PR open a private tab
* Open the network debugger
* Load the mining dashboard and confirm the API response for `/api/v1/lightning/channels-geo` is much lighter than before
* Confirm that the dashboard loading time "feels" faster

### DIFF
--- a/backend/src/api/explorer/channels.api.ts
+++ b/backend/src/api/explorer/channels.api.ts
@@ -17,32 +17,60 @@ class ChannelsApi {
     }
   }
 
-  public async $getAllChannelsGeo(publicKey?: string): Promise<any[]> {
+  public async $getAllChannelsGeo(publicKey?: string, style?: string): Promise<any[]> {
     try {
+      let select: string;
+      if (style === 'widget') {
+        select = `
+          nodes_1.latitude AS node1_latitude, nodes_1.longitude AS node1_longitude,
+          nodes_2.latitude AS node2_latitude, nodes_2.longitude AS node2_longitude
+        `;
+      } else {
+        select = `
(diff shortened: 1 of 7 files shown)


nymkappa:
@mononaut maybe this commit will break the unfurl layout so feel free to double check https://github.com/mempool/mempool/pull/2354/commits/fd46ea82bf783966f7b08c54bd400de41caba328

wiz (approved):
Tested ACK @ v2.5.0-dev [[fd46ea82](https://github.com/mempool/mempool/commit/fd46ea82)]

However there is a bug where the loading spinner animation appears forever on some pages such as `/lightning/channel/817151544242470912`

PR #2355, opened 2022-08-22
Fix network regex matching

softsimon:
This fixes the the bug in the URL network matching in the scenario of the name "liquid" for example appears later in the URL as in:

/lightning/nodes/rankings/**liquid**ity

Now matching from the root of the url only.

Make sure subnetworks still work like /testnet and /liquid


wiz (approved):
Tested ACK @ v2.5.0-dev [[9af2d478](https://github.com/mempool/mempool/commit/9af2d478)]

PR #2360, opened 2022-08-22
Alternative transaction unfurl design

mononaut:
This PR implements a different take on the transaction link preview design.

The design centers around an SVG "bow-tie"-shaped sankey diagram of inputs and outputs, which hopefully looks more striking than tables of raw data, while still conveying some useful information.

lightning-pool

e.g, it's easy to see which of the following transactions are simple spends, exchange payouts, whirlpool or wabisabi coinjoins etc:

| simple-spend | wabisabi |
|-|-|
| whirlpool | payout |

There's a limited amount of 'clear' space towards the top of the diagram, currently used to show total value and fee rate fields.

The bottom of the diagram (potentially overlapping some input/output lines) is used to show coinbase data, or up to three OP_RETURN messages. Perhaps this space could also be used to display other special transaction info, like details of lightning channel open/closes.

| coinbase | opreturn |
|-|-|

There isn't much room for anything else. Notably, confirmation status/time/block height is missing, as well as weight/vbyte size. We could possibly squeeze one of those fields in next to the txid at the top.


softsimon:
Some thoughts

* TXID migtht noe be required since it's in the link posted
* A transaction preview card need to contain the base details like Fee rate, date, Size, Weight, 
* The White graphis stands out too much. It should be in a mempool color like brighter purple or blue for example

mononaut:
Here's a quick mockup with the diagram in mempool blue, and the txid replaced with some extra data fields:

restyled

or using the mempool block gradient colors:

restyled-gradient-2

mononaut:
> This graphic might not be relatable since our TX page doesn't have it (yet) .. and a mini format of the lists of inputs and outputs might be better

Good point, yes, it might be confusing if there's a diagram on the link preview but not the actual page.

I did briefly play with some designs for just a simplified list of inputs and outputs, instead of a diagram, but I couldn't figure out how to make it look acceptable for transactions involving more than a couple of UTXOs.

wiz:
The gradient colors look cool, wanna ship this unfurler implementation PR first and then add it into the TX page later?

mononaut:
I just pushed a change updating the layout to match the mockups above, plus a truncated txid at the top.

The diagram now uses color gradients borrowed from the mined blocks graphics for each network, adjusted to make the relative contrast more consistent (and hue shifted heavily towards green for bitcoin testnet).

I'm not totally sure about all of these colors, maybe it would be better to stick to the mempool purple & blue for everything?

| Mainnet | Testnet |
|-|-|
| mainnet | testnet |
| Signet | Liquid |
| signet | liquid |
| Liquid Testnet | |
| liquid-testnet | |

softsimon:
Latest revision looks great and the overall design is growing on me. The great thing is that it pops with mempool colors which will make it very visual in a Twitter feed.. as compared to a simple information grid.

wiz (approved):
Beautiful! Tested ACK @ v2.5.0-dev [[6ae05842](https://github.com/mempool/mempool/commit/6ae05842)]

9ebc1e1e9950b586606d986d3391a6a97205fb5031ebe37f63ff7ed15ee5efc9

Issue #2359, opened 2022-08-22
Nodejs error in lightning backend


nymkappa:
Aug 22 18:52:51 node201.fra  mempool[91028]: ERR <lightning> Cannot update channel list. Reason: channelId.indexOf is not a function

PR #2361, opened 2022-08-22
Fix "cannot update channel list" error

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2359
Hopefully this will therefore also fixe the ISP ranking


wiz (approved):
Tested ACK @ v2.5.0-dev [[f24d4124](https://github.com/mempool/mempool/commit/f24d4124)]

Fixes the ISP pie chart 👍

Issue #2358, opened 2022-08-22
mysql lock error


nymkappa:
Aug 22 21:34:53 [39761] ERR: <lightning> $setNodesInactive() error: Deadlock found when trying to get lock; try restarting transaction

PR #2362, opened 2022-08-22
Await mysql queries in order to avoid deadlock

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2358

I'm not sure what's the reason for such deadlocks but the only two occurrences that I saw on keybase are from those two functions, so I'm feeling confident running them asyncronously is actually the cause of the problem. Way to go nodejs!


softsimon (approved):
utACK

wiz (approved):
Tested ACK @ v2.5.0-dev [[23fab652](https://github.com/mempool/mempool/commit/23fab652)]

Never experienced the reported issue but I tested it and it ran fine for 16 hours

Issue #2348, opened 2022-08-21
Esplora: Cannot read properties of undefined


nymkappa:
Investigate regular `Aug 21 13:16:02 node201.fra  mempool[67690]: ERR <lightning> $runClosedChannelsForensics() error: Cannot read properties of undefined (reading 'length')` errors, make sure they're handled properly and fix the log so we know which esplora API failed

PR #2363, opened 2022-08-22
Wrap esplora call into try/catch when scanning channel forensics

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2348

Completely untested.

The expected behavior is that when `ERR <lightning> $runClosedChannelsForensics() error: Cannot read properties of undefined (reading 'length')` error happen, instead of cutting off the loop we simply print the message and move on to the next channel.

If there is an issue with esplora, I expect to see those messages printed very often, so we can debug further if needed.


nymkappa:
@softsimon Please re-review the PR as I added try/catch granularity for each esplora call so we can log the faultry api calls.

softsimon (approved):
utACK

wiz (approved):
Tested ACK @ v2.5.0-dev [[a30bb4e6](https://github.com/mempool/mempool/commit/a30bb4e6)]

Seems OK so far, no errors yet:
```
Aug 23 14:34:21 node201.fra  mempool[61654]: INFO <lightning> Generating LN network stats for Sat, 28 Aug 2021 00:00:00 GMT (1630108800). Processed 12/1228 files
```

Issue #2376, opened 2022-08-23
Full width node history if location is not available


nymkappa:
<img width="906" alt="image" src="https://user-images.githubusercontent.com/9780671/186204761-da45523b-7d72-481a-8b8e-73c3c0ae5256.png">

PR #2377, opened 2022-08-23
Node history chart full width if we can't display channel map

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2376

Screen Shot 2022-08-23 at 6 01 04 PM
Screen Shot 2022-08-23 at 6 00 59 PM


wiz (approved):
Tested ACK @ v2.5.0-dev [[0add42c5](https://github.com/mempool/mempool/commit/0add42c5)]

PR #2366, opened 2022-08-23
Create active channel tree map component

nymkappa:
This PR adds a new component on the node page to show active channels in a tree chart.
It will require a bit of polishing to look great in the future.

We only show the top 1000 biggest channels for a node. If you click on a square it will open the channel page.

Some examples:

Screen Shot 2022-08-23 at 10 46 24 AM
Screen Shot 2022-08-23 at 10 45 56 AM
Screen Shot 2022-08-23 at 10 46 38 AM

#### Testing

* Once you deploy the PR, just open any node page and you'll see the new chart right above the channel list

PR #2381, opened 2022-08-23
Do not fetch node stats for channel tree graph

nymkappa:
Fixes an issue described by @softsimon which make the loading time for the node page very long.
The issue was introduced in https://github.com/mempool/mempool/pull/2366.

The issue is that [we fetch extra node stats](https://github.com/mempool/mempool/blob/8c885e87d41bc23e91811be09109f0b97b1208d3/backend/src/api/explorer/channels.api.ts#L291) for 1000 channels, which represent 1 db query for each of them in the backend, before returning the API response.

This PR makes sure [we don't run those extra query for the tree chart](https://github.com/mempool/mempool/pull/2381/files#diff-8fc84f80a2d1d6fd7c59b59be37355bdae8054c086a9d3148fa6538216d80ed3R308) (since we don't need that data at all).

Also, I [trimmed the fields returned by the API](https://github.com/mempool/mempool/pull/2381/files#diff-8fc84f80a2d1d6fd7c59b59be37355bdae8054c086a9d3148fa6538216d80ed3R309) for the tree chart because we don't use those, so the api response will be lighter as well.

The API endpoint in question is `/api/v1/lightning/channels?public_key=03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f&index=-1&status=active`

### DIFF
--- a/backend/src/api/explorer/channels.api.ts
+++ b/backend/src/api/explorer/channels.api.ts
@@ -288,21 +288,36 @@ class ChannelsApi {
 
       const channels: any[] = []
       for (const row of allChannels) {
-        const activeChannelsStats: any = await nodesApi.$getActiveChannelsStats(row.public_key);
-        channels.push({
-          status: row.status,
-          closing_reason: row.closing_reason,
-          capacity: row.capacity ?? 0,
-          short_id: row.short_id,
-          id: row.id,
-          fee_rate: row.node1_fee_rate ?? row.node2_fee_rate ?? 0,
-          node: {
-            alias: row.alias.length > 0 ? row.alias : row.public_key.slice(0, 20),
-            public_key: row.public_key,
-            channels: activeChannelsStats.active_channel_count ?? 0,
-            capacity: activeChannelsStats.capacity ?? 0,
-          }
-        });
+        let channel;
+        if (index >= 0) {
+          const activeChannelsStats: any = await nodesApi.$getActiveChannelsStats(row.public_key);
+          channel = {
+            status: row.status,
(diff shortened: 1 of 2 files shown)


backend/src/api/explorer/channels.api.ts
```diff
@@ -288,21 +288,36 @@ class ChannelsApi {
...
+            id: row.id,
+            fee_rate: row.node1_fee_rate ?? row.node2_fee_rate ?? 0,
+            node: {
+              alias: row.alias.length > 0 ? row.alias : row.public_key.slice(0, 20),
+              public_key: row.public_key,
+              channels: activeChannelsStats.active_channel_count ?? 0,
+              capacity: activeChannelsStats.capacity ?? 0,
+            }
+          };
+        } else if (index === -1) {
```
softsimon, line 308: Maybe just an `else` here instead of `else if ` since channels will push undefined if index is -2 or something else .. maybe not important but.

nymkappa: The correct fix is to return an appropriate http error code if the query parameter is invalid. I will add this test in the route.

nymkappa: Added query parameter test in https://github.com/mempool/mempool/pull/2381/commits/43cc9499b1dbfdd18ca0c12a1e4e07aa2e9462ef

wiz (approved):
Tested ACK @ v2.5.0-dev [[a9e1ce42](https://github.com/mempool/mempool/commit/a9e1ce42)]

Issue #2382, opened 2022-08-24
Incorrect log for node location update


nymkappa:
`Aug 24 11:06:18 [62733] INFO: <lightning> 3504 nodes location data updated`

This number is not correct

PR #2385, opened 2022-08-24
Log correct maxmind mysql updates - fix stats import processed files counter

nymkappa:
Very low priority PR

- Fixes https://github.com/mempool/mempool/issues/2382
- Also, when we skip some files for various reason the stats importer, we now properly increment the number of processed file so the log will correct progress


wiz (approved):
Partially Tested ACK @ v2.5.0-dev [[cc79b2b2](https://github.com/mempool/mempool/commit/cc79b2b2)]

Issue #2384, opened 2022-08-24
Cleanup up imported gossip data with hardcoded condition or queries


nymkappa:
Once the latest backup is imported, I'll code the automatic cleanup

PR #2387, opened 2022-08-24
Hardcode some condition to invalidate imported topology files

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2384

During the topology import, we ignore some files based on their timestamp and node count. If the node count is too low based on our hard coded estimates, we skip the file.

#### Testing

* You need to delete the historical data from your db. Make sure to keep non recoverable historical data. You can run the following queries to delete only historical data:
```mysql
# Delete data before July 16th 2022
delete from lightning_stats where UNIX_TIMESTAMP(added) < 1657929600;
delete from node_stats where UNIX_TIMESTAMP(added) < 1657929600; # This will take some time
```
* Restart the lightning nodejs backend and wait for the historical import to complete
* Visually confirm the expected following result
* In node pages, the same "interpolated gaps" in the channels/capacity charts should be visually present

#### Result

Raw data
image
Cleaned up
Screen Shot 2022-08-28 at 7 50 29 AM

Raw data
image
Cleaned up
Screen Shot 2022-08-28 at 7 50 37 AM

Raw data
image
Cleaned up
Screen Shot 2022-08-28 at 7 50 55 AM

### DIFF
--- a/backend/src/tasks/lightning/sync-tasks/stats-importer.ts
+++ b/backend/src/tasks/lightning/sync-tasks/stats-importer.ts
@@ -367,6 +367,12 @@ class LightningStatsImporter {
           continue;
         }
     
+        if (this.isIncorrectSnapshot(timestamp, graph)) {
+          logger.debug(`Ignoring ${this.topologiesFolder}/${filename}, because we defined it as an incorrect snapshot`);
+          ++totalProcessed;
+          continue;
+        }
+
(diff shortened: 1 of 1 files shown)


wiz (approved):
Mostly Tested ACK @ v2.5.0-dev [[6b049f2c](https://github.com/mempool/mempool/commit/6b049f2c)]

PR #2388, opened 2022-08-24
Fix path for newsyslog configs in prod installer

Emzy:


wiz (approved):
utACK - looks good but not sure if `newsyslog -C` works on linux

Emzy:
> utACK - looks good but not sure if `newsyslog -C` works on linux

fixed in da4e6fc

Emzy:
Added Debian rsyslog configuration.

wiz (approved):
utACK

PR #2398, opened 2022-08-25
Handle languages in network regex detection

### LINKED ISSUE #2396: Network regex broken when on different languages
This bug got introduced in https://github.com/mempool/mempool/pull/2355

Network is not handled when using language **/sv/signet**

softsimon:
fixes #2396

Going to` /sv/testnet` was broken as it expected the network name to be the first.

Now this should detect "testnet" 

`/testnet/`
`/jp/testnet/`
`/pt-BR/testnet/`


wiz:
Replaced by #2413

PR #2401, opened 2022-08-27
[Feature]: Improve Search bar with better regex and network based address matches

junderw:
Currently WIP.

[Example of good Regex commenting](https://github.com/bisq-network/bisq/blob/4d62136d6218d4785bc678bf10daf22c4d85f960/desktop/src/main/java/bisq/desktop/components/paymentmethods/data/JapanBankData.java#L859) given by @wiz 

TODOs:
- [ ] Find exhaustive list of Liquid Testnet version bytes (missing P2PKH and potentially one other bech32/blech32 HRP)
- [x] (From @softsimon) Limit blockheight regex to a length smaller than lightning channel id. (around 17 digits) (Solved: picked 13 max)
- [ ] (From @softsimon) Add Regex for compressed pubkey (to search for lightning node by node pubkey)


frontend/src/app/shared/regex.utils.ts
```diff
@@ -0,0 +1,224 @@
...
+const HEX_CHARS = `[a-fA-F0-9]`;
+
+// A regex to say "A single 0 OR any number with no leading zeroes"
+// Capped at 13 digits so as to not be confused with lightning channel IDs (which are around 17 digits)
+// (?:             // Start a non-capturing group
+//   0             // A single 0
+//   |             // OR
+//   [1-9][0-9]{0,12} // Any succession of numbers up to 13 digits starting with 1-9
+// )               // End the non-capturing group.
+const ZERO_INDEX_NUMBER_CHARS = `(?:0|[1-9][0-9]{0,12})`;
```
antonilol, line 18: this could match a lightning channel id (in regtest). a channel id can be anywhere from 2^40 (no channels can be confirmed in the genesis block, so height > 0) to 2^64-1, with the lowest being 13 characters long, this regex is used for blockheight and transaction output index, which are both limited to 32 bit integers. the highest 32 bit int is only 10 characters long

natsee:
> * [ ]  (From @softsimon) Add Regex for compressed pubkey (to search for lightning node by node pubkey)

It looks like this has been implemented: 

Screenshot 2023-12-28 at 16 42 27

****************************

I am willing to work on this branch if needed, to resolve conflicts (mainly adding the support to search for P2PK and date / timestamp), and also to integrate a fix for https://github.com/mempool/mempool/issues/4529 (bug when searching a block by hash on signet).

PR #2402, opened 2022-08-27
Fix for broken SQL to load Lightning nodes on ISP page

softsimon:
My mistake


wiz (approved):
Tested ACK @ v2.5.0-dev [[d051538c](https://github.com/mempool/mempool/commit/d051538c)]

PR #2403, opened 2022-08-27
Fix for difficulty adjustment throwing error before in sync

softsimon:
This fixes the following error that happens when the backend restarts before it syncs with the chain:

`Aug 27 08:01:53 [28512] DEBUG: Error parsing websocket message: Cannot read properties of undefined (reading 'timestamp')`

and
Screen Shot 2022-08-27 at 10 19 46

This PR:

The `/api/v1/difficulty-adjustment` will send 503 Service Unavailable before it has valid data.

Early websocket connections will receive a null value for Difficulty adjustment but will be corrected in the frontend the very next second on the next valid websocket push.


wiz (approved):
Tested ACK @ v2.5.0-dev [[d9320574](https://github.com/mempool/mempool/commit/d9320574)]

PR #2408, opened 2022-08-27
Fixes multiple bugs with outspends and channels

### LINKED ISSUE #412: UI: When address receives a new transaction, the UTXO links stay the same as in previous tx
### Description

Just noting this reproducible UI issue.

Open [mv4rnyY3Su5gjcDNzbMLKBQkBicCtHUtFB address](https://mempool.space/testnet/address/mv4rnyY3Su5gjcDNzbMLKBQkBicCtHUtFB) which belongs to [coinfaucet.eu btc-testnet](https://coinfaucet.eu/en/btc-testnet/) in a modern browser and send a small transaction there (so that it will have two outputs). If the previous last transaction also had two outputs one of which was already used (red arrow instead of green), the newly-appearing transaction will get one red arrow even though it is very fresh (even not confirmed yet) and non of the UTXOs has been spent.

#### Version

Current production `mempool.space`, i.e. `v2.2-dev (21bb879f)`

### Steps to reproduce

Described above.

### Expected behaviour

The arrows on newly arrived transaction should point to actual UTXOs and turn red only if the UTXO is spent.

### Actual behaviour

On dynamic update (probably done by JavaScript) the arrow states of the OUTPUTs are kind of inherited from the recently-shown last transaction (just before arriving the new one).

### Screenshots

No need for screenshot. Have a look at https://mempool.space/testnet/address/mv4rnyY3Su5gjcDNzbMLKBQkBicCtHUtFB and send a testnet transaction there.

#### Device or machine

Linux x86_64, Google Chrome 89.0.4389.90

softsimon:
fixes #412

**Test cases**
1. On an address page with more than 25 transactions. Load more (on scroll) should only fetch the outspends and channel data of the newly loaded transactions and update the UI correctly
2. On an address page, if a new transaction pops in to the address, the outspend should be fetched correctly and update the page and not affect the rest of the transaction outspends (red or green arrows)
3. A transaction page should now display all the existing channel openings, not just 1 (example below)

**Description**
When loading more transactions on an address page, or if new transactions pop in, outspends and channel data should be fetched only on these new transactions and update in the UI.

Current bug was that outspends loadmore loaded more data than it required, and new address transactions coming in broke the Outspend arrow links index order.

Another bug was that we assumed only 1 channel open in one transaction, this is not the case.

Example on mainchain TX: 93163493e65c52d66a0fbafb7c97322a434ab0bbd9554a8c5b7b70640eb4b931

Before:
Screen Shot 2022-08-27 at 16 04 59

After:
Screen Shot 2022-08-27 at 16 05 07

### DIFF
--- a/backend/src/api/explorer/channels.routes.ts
+++ b/backend/src/api/explorer/channels.routes.ts
@@ -70,7 +70,7 @@ class ChannelsRoutes {
     }
   }
 
-  private async $getChannelsByTransactionIds(req: Request, res: Response) {
+  private async $getChannelsByTransactionIds(req: Request, res: Response): Promise<void> {
     try {
       if (!Array.isArray(req.query.txId)) {
         res.status(400).send('Not an array');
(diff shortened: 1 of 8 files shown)


softsimon:
tested ACK @ v2.5.0-dev [[40dc4764](https://github.com/mempool/mempool/commit/40dc4764)]

Verified the 3 test cases mentioned now working as compared to before

PR #2413, opened 2022-08-27
Handle locale & preview in network detection regex

mononaut:
PR #2355 changed the network regex to only match network strings at the start of the url, to avoid accidentally matching the wrong network when a string appears later.

so it correctly identifies e.g. `/lightning/nodes/rankings/liquidity` as a bitcoin mainnet route instead of a liquid route.

Unfortunately, routes can also have language and/or 'preview' prefixes, and the new regex fails to match the network in these urls. e.g:

`/ja/testnet/...`
`/preview/testnet/block/...`
`/pt-BR/preview/testnet/address/...`

This PR extends the network regex to handle these cases.


mononaut:
I just noticed issue #2396, and realized the language handling part of this fix duplicates PR #2398

softsimon:
> I just noticed issue #2396, and realized the language handling part of this fix duplicates PR #2398

Yours is arguably a bit better and detailed

wiz (approved):
Tested ACK @ e7aa862e43979e3d137c5b8353ba2f24cbc7d800

PR #2414, opened 2022-08-27
Fix unfurls for lightning pages with no geodata

mononaut:
Fixes an unfurler bug following a change in the behavior of the lightning map component.

The map is now not rendered by default when a node/channel has no geolocation, and so does not trigger the 'ready' event expected by the unfurler.

This PR adds an optional 'placeholder' input to the map component, which causes the map to always render.

node-no-geodata

channel-no-geodata


wiz (changes requested):
Tested NACK @ v2.5.0-dev [[a2e4da84](https://github.com/mempool/mempool/commit/a2e4da84)]

The `/preview` page now shows the placeholder image but for some reason I get a `failed to render page preview` error:
```
failed to render page preview for screenshot due to client-side error. probably requested an invalid ID
```

wiz (approved):
Sorry disregard previous review, was a problem with my test env

Tested ACK @ v2.5.0-dev [[a2e4da84](https://github.com/mempool/mempool/commit/a2e4da84)]

PR #2415, opened 2022-08-28
Disable LN historical import if not mainnet

nymkappa:
Disable LN stats historical import if non mainnet

#### Testing that stats import is not triggered on non mainnet

* Set your `config.MEMPOOL.network.NETWORK` to `testnet` or `signet`
* `delete from lightning_stats where UNIX_TIMESTAMP(added) = 1657836000` (Query executed OK, 1 row affected)
* Restart the lightning nodejs backend
* Wait for the following info message `INFO: <lightning> Starting Lightning Stats service`
* Confirm the previously deleted row has not been re-imported `select * from lightning_stats where UNIX_TIMESTAMP(added) = 1657836000` (0 row)

#### Testing that stats import still works on mainnet

* `delete from lightning_stats where UNIX_TIMESTAMP(added) = 1657836000` (Query executed OK, 1 row affected)
* Restart the lightning nodejs backend
* Wait for the following info message `INFO: <lightning> Founds a topology file that we did not import. Importing historical lightning stats now.`
* Confirm the previously deleted row has properly been re-imported `select * from lightning_stats where UNIX_TIMESTAMP(added) = 1657836000` (1 row)


wiz (approved):
Tested ACK @ v2.5.0-dev [[28b9205e](https://github.com/mempool/mempool/commit/28b9205e)]

PR #2418, opened 2022-08-28
Fix js crash in node page and add loading spinner to channel tree chart

nymkappa:
Fixes js crash when there is no active channels

Screen Shot 2022-08-28 at 11 13 30 AM

### Spinner

@softsimon A spinner you won't need to fix

https://user-images.githubusercontent.com/9780671/187066879-ff50355f-1de1-44ea-9ba3-a4b914216bc0.mov


wiz (approved):
Tested ACK @ v2.5.0-dev [[8d3a1c2d](https://github.com/mempool/mempool/commit/8d3a1c2d)]

PR #2420, opened 2022-08-28
Custom lazy loading strategy

softsimon:
Adding a custom lazy loaded strategy so that we can customize which modules should be lazy loaded and when, using the `data: { preload: true },` config on the lazy-routes. I also added a slight 150ms delay for the preloading so all the requests for the current page is prioritized.

Solves the current issue with always-on preloading:

* Disabled pre-loading for the Preview cards module because it's never being used by users
* Disabled pre-loading of Lightning module IF the Lightning enabled is set to `false` in the frontend


wiz (approved):
Tested ACK @ v2.5.0-dev [[b7662347](https://github.com/mempool/mempool/commit/b7662347)]

PR #2421, opened 2022-08-28
Fix clipboard for node pubkey on channel page

hunicus:
Copy-to-clipboard doesn't work because of a misnamed variable in the channel template.

Test by trying to copy node pubkey on any channel page using the click-to-copy button. Example I tested with was `lightning/channel/824803045759582209`.

copy-node-pubkey


softsimon (approved):
tested ACK @ v2.5.0-dev [[bf0244f9](https://github.com/mempool/mempool/commit/bf0244f9)]

PR #2423, opened 2022-08-28
Save proper node socket format into db

nymkappa:
This PR updates the historical import to support the new topology output for node addresses (which are now encoded in base32 format instead of raw bytes, see https://github.com/mempool/topology/pull/4).

This also fixes the missing geolocation for historical nodes with ipv4 addresses.

#### Testing

- Truncate the historical lightning stats
```mysql
# Delete data before July 16th 2022
delete from lightning_stats where UNIX_TIMESTAMP(added) < 1657929600;
delete from node_stats where UNIX_TIMESTAMP(added) < 1657929600; # This will take some time
```
- Run the lightning nodejs backend with the new topology files. Nodes sockets will be updated so you don't have to truncate this table.


wiz:
ready for testing?

wiz (approved):
Tested ACK @ v2.5.0-dev [[1a5ee325](https://github.com/mempool/mempool/commit/1a5ee325)]

PR #2429, opened 2022-08-29
Fix overflowing tx on mobile

nymkappa:
This PR should be carefully reviewed by at least two persons because it touches the transaction component.

#### Currently the tx list overflows because of the tx tags

image

#### PR result

**Mobile channel page**
image

**Desktop channel page**
image


softsimon:
Added some margin below address labels as I think it becomes too tight and merged.

softsimon:
Added some margin below address labels as I think it becomes too tight and merged.

softsimon:
I am not a fan of this new compressed look without margin between each address and their labels. 

Screen Shot 2022-08-29 at 12 33 55

wiz (changes requested):
NACK for reasons Simon listed

nymkappa:
@softsimon Please give this bug a shot

PR #2433, opened 2022-08-29
Node and Channel page skeleton loaders

softsimon:
Skeleton loaders for Node and Channel pages. UPDATED.

Node skeleton loader preview:

Screen Shot 2022-08-30 at 13 47 27

Channel skeleton loader preview:

Screen Shot 2022-08-30 at 13 47 34


nymkappa:
Thanks for that!

Some feedbacks:

- In the node page, how about having darker background for the socket instead of white? Actually I just realize but why is it even white in the first place? I don't remember seeing this style anywhere else on the site.
- In the node page, I'm not fan of the giant "scan" effect for charts. I think it would look cleaner to show a darker area with a loading spinner for each chart, see the mining dashboard for example
Screen Shot 2022-08-29 at 6 02 07 PM
- In the node page, There is no skeleton for the channel tree chart in the node page, so the channel list jumps
- In the channel page, the skeleton hides too early, so between the time the skeleton is hidden and the time the map is renderer, there is nothing, and the pages jump
image

nymkappa (changes requested):
Thanks for that!

Some feedbacks:

- In the node page, how about having darker background for the socket instead of white? Actually I just realize but why is it even white in the first place? I don't remember seeing this style anywhere else on the site.
- In the node page, I'm not fan of the giant "scan" effect for charts. I think it would look cleaner to show a darker area with a loading spinner for each chart, see the mining dashboard for example

Screen Shot 2022-08-29 at 6 02 07 PM

- In the node page, There is no skeleton for the channel tree chart in the node page, so the channel list jumps
- In the channel page, the skeleton hides too early, so between the time the skeleton is hidden and the time the map is renderer, there is nothing, and the pages jump

image

- In the node page, the page jumps between the skeleton and the content, probably because of the multi lines text

https://user-images.githubusercontent.com/9780671/187244920-a017dd0c-9228-4b78-8acb-4da920af783a.mov

softsimon:
One issue left is that the Location and ISP can be missing, or Location take a separate row when Country is present causing a jump.

wiz (approved):
Tested ACK @ v2.5.0-dev [[ef16f3bd](https://github.com/mempool/mempool/commit/ef16f3bd)]

PR #2435, opened 2022-08-29
Replacing all PNGs with inline SVG

softsimon:
* All network dropdown icons are now inline SVGs (Dropdown paddings has changed slightly)
* Mempool (without .space), Liquid Network and Bisq Markets logo is now SVG,
* Dashed divider is now a SVG dashed line (UX has changed slightly)
* Clippy copy button is now inline SVG

**How to test**
Check out so that the icons render correctly and is not misaligned on all Master Pages including Liquid, Bisq AND Preview Cards for all networks.

Screen Shot 2022-08-29 at 19 03 07

Screen Shot 2022-08-29 at 19 03 11

Screen Shot 2022-08-29 at 19 03 20


wiz (approved):
Tested ACK @ v2.5.0-dev [[a0c54531](https://github.com/mempool/mempool/commit/a0c54531)]

PR #2439, opened 2022-08-29
Add loading animation for channel list

nymkappa:
This PR adds a loading spinner when we change between open/close channels, and when we use the pagination controls. It is the exact same idea as the blocks list loading effect. During the loading, the list opacity is also set to 0.75 instead of 1.0.

image

image


softsimon (approved):
tested ACK @ v2.5.0-dev [[90c0ece9](https://github.com/mempool/mempool/commit/90c0ece9)]

Issue #2291, opened 2022-08-13
Fix wrong ASN


nymkappa:
if the IP is in the 170.75.160.0/20 or 172.81.176.0/21 netblocks, set the ASN to 394745
Use https://github.com/gosquared/ipcheck/blob/master/ipcheck.js
Example: `IPCheck.match('192.168.0.1', '192.168.0.1/32'); //= true`

PR #2441, opened 2022-08-29
Fix wrong ASN for Lunanode ip ranges

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2291

**Note: It seems to create some unintended side effects as described in matrix, I may need some help to debug this**

#### Testing

* Delete wrong lunanode as number from the db
```mysql
delete from geo_names where names = '"Lunanode"'
```
* Restart the lightning nodejs backend
* Confirm Lunanode as number is correct
```mysql
select id as as_number from geo_names where names = '"Lunanode"'
```
image


wiz (approved):
Tested ACK @ v2.5.0-dev [[56ec8b90](https://github.com/mempool/mempool/commit/56ec8b90)]

It's better but AS 174 is not Lunanode so we need to fix that next :(

PR #2442, opened 2022-08-29
Option to disable puppeteer in unfurler

mononaut:
Adds an optional `PUPPETEER.DISABLE` config setting, which disables puppeteer and /render api endpoints (defaults to false).


wiz (approved):
Tested ACK @ 034b7fb516d91d5fe17431efd7ddfc41106df5d7

PR #2443, opened 2022-08-30
Update channel map color

nymkappa:
This PR changes the default world map background color from the current grey to dark blue.

Color is now `#272b3f`.
`#212536` was also suggested by pedro but it seems too dark to me when screen luminosity is low.
Feel free to play around with the color.

image

image


softsimon (changes requested):
ACK for the color change.. I think it looks better

NACK for including the Luna Node AS commits in the same PR

wiz (approved):
Tested ACK @ v2.5.0-dev [[56ec8b90](https://github.com/mempool/mempool/commit/56ec8b90)]

PR #2444, opened 2022-08-30
Fix infinite loading spinner in channel map when no active channel exists

nymkappa:
`If there are no open channels then the channel`~map~`tree chart fails to load and loading spinner stays forever`

### Testing

* Load /lightning/node/0283de1355ab232080de2904edc77388c6651e2d81a0336bfb739f893b35aa76ce
* Confirm that the channel tree chart is not shown and the loading spinner is gone
* Load /lightning/node/026c795878725bed713c36511857ecb67557516376174fbb8c8c54e117f02d8143
* Confirm that the channel tree chart is shown and the loading spinner is gone


wiz (approved):
Tested ACK @ v2.5.0-dev [[f4f60a6b](https://github.com/mempool/mempool/commit/f4f60a6b)]

PR #2445, opened 2022-08-30
When there is no stats for the past 7 days, don't show % changes

nymkappa:
The percentage change in the lightning dashboard aim to show the change for the past 7 days. Currently it's bugged if you don't run your lightning backend for at least 7 days, because if we will old historical data, therefore showing huge % increase.

This PR fixes:
* The mysql query to select the current day and current day - 7 days instead of the last 7th rows
* If there is not point of comparison, we hide the percentage changes and tooltips
* Fix some i18n

Screen Shot 2022-08-30 at 9 06 54 AM

Screen Shot 2022-08-30 at 9 08 59 AM

### Testing
* Run `SELECT * FROM lightning_stats WHERE DATE(added) = DATE(NOW() - INTERVAL 7 DAY)`
  * If you have a result  
    * **Take note of the `added` field and `id`**
    * Confirm that you see the % change in the dashboard
    * Change the date of the previously found row so that the previous query does not return any result
    * Confirm that you don't see the % change in the dashboard (make sure to hard reload the page)
    * Don't forget to set the manually modified `added` field back to its previous value
  * If you don't have a result
    * Confirm you don't see % change in the dashboard 
    * Create a new row and set the `added` field manually so that it is 7 days before the most recent one
    * Confirm you see the % change in the dashboard
    * Don't forget to delete the manually inserted row

### DIFF
--- a/backend/src/api/explorer/statistics.api.ts
+++ b/backend/src/api/explorer/statistics.api.ts
@@ -27,7 +27,7 @@ class StatisticsApi {
   public async $getLatestStatistics(): Promise<any> {
     try {
       const [rows]: any = await DB.query(`SELECT * FROM lightning_stats ORDER BY added DESC LIMIT 1`);
-      const [rows2]: any = await DB.query(`SELECT * FROM lightning_stats ORDER BY added DESC LIMIT 1 OFFSET 7`);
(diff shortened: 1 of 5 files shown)


softsimon (approved):
tested ACK @ v2.5.0-dev [[aec6d57f](https://github.com/mempool/mempool/commit/aec6d57f)]

wiz:
I don't think this is good - we should show the percentages or some placeholder otherwise the layout changes randomly

PR #2448, opened 2022-08-30
Typeahead loading spinner and regex fixes

softsimon:
* Added a loading-spinner in the search bar while the user is doing "typeahead-search"
* Fixed the Block Height regex so it only accepts 9 digits long block heights. Fixes how long Channel IDs are interpreted as channels. 
* Fixed the Transaction ID regex that incorrectly detected Node PublicKeys are TXIDs due to the output index matching (eg TXID:OUTPUT)

New loading spinner:
Screen Shot 2022-08-30 at 11 48 40

Previously searching quickly for a NodeID
Screen Shot 2022-08-30 at 11 49 14

Previously searching quickly for a Channel ID
image


softsimon:
2 new fixes

* When the loading spinner is spinning you can still click on it to search
* When you click on the Search button it will search on the item that is selected on top in the list

PR #2460, opened 2022-08-30
Harcode lunanode, FDCservers and cogent asn

nymkappa:
Hardcode some asn to fix buggy maxmind responses.
You need to truncate `geo_names` table.
Nodes asn will be updated automatically.

image


wiz (approved):
Tested ACK @ v2.5.0-dev [[51d35ec7](https://github.com/mempool/mempool/commit/51d35ec7)]

PR #2467, opened 2022-08-30
Fix unfurler navigation race condition

mononaut:
Fixes a race condition in the unfurler, which could sometimes cause previews (usually lightning previews) to render incorrectly on the first request after a restart or puppeteer page crash.


wiz (approved):
utACK

Issue #2465, opened 2022-08-30
ipv6 has emty text and the brackets are missing


Emzy:
### Description

At the node address field the you can select to show ipv6 but it has no text.
Also ipv6 needs [] around the address.

#### Version

v2.5.0-dev d578353f

### Steps to reproduce

Open a node like 03e399589533581e48796e29a825839a010036a61b20744fda929d6709fcbffcc5
select the addess type (ipv4) and click below on the empty field.

### Expected behaviour

"ipv6"
Address: ...cbffcc5@2401:b140:1::92:206:9735
### Actual behaviour

""
Address: Address: ...cbffcc5@[2401:b140:1::92:206]:9735
### Screenshots

<img width="1444" alt="Screen Shot 2022-08-30 at 20 15 02" src="https://user-images.githubusercontent.com/1223908/187514152-8fc9d699-9edf-48cd-a8fb-3febcc7ac824.png">
<img width="1453" alt="Screen Shot 2022-08-30 at 20 15 13" src="https://user-images.githubusercontent.com/1223908/187514163-b6eb7e89-684f-48e9-b1a8-2cca43dec2de.png">
<img width="1452" alt="Screen Shot 2022-08-30 at 20 17 24" src="https://user-images.githubusercontent.com/1223908/187514174-4c99a5e1-3d0b-478e-9d0e-9e9289830448.png">

#### Device or machine

FF on macOs

PR #2468, opened 2022-08-30
CLN - Add brakets around ipv6

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2465

image

image


Emzy (approved):
tested ACK v2.5.0-dev b690dcaa

PR #2469, opened 2022-08-30
Only show active channels on world map

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2498

This node `034dee6a84aadb411a7ab40ece6f6777ba20d4e7a575035b49c374b4e112466cc6` has currently 3 active channels (display in the header, tree chart and channel list).

But if you zoom in the node in the channel map, you actually see 5
image

Some of those channels are actually closed channels. So this PR updated the mysql query to make sure we only show active channels in the world map.

### DIFF
--- a/backend/src/api/explorer/channels.api.ts
+++ b/backend/src/api/explorer/channels.api.ts
@@ -39,7 +39,8 @@ class ChannelsApi {
         FROM channels
         JOIN nodes AS nodes_1 on nodes_1.public_key = channels.node1_public_key
         JOIN nodes AS nodes_2 on nodes_2.public_key = channels.node2_public_key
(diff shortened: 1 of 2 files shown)


wiz (changes requested):
NACK - if node has no active channels then map is broken forever loading

nymkappa:
> Maybe this should be a toggle, i.e. switch showing closed channels, and if the node has no open channels then enable the show closed channels, also maybe the closed channels can be a different color

Yeah we need a better UX to tie the map, the channel tree chart and the channel list all together.

If they all follow the open/close channel control that could be nice.

That is not the goal of that PR tho so this can be a feature request.

softsimon (approved):
tested ACK @ v2.5.0-dev [[a75262d7](https://github.com/mempool/mempool/commit/a75262d7)]

Issue #2463, opened 2022-08-30
Node Last Update date is wrong


softsimon:
Last Update date should be when the node last published a change to either a channel or the node info.

Currently:

Node01.Fulmo.Org say Last update _‎2022-05-09 04:50 (3 months ago)_ which is very confusing given the node is online and active.

PR #2470, opened 2022-08-30
Consider channels updates as well for node updated at field

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2463

Currently we're only showing latest node update (color, alias etc...)

image

With this PR, we also consider channels update timestamps

image

### DIFF
--- a/backend/src/api/explorer/channels.api.ts
+++ b/backend/src/api/explorer/channels.api.ts
@@ -522,6 +522,23 @@ class ChannelsApi {
       logger.err('$setChannelsInactive() error: ' + (e instanceof Error ? e.message : e));
     }
   }
+
+  public async $getLatestChannelUpdateForNode(publicKey: string): Promise<number> {
+    try {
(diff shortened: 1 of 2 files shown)


softsimon:
I think this fix works for now.. But I believe we should update the actual updated_at or add a new updated-field in the database where we actually update the node last updated date when the channel updates are coming in, not making an extra query for the latest channel update.

nymkappa:
> I think this fix works for now.. But I believe we should update the actual updated_at or add a new updated-field in the database where we actually update the node last updated date when the channel updates are coming in, not making an extra query for the latest channel update.

Fixed in https://github.com/mempool/mempool/pull/2470/commits/abff8488312f6c0c79bb175d43b4e208b5980336

wiz (approved):
Tested ACK @ v2.5.0-dev [[ff1aae85](https://github.com/mempool/mempool/commit/ff1aae85)]

PR #2471, opened 2022-08-30
Change channel api arg from public_key to publicKey

hunicus:
None of the other API items use snake_case for arguments, but `getChannelsForNode()` does for some reason...would be nice to switch it to camelCase for consistency.


softsimon (changes requested):
tested NACK @ v2.5.0-dev [[806adc3c](https://github.com/mempool/mempool/commit/806adc3c)]

Frontend client is still requesting `public_key` so it breaks.

hunicus:
Shoot, of course. I'm going to close this PR as it's not that important and I'm not familiar enough with the frontend code to make the changes without breaking anything...will open an issue instead.

PR #2474, opened 2022-08-30
Grey out inactive search results

softsimon:
Inactive nodes and closed channels will now be shown as greyed out. (But you can still navigate and click on them.)

Example:

Screen Shot 2022-08-30 at 23 58 12


wiz (approved):
Tested ACK @ v2.5.0-dev [[38324575](https://github.com/mempool/mempool/commit/38324575)]

PR #2483, opened 2022-08-31
Fix unfurler race condition again

mononaut:
Second attempt at fixing the race condition, after #2467 was apparently incompatible with the production server.

This should better address the root cause of the bug, and avoids the incompatible puppeteer command that broke in production before.

The main symptom is occasional blank templates on lightning node and channel previews, e.g.
blank-channel

It's tricky to reproduce, because it seems to occur only when a lightning preview request immediately follows a session restart with exactly the wrong timing.

This PR also fixes a separate bug where transaction previews could sometimes render with null timestamps (1970-01-01...).


wiz (approved):
Tested ACK @ 80dfa0e937ba5a3f406548f97fdc0e8d3ad8ac64

PR #2484, opened 2022-08-31
Use new mempool preview image as default

mononaut:
Replaces the old default mempool preview image with the new dashboard preview, to be more consistent.

before:
| dashboard only (https://mempool.space/, https://mempool.space/testnet/ etc) | default/fallback for other routes (https://mempool.space/docs or https://mempool.space/block/<invalid>, etc) |
|-|-|
| dashboard | mempool-space-preview |

after:
| both dashboards and general default/fallback |
|-|
| dashboard |

### DIFF
--- a/unfurler/src/routes.ts
+++ b/unfurler/src/routes.ts
@@ -56,8 +56,7 @@ const routes = {
 
 const networks = {
   bitcoin: {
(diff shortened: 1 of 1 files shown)


wiz (approved):
utACK

PR #2485, opened 2022-08-31
Restyle lightning preview titles to match main pages

mononaut:
This PR slightly restyles the lightning preview headers to match the latest design of the main node and channel pages.

| Before | After |
|-|-|
| node-before | node-after |
| channel-before | channel-after |

### DIFF
--- a/frontend/src/app/components/address/address-preview.component.html
+++ b/frontend/src/app/components/address/address-preview.component.html
@@ -1,12 +1,12 @@
 <div class="box preview-box" *ngIf="address && !error">
+  <h2 class="preview-header" i18n="shared.address">Address</h2>
   <div class="row">
     <div class="col-md">
-      <div class="title-address">
-        <h1 i18n="shared.address">Address</h1>
+      <div class="row d-flex justify-content-between">
+        <div class="title-wrapper">
+          <h1 class="title truncated"><span class="first">{{addressString.slice(0,-4)}}</span><span class="last-four">{{addressString.slice(-4)}}</span></h1>
+        </div>
       </div>
-      <a [routerLink]="['/address/' | relativeUrl, addressString]" class="address-link" >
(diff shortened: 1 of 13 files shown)


wiz:
Hmm, with this change the "lightning node" becomes quite small.
How about moving the title, i.e. "lightning node" into the top center empty part?

softsimon (approved):
utACK

Looks great!

wiz:
After discussing with Simon, we think we should keep the logo and network consistent with the mempool.space website, so in this case "(B) Mainnet" instead of "(B) Mainnet Lightning" and then put "Lightning Node" as the title in the center, then you can increase the font size of the pubkey or channel ID a bit as well

mononaut:
like this?

| node-w-header | channel-w-header |
|-|-|

mononaut:
do we want similar headers on the block, address and transaction previews?

wiz:
Yes

softsimon:
> do we want similar headers on the block, address and transaction previews?

Yes. absolutely  It would make it consistent and give more room for block hash and TXID.

softsimon:
1. Change to Lightning Node and Lightning Channel .. capitalized titles
2. Add the node pubkey below it.. similar to the website.

mononaut:
oh, I haven't rebased this PR on the fixed master branch yet. that's probably it 😅

mononaut:
ok, here we go, consistent headers & title/subtitle sizes everywhere.

I had to shift the channel layout around a bit to fit in the long channel id, otherwise there was a lot of dead space at the top right and a very small map.

this PR is also rebased on master now.

| node | channel |
|-|-|
| node | channel |
| block | transaction |
| block | transaction  |
| address | |
| address | |

softsimon:
Looks much better now. So I will ACK this.

There is some room for improvements regarding padding/margin around certain texts and elements to make it look perfectly symmetric, but we'll get there slowly

PR #2486, opened 2022-09-01
Add docker lightning backend config

WesVleuten:
Added missing lightning configuration in the backend config for docker


docker/backend/mempool-config.json
```diff
@@ -67,6 +67,22 @@
...
+    "LOGGER_UPDATE_INTERVAL": __LIGHTNING_LOGGER_UPDATE_INTERVAL__,
+    "TOPOLOGY_FOLDER": "__LIGHTNING_TOPOLOGY_FOLDER__"
+  },
+  "LND": {
+    "TLS_CERT_PATH": "__LND_TLS_CERT_PATH__",
+    "MACAROON_PATH": "__LND_MACAROON_PATH__",
+    "REST_API_URL": "__LND_REST_API_URL__"
+  },
+  "CLIGHTNING": {
+    "SOCKET": "__CLN_SOCKET__"
```
knorrium, line 84:
```suggestion
    "SOCKET": "__CLIGHTNING_SOCKET__"
```

knorrium (changes requested):

docker/backend/mempool-config.json
WesVleuten: `__CLN_SOCKET__` is defined in start.sh which has been approved before. Your suggestion would require a change in start.sh as well, is that preferred?

https://github.com/mempool/mempool/blob/master/docker/backend/start.sh#L108

knorrium: Yes. `__CLN_SOCKET__`  was a mistake because the key in the config.json file is under `CLIGHTNING`. Ideally we should keep all the keys in sync. There is one exception (`BACKEND_HTTP_PORT`) that I'm going to fix before we tag the next release.

WesVleuten:
Key has been updated!

knorrium (approved):
utACK

Issue #2492, opened 2022-09-02
Lightning: ISP View - City column should probably be called Location instead


knorrium:
Since we can't provide the exact location for all nodes, consider changing the label to a more generic term "Location".

![Screenshot from 2022-09-01 21-21-29](https://user-images.githubusercontent.com/100320/188058695-35e82022-9cda-48ec-880d-da78a80e44bc.png)

nymkappa:
Oh yeah, before it was only the city and I forgot to change the label.

PR #2500, opened 2022-09-02
Renamed "City" to "Location"

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2492
I also use the new component so it's consistent with the isp nodes list.

**For example**

Master
image

PR
Screen Shot 2022-09-02 at 10 02 57 AM


wiz (approved):
Tested ACK @ v2.5.0-dev [[01588305](https://github.com/mempool/mempool/commit/01588305)]

Also, just to double check, but I assume you're intentionally omitting the flag emoji on the country page because it would be the same flag for all locations, right?

PR #2506, opened 2022-09-03
Fix sed command for the pools json urls

knorrium:
These were causing the following errors during the backend startup:

```
sed: -e expression #1, char 103: unterminated `s' command
sed: -e expression #1, char 102: unterminated `s' command
```


wiz (approved):
utACK

Issue #2503, opened 2022-09-02
Missing `cltv_expiry_delta` on lightning channels


t-bast:
Great job on the lightning integration, this is slick! I noticed that you seem to be missing the `cltv_expiry_delta` on channels though, it appears as `0 blocks` for all channels.

![cltv_delta](https://user-images.githubusercontent.com/31281497/188108284-00348d9c-d155-49b7-a90e-fa047059a541.png)

PR #2507, opened 2022-09-04
When using clightning, use listchannels.delay as cltv_delta

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2503

image


wiz (approved):
Tested ACK @ v2.5.0-dev [[3a633754](https://github.com/mempool/mempool/commit/3a633754)]

Screen Shot 2022-09-10 at 1 26 48

Issue #2498, opened 2022-09-02
Lightning: Node map doesn't load when there's no channel data


knorrium:
`/lightning/node/03aaa60766051e209bac196b5f8dad55fe00388e29480639935fa5443f4fe79cde` 

![Screenshot from 2022-09-01 22-08-27](https://user-images.githubusercontent.com/100320/188063124-b310f652-763d-418e-9614-b23ad882e3d8.png)

nymkappa:
Duplicate of https://github.com/mempool/mempool/pull/2469

PR #2508, opened 2022-09-04
Don't show channel map is there is no active channels

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2498


softsimon:
Is this PR still relevant/needed? Given the pointed issues are closed.

Issue #2493, opened 2022-09-02
Lightning: Add hover with full location on truncated entries


knorrium:
We are truncating the location in some cases, so add a hover to display the full location name ("North Charleston, SC")
![Screenshot from 2022-09-01 21-25-29](https://user-images.githubusercontent.com/100320/188059038-94a1b3c2-0b9c-4ba5-9612-77de70b6a319.png)

PR #2509, opened 2022-09-04
Show tooltip on location is truncated

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2493
May conflict with https://github.com/mempool/mempool/pull/2500

image

image


knorrium (approved):
Tested ACK @ v2.5.0-dev [[b53bd514](https://github.com/mempool/mempool/commit/b53bd514)]

However I think that now that the tooltip works, we could show City and State on the Nodes per country table

Issue #2490, opened 2022-09-01
Make green badge for zero base fee channels


softsimon:
On the channels page. Add a "feature" badge after the Base Fee .. and it should be GREEN when the channel is zero fee base fee and yellow if not.

nymkappa:
Following the same idea we can create a zero base fee node list.
You could either access it through the dashboard (a new widget) or when you click on the badge described in this feature request.

softsimon:
> Following the same idea we can create a zero base fee node list. You could either access it through the dashboard (a new widget) or when you click on the badge described in this feature request.

The Top 100 node list should have a Zero base fee toggle that grey out the ones that are not :)

PR #2510, opened 2022-09-04
Show zero base fee tag on channels

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2490
Note that I used red instead of yellow.

### Desktop

image

image

### Mobile

image

image

Note that I've shortened huge values, also showing `B` instead of `G` for money units like sats, maybe other units also need to be changed.


softsimon (changes requested):
tested NACK

* The tooltip claims that there is no base fee when there is. 
* Spelling should probably be "charge* any base fee.."
* Consider hiding the tag completely when there are null / - value base fee.

Channel ID: 827304434701107200

Screen Shot 2022-09-08 at 14 14 44

nymkappa:
> tested NACK
> 
>     * The tooltip claims that there is no base fee when there is.
> 
>     * Spelling should probably be "_charge_ any base fee.."
> 
>     * Consider hiding the tag completely when there are null / - value base fee.
> 
> 
> Channel ID: 827304434701107200
> 

Rebased on master and fixed in https://github.com/mempool/mempool/pull/2510/commits/ef873f48052d4362cace1a1a7aa8a1bc172e9e5f

image
image

softsimon:
Rebase please

nymkappa:
image

image

wiz:
Shouldn't the red tag say "non-zero base fee" ?

softsimon:
> Shouldn't the red tag say "non-zero base fee" ?

We don't say "No RBF", "No Taproot" or "No Segwit"  just "RBF" etc, but in red .. So this is consistent. But I would be OK with red Non-zero base fee in this case.

wiz:
That's because everyone knows RBF is good and desirable. But hardly anyone knows if zero base fee is good or not.

nymkappa:
- Renamed to `Non-zero base fee`
- Also, I show the base fee in msats because that's you configure it in your LN node. This also fixes the issue where we show the red badge and display "0 sats" due to the rounding error, when the base fee is very low yet not 0

image
image

softsimon (approved):
tested ACK @ v2.5.0-dev [[2c370ffc](https://github.com/mempool/mempool/commit/2c370ffc)]

Issue #2479, opened 2022-08-31
Make channel closing date more prominent


softsimon:
**A)**

In the Closed Channels list. Replace the Fee Rate with Closed Date and maybe only show it humanly readable like "< 1 hour ago"

**B)**

On a closed channels page, make it more apparent when the channel was closed on top by adding Closed date: <date> <date ago>

**C)**

When fee rate is missing, instead write `N/A` on the channels page

nymkappa:
C is fixed in https://github.com/mempool/mempool/pull/2504

PR #2512, opened 2022-09-04
If a channel is closed, show closing date instead of last update

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2479

**If the channel is closed, show closing date instead of last update**

image

**If we currently browser closed channel list, show closing date instead of fee rate**

image

image


softsimon (approved):
tested ACK @ v2.5.0-dev [[5389928c](https://github.com/mempool/mempool/commit/5389928c)]

PR #2513, opened 2022-09-04
UID/GID configurable with ARGs when building docker images

Oberfeldwedler:
Makes it possible to configure the UID/GID via build Args when building images. If not specified it defaults to UID/GID=1000.


knorrium:
Tested ACK

Building locally with `--build-arg UID=1001 --build-arg GID=1001`:

Frontend
```
/patch $ id -g
1001
/patch $ id -u
1001
```

Backend
```
$ id -u
1001
$ id -g
1001
```

knorrium (approved):

docker/backend/Dockerfile
```diff
@@ -12,14 +15,18 @@ RUN npm install --omit=dev --omit=optional
...
-RUN chown 1000:1000 ./
-COPY --from=builder --chown=1000:1000 /build/package ./package/
-COPY --from=builder --chown=1000:1000 /build/mempool-config.json /build/start.sh /build/wait-for-it.sh ./
+RUN chown $UID:$GID ./
+COPY --from=builder --chown=$UID:$GID /build/package ./package/
+COPY --from=builder --chown=$UID:$GID /build/mempool-config.json /build/start.sh /build/wait-for-it.sh ./
+
 
-USER 1000
+RUN usermod -u $UID -o node && groupmod -g $GID -o node
```
wiz, line 28: Why is this needed?

docker/frontend/Dockerfile
```diff
@@ -24,15 +29,19 @@ COPY --from=builder /build/nginx-mempool.conf /etc/nginx/conf.d/
...
+        chown -R $UID:$GID /var/log/nginx && \
+        chown -R $UID:$GID /etc/nginx/nginx.conf && \
+        chown -R $UID:$GID /etc/nginx/conf.d
 RUN touch /var/run/nginx.pid && \
-        chown -R 1000:1000 /var/run/nginx.pid
+        chown -R $UID:$GID /var/run/nginx.pid
+
+RUN deluser nginx \
+    && addgroup -g $GID -S nginx \
+    && adduser -S -D -H -u $UID -h /var/cache/nginx -s /sbin/nologin -G nginx -g nginx nginx
```
wiz, line 42: Why is this needed?

Oberfeldwedler: Alpine comes with no way that I know of, to modify the UID/GID of an existing user, besides directly changing config-files. 
In the backend-image I used `usermod`, which exists in Debian but not Alpine.
As I don't want to write directly to config-files, I simply delete the old user and create a new one with the correct IDs.

docker/backend/Dockerfile
Oberfeldwedler: When you look at the Dockerfile of your backend base image `node:16.16.0-buster-slim`, you will notice that the user `node` is created with default IDs.

This user has to be modified, so the process called at the Entrypoint can run under the user specified when building.

```
FROM debian:buster-slim

RUN groupadd --gid 1000 node \
  && useradd --uid 1000 --gid node --shell /bin/bash --create-home node
...

```

Oberfeldwedler:
What do I have to do to make that sign message go away? 
I created the file like it was specified in `CONTRIBUTING.md`.

> Thanks for your contribution to The Mempool Open Source Project. We require all contributors to agree to our Contributor >License Agreement and don't seem to have one for you on file. Please see the instructions in CONTRIBUTING.md to proceed >with this pull request.

wiz:
Sorry for the delay I will review and merge your PR soon. The CLA error will go away after merging

docker/backend/Dockerfile
wiz: I think we should enclose $UID and $GID in quotes like this `"${UID}"`

docker/frontend/Dockerfile
```diff
@@ -24,15 +29,19 @@ COPY --from=builder /build/nginx-mempool.conf /etc/nginx/conf.d/
...
-        chown -R 1000:1000 /etc/nginx/nginx.conf && \
-        chown -R 1000:1000 /etc/nginx/conf.d
+RUN chown -R $UID:$GID /patch && chmod -R 755 /patch && \
+        chown -R $UID:$GID /var/cache/nginx && \
+        chown -R $UID:$GID /var/log/nginx && \
+        chown -R $UID:$GID /etc/nginx/nginx.conf && \
+        chown -R $UID:$GID /etc/nginx/conf.d
 RUN touch /var/run/nginx.pid && \
-        chown -R 1000:1000 /var/run/nginx.pid
+        chown -R $UID:$GID /var/run/nginx.pid
```
wiz, line 38: I think we should enclose $UID and $GID in quotes like this `"${UID}"`

docker/backend/Dockerfile
```diff
@@ -12,14 +15,18 @@ RUN npm install --omit=dev --omit=optional
...
+ARG GID
 
 WORKDIR /backend
 
-RUN chown 1000:1000 ./
-COPY --from=builder --chown=1000:1000 /build/package ./package/
-COPY --from=builder --chown=1000:1000 /build/mempool-config.json /build/start.sh /build/wait-for-it.sh ./
+RUN chown $UID:$GID ./
+COPY --from=builder --chown=$UID:$GID /build/package ./package/
+COPY --from=builder --chown=$UID:$GID /build/mempool-config.json /build/start.sh /build/wait-for-it.sh ./
```
wiz, line 25: I think we should enclose $UID and $GID in quotes like this `"${UID}"`

docker/backend/Dockerfile
Oberfeldwedler: `${UID}` and `${GID}` work fine. Tested with `docker build --build-arg UID=1001 --build-arg GID=1001 .`

What would be the benefit of changing it to `${UID}`? 

I checked [Dockerfile Reference](https://docs.docker.com/engine/reference/builder/#using-arg-variables). They didn't use this syntax either.

Oberfeldwedler:
What is still to be done here? Can we get this merged? 

> I think we should enclose /$UID and /$GID in quotes like this "${UID}"

I do not think this is necessary, but if you insist I can change it anyway although imo it does nothing.

wiz:
Yes, please add the quotes.
Can you explain why you need to be able to configure the UID / GID ?

PR #2514, opened 2022-09-05
Feature: Add endpoint for PSBT nonWitnessUtxo inclusion

junderw:
# TODO
- [ ] Add API document.
- [ ] Add UI for posting to this endpoint. (Maybe similar to https://mempool.space/tx/push)

# Feature
Takes the PSBT as text/plain body, parses it, and adds the full parent transaction to each input that doesn't already have it. This is used for BTCPayServer / Trezor users which need access to the full parent transaction even with segwit inputs. It will respond with a text/plain PSBT in the same format (hex|base64).

Requested by @Kukks of BTCPay


backend/src/api/bitcoin/bitcoin.routes.ts
```diff
@@ -241,6 +243,48 @@ class BitcoinRoutes {
...
+          const transaction: IEsploraApi.Transaction = await bitcoinApi.$getRawTransaction(txid, true);
+          if (!transaction.hex) {
+            throw new Error(`Couldn't get transaction hex for ${txid}`);
+          }
+          psbt.updateInput(index, {
+            nonWitnessUtxo: Buffer.from(transaction.hex, 'hex'),
+          });
+        }
+      }
+      res.send(format === 'hex' ? psbt.toHex() : psbt.toBase64());
```
knorrium, line 282: @junderw I don't know if there's a desire to indicate whether the input has been altered but without more context I find it a bit weird that an `/addparents` endpoint returns the same result and doesn't actually add any parents.

```
curl -vvv --request POST \
  --url http://localhost:8999/api/v1/psbt/addparents \
  --header 'Content-Type: text/plain' \
  --data 'cHNidP8BAHUCAAAAASaBcTce3/KF6Tet7qSze3gADAVmy7OtZGQXE8pCFxv2AAAAAAD+////AtPf9QUAAAAAGXapFNDFmQPFusKGh2DpD9UhpGZap2UgiKwA4fUFAAAAABepFDVF5uM7gyxHBQ8k0+65PJwDlIvHh7MuEwAAAQD9pQEBAAAAAAECiaPHHqtNIOA3G7ukzGmPopXJRjr6Ljl/hTPMti+VZ+UBAAAAFxYAFL4Y0VKpsBIDna89p95PUzSe7LmF/////4b4qkOnHf8USIk6UwpyN+9rRgi7st0tAXHmOuxqSJC0AQAAABcWABT+Pp7xp0XpdNkCxDVZQ6vLNL1TU/////8CAMLrCwAAAAAZdqkUhc/xCX/Z4Ai7NK9wnGIZeziXikiIrHL++E4sAAAAF6kUM5cluiHv1irHU6m80GfWx6ajnQWHAkcwRAIgJxK+IuAnDzlPVoMR3HyppolwuAJf3TskAinwf4pfOiQCIAGLONfc0xTnNMkna9b7QPZzMlvEuqFEyADS8vAtsnZcASED0uFWdJQbrUqZY3LLh+GFbTZSYG2YVi/jnF6efkE/IQUCSDBFAiEA0SuFLYXc2WHS9fSrZgZU327tzHlMDDPOXMMJ/7X85Y0CIGczio4OFyXBl/saiK9Z9R5E5CVbIBZ8hoQDHAXR8lkqASECI7cr7vCWXRC+B3jv7NYfysb3mk6haTkzgHNEZPhPKrMAAAAAAAAA'
Note: Unnecessary use of -X or --request, POST is already inferred.
*   Trying 127.0.0.1:8999...
* Connected to localhost (127.0.0.1) port 8999 (#0)
> POST /api/v1/psbt/addparents HTTP/1.1
> Host: localhost:8999
> User-Agent: curl/7.81.0
> Accept: */*
> Content-Type: text/plain
> Content-Length: 740
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< X-Powered-By: Express
< Access-Control-Allow-Origin: *
< Content-Type: text/plain; charset=utf-8
< Content-Length: 740
< ETag: W/"2e4-OLYF1fJM5WHxA45fGZ+2IKi2e1c"
< Date: Sat, 10 Sep 2022 06:08:47 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< 
* Connection #0 to host localhost left intact
cHNidP8BAHUCAAAAASaBcTce3/KF6Tet7qSze3gADAVmy7OtZGQXE8pCFxv2AAAAAAD+////AtPf9QUAAAAAGXapFNDFmQPFusKGh2DpD9UhpGZap2UgiKwA4fUFAAAAABepFDVF5uM7gyxHBQ8k0+65PJwDlIvHh7MuEwAAAQD9pQEBAAAAAAECiaPHHqtNIOA3G7ukzGmPopXJRjr6Ljl/hTPMti+VZ+UBAAAAFxYAFL4Y0VKpsBIDna89p95PUzSe7LmF/////4b4qkOnHf8USIk6UwpyN+9rRgi7st0tAXHmOuxqSJC0AQAAABcWABT+Pp7xp0XpdNkCxDVZQ6vLNL1TU/////8CAMLrCwAAAAAZdqkUhc/xCX/Z4Ai7NK9wnGIZeziXikiIrHL++E4sAAAAF6kUM5cluiHv1irHU6m80GfWx6ajnQWHAkcwRAIgJxK+IuAnDzlPVoMR3HyppolwuAJf3TskAinwf4pfOiQCIAGLONfc0xTnNMkna9b7QPZzMlvEuqFEyADS8vAtsnZcASED0uFWdJQbrUqZY3LLh+GFbTZSYG2YVi/jnF6efkE/IQUCSDBFAiEA0SuFLYXc2WHS9fSrZgZU327tzHlMDDPOXMMJ/7X85Y0CIGczio4OFyXBl/saiK9Z9R5E5CVbIBZ8hoQDHAXR8lkqASECI7cr7vCWXRC+B3jv7NYfysb3mk6haTkzgHNEZPhPKrMAAAAAAAAA
```

backend/src/api/bitcoin/bitcoin.routes.ts
```diff
@@ -241,6 +243,48 @@ class BitcoinRoutes {
...
+    } catch (e: any) {
+      res.status(500).send(e instanceof Error ? e.message : e);
```
knorrium, line 284: @junderw For a case like `No such mempool or blockchain transaction. Use gettransaction for wallet transactions.`, maybe a 400 status code would be better than 500.

junderw: Fixed. I return a 404 status code to indicate the data was missing.

backend/src/api/bitcoin/bitcoin.routes.ts
junderw: It now returns a 422 status code ([422 Unprocessable Entity](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422)) with the text `Psbt had no missing nonUtxoWitnesses.`

backend/src/api/bitcoin/bitcoin.routes.ts
```diff
@@ -241,6 +243,74 @@ class BitcoinRoutes {
...
+            if (!transaction.hex) {
+              throw new Error('');
+            }
+          } catch (err) {
+            throw new Error(`${notFoundError} #${index} @ ${txid}`);
+          }
+
+          psbt.updateInput(index, {
+            nonWitnessUtxo: Buffer.from(transaction.hex, 'hex'),
+          });
```
antonilol, line 291: according to bip 174, for witness inputs the witnessUtxo must be filled in, and for legacy inputs the nonWitnessUtxo, but also allows witness inputs to have nonWitnessUtxo.

when i `walletprocesspsbt` a psbt with nonWitnessUtxo and no witnessUtxo for a witness input (from electrum), bitcoind fills in the witnessUtxo, and strips the input witness from nonWitnessUtxo. This way wallets that need witnessUtxo can read it, wallets that dont understand segwit dont encounter witness data in nonWitnessUtxo and wallets that check the txid can still do this because witness data does not affect the txid. i thinks it is best to stay in line with bitcoind's implementation

backend/src/api/bitcoin/bitcoin.routes.ts
```diff
@@ -241,6 +243,74 @@ class BitcoinRoutes {
...
+        psbt = bitcoinjs.Psbt.fromBase64(req.body);
+        format = 'base64';
+      } catch (e1) {
+        try {
+          psbt = bitcoinjs.Psbt.fromHex(req.body);
+          format = 'hex';
+        } catch (e2) {
+          throw new Error(`Unable to parse PSBT`);
+        }
+      }
```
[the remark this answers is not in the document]
junderw: Yes, but there are other parsing errors that can occur other than the headers. Adding a header check is superfluous since these check both the headers and everything else.

backend/src/api/bitcoin/bitcoin.routes.ts
junderw: Thoughts @NicolasDorier ?

This feature is a request from BTCPayServer, so I delegate this design decision to them.

backend/src/api/bitcoin/bitcoin.routes.ts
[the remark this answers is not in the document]
antonilol: true, it will work this way

(if error messages from bitcoinjs would be sent back to the client this might get confusing. a base64 psbt that can't get parsed as base64 will be parsed as hex, which will obviously fail. which error is sent to the client, `e1` or `e2`?)

backend/src/api/bitcoin/bitcoin.routes.ts
NicolasDorier: I didn't know that some implementation were removing the witness from nonWitnessUTXO. I never did it and never got problems, I guess this is because any wallet sufficiently advanced to know about PSBT already know what is witness data. I would prefer you don't strip the witness it as I don't have experience on what happen on hardware wallets. (btw, having both witnessUTXO and nonWitnessUTXO wasn't supported by some wallet at one point, but I guess it's ok now)
I also prefer not stripping witness data, because I plan to save the transaction in our cache, and I don't expect not having the full tx.

An idea if you can't get your mind on that would be to allow query parameters to the URL to control this, so on BTCPay side we would be able to redirect users on exactly the right way to do the PSBT. (for example, if we know a wallet don't support both witnessUTXO and nonWitnessUTXO we could redirect on the form where witnessUTXO isn't filled)

But this is maybe too much for this PR at this point, better to experiment with it and act on the feedback.

softsimon (changes requested):
It doesn't work with blockstream/elects set as backend because the transactions returned doesn't contain the '`transaction.hex`' field.

Use the API to fetch a transaction as HEX instead.

softsimon (approved):
tested ACK @ 74dbd6ce

PR #2519, opened 2022-09-06
Show tor+clearnet node series in chart

nymkappa:
In a previous PR, I've split tor only nodes and hybrid nodes (tor + clearnet), but I did not updated the chart. So the LN dashboard `Nodes` count does not match the `Total` in the chart:

image

This PR adds a new line series to the node network charts to also show the share of nodes running on both Tor and clearnet at the same time, so that the `Total` count matches the one in the LN dashboard.

image


softsimon (changes requested):
No legend to show/hide Hybrid nodes

Screen Shot 2022-09-08 at 16 03 30

wiz:
Concept NACK for the phrase "Clearnet & Tor" being ambiguous: it can be interpreted as either "the sum of (nodes that have clearnet connectivity) and (nodes that have Tor connectivity)" or "nodes that have both clearnet and Tor connectivity" 

Instead how about thinking of a phrase that best describes nodes with such connectivity, for example within clearnet we refer to nodes having both IPv4 and IPv6 as "dual stack" - maybe there is a similar term for nodes that have both clearnet and Tor connectivity

wiz:
Maybe something like "Clearnet only", "Tor only", "Hybrid clearnet/tor"

wiz:
Maybe something like:
* Clearnet-only nodes
* Tor-only nodes
* Hybrid clearnet+tor nodes

hunicus:
Another suggestion:
- Only reachable on Tor
- Only reachable on clearnet
- Reachable on Tor & clearnet

wiz:
Just like we use "Clearnet" to group IPv4/IPv6, we should probably use its opposite "Darknet" to group Tor/I2P/CJDNS... then we can say "Clearnet capacity" and "Darknet capacity" or as you suggested:

* Reachable on Clearnet Only
* Reachable on Darknet Only
* Reachable on Clearnet and Darknet

nymkappa:
Screen Shot 2022-09-22 at 4 09 37 PM
Screen Shot 2022-09-22 at 4 09 51 PM

wiz (approved):
Tested ACK @ v2.5.0-dev [[cce82c12](https://github.com/mempool/mempool/commit/cce82c12)]

PR #2521, opened 2022-09-06
Fix preview tx diagram for zero value transactions

mononaut:
[Some Liquid coinbase transactions](https://liquid.network/tx/1a015575cd3d46c086efd84d1e516de05b40d41ac1e61714bff383451be24f8c) can have zero value, which breaks the visualization, since technically no coherent diagram exists for such a transaction.

This PR improves the appearance of that edge case, by drawing the inputs/outputs with equal weight.

| Before | After |
|-|-|
| before | after |


wiz (approved):
Tested ACK @ v2.5.0-dev [[c92fcd20](https://github.com/mempool/mempool/commit/c92fcd20)]

PR #2534, opened 2022-09-09
Removing extra capacity row

softsimon:


wiz (approved):
Tested ACK @ v2.5.0-dev [[a8188a35](https://github.com/mempool/mempool/commit/a8188a35)]

PR #2541, opened 2022-09-10
Refactor preview routes into separate module

mononaut:
This PR moves the various link preview components & pages to their own routing module, which should reduce the amount of repetition and clutter in the main routing modules, especially as the number of different previews grows.


softsimon:
Nice. I guess all the preview related code will be lazy loaded now and not integrated in the main bundle.

wiz (approved):
Tested ACK @ v2.5.0-dev [[da8044c0](https://github.com/mempool/mempool/commit/da8044c0)]

PR #2542, opened 2022-09-10
Add Lightning ISP preview

mononaut:
This PR adds a new link preview for the Lightning ISP page (`/lightning/nodes/isp/:asn`)

There's some dead space to the upper right, because the fully zoomed out map doesn't look good at taller aspect ratios. Maybe this could be used for ISP logos in future?

amazon-isp


mononaut:
@nymkappa this touches the `nodes-map` component to make it reactive to `nodes` input changes (needed for the unfurler).

I don't think that breaks any of the other pages that use the map, but you might want to double check :)

wiz (approved):
Tested ACK @ v2.5.0-dev [[047843b1](https://github.com/mempool/mempool/commit/047843b1)]

PR #2543, opened 2022-09-10
Adjust block preview layout

mononaut:
Adjusts the block preview layout to make the block hash more prominent

| before | after |
|-|-|
| hash-before | hash-after |

### DIFF
--- a/frontend/src/app/components/block/block-preview.component.html
+++ b/frontend/src/app/components/block/block-preview.component.html
@@ -10,11 +10,10 @@
             <ng-template [ngIf]="blockHeight === 0"><ng-container i18n="@@2303359202781425764">Genesis</ng-container></ng-template>
             <ng-template [ngIf]="blockHeight" i18n="shared.block-title">{{ blockHeight }}</ng-template>
           </h1>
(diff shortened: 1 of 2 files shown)


softsimon (approved):
utACK

wiz:
How does this look on Liquid with 7 digit block hashes or how high can we go?

mononaut:
good point, the rightmost digits were clipped awkwardly at higher block heights.

I just pushed a change to truncate dynamically in CSS instead, which should also use the space better at lower block heights. It means we gain an ellipsis on the top line, and lose perfect character alignment

| liquid-block | bitcoin-short-block |
|-|-|

wiz (approved):
Tested ACK @ v2.5.0-dev [[deae7b28](https://github.com/mempool/mempool/commit/deae7b28)]

Looks good but there seems to be extra space where the block hash used to be, below the block height / above the table rows, maybe we should move that up a little? Also, currently lightning node pubkeys are blue, should the block hash also be blue or should we change the lightning pubkey to be white? I think the blue looks cool for a pubkey/hash, @softsimon wdyt?

PR #2548, opened 2022-09-12
Suggest block height in search result

softsimon:
The problem now on master is that if you want to go to a block height like "5" and click Enter you will be taken to the first search result which is a Lightning Node.

Screen Shot 2022-09-12 at 19 20 42

This PR adds matching block heights as first search result.
Screen Shot 2022-09-12 at 19 18 45


wiz (approved):
Tested ACK @ v2.5.0-dev [[b6296fcb](https://github.com/mempool/mempool/commit/b6296fcb)]

PR #2550, opened 2022-09-13
Add python example for websocket api docs

hunicus:
Shows only on WebSocket API docs for Bitcoin networks.


wiz (approved):
Tested ACK @ v2.5.0-dev [[666a03ba](https://github.com/mempool/mempool/commit/666a03ba)]

PR #2551, opened 2022-09-13
Fix spacing on websocket api examples

hunicus:
This PR adjusts spacing on WebSocket API examples to look more presentable.

The PR diff looks weird so I suggest looking at the individual diffs for each commit instead.

## Current

Screenshot from 2022-09-13 07-40-33

## Proposed

Screenshot from 2022-09-13 07-40-57


wiz (approved):
utACK since just spacing changed

PR #2561, opened 2022-09-18
Optimize transaction diagram component

mononaut:
The current transaction diagram implementation naively draws an SVG line for every supplied input and output. This works fine for typical transactions, but performance suffers in extreme examples with many thousands of inputs/outputs. This could conceivably cause unfurler timeouts, and limits its suitability for directly user-facing pages.

This PR enforces a cap of 250 SVG lines on each side of the diagram, and condenses the excess into a single SVG element, without significantly affecting the appearance of the final graphic.

Example of an extreme transaction with ~13k outputs, before and after applying the optimization:
| Before | After |
|-|-|
| out-all | out-250 |
| rendering time up to 60 seconds | almost instant |

### DIFF
--- a/frontend/src/app/components/tx-bowtie-graph/tx-bowtie-graph.component.ts
+++ b/frontend/src/app/components/tx-bowtie-graph/tx-bowtie-graph.component.ts
@@ -7,6 +7,13 @@ interface SvgLine {
   class?: string;
 }
 
+interface Xput {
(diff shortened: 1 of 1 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[287756ea](https://github.com/mempool/mempool/commit/287756ea)]

Rendering time is overall improved for large TX

PR #2562, opened 2022-09-18
Tx page diagram

mononaut:
(builds on PR #2561)

This PR explores adding the sankey diagram from the transaction previews to the main transaction page.

### Summary

By default the diagram works the same way as in the previews, showing up to 24 inputs/outputs in full, with the remainder spilling off the bottom of the graphic.

If there are >24 inputs or outputs, a toggle button is shown which can expand the graphic to fit up to 250 inputs/outputs.

Hovering the mouse over a line highlights it, and displays a tooltip with a simple summary.

show-more

### Limitations

There are a few obvious areas for improvement, which I haven't focused much effort on yet:

##### Mouseover UX
The SVG lines can be as little as 1 or 2 pixels thick, so it can be awkward to select the right line because the hover event is only triggered by an exact hit.

##### Mouseover highlight effect
The current effect is mostly a placeholder. It's tricky to restyle line endcaps on hover, so I've sidestepped that for now by applying a gradient to the middle of the line instead.

##### Expand/collapse UX
A bit jumpy - we probably want to adjust the page scroll after changing the diagram height.

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -196,7 +196,7 @@
 
     <div class="box">
       <div class="graph-container" #graphContainer>
-        <tx-bowtie-graph [tx]="tx" [width]="graphWidth" [height]="graphExpanded ? (maxInOut * 15) : 360" [maxStrands]="graphExpanded ? maxInOut : 24" [network]="network"></tx-bowtie-graph>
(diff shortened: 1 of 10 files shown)


wiz:
🤩

wiz (approved):
Tested ACK @ v2.5.0-dev [[78cc33ab](https://github.com/mempool/mempool/commit/78cc33ab)]

Everything about this PR is beautiful, I love how the codes dynamically generates SVG using the Angular container....

Few suggestions tho:
* Maybe change the word "Diagram" to "Flow"
* Clicking on the diagram's inputs or outputs should probably scroll down to the inputs / outputs section and highlight the selected thing
* Perhaps being able to go to the next TX like clicking an arrow would be another cool integration

PR #2564, opened 2022-09-18
[Feature] Search for block by timestamp

### LINKED ISSUE #2555: [Feature] Search for block by timestamp
https://twitter.com/mempool/status/1570537517585408000

I think, therefore, I am.

junderw:
Fixes #2555 

## TODO

- [ ] Decide on final API
- [ ] Write API docs
- [ ] Decide frontend UX
  - [ ] Where to put it?
  - [ ] What should it look like?
  - [ ] How does the user interact?
- [ ] Write frontend code (depending on difficulty of design, will probably need assistance)

@knorrium I know you were asking about this, so what do you think about the API currently?


wiz (approved):
Tested ACK @ 48e2df3f7a2b2701ca2f5c2d784f266368c9b872

Should probably check if database is disabled and add a return for that

PR #2569, opened 2022-09-19
Clipboard buttons fix

### LINKED ISSUE #2566: Buttons with clipboard component doesn't trigger
When the clipboard component is within a button, clicking on the actual button (outside the pixels of the clippy icon) doesn't trigger a clipboard copy.

Make it so that the button itself trigger the clippy, maybe by making the clipboard component into a directive style.

softsimon:
fixes #2566

Now you can click on the whole button, not just the icon, to copy text message.

This was mostly relevant for the Copy button on the Lightning Node page.

Screen Shot 2022-09-19 at 18 22 13


wiz (approved):
Tested ACK @ v2.5.0-dev [[7c13f5d8](https://github.com/mempool/mempool/commit/7c13f5d8)]

Issue #2556, opened 2022-09-16
Lightning Liquidity ranking page doesn't handle Tor nodes properly


wiz:
https://mempool.space/signet/lightning/nodes/rankings/liquidity
<img width="126" alt="Screen Shot 2022-09-17 at 2 48 53" src="https://user-images.githubusercontent.com/232186/190701486-fd23f155-258c-4823-858d-21fe15a7ce0e.png">

PR #2573, opened 2022-09-22
Fix "undefined" location in node ranking

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2556

Shows:
* Flag + city OR
* Flag + country OR
* `-` when we don't know

Screen Shot 2022-09-22 at 3 11 35 PM


softsimon (approved):
tested ACK @ v2.5.0-dev [[387cebeb](https://github.com/mempool/mempool/commit/387cebeb)]

PR #2578, opened 2022-09-24
Revert info button color change

softsimon:


wiz (approved):
Tested ACK @ v2.5.0-dev [[16cb3de2](https://github.com/mempool/mempool/commit/16cb3de2)]

Details button is back to correct color but loading spinners are now purple, not sure if that was intention or not but will merge for now

PR #2587, opened 2022-09-29
Handle network url ending matching better

softsimon:
Before
`/testnethello` would be interpreted as testnet.

After
only
`/testnet`
and
`/testnet/`
`/testnet/***`
will match to testnet

Make sure to test with and without language so nothing broke.


mononaut:
Tested ACK @ v2.5.0-dev [4154d308]

Works great. Tested on bitcoin & liquid, with and without language in the url, and nothing seems to be broken in the main site or the unfurler.

mononaut (approved):
Tested ACK @ v2.5.0-dev [4154d308]

Works great. Tested on bitcoin & liquid, with and without language in the url, and nothing seems to be broken in the main site or the unfurler.

PR #2590, opened 2022-09-29
Move logo to LHS on mining pool preview

mononaut:
Moves logos to the left hand side of the pool name on the mining pool preview, to better match the main page.

| before | after |
|-|-|
| pool-before | pool-after |


wiz (approved):
utACK

PR #2591, opened 2022-09-29
Toggle option for transaction flow diagram with query param

mononaut:
This PR adds toggle buttons to completely hide/show the new flow diagram on the transaction page, and enables a new url query parameter `showFlow` to provide control over this in links.

The flow diagram is shown by default. Adding `?showFlow=false` to a transaction url will make the diagram hidden.

flow-shown

flow-hidden

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -190,39 +190,55 @@
 
     <br>
 
-    <div class="title">
-      <h2 i18n="transaction.flow|Transaction flow">Flow</h2>
-    </div>
-
-    <div class="box">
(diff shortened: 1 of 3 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[619a6bd3](https://github.com/mempool/mempool/commit/619a6bd3)]

Works well, but I think we could polish the button strings, i.e. instead of "Show flow diagram" something more brief like simply "Details" - @hunicus any suggestions?

hunicus:
I think making button texts "Hide Diagram" and "Show Diagram" should work fine. The section titles explain the rest ("Flow" and "Inputs & Outputs").

PR #2594, opened 2022-09-30
Add a time turner toggle

### LINKED ISSUE #2588: [UI] Reverse the arrow of time in the block view
### Description

Currently, the arrow of time is pointing to the left in the block view. This is counter-intuitive, as time normally flows in the other direction.

image

### Problem to be solved

The more intuitive way of reading the arrow of time would be to go to the right. This is in alignment with the other graphs.

### Proposed solution

Put mined blocks to the left of the divider, and future blocks (mempool) on the right side.

Ugly mock-up:

image

#### Additional info

For right-to-left reading countries, this could be a switchable toggle.

Idea is from @dergigi, see https://twitter.com/dergigi/status/1575214771846402048.

mononaut:
This PR adds a small toggle to "reverse the arrow of time" (per #2588) and switch between left-to-right and right-to-left block layouts.

The default direction remains right-to-left, but the user's preference is saved to localStorage and so should be persistent across visits.

https://user-images.githubusercontent.com/83316221/193168329-6526f1fa-c8ea-4680-8fe3-07f4a239bb13.mp4

#### Blockchain blocks

The entire component is reversed, including projected block gradients and the marker arrow. 

Where the mined/projected divider is usually not centered (i.e. on Liquid and on small screens), the position of the divider is mirrored.

| RTL | LTR |
|-|-|
| mobile-block-rtl | mobile-block-ltr |

The left-to-right layout is achieved mainly through css transforms of the default layout.
1. This gives us nice smooth compositor-only transitions, which should be much more performant than e.g. changing absolute positions.
2. It simplifies the layout logic considerably.

#### Block navigation

The next/previous block navigation arrows and keyboard shortcuts are swapped.

#### Projected block visualization

Transactions are aligned to match the mini block at the top of the page, and the transition between directions is animated as usual.

The directions of transitions between projected blocks, and transaction entry/exit animations are also reversed.

---

(resolves #2588)

### DIFF
--- a/frontend/src/app/components/block-overview-graph/block-overview-graph.component.ts
+++ b/frontend/src/app/components/block-overview-graph/block-overview-graph.component.ts
@@ -1,4 +1,4 @@
-import { Component, ElementRef, ViewChild, HostListener, Input, Output, EventEmitter, NgZone, AfterViewInit, OnDestroy } from '@angular/core';
+import { Component, ElementRef, ViewChild, HostListener, Input, Output, EventEmitter, NgZone, AfterViewInit, OnDestroy, OnChanges } from '@angular/core';
 import { TransactionStripped } from '../../interfaces/websocket.interface';
 import { FastVertexArray } from './fast-vertex-array';
 import BlockScene from './block-scene';
@@ -11,7 +11,7 @@ import { Position } from './sprite-types';
   templateUrl: './block-overview-graph.component.html',
   styleUrls: ['./block-overview-graph.component.scss'],
 })
-export class BlockOverviewGraphComponent implements AfterViewInit, OnDestroy {
+export class BlockOverviewGraphComponent implements AfterViewInit, OnDestroy, OnChanges {
   @Input() isLoading: boolean;
   @Input() resolution: number;
   @Input() blockLimit: number;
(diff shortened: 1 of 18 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[15caef10](https://github.com/mempool/mempool/commit/15caef10)]

Great work on turning back time :)
Few minor issues:
* If you turn back time, then you can no longer drag (on mobile) or scroll horizontally (on desktop like macbook where you have a trackpad) to scroll the timechain - however since it still works fine on the current orientation I think it's fine to merge as-is for now and fix this in a future PR later
* For locales that use RTL like ar or fa, we should also default to RTL timechain unless the user has a preference overriding it
* There is a JS console error:
```
ERROR DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty.
```

wiz:
Since the JS console error doesn't seem to affect anything I'll go ahead and merge now, but it would be nice if @mononaut or @softsimon can fix it today

dergigi:
Not all is lost. At least in this small corner of the universe the arrow of time has been restored, history goes from left to right, the future follows the present, and everything—finally—makes sense again. Thank you 🙏🧡

PR #2595, opened 2022-09-30
Fix for add class js error

softsimon:


wiz (approved):
Tested ACK @ v2.5.0-dev [[f2889fc0](https://github.com/mempool/mempool/commit/f2889fc0)]

PR #2601, opened 2022-10-03
Fix transaction marker boundary condition

### LINKED ISSUE #2598: The arrow should point to where in the mempool the TX is located
This is a high fee transaction and the arrow should be furthest to the right.
The arrow should only be within the colored area of transactions.

Screen Shot 2022-09-30 at 16 43 50

mononaut:
Fixes #2598

### Issue

Caused by a bad boundary condition when finding the position of a tx within the fee ranges of the projected blocks.

We were looking for a range where `min <= feerate < max`, but for the highest effective feerate tx in each projected block, `feerate == max` so no matching range is found, and the marker arrow defaults to the center of the selected block.

This can also happen if a transaction has an effective fee rate that lies between the maximum fee of one projected block and the minimum fee of the next. In theory this shouldn't be possible, but presumably sometimes the two sources of data (the mempool block fee ranges, and effective rates calculated from `getCPFPinfo` data) don't perfectly align.

### Steps to reproduce

Go to /mempool-block/0 and select the highest effective feerate transaction (usually the very top-right corner, unless that tx has lower rate descendants). The marker arrow incorrectly displays in the center of the block, instead of the far right hand side.

### DIFF
--- a/frontend/src/app/components/mempool-blocks/mempool-blocks.component.ts
+++ b/frontend/src/app/components/mempool-blocks/mempool-blocks.component.ts
@@ -287,11 +287,12 @@ export class MempoolBlocksComponent implements OnInit, OnDestroy {
 
     this.arrowVisible = true;
 
-    for (const block of this.mempoolBlocks) {
-      for (let i = 0; i < block.feeRange.length - 1; i++) {
+    let found = false;
+    for (let txInBlockIndex = 0; txInBlockIndex < this.mempoolBlocks.length && !found; txInBlockIndex++) {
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK @ v2.5.0-dev [[1af38456](https://github.com/mempool/mempool/commit/1af38456)]

wiz (approved):
Tested ACK @ v2.5.0-dev [[6005bbea](https://github.com/mempool/mempool/commit/6005bbea)]

PR #2603, opened 2022-10-03
Reverse time by default for RTL languages

mononaut:
Reverses the default time direction for RTL language users, so that the mempool blocks are at the 'start' of the line, like they are for other languages in the original layout.

_(remember to clear localStorage before testing, otherwise the default direction will be overridden by any cached preference)_


wiz (approved):
Tested ACK @ v2.5.0-dev [[66c99e2f](https://github.com/mempool/mempool/commit/66c99e2f)]

PR #2605, opened 2022-10-04
Transaction diagram interactivity & navigation

mononaut:
This PR adds click events to the input/output lines in the transaction flow diagram.

When clicked, the corresponding entry in the table is highlighted & scrolled into view (and the table extended to make it visible, if necessary)

https://user-images.githubusercontent.com/83316221/193927821-963cd6b7-f82b-49bb-aeae-325c3975870f.mp4

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -208,7 +208,10 @@
             [lineLimit]="inOutLimit"
             [maxStrands]="graphExpanded ? maxInOut : 24"
             [network]="network"
-            [tooltip]="true">
+            [tooltip]="true"
+            (selectInput)="selectInput($event)"
+            (selectOutput)="selectOutput($event)"
+          >
           </tx-bowtie-graph>
         </div>
         <div class="toggle-wrapper" *ngIf="maxInOut > 24">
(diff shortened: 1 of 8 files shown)


softsimon:
I personally wanted the clicking on the UTXO actually takes you to the next transaction, if it was spent, otherwise it could take you to the utxo below like this.

Because with this way, you could use the diagram to navigate back and forth to track coins between transactions.

mononaut:
I've pushed some changes:
 - Clicking on an input or spent output now navigates to the input/spending transaction
 - Clicking on an unspent output highlights that output, and scrolls to it in the list as before

Navigating like this using the flow diagram was quite awkward, because the page jumps down to the selected output in the table when going to an input tx, and it's not obvious where you came from.

So to improve the experience a bit, I also:
 - disabled the scroll-into-view behavior on loading a `/tx/txid:vout` url
 - added highlighting for selected inputs/outputs in both the diagram and the table
 - added support for selecting/highlighting specific inputs by url, like `/tx/vin:txid` (is this the best syntax to use?)
 
https://user-images.githubusercontent.com/83316221/193952296-672c885d-ea94-4d22-b535-f1217820bb69.mp4

The new transaction input url format is maybe a bit out of scope here. If so, I can break it out into a separate PR.

wiz (approved):
Tested ACK @ v2.5.0-dev [[42dec957](https://github.com/mempool/mempool/commit/42dec957)]

Looks great using TX with lots of outputs like this one: `/tx/0a96c2ef88437436422f203863d6f293267e773bdc8736fabe559ce31261856f:85`

However I agree that `/tx/vin:txid` might not be the best syntax to use - for backwards compatibility I think we should append to the end of `/tx/:txid` - I'm not sure what's best off the top of my head, but perhaps using an anchor style link something like `#vin=4` or `#vout=4` might be a good idea?

PR #2607, opened 2022-10-04
Fix transaction confirmation badge layout bug

mononaut:
Fixes a layout glitch on the transaction page, on screens between ~590px and 768px wide:

layout-bug


wiz (approved):
Tested ACK @ v2.5.0-dev [[001f7a4f](https://github.com/mempool/mempool/commit/001f7a4f)]

Nice catch

PR #2611, opened 2022-10-06
Updating time toggle size

softsimon:
Reducing the size of the blockchain flip toggle to make it little less prominent.

Before
Screen Shot 2022-10-06 at 20 33 03

After
Screen Shot 2022-10-06 at 20 33 06


wiz (approved):
Tested ACK @ v2.5.0-dev [[52b2ee4f](https://github.com/mempool/mempool/commit/52b2ee4f)]

PR #2612, opened 2022-10-06
Display LN capacity on mobile

### LINKED ISSUE #2610: Display capacity instead of channel count on mobile
Currently on mobile, we can see node count and channel count.

Arguably, network capacity is the most interesting metric, and channel count is the least interesting, so we should instead display **Capacity** and **Node Count** on mobile.

Screen Shot 2022-10-06 at 20 10 08

softsimon:
fixes #2610

Before
Screen Shot 2022-10-06 at 21 13 35

After
Screen Shot 2022-10-06 at 21 13 53


wiz (approved):
Tested ACK @ v2.5.0-dev [[b638719e](https://github.com/mempool/mempool/commit/b638719e)]

Shellnco (approved):

PR #2615, opened 2022-10-06
Inverting med/avg toggle

softsimon:
Grey should mean it is selected. Blue link means it is clickable.

In this example, "med" is the selected view and you can click on "avg" to switch.

Before
Screen Shot 2022-10-07 at 01 19 14

After
Screen Shot 2022-10-07 at 01 19 10


wiz (approved):
Tested ACK @ v2.5.0-dev [[5365f611](https://github.com/mempool/mempool/commit/5365f611)]

PR #2617, opened 2022-10-08
Initial frontend runtime config support

knorrium:
This is the first step to use runtime configuration on the frontend. 

The build system still relies on the `mempool-frontend-config.json`  file to read the settings for backwards compatibility, but generates a template and an actual config file to be served by the `resources` folder. If no config file is provided, we still generate the files but with no placeholder values on the template, as it will respect the defaults defined in the configuration class.

Future PRs should include:
- Providing default values for all variables on the Docker build
- Updating the Docker scripts to read and replace values before the container starts
- Cleanup the test commands in the `package.json` file


softsimon (approved):
It works and the new way of loading the config separately should hopefully not break if there's some race condition in the loading of the scripts.

I don't know what the template is for, but I suppose Docker will use it later?

Also adding the unique ID to the file causing it to reload every time might be a problem when everything else is cached. It should be able to be cached by our nginx just the same way every other file is cached no?

tested ACK @ v2.5.0-dev [[ad7e7795](https://github.com/mempool/mempool/commit/ad7e7795)]

frontend/src/index.liquid.html
```diff
@@ -1,10 +1,15 @@
...
+
 <head>
   <meta charset="utf-8">
   <title>mempool - Liquid Network</title>
   <base href="/">
 
+  <script>
+    document.write("<script src=\"/resources/config.js?v=" + new Date().getTime() + "\"><\/script>");
+  </script>
+
```
wiz, line 12: What is going on here exactly? lol

frontend/src/index.mempool.html
```diff
@@ -1,10 +1,15 @@
...
+
 <head>
   <meta charset="utf-8">
   <title>mempool - Bitcoin Explorer</title>
   <base href="/">
 
+  <script>
+    document.write("<script src=\"/resources/config.js?v=" + new Date().getTime() + "\"><\/script>");
+  </script>
+
```
wiz, line 12: What is the intention of this ugly hack, to bypass cache or something?

knorrium: Yes - appending the timestamp so we always get a fresh config

frontend/src/index.bisq.html
```diff
@@ -1,10 +1,15 @@
...
+
 <head>
   <meta charset="utf-8">
   <title>mempool - Bisq Markets</title>
   <base href="/">
 
+  <script>
+    document.write("<script src=\"/resources/config.js?v=" + new Date().getTime() + "\"><\/script>");
+  </script>
+
```
wiz, line 12: wut

frontend/src/index.mempool.html
wiz: Instead of this hack, let's serve the config.* with no-cache and expires -1 in nginx.conf ?

knorrium: We can try that. 

Can you share the actual change needed to the nginx config so I can make the docker changes to apply the overrides and try on a real image?

wiz: In the `production/nginx/server-common.conf` there is a block like this:
```
# cache /resources/** for 1 week since they don't change often
location /resources {
        try_files $uri /en-US/index.html;
        expires 1w;
}
```
below it I would add a more specific match block:
```
# only cache /resources/config.* for 10 seconds since it changes often
location /resources/config. {
        try_files $uri =404;
        expires 10s;
}
```

wiz: Since the `index.html` is cached for 5 minutes, that seems like a reasonable time to cache the `config.js` files for as well

wiz: You'll also need to modify the top-level `nginx-mempool.conf` with a similar block, since they are still separate until @erikarvstedt merges them in his PR for that later

wiz (approved):
Tested ACK @ v2.5.0-dev [[235ac204](https://github.com/mempool/mempool/commit/235ac204)]

PR #2621, opened 2022-10-10
WIP: new transaction selection algorithm & scoring for block audits

mononaut:
This draft PR implements a version of Bitcoin Core's [transaction selection algorithm](https://github.com/bitcoin/bitcoin/blob/master/src/node/miner.cpp#L292-L430), and uses it to generate expected block templates for the block audit feature, which should hopefully be more closely aligned with the results of Core's `getBlockTemplate`.

The implementation is designed to work either standalone just for the block audit feature, or (perhaps in future) as a drop-in replacement for the current `updateMempoolBlocks` algorithm for constructing mempool blocks.

### Transaction selection algorithm

The algorithm works by assigning each transaction an 'ancestor score' based on the effective fee rate of the transaction plus all of its unconfirmed ancestors.

Blocks are constructed by greedily selecting the 'package' of transaction + ancestors with the highest score. After selecting each package, the scores of any remaining unselected descendants are updated.

To avoid having to re-sort the entire mempool every time descendant scores are updated, the implementation maintains a smaller separately sorted list of transactions with modified scores. ~If this approach turns out to be too slow, it might be necessary to swap the list for a max-heap or some other fast priority queue structure.~ _(**edit:** simple list now replaced with a pairing heap priority queue structure, which performs very well)_

### Comparison with `updateMempoolBlocks`

`updateMempoolBlocks` uses a single pass of `setRelativesAndGetCpfpInfo`, which identifies lower-fee-rate unconfirmed ancestors for each transaction, and sets a 'best descendant', in order to calculate approximate effective fee rates.

This is fast, but it turns out a single pass is insufficent to identify optimal CPFP descendants, and can't account for competing claims to CPFP ancestors. The result is that this approach tends to significantly underestimate the true effective fee rates of CPFP-related transactions.

In the context of the block audit feature, this means CPFP transactions are often incorrectly excluded from the projected block template.

The new algorithm is slightly more complex in the worst case, but should correctly prioritize most CPFP transactions.

#### ~Complexity~ _(should be resolved with new optimizations - monitoring a live deployment to confirm)_

Need to confirm that this implementation runs fast enough when the mempool is congested.

#### ~Bad match rates~ _(resolved with new scoring heuristics)_
Despite modeling projected blocks _more accurately_, this implementation sometimes produces much _worse_ 'match rate' scores.

We usually have some extra transactions in our projected template that are missing from the mined block, simply because miners update their templates less frequently. When these transactions are large, they push many other lower fee transactions out of our projected block.

At the same time, the old `updateMempoolBlocks` algorithm tends to wrongly exclude CPFP packages, (which are usually relatively high-weight). These two errors cancel out, and raise the match rate score.

extreme example (notice the large blue 'missing' transactions, the wrongly excluded CPFP packages in pink on the left, and the many small 'displaced' transactions in pink at the top of the block on the right):

| old algorithm | new algorithm |
|-|-|
| before | after |

##### Possible solutions:

The way 'missing' transactions displace others into the 'added' category, means we're sort of double-counting the effect. This could be mitigated by:
- Only counting either 'added' _or_ 'missing' transactions.
- Selectively ignoring 'added' transactions that only narrowly fall outside of our predicted block.

We could also calculate the match rate score based on transaction weight rather than transaction count (or offer both?).

### DIFF
--- a/backend/src/api/audit.ts
+++ b/backend/src/api/audit.ts
@@ -1,3 +1,4 @@
+import logger from '../logger';
 import { BlockExtended, TransactionExtended, MempoolBlockWithTransactions } from '../mempool.interfaces';
 
 const PROPAGATION_MARGIN = 180; // in seconds, time since a transaction is first seen after which it is assumed to have propagated to all miners
--- a/backend/src/api/audit.ts
+++ b/backend/src/api/audit.ts
@@ -1,0 +1,114 @@
+import { BlockExtended, TransactionExtended, MempoolBlockWithTransactions } from '../mempool.interfaces';
+
+const PROPAGATION_MARGIN = 180; // in seconds, time since a transaction is first seen after which it is assumed to have propagated to all miners
+
+class Audit {
+  auditBlock(block: BlockExtended, txIds: string[], transactions: TransactionExtended[],
+    projectedBlocks: MempoolBlockWithTransactions[], mempool: { [txId: string]: TransactionExtended },
+  ): { censored: string[], added: string[], score: number } {
+    const matches: string[] = []; // present in both mined block and template
+    const added: string[] = []; // present in mined block, not in template
+    const fresh: string[] = []; // missing, but firstSeen within PROPAGATION_MARGIN
+    const isCensored = {}; // missing, without excuse
+    const isDisplaced = {};
+    let displacedWeight = 0;
+
+    const inBlock = {};
+    const inTemplate = {};
+
+    const now = Math.round((Date.now() / 1000));
+    for (const tx of transactions) {
+      inBlock[tx.txid] = tx;
+    }
+    // coinbase is always expected
+    if (transactions[0]) {
+      inTemplate[transactions[0].txid] = true;
+    }
+    // look for transactions that were expected in the template, but missing from the mined block
+    for (const txid of projectedBlocks[0].transactionIds) {
+      if (!inBlock[txid]) {
+        // tx is recent, may have reached the miner too late for inclusion
+        if (mempool[txid]?.firstSeen != null && (now - (mempool[txid]?.firstSeen || 0)) <= PROPAGATION_MARGIN) {
+          fresh.push(txid);
+        } else {
+          isCensored[txid] = true;
+        }
+        displacedWeight += mempool[txid].weight;
+      }
+      inTemplate[txid] = true;
+    }
+
+    displacedWeight += (4000 - transactions[0].weight);
+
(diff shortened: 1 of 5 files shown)


wiz:
Thanks for working on this. I'm sure @thebluematt will probably have some input, but in any case we need @softsimon @nymkappa @knorrium to help review this PR as well

mononaut:
I've pushed a commit which updates the scoring method in line with TheBlueMatt's suggestions, and both the overall scores and false positive rate are significantly improved.

| Old tx selection & scoring method | New tx selection & scoring method |
|-|-|
| Audit scores over 24hrs | |
| before-24hr | after-24hr |
| Typical lower-scoring block | |
| before-block | after-block |

The score is now calculated as `#matches / (#matches + #missing)`.

### Transaction classification
**"Match"**
Transaction present in both our projected next block and the mined block.

**"Missing"** (possibly censored)
Transactions which were expected in our projected block, but absent from the mined block, except those which:
 - were first seen within the last 3 minutes.
    (allows for tx propagation time, and mining template update frequency)
 - occupy the bottom N weight units of our projected block by fee rate.
    (transactions which the miner would likely have included if they had space. Reasonable values for N vary depending on how many weight units the miner spent on other transactions.)

**"Added"**

These do not directly contribute to the score, but are interesting to identify in the audit visualization, and affect the heuristics for "missing" transactions.

A transaction is considered "added" if it is included in the mined block, but does not appear in our projected next block or in the first M weight units of subsequent projected blocks.

**Other**

The block audit visualization now only highlights transactions which meet these definitions of "missing" and "added".

It might be useful to continue to highlight the broader differences between projected and mined blocks in other colors, although this is probably out of scope for this PR.

mononaut:
My first attempt at the transaction selection algorithm focused on getting the template results correct, and was not at all optimized, so performance was terrible with larger mempools. 

I've now spent some time refactoring it to improve complexity and runtime, and it now consistently runs in a similar time to `updateMempoolBlocks` on my machine, when tested against simulated mempools of varying sizes (from 2MB - 300MB).

I'm going to leave it running on my dev server for a while, to confirm how it handles real mempool conditions, but the performance seems reasonable so far. (at the time of writing, the mempool is ~25k txs and 60MB deep, and `makeBlockTemplates` is running in ~0.08s vs ~0.06s for `updateMempoolBlocks`).

wiz (approved):
Tested ACK @ v2.5.0-dev [[bee573fd](https://github.com/mempool/mempool/commit/bee573fd)]

As for CPU time, currently getting around 1ms for a near-empty mempool:
```
Oct 31 05:43:51 [42245] DEBUG: Mempool blocks calculated in 0.007 seconds
```

PR #2627, opened 2022-10-11
Fix flow diagram navigation for coinbases & peg-ins

mononaut:
_(builds on #2628) (resolves #2640)_

Fixes a bug where clicking on a coinbase or liquid peg-in input wrongly attempted to navigate to a non-existent origin transaction.

Now, clicking such an input simply highlights it in the input table below, like clicking an unspent output.

PR #2628, opened 2022-10-11
transaction vin/vout selection syntax via url fragments

mononaut:
This PR switches to a new syntax for referencing specific transaction inputs and outputs as suggested in [#2605](https://github.com/mempool/mempool/pull/2605#pullrequestreview-1135240883).

Rather than modifying the txid field with urls like `/tx/<vin>:<txid>` or `/tx/<txid>:<vout>`, the new syntax uses query-string-formatted fragment parameters, e.g. `/tx/<txid>#vin=3` or `/tx/<txid>#vout=1`.

Using fragments preserves user privacy (since, unlike query parameters or url paths, they are not sent to the server in the HTTP request), and avoids breaking things that assume the txid field is unmodified (like the unfurler).

Abusing the fragment identifier breaks normal browser handling of anchor references, so we simulate this by checking for any element matching the first blank fragment parameter and scrolling it into view after page load. For example, after navigating to `/tx/<txid>#flow&vout=1`, the page loads scrolled to the flow diagram section, with the second output highlighted.

For backwards compatibility with previously shared urls, we continue to accept and correctly parse the old syntax, but quietly rewrite the location in the new format.


wiz:
Please rebase onto current master

mononaut:
rebased

wiz:
@softsimon why was this closed?

mononaut:
the commit for this PR was inadvertently merged as part of #2627

nymkappa:
> the commit for this PR was inadvertently merged as part of #2627

Did it fixes https://github.com/mempool/mempool/issues/2648 then?

antonilol:
> Did it fixes #2648 then?

not completely, see the comment there.
i am still able to reproduce that on my own mempool (c8e84ec0) and on mempool.space (cc99dd07)

PR #2631, opened 2022-10-12
Maintain routing when switching network

### LINKED ISSUE #2623: Stay on same page when switching network
When switching network (to Signet, Testnet, Liquid) in the Network Dropdown, the user should stay on the same page, for example the `/lightning` explorer tab.

* Only if the network you navigate to has that page. (Liquid doesn't have /lightning etc)
* Ignore all network specific pages like a specific TX, in this case just go to the start page /.
* Perhaps it's better to have a white list of the pages that are common and change the URLs in the dropdown on the fly.

mononaut:
This PR resolves #2623 by augmenting the route tree with network-compatibility hints, and using that to choose the best page to link to for each option in the network selector dropdown.

For example, from the page `mempool.space/testnet/graphs/mining/pools`, the network selector will link to the following:
| network | href |
|-|-|
| Bitcoin mainnet | `/graphs/mining/pools` |
| Bitcoin testnet | `/testnet/graphs/mining/pools` |
| Bitcoin signet | `/signet/graphs/mining/pools` |
| Bisq | `bisq.markets` |
| Liquid mainnet | `liquid.network/graphs` |
| Liquid testnet | `liquid.network/graphs` |

and from `mempool.space/lightning/node/:nodeid`

| network | href |
|-|-|
| Bitcoin mainnet | `lightning/node/:nodeid` |
| Bitcoin testnet | `/testnet/lightning` |
| Bitcoin signet | `/signet/lightning` |
| Bisq | `bisq.markets` |
| Liquid mainnet | `liquid.network` |
| Liquid testnet | `liquid.network` |

### Network hints

Network hints are added using the `data` attribute on Route objects.

`data.networks: ['bitcoin', 'liquid']` means this route is available on both bitcoin and liquid, but not bisq.

`data.networkSpecific: true` means this route should not be preserved across network changes (e.g. a specific transaction page).

If a route has neither `data.networks` nor `data.networkSpecific` attributes set, it is assumed to be compatible with all networks.

### Service

This PR adds a new service `navigationService`, which subscribes to router navigation events.

Each time the route changes, it updates an Observable with the longest compatible path for the current route for each network option by traversing the active route tree.


antonilol:
currently, when on testnet or signet, going to the about page gets you to the mainnet (`/testnet/about` gives an about page, but `/about` is used with the `(i)` button), the same with liquid testnet

this seems related, will this also be fixed?

mononaut:
no, that's not currently fixed by this PR.

I think there might be a few other places where internal links fail to preserve testnet/signet, so it's probably best tackled all at once with a separate PR.

wiz:
> when on testnet or signet, going to the about page gets you to the mainnet

That's simply because `/about` is not related to any network, just like `/terms-of-service` or other such pages, so in that case the correct fix is to preserve the state of the pull-down selector so it doesn't get reset to mainnet cc @knorrium

wiz (approved):
Tested ACK @ v2.5.0-dev [[f3cfc7f8](https://github.com/mempool/mempool/commit/f3cfc7f8)]

Works great!

PR #2639, opened 2022-10-14
Handle instant block, txid and address search

### LINKED ISSUE #2619: Search bar should handle instant TXID and Address search
Currently, pasting a TXID or Address and press enter quickly will force you to wait for the first search happened. 

It should be instant as long as the input is valid.

softsimon:
fixes #2619

Now when user is pasting (or typing in) a Block Height Number, Block Hash, Address or TXID that matches the regular expression, instantly display the search bar with a Go to "search_term" option, so that Paste + Enter can be done immediately without waiting for the server response.

### DIFF
--- a/frontend/src/app/components/search-form/search-form.component.ts
+++ b/frontend/src/app/components/search-form/search-form.component.ts
@@ -24,7 +24,7 @@ export class SearchFormComponent implements OnInit {
   typeAhead$: Observable<any>;
   searchForm: FormGroup;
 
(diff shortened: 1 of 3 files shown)


softsimon:
Have to switch to draft because Address regex is broken

softsimon:
Updated bch1 regex to detect `42-62`, but which might need to be updated this with future segwit version soft forks.

> As a result of the previous rules, addresses are always between 14 and 74 characters long, and their length modulo 8 cannot be 0, 3, or 5. Version 0 witness addresses are always 42 or 62 characters, but implementations MUST allow the use of any version.

wiz (approved):
Partially Tested ACK @ v2.5.0-dev [[ffad5e2a](https://github.com/mempool/mempool/commit/ffad5e2a)]

Seems to work quite well, but only tested the most common cases:
* block height
* block hash
* address
* lightning node
* lightning channel

PR #2649, opened 2022-10-19
Extend flow diagram to differentiate spent and unspent TXOs

mononaut:
This PR adds some visual differentiation between spent and unspent TXOs in the flow diagram.

The general idea is to suggest that the diagram for this specific transaction is part of a larger UTXO graph, so spent outputs 'connect' to the input line for some other transaction, while unspent outputs remain unconnected.

Continuing the concept, regular inputs have a matching connector, while coinbase inputs and liquid peg-ins do not.

This change is limited to the transaction page, and not reflected in unfurler previews.

Examples:

| | |
|-|-|
| Regular tx with one unspent and one spent output | spent-unspent-2 |
| Larger tx, showing appearance with small input/output lines | larger-tx |
| Testnet coinbase (with spent output) | testnet-spent-coinbase |
| Liquid peg-in (with spent output) | liquid-pegin |

### DIFF
--- a/frontend/src/app/components/transaction/liquid-ublinding.ts
+++ b/frontend/src/app/components/transaction/liquid-ublinding.ts
@@ -126,9 +126,13 @@ export class LiquidUnblinding {
   }
 
   async checkUnblindedTx(tx: Transaction) {
-    const windowLocationHash = window.location.hash.substring('#blinded='.length);
-    if (windowLocationHash.length > 0) {
-      const blinders = this.parseBlinders(windowLocationHash);
+    if (!window.location.hash?.length) {
+      return tx;
+    }
+    const fragmentParams = new URLSearchParams(window.location.hash.slice(1) || '');
+    const blinderStr = fragmentParams.get('blinded');
+    if (blinderStr && blinderStr.length) {
+      const blinders = this.parseBlinders(blinderStr);
       if (blinders) {
         this.commitments = await this.makeCommitmentMap(blinders);
         return this.tryUnblindTx(tx);
--- a/frontend/src/app/components/transaction/transaction-preview.component.html
+++ b/frontend/src/app/components/transaction/transaction-preview.component.html
@@ -29,7 +29,7 @@
 
 
   <div class="row graph-wrapper">
-    <tx-bowtie-graph [tx]="tx" [width]="1112" [height]="346" [network]="network"></tx-bowtie-graph>
+    <tx-bowtie-graph [tx]="tx" [width]="1132" [height]="346" [network]="network"></tx-bowtie-graph>
     <div class="above-bow">
       <p class="field pair">
         <span [innerHTML]="'&lrm;' + (tx.size | bytes: 2)"></span>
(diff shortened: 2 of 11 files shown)


softsimon (changes requested):
Could you add the `cursor: pointer` to the new connectors to indicate clickability?

Another thought I had was, now with the connectors, maybe the regular flow-lines links should just jump to reveal the UTXO below, and only the connector will take you to the transaction? (Just an idea if it makes sense)

wiz (changes requested):
Tested NACK @ 6b9ac8a26c9d3615bf5d94b63816a33b365dd2bc

When clicking on a spent TXO on liquid that has confidential transactions, a `Error: Invalid blinding data (invalid number)` error is displayed, i.e. `/tx/d27370573b4bb0ea40f26027373f641b740d66dc029526069ce2143af9d1acbf?showFlow=true#flow=&vin=0`

Screen Shot 2022-10-31 at 11 32 21

mononaut:
Good catch, I've just pushed a fix.

The liquid unblinding code assumed any url fragment must be a blinding factor, so it was throwing errors trying to parse `#vout=...` etc.

I've rewritten that part using URLSearchParams, so it'll play nice with the new vin/vout syntax and other anchor tags.

Here's a Liquid txid with blinding factors to verify unblinding still works: `/tx/1a7f371207ca5a1d13d8ab3dab7fb7a9e05288fe922665e30731adddddd81ecb#blinded=800,6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d,90fb24aab49dfd862e60161ae621f9dc9679f3197b55e6b1b7454155c94a1b7f,4f07afe0d712e80b91319eaadcb3401ee22e6f88caa1f0faadfd9ea7bc95abfe,900,6f0279e9ed041c3d710a9f57d0c02928416460c4b722ae3457a11eec381c526d,6ca03a3942ea11ccec276fe5750e2859fe53b33d0a840f71818161620682cca0,a00df580064426065097de14b92023b0da2e900f8607e42fade2da46235edf6b`

mononaut:
I've also added cursor:pointer and mouse event handling to the line endcaps and connectors.

mononaut:
> Another thought I had was, now with the connectors, maybe the regular flow-lines links should just jump to reveal the UTXO below, and only the connector will take you to the transaction? (Just an idea if it makes sense)

I gave this a try, but the shape & size of the connector element makes it quite difficult to click, especially for smaller utxos.

softsimon (changes requested):
To improve intuitivity

* Clicking on the main bands should just highlight and take you down to the UTXOs both on input and outputs
* The "connector" should be a separate highlighted button that takes you to the previous or next UTXO
* Perhaps something different in the tooltip to clarify the difference

mononaut:
Updated:
 - clicking on inner band scrolls to the input/output in the list
 - clicking on the outer connector navigates to the connected transaction
 - tooltip for the outer connector shows the txid and vin/vout number of the connected transaction
 
Screenshot 2022-11-22 at 10 14 39 AM

wiz (approved):
Tested ACK @ v2.5.0-dev [[4abd77fe](https://github.com/mempool/mempool/commit/4abd77fe)]

softsimon (approved):
tested ACK

PR #2651, opened 2022-10-20
Fix lightning channel close classification logic

### LINKED ISSUE #2637: Force closed transaction classified as Mutual
This transaction in question looks like a force close, but it has been classified as mutual close.

Investigate if this has been mislabeled and what might be the cause.

https://mempool.space/lightning/channel/745541451452710913

mononaut:
(fixes #2637)

This PR fixes a bug in the `runClosedChannelsForensics` close classification logic.

Specifically, we were misidentifying force closes as mutual when the close had:
 - a single spent `to_remote` output
 - no `to_local` output
 - no HTLCs
 - no anchors
 
e.g. https://mempool.space/lightning/channel/745541451452710913

In that case, the two types of close can be distinguished by their sequence numbers & locktime, however this clause prematurely assumed such a close must be mutual:

https://github.com/mempool/mempool/blob/eb18625802995db26db115aa4755227a395af137/backend/src/tasks/lightning/network-sync.service.ts#L329-L332

Channels that have already been misclassified will need to be reconsidered. I'm not sure what the best approach is here - maybe add a database migration to reset `closing_reason=NULL` for force-closed channels, so they'll be picked up on the next run of `runClosedChannelsForensics`?

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -4,7 +4,7 @@ import logger from '../logger';
 import { Common } from './common';
 
 class DatabaseMigration {
-  private static currentVersion = 40;
+  private static currentVersion = 41;
   private queryTimeout = 120000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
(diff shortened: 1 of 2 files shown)


softsimon (approved):
Added migration script to force rescan of closed transactions classified as mutual.

tested ACK

Issue #2536, opened 2022-09-09
Node count in list does not match the node count in header


nymkappa:
# How to repro

- go to https://mempool-dev.fra.mempool.space/graphs/lightning/nodes-per-isp
- toggle switch to nodes
- note the number of nodes for wiz
- go to https://mempool-dev.fra.mempool.space/lightning/nodes/isp/54415
- note the number of nodes for wiz

PR #2654, opened 2022-10-21
Only show active nodes is isp page

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2536

I'd like to see how this PR behave with the up to date production database. My local db is outdated and it will take some time to re-sync with the LN graph 😄.

With this PR, we don't show nodes which have an active channel count of 0 anymore. This should therefore matches the result we have in the ISP ranking when sorting per node count.


wiz (changes requested):
Tested NACK @ v2.5.0-dev [[7232c475](https://github.com/mempool/mempool/commit/7232c475)]

Screen Shot 2022-11-21 at 18 45 41

After deploying this PR, if you check the Amazon.com ISP page, it is missing Bitrefill. It seems this is because Bitrefill is hosted on the 2nd ASN belonging to Amazon.com, so I believe your new SQL query is not properly handling the case of ISPs with more than one ASN.

PR #2662, opened 2022-10-29
Block audit feature

mononaut:
__(depends on PR #2621)__

This PR tidies up some unfinished parts of the Block Audit page, and exposes that feature by linking it from the block page and the /blocks and mining dashboard latest blocks lists.

### Terminology

"Match Rate" didn't seem appropriate now that the scoring is geared more towards "is a miner possibly censoring transactions?".

I've provisionally renamed the metric "Block Health", but I'm sure we can come up with something better. Maybe "Audit Score" or "Prediction Rate"?

### Block Audit page

I tweaked the styling of the audit visualizations a bit:

 - Swapped the colors around: 'missing' txs are now pink/magenta, and 'added' are blue/cyan.
 - Desaturated the colors for normal 'matching' transactions, to make highlighted txs more visible.
 - Different colors for the new categories of tx (see note below).
 - Changed the subtitles & toggle button labels to 'Projected Block' and 'Actual Block'.

The new scoring heuristics give us five categories of transaction, based on the differences between the projected and mined block, and whether those differences appear to be intentional or by chance.

| | in the projected block | in the mined block | intentional? | color |
|-|-|-|-|-|
| *match* | ✅ | ✅ | ~ | normal |
| *removed* | ✅ | ❌ | ✅ | magenta |
| *omitted* | ✅ | ❌ | ❌ | dark pink |
| *added* | ❌ | ✅ | ✅ | cyan |
| *included* | ❌ | ✅ | ❌ | dark blue |

Again, the names could be improved.

audit-page

### Block Page

The score is added to the block details table, and links to the corresponding block audit page.

block-page

### Blocks List

The score is shown in the list on the /blocks page and on the mining dashboard as a "progress bar"-style meter, and again links to the audit page.

The list on the main dashboard is unchanged.

main-block-list

mining-dash

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -20,6 +20,7 @@ import indexer from '../indexer';
 import fiatConversion from './fiat-conversion';
 import poolsParser from './pools-parser';
 import BlocksSummariesRepository from '../repositories/BlocksSummariesRepository';
+import BlocksAuditsRepository from '../repositories/BlocksAuditsRepository';
 import mining from './mining/mining';
 import DifficultyAdjustmentsRepository from '../repositories/DifficultyAdjustmentsRepository';
 import PricesRepository from '../repositories/PricesRepository';
@@ -186,14 +187,18 @@ class Blocks {
       if (!pool) { // We should never have this situation in practise
         logger.warn(`Cannot assign pool to block ${blockExtended.height} and 'unknown' pool does not exist. ` +
           `Check your "pools" table entries`);
-        return blockExtended;
+      } else {
+        blockExtended.extras.pool = {
+          id: pool.id,
+          name: pool.name,
+          slug: pool.slug,
+        };
       }
 
-      blockExtended.extras.pool = {
-        id: pool.id,
-        name: pool.name,
-        slug: pool.slug,
-      };
+      const auditSummary = await BlocksAuditsRepository.$getShortBlockAudit(block.id);
+      if (auditSummary) {
+        blockExtended.extras.matchRate = auditSummary.matchRate;
+      }
     }
 
     return blockExtended;
--- a/backend/src/api/mining/mining-routes.ts
+++ b/backend/src/api/mining/mining-routes.ts
@@ -238,6 +238,12 @@ class MiningRoutes {
   public async $getBlockAudit(req: Request, res: Response) {
     try {
       const audit = await BlocksAuditsRepository.$getBlockAudit(req.params.hash);
+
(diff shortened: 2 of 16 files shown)


wiz:
This is beautiful. I can't wait to ship this

mononaut:
Thanks for the feedback!

The labels correspond to:
 - **Removed**: missing from the mined block, and possibly censored.
 - **Omitted**: missing from the mined block, but either recently broadcast or below the heuristic cut-off.
 - **Added**: wasn't even close to being included in our projected block, therefore possibly prioritized or added by the miner manually.
 - **Included**: wasn't in our projected block, but only missed out by a small margin.
 
So "omitted" and "included" cover differences that are nonetheless consistent with prioritization by feerate, while "added" and "removed" should be the more interesting discrepancies that could imply deliberate intervention by the miner.
 
But yeah, I think trying to convey those concepts in single word labels might be more confusing than helpful 😅️

mononaut:
> Another heuristic that may be worth considering, but dunno if its worth the effort - in general blocks are sorted highest-feerate to lowest-feerate (assuming Bitcoin Core generated them). This should allow you to usually identify a "cutoff" point for the lowest feerate included to compare against (you have to be careful with CPFP here), instead of using the crude "bottom 10% of the block heuristic I'd suggested in the last issue. That one may be too fickle to get right, however.

That's more or less how the heuristic is implemented now. Rather than ignoring a fixed 5% or 10% off the bottom, I choose a cut-off based on the maximum effective feerate from the bottom N weight units of our projected block, where N is the total weight of 'unexpected' transactions in the mined block.

Or, thinking about it the other way around, I only consider transactions that the miner actually had space for, after spending N weight units on their own stuff.

wiz (approved):
Tested ACK @ v2.5.0-dev [[45273f93](https://github.com/mempool/mempool/commit/45273f93)]

Few things:
* The skeleton loaders could use some polish on both desktop and mobile, for example there is a duplicate title, and block shows as "Genesis" while skeleton is visible
* We should probably display unknown for the block health feature for block heights before we deployed this feature to prod, otherwise the old getsimontemplate data suggests the blocks had very low match rates when it was just our algorithm that needed improvement - I suppose we could just delete old MySQL rows for those block heights?

Screen Shot 2022-10-31 at 15 04 11
Screen Shot 2022-10-31 at 15 01 33
Screen Shot 2022-10-31 at 15 00 48

mononaut:
yup, I'll open a new PR to fix up the skeleton loaders and add a database migration to drop the older data.

PR #2664, opened 2022-10-31
db migration to clear obsolete block audit data

mononaut:
Adds a database migration to delete old block audit data generated using getsimontemplate.

Also adds a guard to avoid saving bad audit data while the mempool isn't fully loaded.


softsimon (changes requested):
Truncate instead of Drop

mononaut:
fixed and rebased

wiz (approved):
Tested ACK @ 24dba5a2ef1f5cab603dfacdd50257eda517be09

PR #2665, opened 2022-11-01
backend: enable esplora address lookups

niftynei:
Enable `esplora` lookups; (works with the blockstream/electrs branch?)

Maybe not suitable for all esploras? idk.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -90,7 +90,7 @@ class BitcoinRoutes {
       .get(config.MEMPOOL.API_URL_PREFIX + 'block/:hash/summary', this.getStrippedBlockTransactions);
       ;
 
-      if (config.MEMPOOL.BACKEND !== 'esplora') {
+      if (config.MEMPOOL.BACKEND !== 'none') {
         app
           .get(config.MEMPOOL.API_URL_PREFIX + 'mempool', this.getMempool)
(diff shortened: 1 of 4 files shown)


backend/src/api/bitcoin/bitcoin.routes.ts
```diff
@@ -90,7 +90,7 @@ class BitcoinRoutes {
       .get(config.MEMPOOL.API_URL_PREFIX + 'block/:hash/summary', this.getStrippedBlockTransactions);
       ;
 
-      if (config.MEMPOOL.BACKEND !== 'esplora') {
+      if (config.MEMPOOL.BACKEND !== 'none') {
```
softsimon, line 93: We also have a 3rd backend option other than "esplora and "none" (Bitcoin Core only) called "electrum" when running the more limited `romanz/electrs` or Fulcrum where these API endpoints doesn't exist, so the check to enable them must be `esplora` only.

softsimon, line 93: This would disable these requests when running Bitcoin Core only mode which they are made for
none = bitcoin core only
electrum = romanz/electrs / fulcrum compatible
esplora = blockstream/electrs

niftynei: Makes sense, realized after the fact that it was probs ok to remove...

wiz:
nifty, are you using the production/nginx config?

softsimon:
So the correct solution to the problem the PR is trying to fix is to use Nginx (or similar) for the redirects from frontend to esplora. There are nginx configs available for production environments and there is a local dev proxy called local-esplora when running the Angular CLI.

PR #2667, opened 2022-11-02
Rescan unresolved LN channel force closes

### LINKED ISSUE #2638: Scan for penalty transactions
Currently we only scan closed transactions once and flag them as mutual, force closed or penalty closed. But if the force close transaction hasn't been spent yet, we don't know if it was a penalty close or not.

We need to run a new scan periodically on force closed transactions, see if the force close transaction has been spent and look for the Revoked force close / htlc and update the status to closed with penalty.

We might have to mark transactions as "scanned" if the output(s) has been spent and no revocation found.

Example of Forced closed with penalty that we DID detect:
https://mempool.space/lightning/channel/619545115548450817

I checked https://forkmonitor.info/lightning but couldn't find a Penalty transaction from a Channel that Mempool has picked up from the graph since we shipped the Lightning explorer.

mononaut:
resolves #2638

This PR changes `$runClosedChannelsForensics` to also re-scan force closed channels until the outputs are swept, so that we can detect penalty closes.

It uses a new boolean column `closing_resolved` to the `channels` table to distinguish between definite non-penalty force closes (`closing_resolved = true`), and ambiguous force closes (`closing_resolved = false`) which we should keep checking.


mononaut:
Example of a misclassified force close: https://mempool.space/lightning/channel/722689201792221184

wiz (approved):
Tested ACK @ v2.5.0-dev [[eb03fc18](https://github.com/mempool/mempool/commit/eb03fc18)]

PR #2678, opened 2022-11-08
fix error when navigating to huge transactions

### LINKED ISSUE #2672: cant go to large transaction from block/transaction list
### Description

i made a transaction with 69420 outputs.

from the [block view](https://mempool.space/testnet/block/00000000ce4a4666cce2205d760d37b5579cdedf3ac9e4295557e8ac962cde55) it works (you can click it), in the transaction list, clicking the txid throws an error in my console (firefox)
from the [address list](https://mempool.space/testnet/address/tb1pu5lck3nvg5jrqt9aaqe590vuv7w2ysh25pxflc8yv0lz0w8h8mkqvv8gh5), the same error is thrown when clicking the txid

i can get to this transaction by copying the txid from the list where clicking doesn't work and copy it in the search box

i also tried 'Show all', lagged my browser for more than 10 seconds, but it worked. opening the details crashed my browser.

#### Version

mempool.space v2.5.0-dev [ddb1e97c]

### Steps to reproduce

in the description

### Expected behaviour

txid link should be clickable, clicking the big green block in the block preview works

### Actual behaviour

it throws a long error in the console

image

#### Device or machine

firefox 106.0.3-1 on archlinux

mononaut:
This PR resolves #2672, by fixing a bug where navigating to an extremely large transaction can fail with an unrecoverable error.

### Error

The cause is our (ab)use of the browser history API via `routerLink [state]` to stash transaction data and thereby skip an API request when navigating from a list of transactions to an individual transaction page.

Firefox imposes a limit on the amount of data we can store there, and some transactions may exceed that limit when serialized. Because the actual `history.pushState(...)` call happens within Angular's internal navigation logic, there's no way to catch and handle the resulting error.

### Solution

This PR replaces the History API `routerLink [state]` approach with a simple transaction cache managed by `stateService`, which preserves most of the benefits without the data size limit.

**A few minor differences:**

Unlike the history AI, the stateService cache is not preserved across restored tabs or forward/back browser navigation, so those use cases will require an extra API request to fetch transaction data.

If that isn't acceptable, we could cache the data to sessionStorage instead.


wiz (approved):
Tested ACK @ v2.5.0-dev [[fc702435](https://github.com/mempool/mempool/commit/fc702435)]

PR #2679, opened 2022-11-08
"show more" instead of "show all"  button for transaction inputs/outputs

mononaut:
This PR addresses a secondary issue from #2672: clicking "Show all" on a transaction with very many inputs or outputs can lag or crash the browser, as it tries to render all of the (potentially thousands of) new elements.

The PR replaces the simple "Show all" button with a "Show more" button which renders another 40 rows per click, until the inputs/outputs are fully revealed.

The new behavior is reflected everywhere the `transactions-list` component is used, including the transaction, address, and lightning channel pages.

show-more-after |


wiz (changes requested):
40 seems a bit low, as it would hide large batched withdrawal transactions from exchanges like CashApp etc.

if the goal is just to prevent crashing, how about a high default limit of 1000 ?

softsimon:
This breaks the `loadMoreInputs` function that is no longer called. It's a non-esplora only feature.

mononaut:
- increased 'show more' increment to 1000
- change labels to "show more (x remaining)" / "show all (x remaining)"
- fixed missing loadMoreInputs call for non-esplora backends

mononaut:
fixed issues & rebased

wiz (approved):
Tested ACK @ v2.5.0-dev [[6f807b7a](https://github.com/mempool/mempool/commit/6f807b7a)]

softsimon (approved):
tested ACK @ v2.5.0-dev [[6f807b7a](https://github.com/mempool/mempool/commit/6f807b7a)]

PR #2683, opened 2022-11-11
Lightning linked channel forensics

mononaut:
_(builds on PR #2638)_

This PR uses on-chain links between channels to deduce ownership of transaction inputs and outputs, together with some other inferences, and surfaces all of this on the channels page.

The resulting data could be useful for tracing how liquidity flows across the network, or simply to highlight this potential privacy leak.

Specifically, we look for the following cases:
1) where a node funds a new channel with the direct proceeds of previously closed channel - a fairly common behavior for active routing nodes.
2) a node funds a new channel with the change output of a previous channel open.

If there is a single party in common between these two channels, we can conclude that they owned the connecting txos.

From there, we can fully or partially deduce closing balances of each node in any closed channels, as well as funding contributions to the new channel. In the case of a force close, we can sometimes also infer which node initiated the close.

complete-output-attribution

### Implementation

Implemented as another step in the regular lightning network sync tasks, when using an esplora backend (like the existing `runClosedChannelsForensics` task).

The analysis consists of checking each input of the channel open transaction to see if it spends a known channel close output (or spends the result of sweeping such an output, where that step is required by the protocol).

When a matching channel close is found, we deduce what we can using the heuristics described above, and save the data to a new `channels_forensics` db table.

The first run will take several hours, since it involves fetching and examining transactions one or two steps back from every input of every channel open. However, we only analyse each channel once, so subsequent runs should be fast. To avoid stalling other network sync tasks, we break after `GRAPH_REFRESH_INTERVAL` and resume on the next sync.

### Limitations

Complete liquidity flows are only revealed when nodes fully recycle funds on both the opening and closing side of a channel. In other cases, we may only be able to deduce the funding contributions, or the closing balances, or nothing at all.

#### Dual-funded channels
For a channel known to be funded by a single party, discovering the owner of any input to the channel open transaction tells us who contributed all of the capacity.

However, if a channel open uses multiple inputs, (as far as I'm aware) we can't rule out the possiblity that it's actually dual-funded. In that case, the amount contributed by each party is ambiguous, (especially if the transaction generates a lot of change).

partial-attribution


softsimon:
FYI here is an example of dual funded channels

https://twitter.com/alexbosworth/status/1591896718756708352

mononaut:
interesting!

I saw this one from CLN, but didn't realize there were tools for multi-funded channels on LND already.

https://mempool.space/tx/91538cbc4aca767cb77aa0690c2a6e710e095c8eb6d8f73d53a3a29682cb7581

softsimon (changes requested):
`channels_forensics` columns can be merged into the `channels` table

mononaut:
rebased on current master (f73dc59)

wiz:
please rebase and fix DB migration conflict

softsimon (approved):
tested ACK @ ba10df69

wiz (approved):
Tested ACK @ v2.5.0-dev [[ba10df69](https://github.com/mempool/mempool/commit/ba10df69)]

PR #2684, opened 2022-11-11
Only show python example tab on ws tab

hunicus:
The Python example tab is showing for REST endpoints...but it should only show in the WebSocket section for Bitcoin networks.

## Current

python-bad

## Corrected

no-python-good


wiz (approved):
Tested ACK @ v2.5.0-dev [[4bc5de30](https://github.com/mempool/mempool/commit/4bc5de30)]

PR #2694, opened 2022-11-15
calculate & show avg channel distance on node page

mononaut:
This PR calculates the average geographical distance of a node's channels (except those with unknown locations) using the haversine formula, and displays this on the node page.

demo

currently all client-side, since we have to fetch the channel geodata anyway for the map widget.


mononaut:
rebased

wiz (approved):
Tested ACK @ v2.5.0-dev [[ed184824](https://github.com/mempool/mempool/commit/ed184824)]

PR #2663, opened 2022-10-31
Block audit tweaks

mononaut:
various minor improvements to the block audit feature:
- fix skeleton loaders
- small change to the audit score calculation to reduce false positives.
- remove confusing 'omitted' and 'included' counts.
- highlight the corresponding tx in the other block on mouseover.
- shift 'added' color slightly towards blue, to differentiate it from the mouseover color.
- handle `/block-audit/:height` as well as `/block-audit/:hash`.

PR #2698, opened 2022-11-17
enable new tx selection algorithm in own thread with config setting

### LINKED ISSUE #2690: Enable new transaction selection algorithm with caveats
* New transaction selection algorithm used for the mempool calculation loop
* Create a new setting that is disabled by default(?), that enables both the new scoring and mempool loop, default uses the old algorithm both for scoring and mempool refresh (raspberry pi's)
* Run the new algorithm in a asynchronous NodeJS worker thread (`node-worker-threads-pool`) to not lock the main thread

mononaut:
_(requires PR #2663)_

Resolves #2690 by moving the new transaction selection algorithm into its own worker thread, enabled by a config setting `MEMPOOL.ADVANCED_TRANSACTION_SELECTION` (default `false`).

If `MEMPOOL.ADVANCED_TRANSACTION_SELECTION` is set to `true`, the new algorithm completely replaces `updateMempoolBlocks`, both for calculating the block audit score and building mempool blocks on each mempool update.


mononaut:
rebased on current master [547b60f]

mononaut:
rebased

wiz (approved):
Tested ACK @ v2.5.0-dev [[b1d49097](https://github.com/mempool/mempool/commit/b1d49097)]

@knorrium would be nice to get your help testing this one further...

PR #2705, opened 2022-11-19
better representation of zero-value outputs in flow diagram

mononaut:
_(merge PR #2649 first)_

change the shape of zero-valued outputs in the transaction diagram to make OP_RETURNs more visible and the 'flow' of coins more intuitive, as discussed in #2604.

Before:
zero-value-before

After:
zero-value-after


wiz (approved):
Tested ACK @ v2.5.0-dev [[7e01a222](https://github.com/mempool/mempool/commit/7e01a222)]

PR #2716, opened 2022-11-21
Reverse the direction of the flow diagram for RTL locales

mononaut:
_(merge PR #2705 first)_

Reverses the direction of the flow diagram for RTL locales, so that it matches the order of the inputs/outputs in the list.

Before:
before

After:
after


mononaut:
rebased on latest #2705

PR #2722, opened 2022-11-23
Optimize force close penalty scans

### LINKED ISSUE #2719: Force close penalty scan optimization
* Investigate a local in memory cache of previously scanned UTXOs
* Investigate adding a separate scanner service interval that only runs once a day for example for this type of scans.

mononaut:
Resolves #2719 by moving long-running force close penalty rescans into their own lightning tasks service with a separately configurable interval (default every 12 hours).

Regular first-time channel close forensics scans still run during the main lightning network sync task, so that simple channel closes are still classified quickly.

Also avoids DOS-ing the backend by throttling bitcoinApi calls from the forensics task to at most 10/s, and caching relevant transaction data in memory.


backend/src/tasks/lightning/forensics.service.ts
```diff
@@ -0,0 +1,223 @@
...
+    │ locktime starts with 0x20? │      └─────────────────┘
+    └──────────────┬─────────────┘
+                   no
+         ┌─────────▼────────┐
+         │ mutual close = 1 │
+         └──────────────────┘
+  */
+
+  public async $runClosedChannelsForensics(onlyNewChannels: boolean = false): Promise<void> {
+    if (!config.ESPLORA.REST_API_URL) {
```
softsimon, line 64:
```suggestion
    if (config.MEMPOOL.BACKEND !== 'esplora') {
```

softsimon (approved):
tested ACK @ v2.5.0-dev [[6ada8392](https://github.com/mempool/mempool/commit/6ada8392)]

Issue #2710, opened 2022-11-20
Nodes world map crashes sometimes


softsimon:
`https://mempool.space/graphs/lightning/nodes-map` crashes and doesn't render the map.

`ERROR TypeError: Cannot read properties of null (reading 'en')`

PR #2730, opened 2022-11-24
Don't select nodes which do not have country info

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2710

Because I was using `LEFT JOIN` some nodes country data were obviously set to `null` by mysql. I wonder what were the odds of missing this bug in the past.


softsimon (approved):
tested ACK @ v2.5.0-dev [[13b52c42](https://github.com/mempool/mempool/commit/13b52c42)]

PR #2737, opened 2022-11-27
show CPFP info for mined transactions

### LINKED ISSUE #1722: show cpfp information of confirmed txs
### Description

title, when a confirmed transaction is spent or spends a transaction in the same block and the newer transaction has a higher fee rate, a cpfp was done. effective fee rate can be calculated from this

### Problem to be solved

more insight in fees, knowing why a 1 sat/vB tx is in a block with really high fees

### Proposed solution

add a cpfp button just like cpfp'd transactions in the mempool have (note that this button stays there after confirmation and is gone when you refresh the page)

#### Additional info

example:
https://mempool.space/tx/03de81f175033f6cb9d961fc03ad83c9b9f5ca82cfa5e5b2d35f85d2141a62c2 (not my tx)
it got cpfp'd with the 3rd output, found by scrolling through a block and looking for a transaction with lower fees that the one after it

tx 1: 5000 sat / 271.5 vB = 18.4 sat/vB
tx 2: 8903 sat / 190 vB = 46.9 sat/vB
both: (5000 + 8903) sat / (271.5 + 190) vB = 30.1 sat/vB (effective fee rate)

mononaut:
This PR adds support for calculating and indexing CPFP info (effective fee rate, ancestors and descendants) for transactions in mined blocks.

mined-cpfp

The feature can be enabled using a new backend config setting `MEMPOOL.TRANSACTION_INDEXING` (default false), which keeps the last `INDEXING_BLOCKS_AMOUNT` blocks indexed.

Data is stored in two new tables, `cpfp_clusters` and `transactions`. 

The PR also extends the in-mempool CPFP data to include all relevant descendants instead of a single `bestDescendant`.

mempool-descendants

Fixes #1722


antonilol:
#1722

mononaut:
fixes #1722

wiz (changes requested):
Requested changes:
* Use esplora instead of core RPC, it seems "Get Block Transactions" REST API has the necessary data
* Unrelated to this PR, but also investigate if we can eliminate the `blocks_summaries` table and just query esplora on the fly - would save a lot of disk space and ops indexing time :)

mononaut:
Updated to use a combination of Core RPC `getBlock(hash, 0)` (raw hex) and fee data from the block summaries database, which is a little faster than fetching verbose blocks from core.

backend/src/api/blocks.ts
```diff
@@ -678,6 +739,62 @@ class Blocks {
   public getCurrentBlockHeight(): number {
     return this.currentBlockHeight;
   }
+
+  public async $indexCPFP(hash: string, height: number): Promise<void> {
+    let transactions;
+    if (Common.blocksSummariesIndexingEnabled()) {
```
wiz, line 745: This code seems to not work - my `cpfp_clusters` and `transactions` tables are completely empty. Going to disable it for now.
```suggestion
    if (false/*Common.blocksSummariesIndexingEnabled()*/) {
```

wiz (approved):
Tested ACK @ 272b6d2437f0fe817b954e71def3c12d70ab1a48

* I had to disabled the optimization code path that tries to use the `blocks_summaries` table data if `BLOCKS_SUMMARIES_INDEXING` is enabled - it seems that doesn't work as my `cpfp_clusters` and `transactions` tables don't get populated despite the `cpfp_indexed` bool getting set to true for all blocks?? - it would be nice if the above isssue fixed in the future later on
* Simon also noted that the API request to get audit data will retry in a loop if it gets 404, this should be fixed in a separate PR

PR #2738, opened 2022-11-28
first seen time for mined transactions

mononaut:
Save transaction first seen times when a new block arrives, and display that data where available:

first-seen


wiz (changes requested):
Requested changes:
* Add a configuration variable to enable/disable this feature
* Add a service to auto-delete "first seen" rows older than X days, otherwise DB will grow by 500K rows per day :)

softsimon:
> Requested changes:
> 
>     * Add a configuration variable to enable/disable this feature
> 
>     * Add a service to auto-delete "first seen" rows older than X days, otherwise DB will grow by 500K rows per day :)

The problem with deleting the first Seen data is that we will not be able to make interesting graph data showing
* How long time does it take for transactions to be confirmed

mononaut:
Rebased and added the requested features:
 - new config setting `FIRST_SEEN_INDEXING_DAYS` (default 0)
   - if `0`, first seen data is not saved at all.
   - otherwise, data for transactions first seen more than N days ago is periodically deleted.
   
I think the best way to keep track of data for graphs etc would be to calculate and store summary statistics (avg waiting time and percentiles, maybe broken down into fee bands). Could either be per-block or added to the statistics service.

wiz:
We could use a KVS like redis that scales :P

wiz:
Okay well from the ops perspective, we probably can't merge this feature due to the fact it will grow our MySQL db by 500K rows per day so I have to Concept NACK it for that unless you can come up with a better way to implement it

softsimon:
The plan is to re-implement this with another database architecture.

PR #2739, opened 2022-11-28
Fix broken tests on liquid

mononaut:
only query blocks_audits on bitcoin networks


wiz (approved):
Tested ACK using Cypress @ 2290

PR #2742, opened 2022-11-28
Upgrade to Angular 14

softsimon:
Upgraded to Angular 14
Upgraded other 3rd party packages.


wiz (approved):
Tested ACK @ v2.5.0-dev [[cc936745](https://github.com/mempool/mempool/commit/cc936745)]

PR #2743, opened 2022-11-28
Disable ON UPDATE for blocks.blockTimestamp field

nymkappa:


wiz (approved):
Partially Tested ACK using copy/paste of SQL query

PR #2744, opened 2022-11-28
don't use block.extras on liquid

mononaut:
Fixes a bug where the block component tries to access non-existent `blocks.extras` property on liquid.


wiz (approved):
utACK

PR #2749, opened 2022-11-30
Mempool GBT config

softsimon:
Before we just had one configuration option `ADVANCED_TRANSACTION_SELECTION` that turns on the new GBT algoritm for block audits and mempool blocks template generation.

This PR changes to 2 configs.

`ADVANCED_GBT_AUDIT`
`ADVANCED_GBT_MEMPOOL`

Both are default false. But now it's possible to only enable the block audit feature but not on the every mempool loop.


wiz (changes requested):
Requested change to also modify production backend configurations for all Bitcoin networks to the following:
```
    "ADVANCED_GBT_AUDIT": true,
    "ADVANCED_GBT_MEMPOOL": false,
    ```

wiz (approved):
Tested ACK @ c7cca500fa9316f8f4f038eea7723e7a57fa5ccc

PR #2753, opened 2022-11-30
Remove annoying frontend console log

softsimon:
Removing the following log every time you switch page

`network paths updated...`


wiz (approved):
utACK

PR #2763, opened 2022-12-01
Fix broken transaction preview unfurler page

mononaut:
Fixes a bug that breaks the transaction preview unfurler page for confirmed transactions with no CPFP data available, and a related issue where the main transaction page continually retries CPFP requests.


wiz (approved):
Tested ACK @ 92807dbddedfb8bb443b226e3b28a31b14a6117f

PR #2768, opened 2022-12-01
Block page audit fallback

softsimon:
This PR introduces a new Frontend Config variable `BLOCK_AUDIT_START_HEIGHT` that should be set to the Block Height number that the block audit is available, this default fallbacks all the older block to the old legacy design and skeleton loader. Also if no block audit template is available, it will fallback to the old view instead of displaying "No data"

Recent block with block audit enabled.
Screenshot 2022-12-01 at 19 19 22

Old block
Screenshot 2022-12-01 at 19 19 50

WebGL Disabled
Screenshot 2022-12-01 at 19 20 34


wiz (approved):
Tested ACK @ v2.5.0-dev [534f2e27]

Issue #2767, opened 2022-12-01
Possible bug in the price updater


nymkappa:
Dec  1 16:11:20 [84250] ERR: <lightning> Cannot save BTC prices in db. Reason: Cannot read properties of undefined (reading 'XXBTZUSD')

PR #2770, opened 2022-12-01
Make sure exchange API response format is valid before using it

nymkappa:
Possibly fixes https://github.com/mempool/mempool/issues/2767
I've added checks to the response format sent by the exchanges API to avoid reading undefined variables.


softsimon (approved):
code ACK

softsimon (approved):
tested ACK  @ 71b3734

PR #2771, opened 2022-12-01
Fix for javascript errors when navigating blocks

softsimon:
This fixes 2 separate JS errors happening when navigating blocks quickly.

`ERROR TypeError: Cannot read properties of null (reading 'filter')`

and

`ERROR TypeError: this.initCanvas is not a function`


wiz (approved):
Tested ACK @ v2.5.0-dev [[3572ba83](https://github.com/mempool/mempool/commit/3572ba83)]

PR #2772, opened 2022-12-01
Fix possible race condition with block summaries insert/update

nymkappa:
Test https://github.com/mempool/mempool/pull/2779 before this one.
This PR can still be tested if we have further issues after testing 2779.


nymkappa:
@wiz If the racing condition issue has been solved with https://github.com/mempool/mempool/pull/2779 feel free to close this PR. I won't rebase it unless it's actually needed.

PR #2777, opened 2022-12-03
Make forensics node backend call rate limiting configurable

mononaut:
Adds a new config option `config.LIGHTNING.FORENSICS_RATE_LIMIT`, which sets the minimum delay (in milliseconds) between node backend API calls during lightning forensics scans.


softsimon (approved):
tested ACK @ v2.5.0-dev [[3126a559](https://github.com/mempool/mempool/commit/3126a559)]

PR #2778, opened 2022-12-03
Fix bugs related to advanced GBT transaction selection

mononaut:
This PR fixes some bugs in the advanced transaction selection implementation & usage:
 - a regression causing descendants not to be updated properly after selecting a transaction cluster, which could cause underfilled blocks and incorrect CPFP calculations.
 - wrong parameters passed to makeBlockTemplates immediately after calculating the block audit.


softsimon (approved):
tested ACK @ v2.5.0-dev [[79f6ae3b](https://github.com/mempool/mempool/commit/79f6ae3b)]

Compared CPFPs in the mempool with `ADVANCED_GBT_MEMPOOL` enabled and on confirmed transactions and result looks OK.

PR #2779, opened 2022-12-03
Fix block summaries repo upsert race condition

mononaut:
Hopefully fixes the root cause of the possible race condition mentioned in PR #2772.

`BlocksSummariesRepository.$saveSummary` asynchronously checks whether a database row exists, before either inserting a new summary or updating the row. These two DB operations could potentially become interleaved when `$saveSummary` is called multiple times in quick succession, causing failed attempts to insert duplicate entries and loss of data.

This PR replaces `BlocksSummariesRepository.$saveSummary` with two separate functions `$saveSummary` and `$saveTemplate` which use `ON DUPLICATE KEY UPDATE` syntax to save block summary and template data atomically.

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -439,7 +439,7 @@ class WebsocketHandler {
         };
       }) : [];
 
-      BlocksSummariesRepository.$saveSummary({
+      BlocksSummariesRepository.$saveTemplate({
         height: block.height,
         template: {
           id: block.id,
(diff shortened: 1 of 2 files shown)


nymkappa (approved):
utACK @ 132e848fdc7bf2457f44f19f50c33933e3209d7a

softsimon (approved):
tested ACK @ v2.5.0-dev [[132e848f](https://github.com/mempool/mempool/commit/132e848f)]

Good improvement with removing of the extra SELECT query.

Issue #2780, opened 2022-12-03
Lightning: SQL Error on ISP query


knorrium:
`/api/v1/lightning/nodes/isp/27773`

`You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ') ORDER BY capacity DESC' at line 11`

PR #2790, opened 2022-12-05
Handle ISP with no nodes

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2780

image


softsimon (approved):
tested ACK @ v2.5.0-dev [[685433fe](https://github.com/mempool/mempool/commit/685433fe)]

Issue #2754, opened 2022-11-30
Data Migration Rollback Support


softsimon:
For every database migration there should be a reverse migration that restores to previous state.

There should be a new NPM script that can be manually executed to trigger a rollback to a desired version, and the script must have a strong confirmation stage.

For example: npm rollback 45

nymkappa:
Do you have more details about the use cases for this?

softsimon:
> Do you have more details about the use cases for this?

Wiz has some trouble with testing multiple open PRs that all migrate to the same version. Switching between them doesn't trigger migration and the versions conflict.

nymkappa:
> > Do you have more details about the use cases for this?
> 
> Wiz has some trouble with testing multiple open PRs that all migrate to the same version. Switching between them doesn't trigger migration and the versions conflict.

Okay so we dont need those rollback queries for older migrations in that case since they already have been deployed for a long time. Only most recent ones and future ones.

PR #2791, opened 2022-12-05
Add support for db schema rollback down to version 40

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2754

@mononaut if this is merged, make sure to add rollback queries if you increment the db schema version in the future 👍🏻 

### Usage

Example: Rollback db schema to version 40. This is a destructive operation, all deleted data is lost forever.

`npm run build && npm run rollback 40`


wiz:
Thanks, this is very helpful for testing PRs while still in the review process - previously had to manually rollback 😅

nymkappa:
@softsimon feel free to write a spec of what system you're thinking about instead of my implementation, and I will update the PR accordingly.

softsimon:
Already too late for this one. We might come back to this issue in another form at a later time.

PR #2793, opened 2022-12-05
Refactor advanced GBT implementation to minimize inter-thread data transfer

mononaut:
_(depends on PR #2778)_

This PR refactors the advanced transaction selection implementation to minimize the amount of data copied back and forth from the worker thread and thereby avoid stalling the main thread.

### Implementation

Instead of sending the entire mempool to the worker thread on each update, and then copying the whole thing + the constructed blocks back into the main thread after processing, we use a single long-lived thread process and send only essential data.

When the mempool is updated, we send a very condensed summary of new transactions (only txid, vin txids, fee, and weight and feerates) and a list of removed txids.

The worker thread uses that to update its own copy of the mempool in-place, performs the transaction selection algorithm, and returns a condensed summary of the resulting blocks.

The main thread efficiently updates it's own mempool with the new effective feerate and CPFP info, and rehydrates the blocks.

Testing with an artificial 100MB+/ mempool, this does seem to significantly reduce the main thread overhead.

### Alternatives

An alternative solution to avoid copying lots of data between threads would be to store the mempool in a SharedArrayBuffer, which can be accessed from either thread (with appropriate mutex locks).

Unfortunately, SharedArrayBuffer only supports fixed-length byte arrays, so we'd need to serialize and deserialize the data from JSON or Protocol Buffers or something similar in both threads on every update.

That seems likely to come with it's own significant overhead, as well as a lot of extra complexity.

### DIFF
--- a/backend/src/api/mempool-blocks.ts
+++ b/backend/src/api/mempool-blocks.ts
@@ -158,8 +158,11 @@ class MempoolBlocks {
       };
     });
 
+    // (re)initialize tx selection worker thread
     if (!this.txSelectionWorker) {
       this.txSelectionWorker = new Worker(path.resolve(__dirname, './tx-selection-worker.js'));
+      // if the thread throws an unexpected error, or exits for any other reason,
+      // reset worker state so that it will be re-initialized on the next run
       this.txSelectionWorker.once('error', () => {
         this.txSelectionWorker = null;
       });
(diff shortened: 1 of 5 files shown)


wiz:
@softsimon test case for benchmarking
```
until false;do curl -w "\n\n%{time_connect} + %{time_starttransfer} = %{time_total}\n" -s -i http://127.0.0.1:8999/api/v1/fees/recommended|tail -1;done
```

wiz:
From my quick testing with an empty mempool, the performance is basically the same with ADVANCED_GBT_MEMPOOL = true or false, so I would request @softsimon to test with a full mempool in his testing environment and graph the numbers from the above TC to show the difference between enabled and disabled

softsimon (approved):
tested ACK @ v2.5.0-dev [[5e633344](https://github.com/mempool/mempool/commit/5e633344)]

Before this PR, with a ~25 block sized mempool I always get a ~1.7s stuck delay when hammering queries.

`0.002785 + 1.720782 = 1.720805`

With this PR I almost doesn't notice any delay at all when the mempool refreshes.
Occasionally I get a 0.2 to 0.3s delay.

`0.002833 + 0.385104 = 0.385143`

And block audit seems to work as expected still.

wiz:
can you please re-test with a 1GB and 1.5GB mempool? I am concerned we won't be able to enable this feature in production on our Intel servers

Issue #2792, opened 2022-12-05
Nodejs crash while fetching funding txs


nymkappa:
### Reproduce

Insert an invalid channel id in the `channels` table. The script assumes the short_id is valid and crash.

```
Dec  5 16:46:14 [24296] INFO: <lightning> Updated latest network stats
Dec  5 16:46:14 [24296] INFO: <lightning> Caching funding txs for currently existing channels
/Users/nymkappa/Documents/mempool/backend/dist/tasks/lightning/sync-tasks/funding-tx-fetcher.js:95
            value: tx.vout[outputIdx].value,
                                      ^

TypeError: Cannot read properties of undefined (reading 'value')
    at FundingTxFetcher.$fetchChannelOpenTx (/Users/nymkappa/Documents/mempool/backend/dist/tasks/lightning/sync-tasks/funding-tx-fetcher.js:95:39)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async FundingTxFetcher.$fetchChannelsFundingTxs (/Users/nymkappa/Documents/mempool/backend/dist/tasks/lightning/sync-tasks/funding-tx-fetcher.js:43:13)
    at async LightningStatsImporter.$run (/Users/nymkappa/Documents/mempool/backend/dist/tasks/lightning/sync-tasks/stats-importer.js:21:9)
```

PR #2802, opened 2022-12-06
Fix crash when channel short id is not valid

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2792

### Testing

- Change any `channels.short_id` value in the db to an invalid output index, for example `505149x622x1234`
- Without this patch, the backend will crash at some point
- With this patch, confirm it does not crash anymore
- Don't forget to revert your db change after testing

Example output
```
﻿﻿Dec  6 10:40:59 [97243] INFO: <lightning> Caching funding txs for currently existing channels
Dec  6 10:40:59 [97243] ERR: <lightning> Cannot find blockchain funding tx for channel id 505149x622x1234. Possible reasons are: bitcoin backend timeout or the channel shortId is not valid
```


softsimon (approved):
code ACK

softsimon (approved):
tested ACK @ v2.5.0-dev [[3a7dffbe](https://github.com/mempool/mempool/commit/3a7dffbe)]

`Jan 26 17:12:35 [23114] ERR: <lightning> Cannot find blockchain funding tx for channel id 764621x1790x123. Possible reasons are: bitcoin backend timeout or the channel shortId is not valid`

PR #2809, opened 2022-12-07
fix unstable miner tag styles in block preview

### LINKED ISSUE #2781: Unfurler: normalize and define CSS styles
Different environments now have different styling for the `Unknown` miner tag. 

We should define what the expected style should be so we can add regression tests for the previews:

tk7
fra
prod

mononaut:
Resolves #2781 by fixing ambiguous miner tag style and removing css transitions on that class in the block preview.


softsimon (approved):
CODE ACK

wiz (approved):
Tested ACK @ v2.5.0-dev [[39d92f57](https://github.com/mempool/mempool/commit/39d92f57)]

> I can't explain why the tag would have ever rendered in a larger font, and haven't been able to reproduce that yet.

This is likely due to the actual SSR environment, i.e. FreeBSD Chromium xorg fonts etc.

PR #2810, opened 2022-12-08
Fix & reenable cpfp indexer optimized path

mononaut:
Resolves a lingering issue from #2737 by fixing the broken optimized code path in the CPFP indexer and reenabling it when the `blocks_summaries` table is available.

This option speeds up CPFP indexing by extracting block transactions from raw block data and augmenting with fees from the `blocks_summaries` table, to avoid the expensive verbose block RPC call.

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -742,7 +742,7 @@ class Blocks {
 
   public async $indexCPFP(hash: string, height: number): Promise<void> {
     let transactions;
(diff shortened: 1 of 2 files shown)


wiz:
thank you, currently it takes weeks to run the slow code path 😅

wiz (approved):
Partially Tested ACK @ e986aaf1d906c652b077c2db26fbd8a4374a83fa

Seems OK so far, about 3x or maybe 5x faster than before 👍

PR #2824, opened 2022-12-09
cache, serve & display more comprehensive RBF info

mononaut:
This PR extends the backend RbfCache to
 - track replacement relations in both directions.
 - keep a copy of replaced transactions in memory.
 
and then serves and displays this extra info on the transactions page:

rbf-info

This makes it possible to view details of replaced transactions _after_ they have been replaced, and to follow the history of a chain of RBF replacements back through previous versions.

Previously, RBF data was kept cached for one week from time of replacement. Now data expires 24 hours after the most recent version of a transaction leaves the mempool (either because it was evicted or confirmed).

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -51,8 +51,6 @@ export class Common {
   static findRbfTransactions(added: TransactionExtended[], deleted: TransactionExtended[]): { [txid: string]: TransactionExtended } {
     const matches: { [txid: string]: TransactionExtended } = {};
     deleted
-      // The replaced tx must have at least one input with nSequence < maxint-1 (That’s the opt-in)
-      .filter((tx) => tx.vin.some((vin) => vin.sequence < 0xfffffffe))
       .forEach((deletedTx) => {
         const foundMatches = added.find((addedTx) => {
           // The new tx must, absolutely speaking, pay at least as much fee as the replaced tx.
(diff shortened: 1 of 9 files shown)


0xB10C:
Cool!



nit: Might be clearer to say "This transaction is a replacement of txid" here and not to use "replaced" twice.
image

backend/src/api/common.ts
```diff
@@ -51,8 +51,6 @@ export class Common {
   static findRbfTransactions(added: TransactionExtended[], deleted: TransactionExtended[]): { [txid: string]: TransactionExtended } {
     const matches: { [txid: string]: TransactionExtended } = {};
     deleted
-      // The replaced tx must have at least one input with nSequence < maxint-1 (That’s the opt-in)
-      .filter((tx) => tx.vin.some((vin) => vin.sequence < 0xfffffffe))
```
softsimon, line 55: What implications does this change has? That we are in practice enabling detection of Full RBF transactions?

mononaut: Yes, otherwise we miss Full RBF replacements

mononaut: Yes, otherwise we miss Full RBF replacements

softsimon:
Please rebase

softsimon (approved):
works very well.

tested ACK @ v2.5.0-dev [[e4fcac93](https://github.com/mempool/mempool/commit/e4fcac93)]

PR #2825, opened 2022-12-10
Fix the dependabot ignore settings - RTFM

knorrium:
Updating per [the docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#specifying-dependencies-and-versions-to-ignore)


wiz (approved):
ty sir

PR #2833, opened 2022-12-12
Fix routing for top-nav doc button

hunicus:
Doc link in top nav doesn't work for Liquid and Bisq sites...think it might have something to do with the Angular update, not sure.

To reproduce: run master and click one of these links.

bisq-link-bad

liquid-link-bad


softsimon (approved):
tested ACK @ v2.5.0-dev [[a6110637](https://github.com/mempool/mempool/commit/a6110637)]

Confirmed it was broken, now works on all the sites.

PR #2845, opened 2022-12-14
dual-node support for different RBF policies

mononaut:
_(builds on #2824)_

This PR adds frontend config settings & support for fetching missing transactions from an alternate backend (either a Mempool instance, or esplora server).

e.g.
```
...
"FULL_RBF_ENABLED": false,
"ALT_BACKEND_URL": "https://fullrbf.mempool.space"
...
```

This creates a unified experience for looking up transactions which may exist in either the FullRBF or default mempools.

When a transaction is fetched from the alternate backend, mempool-specific info like estimated confirmation time and position in the mempool is not shown, and an alert message is displayed. 

---

_example: Mempool running FullRBF, transaction only available from the alternate backend with mempoolfullrbf=0_
tx from alt backend

---

_example: Mempool running default policy, transaction only available from the alternate backend with mempoolfullrbf=1_
Screenshot 2022-12-14 at 9 39 40 AM


mononaut:
Up to you! the implementation is fairly flexible, the frontend just needs a `tx/:txid` endpoint to ping for transactions in the alternate mempool in esplora format. (and we only hit that endpoint for missing or unconfirmed transactions, so traffic should be comparatively light).

that could be a standalone esplora instance running the other mempool policy, or a minimal Mempool backend on a pruned node, or a full-blown Mempool deployment.

I think the setup that makes the most sense is to run the main production servers with FullRBF, and add a pruned node Mempool backend per network with database etc disabled to serve the non-FullRBF mempool.

In that case, yes, probably best to just expose that backend on a new `/fullrbf/` url, without the separate hostname.

wiz:
I still don't understand why you are suggesting we should duplicate our entire infrastructure tho - i.e. I don't want to add more mempool instances or esplora backends when we already have a 2nd bitcoin node in place (currently called the minfee node), can't you just query that over bitcoind RPC from the same mempool instance? or am I missing something about how this works?

mononaut:
Yeah, maybe I'm overcomplicating this unnecessarily.

IIRC there was a good reason why we couldn't just use the minfee node, but my original designs were a bit more ambitious, so it's possible that limitation no longer applies with the current feature set.

I'll see if I can make it work.

mononaut:
ok, so I remembered a big part of why I initially designed this to run off a second Mempool backend, which is that getting full transaction information out of another raw core node in the backend is a huge PITA. It means we need to handle fetching prevouts to calculate fees & extend the basic tx info, and maintain another whole in-memory mempool.

I've updated this PR with a first attempt at that, which subclasses `Mempool` to fetch and track the second node's mempool, and serves those transactions at a new `/api/v1/tx/alt/:txid` endpoint.

This requires some high-level changes to the `BitcoinApi` and `Mempool` classes which make me a little nervous, but it seems to work ok so far.

Since we're only interested in mempool transactions, and can fetch any confirmed inputs from the main node for fee calculations, the second core node can be pruned and run with `txindex=0`.

The frontend treatment is the same as before.

wiz:
Thanks for restructuring it. I'll wait for @softsimon's review before testing

PR #1180, opened 2022-01-23
add log priority option for stdout log

antonilol:
add log priority option to stdout log (`console.log`) just like the one for syslog
by default, no logs are suppressed. users can manually suppress debug log by setting this option to `info`

maintainers, if you want another name than `STDOUT_LOG_MIN_PRIORITY`, feel free to change it

got the idea from #569

Issue #2840, opened 2022-12-13
Add [SECOND_]CORE_SSL option.


LimpSquid:
### Description

Currently it is not possible to use Bitcoin Core's RPC API over SSL. 

### Problem to be solved

Allow backend to connect to Bitcoin Core's RPC API over SSL.

### Proposed solution

As far as I can tell the RPC api already has [support](https://github.com/mempool/mempool/blob/a292745ea7bdfe7365b844c23aa8b4f4dc126efb/backend/src/rpc-api/jsonrpc.ts#L8) for SSL, however the backend is just lacking the ability to configure it. We should add a boolean field `ssl` to [BitcoinRpcCredentials](https://github.com/mempool/mempool/blob/6d6dd09d116a5e0d66c363b120eda469e0c90acd/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts#L22) and allow it to be configured with the environment variables `CORE_RPC_SSL` and `SECOND_CORE_RPC_SSL`.

Additionally, we could implement the `sslCa` option. But that's beyond the scope of this issue.

I'm willing to implement the `ssl` option and make a PR if this issue is something that we want to have fixed.

PR #2846, opened 2022-12-14
Add Bitcoin Core RPC cookie authentication option

### LINKED ISSUE #2806: Add support for rpccookiefile in bitcoind.conf
### Description

In order to avoid username/passwords which are not the recommended config for bitcoind it would be great if i could import the rpccookiefile for bitcoind instead of using user/pass

### Problem to be solved

lnd, electrs, and most other tools i use utilize the cookie file to avoid using a static user:pass, but I'm unable to use mempool with this configuration because it appears as though there is no support for the cookie file

### Proposed solution

allow mempool to use the cookie file for bitcoind rpc authentication

antonilol:
still a draft, i know from #1180 that config changes also need changes in docker related files

now setting `COOKIE` to true should be enough on most systems, where bitcoind places the cookie in the expected (network specific) dir (`~/.bitcoin/<network>/.cookie`), otherwise `COOKIE_PATH` will point mempool in the right direction

user/pass auth is still the default though

the default path of the cookie is not set using the normal way (like every other config entry) because it is network specific, it looks like a dirty workaround.

i also refractored the `rpc-api` a bit, this helped me and i think this will also help other working on this project, typescript can now tell you if the rpc name is correct (the diff of jsonrpc.ts is unreadable because it ended up one level of indentation deeper)

@ShadwDrgn @federicociro @BeardMatt (from the linked issue) can you review my choices/code or test it and tell me if this suits your needs?

feedback/tips on how to improve this are all appreciated!

fixes #2806


ShadwDrgn:
haven't tested but based on what i'm  r eading it looks like we just add COOKIE and COOKIE_PATH variables to our compose environment section as needed and ensure the cookie file is mounted. should work for my purposes at least

federicociro:
Will do a review of the change after my holidays

federicociro (commented):
Change approved and performed as expected. Mempool backend connects to bitcoind correctly through .cookie authentication.

antonilol:
just squashed and rebased this, happy new year everyone! do not forget to update your open source licenses' dates :)

federicociro (commented):
How can we push this? Everything is ok.

antonilol:
this needs review from a maintainer
there is a branch conflict, i will rebase in a second

wiz (changes requested):
Concept ACK for adding cookie support, but Concept NACK for rewriting the entire RPC code... most importantly, that is an unrelated change that should not be part of this PR

antonilol:
i separated the two changes in two separate prs, the second pr contains the start of #2840 as this is not necessary for cookie auth

backend/mempool-config.sample.json
```diff
@@ -33,7 +33,9 @@
     "HOST": "127.0.0.1",
     "PORT": 8332,
     "USERNAME": "mempool",
-    "PASSWORD": "mempool"
+    "PASSWORD": "mempool",
+    "COOKIE": false,
+    "COOKIE_PATH": "/path/to/bitcoin/.cookie"
```
antonilol, line 38: it would be great to have cookie work by default, but setting cookie to true here will turn it on for everybody that did not set in in his config, potentially breaking mempool if the cookie is not in the path set here, note that this is also different for different networks (mainnet cookie is in .bitcoin/.cookie, but testnet in .bitcoin/testnet3/.cookie, and not everyone has his .bitcoin in the home directory)

i think i can change the config and sample config merging to allow for some special logic to use username and password if that is configured, cookie if that is configured and the default cookie path for that network if no auth method is set in the config, or maybe even try some known .bitcoin locations to try to find it (some of these full node distros have it on /mnt for example)

backend/src/api/bitcoin/bitcoin-client.ts
```diff
@@ -2,12 +2,15 @@ import config from '../../config';
 const bitcoin = require('../../rpc-api/index');
 import { BitcoinRpcCredentials } from './bitcoin-api-abstract-factory';
 
+export const defaultCookiePath = `${process.env.HOME}/.bitcoin/${{mainnet:'',testnet:'testnet3/',signet:'signet/'}[config.MEMPOOL.NETWORK]}.cookie`;
+
```
wiz, line 6: I don't think we want to do this, you can configure your cookie path in the configuration file and the default should be `/bitcoin/.cookie`

antonilol: this is the cookie location on my device, i run bitcoind and mempool from the same unix account without docker

on which systems is `/bitcoin` the bitcoind data dir?

ShadwDrgn:
I'm curious if this is just being held up about confusion as to where the bitcoin configs are/should be? This merge seems like it's been ready for some time now. Anything I or anyone else can do to help?

backend/src/api/bitcoin/bitcoin-client.ts
wiz: the official mempool.space servers

antonilol: ok, changing this

antonilol:
can this be reviewed/merged?

knorrium:
The tests are failing due to missing docker config keys and script update

```
Mempool Backend Config › should ensure that the mempool-config.json Docker template has all the keys

    expect(received).toContain(expected) // indexOf

    Expected substring: "\"COOKIE\": \"__CORE_RPC_COOKIE__\""
```

```
 ● Mempool Backend Config › should ensure the docker start.sh script has default values

    expect(received).toMatch(expected)

    Expected pattern: /__CORE_RPC_COOKIE_PATH__=\${CORE_RPC_COOKIE_PATH:=(.*)}/
```

antonilol:
fixed and tests pass now

softsimon (approved):
tested ACK @ [06d1bbdf0]

Tested using with and no cookie and with and without on the second core node.

Issue #2849, opened 2022-12-16
Block page layout is broken when `MINING_DASHBOARD: false` in frontend config


nymkappa:
`MINING_DASHBOARD: true`

<img width="979" alt="image" src="https://user-images.githubusercontent.com/9780671/208142293-7800e9fa-a80d-469f-860c-9de4e1e8d9d8.png">

`MINING_DASHBOARD: false`

<img width="991" alt="Screenshot 2022-12-16 at 17 16 13" src="https://user-images.githubusercontent.com/9780671/208142042-b91fe725-1b73-4e9c-afff-31a616e7fc0b.png">

PR #2850, opened 2022-12-16
Remove duplicated html

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2849


softsimon:
This might not be the solution because we still need to support both the old and new view, depending on audit is available or not.

nymkappa:
@mononaut feel free to take a shot at it, you spent more time than me in there I cannot wrap my head around the multiple flags condition

Issue #2735, opened 2022-11-25
Remove pools.json and logos from frontend


softsimon:
We are no longer using `pools.json` and mining pool icons i the frontend

Remove this legacy code from the `sync-assets.json` script.

But make sure it's not used, when mining dashboard is disabled for example.

nymkappa:
The mining pools logo download needs to stay, otherwise where would we fetch those images from?

PR #2852, opened 2022-12-16
Remove pools.json download from the frontend

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2735.

Today we fetch the mining pools.json directly in the backend. With, or without mining dashboard enabled, we can tag pools on the fly when blocks are queried through the nodejs backend. It also works if you have `DATABASE.ENABLED: false`.

Tested with mining dashboard disabled as such:
- Fresh database
- `"MINING_DASHBOARD": false,` in the frontend config
- `"INDEXING_BLOCKS_AMOUNT": 0,` in the backend config


softsimon (approved):
tested ACK @ v2.5.0-dev [[908b92af](https://github.com/mempool/mempool/commit/908b92af)]

PR #2857, opened 2022-12-17
Added syntax highlight

Piterden:


softsimon (approved):
tested ACK ... Looks good on https://github.com/Piterden/mempool/blob/patch-1/docker/README.md

PR #2859, opened 2022-12-20
display block health as badge & add loaders

### LINKED ISSUE #2830: Add skeleton loader to the Block Health field on the block view
When loading a block for the first time, the Block Health field will show `Unknown` briefly then show the actual data when it's available, however when navigating between blocks with the arrow keys, there's a 1-3s delay to update from the previous value to the new one, which can be confusing (i.e, 88.7% <> 100%).

A skeleton loader will indicate that the data is refreshing.

### LINKED ISSUE #2834: Block Health UX change
I am suggesting this change to the block health number, making it pop more.

Screenshot 2022-12-12 at 20 24 13

Screenshot 2022-12-12 at 20 23 59

I am not as sure about this change... a progress bar might make more sense here.. but it would be best to be consistent. Either do the badges here as well, or use progress bars on the block page as well.

Screenshot 2022-12-12 at 20 26 43

mononaut:
Resolves #2834 and resolves #2830 by changing the `block` and `blocks` pages to display the block health field as a colored badge, and adding skeleton loaders.

Badges are color-coded:
 - green >= 99%
 - yellow >= 75%
 - red < 75%

Screenshot 2022-12-19 at 7 04 26 PM
Screenshot 2022-12-19 at 7 04 52 PM

### DIFF
--- a/frontend/src/app/components/block/block.component.html
+++ b/frontend/src/app/components/block/block.component.html
@@ -57,8 +57,21 @@
               <tr *ngIf="auditEnabled">
                 <td i18n="block.health">Block health</td>
                 <td>
-                  <span *ngIf="blockAudit?.matchRate != null">{{ blockAudit.matchRate }}%</span>
-                  <span *ngIf="blockAudit?.matchRate === null" i18n="unknown">Unknown</span>
+                  <span
+                    class="health-badge badge"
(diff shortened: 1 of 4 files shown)


softsimon (approved):
tested ACK @ v2.5.0-dev [[597536ef](https://github.com/mempool/mempool/commit/597536ef)]

PR #2875, opened 2022-12-26
Go to block optimization

softsimon:
Go to `/blocks` and navigate to another page.

Click to go to a random block.

It will now be instant because the block data is passed in the link.


wiz (approved):
Tested ACK @ v2.5.0-dev [[3317f5e6](https://github.com/mempool/mempool/commit/3317f5e6)]

Issue #2877, opened 2022-12-26
Support default empty LN topology


softsimon:
By default. We don't have any "topology data", and the setting `TOPOLOGY_FOLDER` is an empty string by default.

We should simply ignore any attempt of doing a historical data import if this is not specifically set. Otherwise this will always be the default error log:

```
Dec 27 03:03:54 [84832] ERR: <lightning> Unable to open topology folder at 
Dec 27 03:03:54 [84832] ERR: <lightning> Lightning network stats historical failed. Reason: ENOENT: no such file or directory, scandir ''
```

PR #2883, opened 2022-12-28
Don't try to import LN historical stats if no topology folder is set

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2877

```
Dec 28 12:11:58 [82601] INFO: <lightning> Lightning topology folder is not set. Not importing historical LN stats
```


softsimon (approved):
tested ACK @ a975936

PR #2884, opened 2022-12-28
Infinitely scrolling blockchain

mononaut:
This PR implements virtual scrolling to extend the blockchain bar beyond the last 8 blocks, and reveal the entire chain.

https://user-images.githubusercontent.com/83316221/209808726-395a32ce-f29f-4000-b31e-2bb5bca803ca.mov

Blocks are rendered in up to three 'pages', each sized to fill the width of the window. As the user scrolls, pages are loaded and unloaded, and the scroll position reset, to create the illusion of a seamless continuously scrollable chain.

Block data is fetched in chunks from the `/blocks/:height` endpoint on demand, with a layer of client-side caching to minimize loading when scrolling back and forth or returning to the tip of the chain.

Navigating to a block or transaction page shows the relevant block in the blockchain bar.


wiz:
video looks cool, but is it possible to make the block appear full or empty using the block size data?

mononaut:
yeah, it uses all of the same display logic as before for the block elements themselves, so they're filled according to block weight. It's not obvious in the video because most of those blocks happened to be completely full.

Screenshot 2022-12-28 at 11 51 52 AM

wiz:
amazing, this is one of those things that feels like it should have been added years ago

wiz (approved):
Tested ACK @ v2.5.0-dev [[b2b89110](https://github.com/mempool/mempool/commit/b2b89110)]

Minor issue found when holding down left arrow and stress test scrolling for a while, seems like some kind of race condition where page crashes with the following stack trace
Screenshot 2023-01-12 at 22 11 17

PR #2899, opened 2023-01-03
toggle to enable/disable block audits

mononaut:
Adds a toggle button to the block page to show/hide block audits (if available).

Audits are shown by default, and user's preference is cached to localStorage.

audit-toggle

### DIFF
--- a/frontend/src/app/components/block/block.component.html
+++ b/frontend/src/app/components/block/block.component.html
@@ -23,12 +23,19 @@
 
     <div class="grow"></div>
 
+    <div class="audit-toggle" *ngIf="webGlEnabled && auditAvailable">
+      <app-toggle
+        textLeft="audit"
+        i18n-textLeft="block.audit-mode"
+        [checked]="auditModeEnabled"
+        (toggleStatusChanged)="toggleAuditMode($event)"
+      ></app-toggle>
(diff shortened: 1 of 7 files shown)


wiz:
please rebase

mononaut:
rebased on current master

softsimon:
When audit is off, the Block Details button is now too far down from the block dialog.

Screenshot 2023-01-17 at 03 05 16

Personally I don't think it looks good enough to see the new audit text and toggle on top of every block page.

May I suggest move it to the bottom, to the left of the details button perhaps. And also even turn the text/toggle into a Bootstrap style toggle button `btn btn-outline-info btn-sm active` when it's enabled. (The details button should also add `active` when it's enabled/expanded

Screenshot 2023-01-17 at 03 10 37

softsimon (changes requested):
See comment above

mononaut:
I've pushed an update with those changes:

Screenshot 2023-01-26 at 11 08 50 AM
Screenshot 2023-01-26 at 11 08 40 AM

softsimon (approved):
tested ACK @ v2.5.0-dev [[714700d8](https://github.com/mempool/mempool/commit/714700d8)]

PR #2901, opened 2023-01-03
Fiat currency options

mononaut:
This PR allows users to choose their preferred currency for fiat conversions & statistics via a dropdown on the dashboard page (next to the language selector).

All real-time conversions are then displayed in the selected currency.

The `mining/block-fees` and `mining/block-rewards` graphs will also display fiat amounts in that currency if the data is available, or in USD otherwise (we support fewer currencies for historic data than real-time rates).

Preferences are cached to localStorage, and default to USD if no other preference is set. We could perhaps guess a better default from HTTP headers & geoIP lookup?


wiz:
@mononaut changing the selector causes the dashboard fee estimates to update, but it seems transactions are still displayed using USD all the time... 
Screenshot 2023-01-15 at 1 38 05

mononaut:
Fixed
Screenshot 2023-01-16 at 11 08 09 AM

wiz:
Could you please resolve the merge conflict

mononaut:
rebased on current master

wiz:
since @nymkappa implemented the new fiat price service I'm going to ask him to review and finish up this PR so it can get ready for merging - i think there are some fiat currencies in this PR that we don't have data for (?)

PR #2914, opened 2023-01-06
dynamic CSS text truncation component

mononaut:
This PR adds a new standardized component for dynamically truncating the middle section of long strings like txids, bitcoin addresses etc, and updates most instances of truncation in the frontend to use the new component.

The component:
 - uses CSS `text-overflow: ellipsis`, so that the full text is preserved for clipboard functionality, screen readers etc.
 - expands to fit available space.
 - correctly handles RTL locales.
 
 
Screenshot 2023-01-06 at 10 33 55 AM


softsimon (changes requested):
On the latest transactions table the ellipsis should be more in the middle instead of only showing ~3 characters on the left side. 

Screenshot 2023-01-12 at 18 17 40

This might be hard or impossible to fix but I wish there was no "gap" appearing in certain resolutions and widths.

Screenshot 2023-01-12 at 18 18 19

mononaut:
I've shifted the truncation a bit closer to the middle:
Screenshot 2023-01-13 at 5 02 41 PM

The gaps are unfortunately unavoidable with this method, especially using a variable-width font.

wiz:
is it possible to center the ellipsis in its own <span> or something

wiz (approved):
Tested ACK @ v2.5.0-dev [[c69f2f2b](https://github.com/mempool/mempool/commit/c69f2f2b)]

softsimon (approved):
tested ACK @ v2.5.0-dev [[c69f2f2b](https://github.com/mempool/mempool/commit/c69f2f2b)]

Issue #2904, opened 2023-01-04
Block indexing should handle Coinbase tx with Esplora


softsimon:
There is a gotcha with Esplora, it cannot fetch the unspendable coinbase transaction, so our block indexing need to handle it.

```
Jan  4 16:25:15 [59721] INFO: <lightning> Indexing 1 blocks from #334 to #0
Jan  4 16:25:15 [59721] DEBUG: <lightning> Indexing block #0 | ~1.00 blocks/sec | total: 770335/770335 (100%) | elapsed: 4 seconds

Jan  4 16:26:12 [59721] ERR: <lightning> Block indexing failed. Trying again in 10 seconds. Reason: Cannot fetch coinbase tx 4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b. Reason: Request failed with status code 404
```

Lost4urluv7:
> There is a gotcha with Esplora, it cannot fetch the unspendable coinbase transaction, so our block indexing need to handle it.
> 
> ```
> Jan  4 16:25:15 [59721] INFO: <lightning> Indexing 1 blocks from #334 to #0
> Jan  4 16:25:15 [59721] DEBUG: <lightning> Indexing block #0 | ~1.00 blocks/sec | total: 770335/770335 (100%) | elapsed: 4 seconds
> 
> Jan  4 16:26:12 [59721] ERR: <lightning> Block indexing failed. Trying again in 10 seconds. Reason: Cannot fetch coinbase tx 4a5e1e4baab89f3a32518a88c31bc87f618f76673e2cc77ab2127b7afdeda33b. Reason: Request failed with status code 404
> ```

#2865

nymkappa:
@mononaut Does this bug impact your different indexing?
If not, maybe https://github.com/mempool/mempool/pull/2916 fixes this issue.

wiz:
I can't remember ever having this issue on the production servers. Was it introduced recently?

PR #2916, opened 2023-01-07
Fetch coinbase tx with core when esplora fails to do so

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2904 for the block indexing

We basically retry to fetch the coinbase tx with core when esplora fails to return it.


softsimon (approved):
tested ACK @ 71d1c3

`Jan  8 16:45:11 [97091] NOTICE: <lightning> [Mining] Block indexing completed: indexed 31 blocks`

PR #2917, opened 2023-01-07
Show `Avg fees per block` instead of `Reward Per Tx`

nymkappa:
Currently we are showing the average `Reward Per Tx` for the past 144 blocks. I don't think this metric gives any useful information and I've put it there months ago due to the lack of better ideas.

One of the oldest debate around Bitcoin is about miner's subsidies over the long term, so I though it would be useful to show the average fees per block instead. It does not require any backend change since we already have the data in the frontend.

### Currently

image

### Updated

image

### DIFF
--- a/frontend/src/app/components/reward-stats/reward-stats.component.html
+++ b/frontend/src/app/components/reward-stats/reward-stats.component.html
@@ -13,21 +13,21 @@
       </div>
     </div>
     <div class="item">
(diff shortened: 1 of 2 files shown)


softsimon (approved):
Tested ACK @ v2.5.0-dev [[8a3bcd3b](https://github.com/mempool/mempool/commit/8a3bcd3b)]

I agree this is more informative

PR #2922, opened 2023-01-08
Remove Cypress from the toplevel package-lock.json deps

knorrium:


softsimon:
Won't a change to package log be overridden next time a change is made to package.json and "npm install" generates a new one?

knorrium:
> Won't a change to package log be overridden next time a change is made to package.json and "npm install" generates a new one?

Cypress is already set as an optional dependency so I'm not sure how it ended up on the top level. I did `npm install` again and it wasn't added back :man_shrugging:

PR #2937, opened 2023-01-10
Fallback alias name to pubkey

### LINKED ISSUE #2935: If the node does not have an alias, we should fallback to a truncated pubkey
image

For example, in the node channel list component, if we don't have an alias, we show the truncated pubkey
image

softsimon:
fixes #2935

Before
Screenshot 2023-01-10 at 21 28 05

After
Screenshot 2023-01-10 at 21 26 59

### DIFF
--- a/frontend/src/app/components/address-labels/address-labels.component.ts
+++ b/frontend/src/app/components/address-labels/address-labels.component.ts
@@ -36,7 +36,9 @@ export class AddressLabelsComponent implements OnChanges {
 
   handleChannel() {
     const type = this.vout ? 'open' : 'close';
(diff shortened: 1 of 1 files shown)


nymkappa (approved):
Tested ACK @ 8ebe04baa7168deaa3348142dcbfd51e4dc26ac4 with tx [ff193215d9c4690e759613110940610ceec03eea93335664f6d9ce952eb6550e](https://mempool.space/tx/ff193215d9c4690e759613110940610ceec03eea93335664f6d9ce952eb6550e)

### Before

Screenshot 2023-01-10 at 18 54 37

### After

Screenshot 2023-01-10 at 18 55 21

PR #2939, opened 2023-01-11
CPFP Optimizations

mononaut:
This PR optimizes various aspects of the CPFP indexing process:
 - New compact db table schemas with ~70% reduction in storage requirements.
 - Slightly faster indexing.
 - Eliminates dependence on the blocks table.
 - Skips clusters of transactions with the same fee rates.
 - Removes unnecessary foreign key constraints which should make ops tasks like truncating/exporting/importing these tables less annoying.

And
 - Replaces the slow JOIN query at the `/api/v1/cpfp` endpoint with a fast two-part lookup.

The new schemas store txids as fixed-length binary columns, and replace the JSON cluster transactions column with a compact binary blob.

CPFP data is now inserted in batches instead of individual rows.

The database migration includes a step to copy over any already-indexed CPFP data from the old tables. This could take several hours, but it's much faster than reindexing from the beginning. 
**However, it will delay the backend coming online and result in some downtime**. If that's not acceptable, I expect I could shift this step into a background process.
Old data should be preserved until the migration completes cleanly, but it might be a good idea to back up the `cpfp_clusters` and `transactions` tables before running this, just in case.

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.ts
+++ b/frontend/src/app/components/transaction/transaction.component.ts
@@ -131,8 +131,11 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
           this.cpfpInfo = null;
           return;
         }
-
-        const relatives = [...cpfpInfo.ancestors, ...cpfpInfo.descendants || [cpfpInfo.bestDescendant]];
+        // merge ancestors/descendants
+        const relatives = [...(cpfpInfo.ancestors || []), ...(cpfpInfo.descendants || [])];
+        if (cpfpInfo.bestDescendant && !cpfpInfo.descendants?.length) {
+          relatives.push(cpfpInfo.bestDescendant);
+        }
         let totalWeight =
           this.tx.weight +
           relatives.reduce((prev, val) => prev + val.weight, 0);
--- a/frontend/src/app/components/transaction/transaction.component.ts
+++ b/frontend/src/app/components/transaction/transaction.component.ts
@@ -131,26 +131,17 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
           this.cpfpInfo = null;
           return;
         }
-        if (cpfpInfo.effectiveFeePerVsize) {
-          this.tx.effectiveFeePerVsize = cpfpInfo.effectiveFeePerVsize;
-        } else {
-          const lowerFeeParents = cpfpInfo.ancestors.filter(
-            (parent) => parent.fee / (parent.weight / 4) < this.tx.feePerVsize
-          );
-          let totalWeight =
-            this.tx.weight +
-            lowerFeeParents.reduce((prev, val) => prev + val.weight, 0);
-          let totalFees =
-            this.tx.fee +
-            lowerFeeParents.reduce((prev, val) => prev + val.fee, 0);
 
-          if (cpfpInfo?.bestDescendant) {
-            totalWeight += cpfpInfo?.bestDescendant.weight;
-            totalFees += cpfpInfo?.bestDescendant.fee;
-          }
+        const relatives = [...cpfpInfo.ancestors, ...cpfpInfo.descendants || [cpfpInfo.bestDescendant]];
+        let totalWeight =
+          this.tx.weight +
+          relatives.reduce((prev, val) => prev + val.weight, 0);
+        let totalFees =
+          this.tx.fee +
+          relatives.reduce((prev, val) => prev + val.fee, 0);
 
-          this.tx.effectiveFeePerVsize = totalFees / (totalWeight / 4);
-        }
+        this.tx.effectiveFeePerVsize = totalFees / (totalWeight / 4);
+
         if (!this.tx.status.confirmed) {
           this.stateService.markBlock$.next({
             txFeePerVSize: this.tx.effectiveFeePerVsize,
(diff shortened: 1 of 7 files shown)


frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -131,26 +131,17 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
...
-            lowerFeeParents.reduce((prev, val) => prev + val.fee, 0);
-
-          if (cpfpInfo?.bestDescendant) {
-            totalWeight += cpfpInfo?.bestDescendant.weight;
-            totalFees += cpfpInfo?.bestDescendant.fee;
-          }
 
-          this.tx.effectiveFeePerVsize = totalFees / (totalWeight / 4);
-        }
+        const relatives = [...cpfpInfo.ancestors, ...cpfpInfo.descendants || [cpfpInfo.bestDescendant]];
```
softsimon, line 135: I get a crash here

`transaction.component.ts:264 ERROR TypeError: cpfpInfo.ancestors is not iterable`

mononaut: oops, looks like I forgot `cpfpInfo.ancestors` could be null.

I've just pushed a fix.

softsimon (approved):
tested ACK @ v2.5.0-dev [[2548d2a5](https://github.com/mempool/mempool/commit/2548d2a5)]

wiz (changes requested):
Tested NACK @ bd30f2eb12f26ccd5e55d874bb771744e7b29572 due to 2 issues below:

1) liquid is broken
```
Jan 11 10:44:01 localhost  mempool[89532]: NOTICE <liquid> Starting Mempool Server... (1c2d05f)
Jan 11 10:44:01 localhost  mempool[89532]: INFO <liquid> Downloading external asset icons.json over clearnet...
Jan 11 10:44:01 localhost  mempool[89532]: INFO <liquid> External asset icons.json saved to ./icons.json
Jan 11 10:44:01 localhost  mempool[89532]: INFO <liquid> Restoring mempool and blocks data from disk cache
Jan 11 10:44:01 localhost  mempool[89532]: DEBUG <liquid> Mempool blocks calculated in 0 seconds
Jan 11 10:44:01 localhost  mempool[89532]: INFO <liquid> Database connection established.
Jan 11 10:44:01 localhost  mempool[89532]: DEBUG <liquid> MIGRATIONS: Running migrations
Jan 11 10:44:01 localhost  mempool[89532]: DEBUG <liquid> MIGRATIONS: Database engine version '10.5.17-MariaDB'
Jan 11 10:44:01 localhost  mempool[89532]: DEBUG <liquid> MIGRATIONS: Current state.schema_version 49
Jan 11 10:44:01 localhost  mempool[89532]: DEBUG <liquid> MIGRATIONS: Latest DatabaseMigration.version is 52
Jan 11 10:44:01 localhost  mempool[89532]: DEBUG <liquid> MIGRATIONS: Execute query: ALTER TABLE `blocks` DROP COL
UMN `cpfp_indexed`
Jan 11 10:44:01 localhost  mempool[89532]: ERR <liquid> MIGRATIONS: Unable to create required tables, aborting in
10 seconds. Error: Can't DROP COLUMN `cpfp_indexed`; check that it exists
```

2) drop table fail
```
Jan 11 10:44:30 localhost  mempool[17589]: DEBUG <testnet-lightning> MIGRATIONS: Execute query: DROP TABLE IF EXISTS `cpfp_clusters`
Jan 11 10:44:30 localhost  mempool[17589]: WARN <testnet-lightning> Cannot delete or update a parent row: a foreign key constraint fails
Jan 11 10:44:30 localhost  mempool[17589]: NOTICE <testnet-lightning> MIGRATIONS: Upgrading database schema version number to 52
Jan 11 10:44:30 localhost  mempool[17589]: DEBUG <testnet-lightning> MIGRATIONS: Execute query: START TRANSACTION;
Jan 11 10:44:30 localhost  mempool[17589]: DEBUG <testnet-lightning> MIGRATIONS: Execute query: UPDATE state SET number = 52 WHERE name = 'schema_version';
Jan 11 10:44:30 localhost  mempool[17589]: DEBUG <testnet-lightning> MIGRATIONS: Execute query: COMMIT;
```

wiz:
Also worth noting, the DB schema version was still incremented for case 2 despite the query failing 😅

wiz:
some more notes from testing so far:

```
root@localhost [mempool]> select hex(cluster) from compact_transactions where txid = unhex('000000197DE4DBD58E07BA8398C13B3BFF8
+------------------------------------------------------------------+
| hex(cluster)                                                     |
+------------------------------------------------------------------+
| 3508140BDBDC3E505E3A50EE934597A5E27D221CD38284FEB6F39EA7C437008E |
+------------------------------------------------------------------+
1 row in set (0.000 sec)
```
speed is instant despite there being 150M rows
disk usage for transactions table is 37G -> 12G
disk usage for cpfp_clusters table is 19G -> 11G

wiz (approved):
Tested ACK @ 2548d2a5e9add8f72d108c4c125b1a88e051f230

Previous liquid issue was due to corrupted DB, can disregard
The `drop table` still fails tho, but this doesn't cause any actual issue so we can merge this PR and fix later in separate PR

PR #2940, opened 2023-01-11
Expose whether Lightning is enabled on the backend

knorrium:
Adds the `lightning` field to the `/backend-info` response, in preparation for the "one-frontend-multiple-backends" Docker configuration.


softsimon (approved):
tested ACK @ v2.5.0-dev [[5379ec0f](https://github.com/mempool/mempool/commit/5379ec0f)]

PR #2943, opened 2023-01-12
Node AS table overflow fix

softsimon:
Fixes the following overflow that happens on Node [ 023b3ec101c9ab4e060b345d7d6b1728f5ca2db1e21bda9a88f2b9f6b1a3825ef9 ] when AS is present but no Sockets.

Screenshot 2023-01-12 at 20 48 34


wiz (approved):
Tested ACK @ v2.5.0-dev [[2187d96c](https://github.com/mempool/mempool/commit/2187d96c)]

PR #2951, opened 2023-01-13
Fix page elements obscured by bottom navbar on mobile

mononaut:
Fixes an issue on mobile devices where the lower navbar covers content at the bottom of various pages.

e.g pagination controls were obscured on the blocks page:

| Before | After |
|-|-|
| localhost_4200_block_00000000000000000003018b8b678f5327b697908ffa30e8ae0fcf9283161056(iPhone SE) (1) | localhost_4200_block_00000000000000000003018b8b678f5327b697908ffa30e8ae0fcf9283161056(iPhone SE)


wiz (changes requested):
While this fixes the issues on the blocks page, it actually makes the issue worse on some places like the graphs page or the Lightning dashboard "Connect to our nodes" are even lower now 😅 

Can you go through all the pages using mobile layout and try to fix them all? 

Screenshot 2023-01-16 at 16 49 52

mononaut:
Sorry! I was inadvertently testing with `OFFICIAL_MEMPOOL_SPACE=false`, so the LN dashboard didn't have "connect to our nodes", and I didn't realize the graph pages all use different wrappers.

I think it should be fixed on every page now.

wiz (approved):
Tested ACK @ 2e7a701ca7bb6be7368e370b73875e7d21b4fd00

PR #2956, opened 2023-01-16
Fix missing clipboard buttons

mononaut:
Restores the clipboard copy buttons on txids, addresses, block hashes etc that were accidentally removed by #2914


softsimon:
It looks like the copy button works, but it is not giving the tooltip feedback like before

softsimon (approved):
tested ACK @ v2.5.0-dev [[dfd1de67](https://github.com/mempool/mempool/commit/dfd1de67)]

PR #2957, opened 2023-01-16
Support inner links in truncated string component

mononaut:
Adds support for wrapping only the truncated text in a `routerLink` tag, instead of the whole `app-truncate` component, and applies this everywhere truncated links are used.

e.g,

**before**: (link surface extends over the clipboard button to the end of the line)

https://user-images.githubusercontent.com/83316221/212778227-240f2b0e-4e5f-4a7a-abfe-ecf90489e9bf.mp4

**after**: (link surface limited to the actual link text)

https://user-images.githubusercontent.com/83316221/212778260-4dc3eced-41d0-42e4-bb8f-5decbacdcfbc.mp4

### DIFF
--- a/frontend/src/app/bisq/bisq-address/bisq-address.component.html
+++ b/frontend/src/app/bisq/bisq-address/bisq-address.component.html
@@ -1,11 +1,9 @@
 <div class="container-xl">
   <h1 i18n="shared.address">Address</h1>
   <span class="address-link">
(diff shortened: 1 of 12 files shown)


softsimon (approved):
works great.

tested ACK @ v2.5.0-dev [[73f2d54a](https://github.com/mempool/mempool/commit/73f2d54a)]

PR #2959, opened 2023-01-18
reduce prominence of marginal fee rate audit txs

mononaut:
Changes the coloring of 'marginal fee rate' transactions in the block audit visualizations from pink/blue to a highly desaturated version of the normal fee colors.

Hopefully this makes the block audits look less dramatic when there are lots of same-fee-rate substitutions (e.g. when an exchange dumps multiple blocks worth of identical txs into the mempool), and draw attention towards more interesting transactions.

| Before | After |
|-|-|
| Screenshot 2023-01-18 at 5 12 52 PM | Screenshot 2023-01-18 at 5 12 37 PM |


softsimon (approved):
Looks better!

tested ACK @ v2.5.0-dev [[ec6c96c9](https://github.com/mempool/mempool/commit/ec6c96c9)]

Screenshot 2023-01-21 at 17 37 40

After


Screenshot 2023-01-21 at 17 37 44

PR #2960, opened 2023-01-19
Fix mempool block gradients

### LINKED ISSUE #2938: Mempool block feeRange error
With `ADVANCED_GBT_MEMPOOL` enabled. The Mempool blocks simetimes get bad feeRange data with a high fee transaction sorted in the middle.

Data from websocket:
Screenshot 2023-01-10 at 22 40 20

Result:
Screenshot 2022-12-26 at 17 29 33

mononaut:
Fixes #2938.

Transactions in mempool blocks projected by the advanced GBT algorithm are not necessarily included precisely in order of decreasing effective fee rate, due to topological ordering constraints and varying transaction sizes.

However, our fee range calculations for the mempool block gradients assume that block transaction effective fee rates are monotonically decreasing.

As a result, sometimes the block gradients have random bands of the "wrong" color, where we've inadvertently sampled a transaction with a much higher or lower effective fee rate than the surrounding neighborhood.

This PR filters out transactions with anomalous fee rates before calculating fee ranges, to ensure that the mempool block gradients are always smooth and continuous.

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -35,16 +35,25 @@ export class Common {
   }
 
   static getFeesInRange(transactions: TransactionExtended[], rangeLength: number) {
-    const arr = [transactions[transactions.length - 1].effectiveFeePerVsize];
+    const filtered: TransactionExtended[] = [];
+    let lastValidRate = Infinity;
+    // filter out anomalous fee rates to ensure monotonic range
+    for (const tx of transactions) {
+      if (tx.effectiveFeePerVsize <= lastValidRate) {
+        filtered.push(tx);
+        lastValidRate = tx.effectiveFeePerVsize;
+      }
+    }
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK @ v2.5.0-dev [[fc0af50a](https://github.com/mempool/mempool/commit/fc0af50a)]

Tried for a day and it seems fixed and doesn't cause any problems.

But let me just ask @mononaut before I merge. Why can't a simple sort by fee an easier solution? Because after the GBT algoritm I believe the transactions are not sorted by effective fee.

mononaut:
Yes, a simple sort would work too. I opted for filtering because

 a) It should be marginally more efficient, and
 
 b) I think it'll represent actual market fee rates slightly better. Some of the anomalies I'm filtering out can bias the fee distribution towards higher rates at the upper end of block because of quirks in how the GBT algorithm works.

PR #2974, opened 2023-01-26
Fix node chart legend position error

softsimon:


wiz (approved):
Tested ACK @ v2.5.0-dev [[1cf3e181](https://github.com/mempool/mempool/commit/1cf3e181)]

PR #2981, opened 2023-01-27
Fix for disabling block audit below block height

softsimon:


wiz (approved):
Tested ACK v2.5.0-dev [2c2003af]

PR #2987, opened 2023-01-28
Fix blockchain gaps when KEEP_BLOCKS_AMOUNT > INITIAL_BLOCKS_AMOUNT

### LINKED ISSUE #2985: Liquid network infinity scrolling blockchain support
Set frontend config "BASE_MODULE": "liquid", to reproduce

Screenshot 2023-01-28 at 03 22 55

mononaut:
Fixes #2985, which produces an unexpected gap in the scrollable blockchain when KEEP_BLOCKS_AMOUNT in the frontend is set higher than INITIAL_BLOCKS_AMOUNT in the backend (as it is on liquid.network, for example).


softsimon (approved):
tested ACK @ v2.5.0-dev [[922dc5bd](https://github.com/mempool/mempool/commit/922dc5bd)]

PR #2988, opened 2023-01-28
Fix navigation-breaking JS error on liquid block page

mononaut:
Fixes an bug in the `ngOnDestroy` hook on the block component in Liquid, which interrupts navigating away from that page.

To reproduce the bug: Running in Liquid mode, open any block page, then try to navigate away to a different (non-block) page. Navigation fails with an error in the console.


softsimon (approved):
tested ACK @ v2.5.0-dev [[05e23b05](https://github.com/mempool/mempool/commit/05e23b05)]

PR #2996, opened 2023-01-29
Block health string update

softsimon:
It's weird to call it "Block health" on the block page. We don't call it "Block weight" etc.

So now re-using the same i18n-string we're using on the block list page... simply "Health".

Screenshot 2023-01-29 at 22 00 22


wiz (changes requested):
atest should be latest, eh?

softsimon:
> atest should be latest, eh?

Yes. Force pushed.

PR #2999, opened 2023-01-30
fix block visualization getting out of sync

mononaut:
Fixes an issue where the projected block visualization could get out of sync over time, e.g. gradually accumulating transactions which should have been removed.

This was due to overwriting the current mempool blocks with templates calculated for block audits. Normally this would have no effect, but in some circumstances (particularly when using different algorithms for audits and regular mempool updates) it could produce inconsistent mempool 'deltas'.

### DIFF
--- a/backend/src/api/mempool-blocks.ts
+++ b/backend/src/api/mempool-blocks.ts
@@ -33,7 +33,7 @@ class MempoolBlocks {
     return this.mempoolBlockDeltas;
   }
 
-  public updateMempoolBlocks(memPool: { [txid: string]: TransactionExtended }): void {
+  public updateMempoolBlocks(memPool: { [txid: string]: TransactionExtended }, saveResults: boolean = false): MempoolBlockWithTransactions[] {
     const latestMempool = memPool;
     const memPoolArray: TransactionExtended[] = [];
     for (const i in latestMempool) {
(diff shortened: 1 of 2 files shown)


wiz (changes requested):
Tested NACK @ v2.5.0-dev [[b2251d99](https://github.com/mempool/mempool/commit/b2251d99)]

Hmm, it still seems broken. For example if I load 2 instances of the same server side-by-side, in lab conditions (direct fiber to server) then it immediately looks out of sync. I would expect them to always be identical:

Screenshot 2023-01-31 at 21 32 41

mononaut:
The sets of transactions included in the visualization should be identical, but the arrangement of those transactions will be different unless both clients managed to start their `'projected-block-transactions'` websocket subscriptions simultaneously.

The issue this PR aims to fix is the client's transaction set getting out of sync with the templates on the server. Usually it manifests as extra transactions that accumulate in the visualization when it's left running for many blocks, and disappear after refreshing the page or closing and reopening the visualization.

wiz:
I see what you mean, if I open 2 tabs at the same time they stay in sync, but it still seems something is broken because if you for example reload the page the layout of the transactions should not change arbitrarily

wiz (approved):
Tested ACK @ v2.5.0-dev [[74c5e4af](https://github.com/mempool/mempool/commit/74c5e4af)]

softsimon (approved):
utACK

PR #3007, opened 2023-02-01
Fix blockchain scroll jumping on resize on mobile

mononaut:
This PR fixes a bug in the scrollable blockchain which causes the position to jump back one block when the window resizes on mobile (as it does when the address bar is shown/hidden while scrolling up and down in some browsers).

Before:

https://user-images.githubusercontent.com/83316221/216127976-b5fae239-0945-42b5-977b-7ddcba21d1d1.mp4

After:

https://user-images.githubusercontent.com/83316221/216128498-5936007a-f7b3-47f6-acc7-a301cd3d54e1.mp4


softsimon (approved):
tested ACK @ v2.5.0-dev [[0147af132](https://github.com/mempool/mempool/commit/0147af132)]

PR #3010, opened 2023-02-02
More robust error checking & handling in CPFP repositories

mononaut:
Adds some missing checks and exception handling code to the `CpfpRepository` and `TransactionRepository` classes, to deal with edge cases which should normally never arise, but _may_ occur after meddling with the database (e.g. manually deleting rows or copying tables in from another instance)


softsimon (approved):
tested ACK @ v2.5.0-dev [[900e66ae](https://github.com/mempool/mempool/commit/900e66ae)]

PR #3017, opened 2023-02-06
Clear querlist in component destructor to prevent memory leaks

Arooba-git:
Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications and their libraries. While running the tool and analyzing the code of mempool, we saw that your project does a very good job of ensuring that all async operations are cancelled when components destroy. However, as per Memlab execution results, we found some uncleared querylists causing the memory to leak (screenshots below).

[before]
before collec Screen Shot 2023-02-07 at 5 30 41 AM

Hence we added the fix by clearing the querlist on **api-docs** component unload, and you can see the # of leaks reducing noticeably:
 <br />

after collec Screen Shot 2023-02-07 at 5 37 10 AM

You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering the maximum # of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in txt form):
[mempool-scenario-memlab.txt](https://github.com/mempool/mempool/files/10669066/mempool-scenario-memlab.txt)

Note that some other reported leaks (in Memlab) originated from Angular's internal objects, and hence were ignored.

### DIFF
--- a/contributors/Arooba-git.txt
+++ b/contributors/Arooba-git.txt
@@ -1,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of January 25, 2022.
+
+Signed: Arooba-git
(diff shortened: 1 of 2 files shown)


softsimon (changes requested):
tested nACK @ v2.5.0-dev [[7343b46e](https://github.com/mempool/mempool/commit/7343b46e)]

I get a new JS error when navigating away from the FAQ page

Screenshot 2023-02-14 at 13 46 23

Arooba-git:
> querylists

Right, it might be that some other component in the module still needs access to it. In that case, this PR can be closed.

PR #3022, opened 2023-02-08
Reuse HTTP connections for esplora backend requests

mononaut:
The wrapper for the Esplora API currently creates a new HTTP connection for each request.

As a result, if the backend makes a very large number of Esplora requests in a short time, some nodejs resources can become saturated. When this happens subsequent requests hang for up to 10 seconds until they timeout, halting progress in the main loop.

This should be a rare occurrence in normal operation, but it's easily triggered during the initial sync when starting up a cold backend if either the mempool or recent blocks are congested.

This PR updates the Esplora wrapper to configure an httpAgent for axios with `Keep-Alive` enabled, which allows HTTP connections to the Esplora backend to be reused and prevents resource saturation.

It also speeds up requests to the Esplora API a bit by eliminating the overhead of constantly opening and closing connections.

### DIFF
--- a/backend/src/api/bitcoin/esplora-api.ts
+++ b/backend/src/api/bitcoin/esplora-api.ts
@@ -1,7 +1,12 @@
 import config from '../../config';
 import axios, { AxiosRequestConfig } from 'axios';
+import http from 'http';
 import { AbstractBitcoinApi } from './bitcoin-api-abstract-factory';
 import { IEsploraApi } from './esplora-api.interface';
+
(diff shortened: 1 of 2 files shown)


wiz:
Excellent find, yeah this is certainly affecting performance on production servers

softsimon (approved):
tested ACK @ v2.5.0-dev [[d240503f](https://github.com/mempool/mempool/commit/d240503f)]

Issue #3029, opened 2023-02-10
Advanced GBT mempool crash on block getting mined


wiz:
Running 241850b in prod crashed just now:
```
Feb 10 10:42:37 [46939] DEBUG: New block found (#775870)!
Feb 10 10:42:38 [46939] DEBUG: Indexing tx 1 of 3032 in block #775870
Feb 10 10:42:38 [46939] DEBUG: 3031 of 3032 found in mempool. 1 fetched through backend service.
/mempool/mainnet/backend/dist/api/websocket-handler.js:416
        const auditMempool = (config_1.default.MEMPOOL.ADVANCED_GBT_AUDIT === config_1.default.MEMPOOL.ADVANCED_GB
T_MEMPOOL) ? _memPool : JSON.parse(JSON.stringify(_memPool));
                                                                                                                  
                                        ^

RangeError: Invalid string length
    at JSON.stringify (<anonymous>)
    at WebsocketHandler.handleNewBlock (/mempool/mainnet/backend/dist/api/websocket-handler.js:416:155)
    at /mempool/mainnet/backend/dist/api/blocks.js:467:78
    at Array.map (<anonymous>)
    at Blocks.$updateBlocks (/mempool/mainnet/backend/dist/api/blocks.js:467:66)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Server.runMainUpdateLoop (/mempool/mainnet/backend/dist/index.js:182:13)


```

nymkappa:
Maybe related? https://stackoverflow.com/questions/29175877/json-stringify-throws-rangeerror-invalid-string-length-for-huge-objects

nymkappa:
Fixed in https://github.com/mempool/mempool/pull/3030 ?

wiz:
Fixed in #3030 !

PR #3030, opened 2023-02-10
safer mempool cloning for different GBT algorithms

mononaut:
Short term fix for #3029 - backend crashing on new block with very large mempool, when ADVANCED_GBT_MEMPOOL and ADVANCED_GBT_AUDIT are different.

This appears to be due to the `JSON.stringify` serialization used to clone the mempool exceeding the maximum nodejs string size of 512mb.

This PR replaces `JSON.parse(JSON.stringify(...))` with a simple recursive deep cloning method. It's much slower than `JSON.parse(JSON.stringify(...))`, but should cope with significantly larger mempool sizes.

Tested on the current mainnet mempool of >285MB / ~34k transactions.


wiz (approved):
Tested ACK @ v2.5.0-dev [[8f3b8276](https://github.com/mempool/mempool/commit/8f3b8276)]

PR #3038, opened 2023-02-13
Fix firefox blockchain scroll reset bug

### LINKED ISSUE #3031: Blocks layout should properly reset when returning to dashboards
# Description

This UI state is not valid. When we return to the mempool or mining dashboard, the blocks layout at the top of the page should properly reset to the default `[Mempool blocks] | [Blockchain blocks]`.

image

# Steps to reproduce

0. I'm using Firefox
1.  Open https://mempool.space/

image

2. Access block 300,000 https://mempool.space/block/300000 

image

3. Click on the mining dashboard button and notice the layout does not reset properly as shown in the bug description

mononaut:
Fixes #3031

### DIFF
--- a/frontend/src/app/components/start/start.component.ts
+++ b/frontend/src/app/components/start/start.component.ts
@@ -298,6 +298,10 @@ export class StartComponent implements OnInit, OnDestroy {
   }
 
   ngOnDestroy() {
+    if (this.blockchainContainer?.nativeElement) {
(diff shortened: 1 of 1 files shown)


nymkappa (approved):
Tested ACK on Firefox @ c85d8cd29d6ff2e5b2da4be753eacbb363f40ca9

PR #3039, opened 2023-02-13
Add audit / block health config feature flag

### LINKED ISSUE #3033: Add feature flag for block health
Block health needs to be a toggle on Docker, in addition to the block height for all networks.

https://github.com/mempool/mempool/blob/cf0897e27ce88eae4a7e546272e2fbbfefca8bd5/docker/frontend/entrypoint.sh#L34

mononaut:
Resolves #3033 by adding `AUDIT` flags to both frontend and backend configurations, to control block audit related features (audits, block health scores, and the prediction accuracy graph).

These settings default to false, but are enabled in the production configuration.


backend/src/api/blocks.ts
```diff
@@ -212,9 +212,11 @@ class Blocks {
         };
       }
 
-      const auditScore = await BlocksAuditsRepository.$getBlockAuditScore(block.id);
-      if (auditScore != null) {
-        blockExtended.extras.matchRate = auditScore.matchRate;
+      if (config.MEMPOOL.AUDIT) {
+        const auditScore = await BlocksAuditsRepository.$getBlockAuditScore(block.id);
+        if (auditScore != null) {
```
softsimon, line 217: It's better to use `if (!auditScore)` if you want to catch both undefined and null, otherwise use `!== null`

mononaut: sure, I tend to avoid coercing to boolean for null checks out of habit, because it's a footgun when working with strings and numbers

frontend/src/app/components/block/block.component.ts
```diff
@@ -269,7 +274,7 @@ export class BlockComponent implements OnInit, OnDestroy {
       this.isLoadingOverview = false;
     });
 
-    if (!this.indexingAvailable) {
+    if (!this.indexingAvailable || !this.stateService.env.AUDIT) {
```
knorrium, line 277: @mononaut is this conditional correct? Does it make sense to have audit enabled but indexing disabled?

frontend/src/app/components/block/block.component.ts
```diff
@@ -58,7 +58,7 @@ export class BlockComponent implements OnInit, OnDestroy {
   overviewError: any = null;
   webGlEnabled = true;
   indexingAvailable = false;
-  auditModeEnabled: boolean = !this.stateService.hideAudit.value;
+  auditModeEnabled: boolean = this.stateService.env.AUDIT && !this.stateService.hideAudit.value;
```
knorrium, line 61: @mononaut I think this could be split into two variables, `auditModeEnabled` and `auditModeDisplayed` (or `shouldDisplayAuditMode`), where `auditModeEnabled` is a shortcut for `this.stateService.env.AUDIT` so you don't repeat that throughout the code and `auditModeDisplayed` controls whether it's shown or not, as you have now before this change.

knorrium:
utACK

knorrium (approved):

frontend/src/app/components/block/block.component.ts
mononaut: it's correct, but maybe overly verbose - you're right that `AUDIT` should imply `indexingAvailable`

mononaut:
rebased on master, and made some small changes per @knorrium's suggestions

knorrium:
@mononaut the code is a lot cleaner now, thanks for refactoring.

wiz:
@hunicus can you modify your block audit related FAQ articles to use the new frontend AUDIT config instead of the OFFICIAL config

wiz (approved):
Tested ACK @ v2.5.0-dev [[27a96228](https://github.com/mempool/mempool/commit/27a96228)]

backend/src/api/blocks.ts
wiz: going to merge this for now so we can proceed, feel free to open a PR to tweak this later

Issue #3006, opened 2023-02-01
[Lightning] Wrong last update data on the channel page


knorrium:
Somehow some channel data ended up as 1969 and caused a few issues on the channel page:

1) The channel is shown as active
2) The order of channels on the the table is different than prod

The node page does not show the incorrect date, just the `-`  which wold be better in the channel page as well.

![Screenshot from 2023-01-31 21-45-19](https://user-images.githubusercontent.com/100320/215961853-86fc53c0-2d70-4303-a234-26001bcfc76c.png)

nymkappa:
I managed to reproduce it. This bug only seems to happen with LND. Using the `describeGraph` command we can see that they set this channel `last_update` value to 0.

```
{
            "channel_id": "741771226150469633",
            "chan_point": "ab86dfed91ceb374ade96bf29b808831d24f968a9eddc86dbb396b125014ae3e:1",
            "last_update": 0,
            "node1_pub": "0247e898d2e72acf6e386472b9b98467b11ca9fba7289caa6d658abfb6c97c0ba6",
            "node2_pub": "03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f",
            "capacity": "100000",
            "node1_policy": null,
            "node2_policy": null
},
```

PR #3040, opened 2023-02-13
[LND] Nullify zeroed timestamps

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3006.

This should not impact prod because it's a LND only bug where LND API returns zeroed timestamp 😮‍💨.

image


softsimon (approved):
tested ACK @ v2.5.0-dev [[7316d7d7e](https://github.com/mempool/mempool/commit/7316d7d7e)]

Issue #3005, opened 2023-02-01
[Lightning] Network capacity missing on the home widget


knorrium:
The clearnet, tor and unknown network capacity fields are coming back as `null` from the API, causing the widget to show only zeroes despite some data being available.

`{"clearnetCapacity":null,"torCapacity":null,"unknownCapacity":null"...`

![Screenshot from 2023-01-31 21-09-27](https://user-images.githubusercontent.com/100320/215957710-7abd81c0-87a9-4c57-97e5-0adff17f7386.png)

PR #3044, opened 2023-02-13
Fix node socket parsing with LND

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3005

image


softsimon (approved):
tested ACK @ v2.5.0-dev [[8630ae068](https://github.com/mempool/mempool/commit/8630ae068)]

Issue #3026, opened 2023-02-09
Links to Block Health Section in FAQ


cryptoquick:
### Description

Maybe it'd make sense to add links to make the [Block Health](https://mempool.space/docs/faq#what-is-block-health) section more more discoverable.

### Problem to be solved

The colors in the blocks seemed interesting, but it was hard to figure out what they meant.

### Proposed solution

I'd recommend adding a prompts to the tx view like, "Why was this transaction included?" to aid in discoverability, similar to the "Why is this block empty?" info button on this block:
https://mempool.space/block/000000000000000000037a742ade26b8ce39bb54abcde129ee8fb30b58f10a26

#### Additional info

There seems to be a lot of stuff written about this in #2662... In particular, this table seemed helpful and might be good to have in the FAQ also:
![Screenshot from 2023-02-08 20-50-53](https://user-images.githubusercontent.com/285690/217713503-8d296b04-e121-492b-93f1-1cbdd3a9687a.png)

Ideally the FAQ would also provide some background on the impact of censorship on the blockchain; why it's done, and why it matters.

hunicus:
> Problem to be solved
>
> The colors in the blocks seemed interesting, but it was hard to figure out what they meant.

#3045 addresses this and is now live.

PR #3045, opened 2023-02-13
Add block audit faq

hunicus:
These changes add a FAQ on block audits that describes what they are and how they work. There are also small updates to the existing block health FAQ.

They should cover topics raised by various users over the past few days:
- what an expected block is and how it is determined
- what a block audit shows
- what block health means
- how the block health score can be so high when the actual block is so different from the expected block
- what the colors in the block audit view mean (#3026)

Separately, I propose that we consider renaming "Projected Block" to "Expected Block" on the block page...the idea being that the next block is "projected" until the moment a block is mined, at which point the snapshot of the next projected block becomes the "expected" block for that particular block height. Might seem like splitting hairs on wording but wiz and I discussed that "expected vs actual" is more idiomatic than something like "projected vs actual".

## Block Audit

Screenshot from 2023-02-13 03-30-38

## Block Health

Screenshot from 2023-02-13 06-31-57


frontend/src/app/docs/api-docs/api-docs.component.html
```diff
@@ -218,17 +218,41 @@ <h3 *ngIf="( item.type === 'category' ) && ( item.showConditions.indexOf(network
...
+  <p>The number of transactions appearing in both <span class='math'>s<sub>expected</sub></span> and <span class='math'>s<sub>actual</sub></span> is used (instead of a block's full transaction count) in order to minimize chances that block health is inadvertently impacted by transactions that were most likely not intentionally excluded:</p>
   <ul>
     <li>recently-broadcast transactions, since the miner may simply not have received them</li>
     <li>certain low-feerate transactions, since the miner may have opted to replace them with more profitable out-of-band transactions</li>
   </ul>
-  <p>Mempool uses a re-implementation of Bitcoin Core's transaction selection algorithm to determine the transactions it expects to see in the next block.</p>
+  <p>As a result, block health is <i>not</i> intended to be a measure of how closely an expected block resembles an actual block. The actual block can be vastly different from the expected block, but if no transactions appear to be intentionally excluded, it will have a high health rating (<a [routerLink]="['/block/0000000000000000000515e202c8ae73c8155fc472422d7593af87aa74f2cf3d']">extreme example</a>).</p>
```
mononaut, line 254: > As a result, block health is <i>not</i> intended to be a measure of how closely an expected block resembles an actual block.

this sentence is repeated verbatim in line 245, I'm not sure if that's unintentional or for deliberate emphasis?

hunicus: Nope, I just forgot to remove that line. Thanks for catching that!

hunicus:
Added a note about block audit and block health only being supported on official instances. The FAQs themselves also only show on official instances.

Screenshot from 2023-02-13 19-35-24

softsimon (approved):
tested ACK @ v2.5.0-dev [[7aca0450](https://github.com/mempool/mempool/commit/7aca0450)]

PR #3047, opened 2023-02-13
Change 'projected' to 'expected' for block audit label

hunicus:
See context explained in https://github.com/mempool/mempool/pull/3045.

## Before

projected

## After

expected


softsimon (changes requested):
Could you also change the i18n strings to "block.expected" and "block.expected-block" so that people need to retranslate.

wiz (approved):
Tested ACK @ v2.5.0-dev [[cbbb36b0](https://github.com/mempool/mempool/commit/cbbb36b0)]

PR #3048, opened 2023-02-14
Update mainnet tests

knorrium:
With infinite scrolling, the initial number of blocks loaded was increased and the tests were outdated.

The locators assumed there was only one block with index 0, but we now have an offset, so they had to be included.

Disabled the test that clicks on the "last" block as that is no longer a thing.


wiz:
It seems the tests still fail? Or is that unrelated?

knorrium:
> It seems the tests still fail? Or is that unrelated?

Mainnet and Bisq are passing, Liquid is still failing due to issue #3052

wiz:
ok will merge master into this PR after 3052 is fixed, and re-test this one, and merge after all green

wiz:
@softsimon did you test this? it modified the frontend code

PR #3058, opened 2023-02-15
Remove colon from block overview table

hunicus:
Tiny adjustment to make the subsidy field more consistent with the other fields.

## Before

remove-colon-2

## After

remove-colon-1


wiz (approved):
utACK

PR #3059, opened 2023-02-15
Shrink og avators to make group more square

hunicus:


wiz (approved):
Tested ACK @ v2.5.0-dev [[40decd15](https://github.com/mempool/mempool/commit/40decd15)]

PR #3062, opened 2023-02-15
Fix database usage when database is disabled

nymkappa:
The price updater and the pool parser had two queries being run even if the database is disabled.


wiz (approved):
Tested ACK @ v2.5.0-dev [[1eb68962](https://github.com/mempool/mempool/commit/1eb68962)]

PR #3066, opened 2023-02-16
Add size per weight graph and ts type for `getHistoricalBlockSizesAndWeights`

antonilol:
this PR adds a 'size per weight' (measured in B/vB, bytes per vbyte) line in the existing size and weights graph

before the segwit activation this number is 1, and after that you see this cyan line going up as more and more people start to use segwit, recently we got jpegs, this adds roughly half a megabyte per block

i used cyan as color, hope this fits the theme

block-sizes-weights-all-1676557399

for reviewers:
did i use the `$localize` tag correctly?
i copied parts of the code of the 'weights' data series for this one, maybe i forgot to change/add/remove some fields

### DIFF
--- a/frontend/src/app/components/block-sizes-weights-graph/block-sizes-weights-graph.component.ts
+++ b/frontend/src/app/components/block-sizes-weights-graph/block-sizes-weights-graph.component.ts
@@ -90,6 +90,7 @@ export class BlockSizesWeightsGraphComponent implements OnInit {
                 this.prepareChartOptions({
                   sizes: data.sizes.map(val => [val.timestamp * 1000, val.avgSize / 1000000, val.avgHeight]),
                   weights: data.weights.map(val => [val.timestamp * 1000, val.avgWeight / 1000000, val.avgHeight]),
+                  sizePerWeight: data.weights.map((val, i) => [val.timestamp * 1000, data.sizes[i].avgSize / (val.avgWeight / 4), val.avgHeight]),
                 });
                 this.isLoading = false;
               }),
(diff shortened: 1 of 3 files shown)


frontend/src/app/services/api.service.ts
```diff
@@ -222,8 +222,8 @@ export class ApiService {
     );
   }
 
-  getHistoricalBlockSizesAndWeights$(interval: string | undefined) : Observable<any> {
-    return this.httpClient.get<any[]>(
```
antonilol, line 226: i think this `any[]` was copied from other API calls to here, but was incorrect. this could also be the case with other calls

nymkappa: That's just me not paying enough care about typescript interfacing feature. Thanks for fixing it 👍🏻

frontend/src/app/components/block-sizes-weights-graph/block-sizes-weights-graph.component.ts
```diff
@@ -124,6 +125,7 @@ export class BlockSizesWeightsGraphComponent implements OnInit {
       color: [
         '#FDD835',
         '#D81B60',
+        '#14EDF5',
```
nymkappa, line 128: For the color, I'd try `#039BE5` so it's less flashy. For future reference you can see the colors we usually pick for charts in https://github.com/mempool/mempool/blob/master/frontend/src/app/app.constants.ts#L34

```suggestion
        '#039BE5,
```

nymkappa (approved):
Tested ACK @ a897aebbc1de9148e1bff71eefa26d0b826ce639

That's a great addition! Make it very clear when segwit was activated and how it's usage increases over time. Good idea, thanks a lot.

Aside the color thing it's good to me. Feel free to push an additional commit to update the color.

softsimon (approved):
tested ACK @ 04d8926

PR #3067, opened 2023-02-17
Show new sponsor tiers on about page

nymkappa:
Maybe we'll need to polish the layout (maybe bigger image for whales for example).

### Without chad and whale sponsors 😭

image

### With chad and whale sponsors

image


nymkappa:
Included in another PR so will close this one now

PR #3068, opened 2023-02-17
Raise production memory limits

mononaut:
Increases production bitcoin.conf `maxmempool` setting to 4GB, and raises `npm run start-production` nodejs memory limit to 16GB.

Tested against simulated mempools, the nodejs backend generally exceeds a peak heap size of 16GB at around ~4.2m transactions with combined size of ~2.9GB (roughly equivalent to 4GB of dynamic memory usage in Bitcoin Core).

Other parts of the system fail before we hit this extreme, so simple memory usage is not the limiting factor at that point.


wiz (approved):

Sjors:
Fine, I'll buy more RAM… :-)

PR #3071, opened 2023-02-18
Limit worst-case GBT performance

mononaut:
_(builds on PR #3037)_

This PR leverages the second "minfee" node to limit the worst case performance of the GBT algorithm when the mempool is very full.

Since Bitcoin Core purges transactions beyond the `maxmempool` limit in order of effective fee rate, the content of the minfee node's default mempool should very closely match the top 300MB of the main node's mempool.

When `config.MEMPOOL.USE_SECOND_NODE_FOR_MINFEE` is enabled, we track the content of the second node's smaller mempool, and perform the GBT algorithm only on that reduced list of transactions.

Any remaining transactions from the main mempool are then inserted at the end of the GBT templates, so that total counts and summary statistics remain correct.

## Pros
- Gives the GBT algorithm reasonable worst-case performance regardless of the size of the main mempool.
- Might improve block audit accuracy and fee estimation for miners using default mempool settings.

## Cons
- Incompatible with #2845, since it depends on the second minfee node running the same mempool acceptance policies as the main node (except for `maxmempool`)
- Might reduce block audit accuracy and fee estimation for miners using larger than default mempools, due to missing CPFP packages involving transactions already purged from the minfee mempool.
- No CPFP information is calculated for unconfirmed transactions falling outside the 300MB cut-off.

### DIFF
--- a/backend/src/api/mempool-blocks.ts
+++ b/backend/src/api/mempool-blocks.ts
@@ -1,5 +1,5 @@
 import logger from '../logger';
-import { MempoolBlock, TransactionExtended, ThreadTransaction, TransactionStripped, MempoolBlockWithTransactions, MempoolBlockDelta, Ancestor } from '../mempool.interfaces';
+import { MempoolBlock, TransactionExtended, ThreadTransaction, TransactionStripped, MempoolBlockWithTransactions, MempoolBlockDelta, Ancestor, GbtCandidates } from '../mempool.interfaces';
 import { Common } from './common';
 import config from '../config';
 import { Worker } from 'worker_threads';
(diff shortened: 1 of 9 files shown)


wiz:
```
Mar 20 09:05:49 [78070] DEBUG: Updating mempool...
Mar 20 09:05:50 [78070] DEBUG: Mempool templates calculated in 0.548 seconds
Mar 20 09:05:51 [78070] DEBUG: Mempool updated in 2.25 seconds. New size: 33089 (+16)
```
Recently as the mempool is growing in size this PR has become higher priority.
Could you please rebase this onto current master?

mononaut:
rebased on current master

mononaut:
After reviewing this again during the rebase, I'm tempted to move the PR back to draft status.

It does seem to run well locally in normal conditions, but the current implementation feels brittle, and I'm nervous about how it will handle nodes crashing or getting out of sync or other exceptional circumstances I haven't thought of yet.

It's probably preferable to `makeBlockTemplates` blowing up the backend when the mempool grows much bigger, but until then I'd like to do some more rigorous testing and work on making the feature more robust before this goes near production.

softsimon:
Closing due to not being relevant anymore

PR #3072, opened 2023-02-18
Optimize mempool block 7 data

mononaut:
When the mempool contains more than 8 blocks worth of transactions, [mempool block 7](https://mempool.space/mempool-block/7) represents a "stack of N blocks" containing the rest of the mempool.

Screenshot 2023-02-17 at 7 09 29 PM

We currently include not only summary statistics for the full "stack of N blocks", but also visualization data for every transaction in the stack. This means calculating visualization deltas for all of those transactions, and sending the data to every client viewing that page, even though only the first ~block worth of transactions will actually be visible.

That seems sub-optimal.

This PR limits the `transactions` field of mempool block 7 to only data actually required for the visualization, while continuing to calculate and send proper summary statistics for the full "stack of N blocks".


softsimon (approved):
tested ACK @ v2.5.0-dev [[b21fd0d37](https://github.com/mempool/mempool/commit/b21fd0d37)]

PR #3073, opened 2023-02-18
Migrated transifex and fetched latest translations

softsimon:


wiz (changes requested):
Tested NACK @ v2.5.0-dev [[2993a286](https://github.com/mempool/mempool/commit/2993a286)]

Unable to view LN channels or nodes while using /ja locale

wiz (approved):
Tested ACK @ v2.5.0-dev [[2993a286](https://github.com/mempool/mempool/commit/2993a286)]

Previous NACK was due to invalid test env

Issue #2991, opened 2023-01-28
Channel map state is kept between navigations


softsimon:
How to reproduce:

1. Go to https://mempool.space/lightning/channel/839634358106521600
2. Type `838675583872532480` in the search bar and go to that channel
3. The GEO map is now missing.
4. Reload the page, and there is in fact a GEO map.

PR #3076, opened 2023-02-18
Fix map not being updated when switching between channel pages

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2991

@softsimon if you have a cleaner fix feel free to apply it. Seems like rxjs observable is triggered BEFORE angular updates the `Input()` and so the Observable works with the wrong data.

Adding a 100ms delay before updating the map do the trick but obviously is not the best fix.


wiz (approved):
Tested ACK @ v2.5.0-dev [[ef8a1c73](https://github.com/mempool/mempool/commit/ef8a1c73)]

PR #2869, opened 2022-12-23
Rewrite mining pools parser

nymkappa:
~This PR is ready for review and you can test the code already with the instruction below, but cannot be merged into master yet until the `Todo` is completed.~

Related to https://github.com/mempool/mining-pools/issues/25

### Description

Rewrite the mining pool parser using the new format defined by @0xB10C in https://github.com/0xB10C/known-mining-pools/pull/75

Full `blocks` table re-indexing is required because we are introducing a new `unique_id` field in the pools JSON file and it would just be too much a hassle to support migrating from the old JSON structure.

Full block re-indexing is automatically done with a database migration (version 54), so you want to take out servers from prod for this upgrade @wiz.

### Todo

~Issue https://github.com/mempool/mempool/issues/2898 **should be fixed first**, as we need to re-index the blocks table with this PR.~

~Also, before this PR can be merged into master, we need to settle on how the new mining pools JSON definition file will be managed (where and who). @wiz and @0xB10C I'd like your opinion on this.~

I see two possible approaches:
- ~@0xB10C contributes directly to https://github.com/mempool/mining-pools repo where he can work on the new JSON pools definition as well as his generator scripts~
- ~@0xB10C continues working on his own repo at https://github.com/bitcoin-data/mining-pools and we fork it for our production servers~

**EDIT: We'll discuss this again in the future. For now we'll simply manage our own copy of the file using the new JSON structure.**

### Testing/Upgrading

0. Take the server out of production
1. Stop the mempool nodejs backend
2. Update your nodejs backend configuration like so (this is the most up to date mining pool definition, pending review)
```
"POOLS_JSON_URL": "https://raw.githubusercontent.com/mempool/mining-pools/nymkappa/feature/add-luxor-address/pools-v2.json",
"POOLS_JSON_TREE_URL": "https://api.github.com/repos/mempool/mining-pools/git/trees/nymkappa/feature/add-luxor-address"
```
3. Clear the mempool nodejs backend cache: `rm ./backend/cache/cache*`. _Note, this will cause the nodejs backend to re-fetch all pending transaction in the currently full mempool, so expect some delay at start._
4. Run the PR (pull this branch and restart the mempool nodejs backend). Confirm the `pools` table and `blocks` table is being truncated and re-indexing is in progress.
5. Wait for about 2000 block to be re-indexed then open the mining dashboard. Confirm it shows correct mining pools (compare with mempool.space).
[...]

PR #3077, opened 2023-02-19
Log warning when pools are out of date

nymkappa:
### UPDATED DESC

Most of the changes of this PR have been merged already in https://github.com/mempool/mempool/pull/2869. The only extra thing left from this PR is the updated README and one additional log from the new `--update-pools` command line.

### _ORIGINAL PR_

_Currently in our production servers, we have `config.MEMPOOL.AUTOMATIC_BLOCK_REINDEXING` set to `false`. This means that when mining pools are updated, we are not automatically re-indexing the corresponding blocks table. This leads to inconsistent result where older blocks are tagged using the older pool definition, while new blocks are tagged using the latest pool definition._

_This PR disable the mining pools automatic update if `config.MEMPOOL.AUTOMATIC_BLOCK_REINDEXING` is set to `false`. A new nodejs command line option is provided to manually trigger the mining pools update `--update-pools`, which will both update the mining pools and re-index appropriate blocks at the same time. See updated backend README for more details._

### DIFF
--- a/backend/src/tasks/pools-updater.ts
+++ b/backend/src/tasks/pools-updater.ts
@@ -17,11 +17,6 @@ class PoolsUpdater {
   treeUrl: string = config.MEMPOOL.POOLS_JSON_TREE_URL;
 
   public async updatePoolsJson(): Promise<void> {
-    if (config.MEMPOOL.AUTOMATIC_BLOCK_REINDEXING === false) {
-      logger.info(`Not updating mining pools to avoid inconsistency because AUTOMATIC_BLOCK_REINDEXING is set to false`)
-      return;
-    }
-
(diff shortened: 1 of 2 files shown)


wiz:
Should this be closed?

backend/src/tasks/pools-updater.ts
```diff
@@ -36,12 +36,6 @@ class PoolsUpdater {
 
     this.lastRun = now;
 
-    if (config.SOCKS5PROXY.ENABLED) {
```
nymkappa, line 39: Moved here https://github.com/mempool/mempool/pull/3077/files#diff-a54fca5317f9bce0a6ceb5bf72cde72f4cf8d6627d54f4dca3e6bd301fd8e166R65

wiz (approved):
Tested ACK @ d0d23035133248524ce0fa56a12a8aaa4eb1136e

PR #3078, opened 2023-02-19
Remove hardcoded mining pools colors as it's not relevant

nymkappa:
The only hardcoded colors left are `unknown` and `other` with two different shade of grey

### All

Screenshot 2023-02-19 at 17 15 20

### 3y

image

### Dashboard

Like @wiz mentioned it also matches the lightning dashboard now since we don't have hardcoded colors in there

image

### DIFF
--- a/frontend/src/app/app.constants.ts
+++ b/frontend/src/app/app.constants.ts
@@ -72,22 +72,10 @@ export const chartColors = [
 ];
 
 export const poolsColor = {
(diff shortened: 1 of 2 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[e2eccf6e](https://github.com/mempool/mempool/commit/e2eccf6e)]

PR #3043, opened 2023-02-13
Fixes blocks api (missing fee range)

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3042.

### Before

Note the missing fee range (which also breaks the new infinite blocks scrolling)

https://user-images.githubusercontent.com/9780671/218390603-aa2d6a9c-94ae-40b1-a204-909348d3bdd7.mov

### After

Smooth like a puppy

https://user-images.githubusercontent.com/9780671/218390720-7ab9fe4d-5625-487e-9d14-e2b54323af90.mov

PR #3080, opened 2023-02-19
Fix database usage when database is disabled

nymkappa:
Fix regression introduced in https://github.com/mempool/mempool/pull/3043


softsimon (approved):
utACK

PR #3081, opened 2023-02-19
Remove included in block

softsimon:
Removing this row from the Transaction page, since we now always see the block height on the blockchain on top.

Screenshot 2023-02-19 at 17 23 14


wiz (approved):
Tested ACK @ v2.5.0-dev [[df578f64](https://github.com/mempool/mempool/commit/df578f64)]

Issue #2998, opened 2023-01-30
Expose median timestamp


AlexLloyd0:
Median timestamp is a useful property of blocks due to being monotonically increasing. It's included in the response from bitcoin-core for [many endpoints](https://developer.bitcoin.org/reference/rpc/getblock.html)

We're building an app using mempool's API to query the blockchain, and having a monotonically increasing timestamp is crucial for us. We're currently computing the median timestamp ourselves, but it would be really useful to pass this through and expose via mempool's API (and maybe in the UI?)

Happy to implement this with a couple of pointers

nymkappa:
Related https://github.com/mempool/mempool/pull/3070. Once this PR is merged I'll update the blocks API to expose the median block timestamp.

PR #3070, opened 2023-02-17
Index coinstatsindex - Add bulk block query api

nymkappa:
### TODO before this PR is complete:

- [x] `medianFeeAmt`
- [x] `feePercentiles`
- [x] `orphaned` (and maybe provide orphaned block hash if possible)
- [x] ~`blockTime`~ - Will not implement

### Description

This PR indexes new block data gathered from the `getblockstats`, and `gettxoutsetinfo` core RPC command. 

`gettxoutsetinfo` requires `coinstatsindex=1` to be enabled in `bitcoin.conf`. However enabling this index is not required to run mempool, fields will simply be nullified. If in the future the index is enabled and core is done with re-indexing, we will automatically detect it and **update** our database.

The `blocks` table is now 778MB.

The new data is served through a new API endpoint which has been provided to query blocks between two block heights: `/api/v1/blocks-bulk/:minHeight/:maxHeight` with `maxHeight` defaulting to the current tip if not provided.  An example can be found at the end of this PR description.

### Test scenarios

A new configuration setting has been added `config.MEMPOOL.MAX_BLOCKS_BULK_QUERY` which default to `0`. When set to `0` or less, the new `blocks-bulk` API endpoint is disabled.
Set it to any positive number to choose how many blocks can be queries in a single query.

- The feature is Bitcoin only, 404 otherwise
- Mempool nodejs backend indexing must be enabled _(`config.DATABASE.ENABLED` set to `true` and `config.MEMPOOL.INDEXING_BLOCKS_AMOUNT` set to a positive value)_, 404 otherwise
- In core, if `coinstatsindex=0` then `utxoset_size` and `total_input_amt` will be set to `null` in the API response
- In core, if `coinstatsindex=1` and core is currently indexing, `utxoset_size` and `total_input_amt` will be set to `null` in the API response until core is done indexing
- In core, if `coinstatsindex=1` and core finished indexing all blocks, `utxoset_size` and `total_input_amt` will be properly returned
- `fee_amt_percentiles` and `median_fee_amt` requires `config.MEMPOOL.INDEXING_BLOCKS_AMOUNT: true`, `null` otherwise

Example response https://github.com/mempool/mempool/pull/3070#issuecomment-1442721036

PR #3082, opened 2023-02-19
Expose median block timestamp in block(s) API(s)

nymkappa:
Depends on  https://github.com/mempool/mempool/pull/3070

Should not impact bisq/liquid.
Should work with or without indexing enabled (see fields which may not be available in different scenarios in https://github.com/mempool/mempool/pull/3070 description)


nymkappa:
Will redo including all new indexed fields

PR #3086, opened 2023-02-20
Expose new indexed block data in block(s) APIs

nymkappa:
**EDIT: Will re-work on that PR after https://github.com/mempool/mempool/issues/3121 is fixed**

Depends on  https://github.com/mempool/mempool/pull/3070

Does not impact bisq/liquid.
Does not impact current response format, we **only add** new fields, so it should not break anything in the frontend.

Works with or without indexing enabled, however some data is not available without indexing, see https://github.com/mempool/mempool/pull/3070 for more details on that.

_This PR can be built upon to show new interesting data in the block component._


nymkappa:
Redo in https://github.com/mempool/mempool/pull/3166

PR #3092, opened 2023-02-21
Use historical price in older blocks / transactions

nymkappa:
Fixes https://github.com/mempool/mempool/issues/175 ❤️ 

Please let me know if there are other places where we need to use the historical price.

image

image

image


wiz (approved):
Tested ACK @ v2.5.0-dev [[70d96583](https://github.com/mempool/mempool/commit/70d96583)]

Seems fine but some tests failing so need to look into that

PR #3098, opened 2023-02-21
Use mempool chart color palette on hashrate history

nymkappa:
Fix the following regression

### Before

image

### New

image


wiz (approved):
Tested ACK @ v2.5.0-dev [[2bfe1923](https://github.com/mempool/mempool/commit/2bfe1923)]

Issue #3094, opened 2023-02-21
Cannot click on "all" in block reward chart


nymkappa:
Need to check other charts for the same bug if present

nymkappa:
Okay so the actual bug is that when the api takes too long to return the data AND we click on a different timespan in the mean time, then you end up with a mixed state (for example 1y button is actived, but chart shows 2y)

PR #3099, opened 2023-02-21
Fix timespan UX in mining charts

nymkappa:
- Make sure we cannot query a new timespan while we're loading the data/chart
- Fix "cannot click on all" in block reward graph

Fixes https://github.com/mempool/mempool/issues/3094


wiz (approved):
Tested ACK @ v2.5.0-dev [[680f7d39](https://github.com/mempool/mempool/commit/680f7d39)]

PR #3100, opened 2023-02-21
Fixes a syntax error with certain keywords

softsimon:
If you did a search for just a + or - it would search for an empty keyword. Added a filter for empty search words.

### DIFF
--- a/backend/src/api/explorer/nodes.api.ts
+++ b/backend/src/api/explorer/nodes.api.ts
@@ -362,7 +362,12 @@ class NodesApi {
   public async $searchNodeByPublicKeyOrAlias(search: string) {
     try {
       const publicKeySearch = search.replace('%', '') + '%';
-      const aliasSearch = search.replace(/[-_.]/g, ' ').replace(/[^a-zA-Z0-9 ]/g, '').split(' ').map((search) => '+' + search + '*').join(' ');
+      const aliasSearch = search
+        .replace(/[-_.]/g, ' ')
+        .replace(/[^a-zA-Z0-9 ]/g, '')
+        .split(' ')
+        .filter(key => key.length)
+        .map((search) => '+' + search + '*').join(' ');
       const query = `SELECT public_key, alias, capacity, channels, status FROM nodes WHERE public_key LIKE ? OR MATCH alias_search AGAINST (? IN BOOLEAN MODE) ORDER BY capacity DESC LIMIT 10`;
       const [rows]: any = await DB.query(query, [publicKeySearch, aliasSearch]);
       return rows;
--- a/backend/src/api/explorer/nodes.api.ts
+++ b/backend/src/api/explorer/nodes.api.ts
@@ -368,6 +368,7 @@ class NodesApi {
         .split(' ')
         .filter(key => key.length)
         .map((search) => '+' + search + '*').join(' ');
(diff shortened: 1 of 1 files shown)


wiz (changes requested):
Please add comments above each regex replacement etc. to explain the intention of each line. We have a ton of undocumented magic regex strings in our code. Example of how to do it properly: https://github.com/bisq-network/bisq/blob/master/desktop/src/main/java/bisq/desktop/components/paymentmethods/data/JapanBankData.java#L860

nymkappa:
I think the problem with the current implementation is that it does not work for all nodes.

Alias are set by the user and can be set to weird stuff like this one `(((d(^_~)b)))` https://mempool.space/lightning/node/03b067d64918fc596f8dcffbefa6a9ae09a1afe269cf038312211f51918448ac84 or this one `🏜⚡🌪️` (https://mempool.space/lightning/node/02db49beaa4d8d229921f1e85370841489981cc0e0f5de4fcb8816e2757232b6a7) and you cannot search it with the current implementation.

Can't we simply have a `%keyword%` query instead, would not that work?

softsimon:
> Can't we simply have a `%keyword%` query instead, would not that work?

%keyword% is wildcard search and can't be indexed so it's slower as the node database grow. keyword% can be indexed but then you can't search for "Nicehash" and get result for ln.nicehash.com. So we use fulltext index for words "ln, nicehash, com" and nicehash* will find it instantly.

backend/src/api/explorer/nodes.api.ts
```diff
@@ -362,7 +362,12 @@ class NodesApi {
...
     try {
       const publicKeySearch = search.replace('%', '') + '%';
-      const aliasSearch = search.replace(/[-_.]/g, ' ').replace(/[^a-zA-Z0-9 ]/g, '').split(' ').map((search) => '+' + search + '*').join(' ');
+      const aliasSearch = search
+        .replace(/[-_.]/g, ' ') // Replace all -_. characters with empty space. Eg: "ln.nicehash" becomes "ln nicehash".  
+        .replace(/[^a-zA-Z0-9 ]/g, '') // Remove all special characters and keep just A to Z, 0 to 9.
+        .split(' ')
+        .filter(key => key.length)
+        .map((search) => '+' + search + '*').join(' ');
       const query = `SELECT public_key, alias, capacity, channels, status FROM nodes WHERE public_key LIKE ? OR MATCH alias_search AGAINST (? IN BOOLEAN MODE) ORDER BY capacity DESC LIMIT 10`;
```
wiz, line 371:
```suggestion
// %keyword% is wildcard search and can't be indexed so it's slower as the node database grow. keyword% can be indexed but then you can't search for "Nicehash" and get result for ln.nicehash.com. So we use fulltext index for words "ln, nicehash, com" and nicehash* will find it instantly.
      const query = `SELECT public_key, alias, capacity, channels, status FROM nodes WHERE public_key LIKE ? OR MATCH alias_search AGAINST (? IN BOOLEAN MODE) ORDER BY capacity DESC LIMIT 10`;```

wiz (approved):
Tested ACK @ v2.5.0-dev [[03be3ab5](https://github.com/mempool/mempool/commit/03be3ab5)]

PR #3102, opened 2023-02-22
Ignore coinbase transaction in block health calculation

mononaut:
Adjusts the block health score calculation to exclude the coinbase transaction.

Perfectly matching blocks will continue to score 100%, while completely censored or empty blocks will now score 0%.


wiz:
I'm testing this on 201 fra overnight since I need an empty block to get mined to test it properly

softsimon (approved):
Looks safe.

utACK

wiz (commented):
Tested ACK @ v2.5.0-dev [[86a0488b](https://github.com/mempool/mempool/commit/86a0488b)]

Verified empty block gets 0% score:

Screenshot 2023-02-27 at 9 35 33

PR #3104, opened 2023-02-22
Fix missing fees in liquid block tooltips

mononaut:
Fixes missing fee data in the mined block visualization on Liquid, due to a difference in `getblock` response format between Bitcoin and Elements Core RPC APIs.

| Before | After |
|-|-|
| before-liquid-fees | after-liquid-fees |

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -143,7 +143,10 @@ class Blocks {
    * @returns BlockSummary
    */
   public summarizeBlock(block: IBitcoinApi.VerboseBlock): BlockSummary {
-    const stripped = block.tx.map((tx) => {
+    if (Common.isLiquid()) {
+      block = this.convertLiquidFees(block);
+    }
+    const stripped = block.tx.map((tx: IBitcoinApi.VerboseTransaction) => {
       return {
         txid: tx.txid,
         vsize: tx.weight / 4,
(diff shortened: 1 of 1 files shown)


wiz (changes requested):
Tested NACK @ v2.5.0-dev [[7ab677fe](https://github.com/mempool/mempool/commit/7ab677fe)]

Cleared mempool backend cache, nginx cache, browser cache, but I'm still getting zero fees - any idea?
Screenshot 2023-02-23 at 13 03 35

mononaut:
I can't reproduce this on [[7ab677fe](https://github.com/mempool/mempool/commit/7ab677fe)].

as far as I can tell I'm running elementsd, esplora, and the mempool backend all with exactly the same config as production, and with the fix my local backend now always serves `/api/v1/block/<hash>/summary` with the correct fees.

Screenshot 2023-02-23 at 9 59 38 AM

mononaut:
ok, I think I figured it out.

apparently [fees were only added to the `getblock` response](https://github.com/ElementsProject/elements/commit/658b2312da0ce8a95dc299e08363a2d85be4bd56) from Elements v22 onward, so the data simply isn't available via that method in older versions.

we can either update the minimum supported version to Elements 22.0, or I can rewrite this PR to fetch Liquid tx fees individually from esplora or maybe calculate them explicitly from the transaction data.

wiz:
ah ok, it's fine, let me upgrade elements on all our servers and I'll re-test this PR

wiz (approved):
Tested ACK @ v2.5.0-dev [[ed46232b](https://github.com/mempool/mempool/commit/ed46232b)]

softsimon (approved):
tested ACK @ v2.5.0-dev [ed46232b]

PR #3109, opened 2023-02-22
Add border to qr code because it's a best practice

nymkappa:
image


wiz:
my hero

softsimon (approved):
tested ACK @ v2.5.0-dev [[273770a63](https://github.com/mempool/mempool/commit/273770a63)]

wiz:
there is already a border on the address page, so if you're adding a border on the QR code component then we should remove the border on the address page

nymkappa:
Okay so the css is weird. Therefore in order to avoid breaking all the things I just added a `border` parameter in the qr code component that I can use in other places in the future.

Rebased on top of master.

softsimon (approved):
tested ACK @ v2.5.0-dev [[6f68c1666](https://github.com/mempool/mempool/commit/6f68c1666)]

PR #3110, opened 2023-02-22
Fix 'NaN' price for unconfirmed transaction since we have no block timestamp

nymkappa:
Fixes unconfirmed transactions

image

image


softsimon (approved):
tested ACK @ v2.5.0-dev [[2246a6f3](https://github.com/mempool/mempool/commit/2246a6f3)]

PR #3105, opened 2023-02-22
Ignore BTC prices that are out of wack

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3095

Tested with hardcoded insane values from the Coinbase price feed.
Example output that may happen if a price API sends us a price that does not make sense.

```
Feb 22 12:31:22 [59810] INFO: [Mining] Ignored Coinbase BTC/USD price: 100000000000000
Feb 22 12:31:34 [59810] INFO: [Mining] Ignored Coinbase BTC/GBP price: 100000000000000
Feb 22 12:32:15 [59810] INFO: [Mining] Ignored historical BTC/USD price: 100000000000
Feb 22 12:32:15 [59810] INFO: [Mining] Ignored historical BTC/GBP price: 100000000000
```

PR #3111, opened 2023-02-23
Redo/Fix completely failed PR #3092 + add PR #3105

nymkappa:
A db migration will truncate `prices` and `blocks_prices`

Sorry about that. I did not test my work properly in #3092 leading to another failed PR #3110 which breaks very old blocks price.

A related PR #3105 was pending review but I included it here with some changes as it is required for making the whole thing work properly.

I've also added support for historical price in the flow chart of the transaction component and block overview graph so @mononaut please take a look at my code changes as well 👍🏻


wiz (approved):
Tested ACK @ v2.5.0-dev [[d2ea9215](https://github.com/mempool/mempool/commit/d2ea9215)]

Seems fine, will run it overnight on staging

Issue #3004, opened 2023-02-01
[Lightning] Charts on the home screen look broken when there are few data points


knorrium:
When we don't have enough historical data to zoom out the chart, the legends overlap the lines:

![Screenshot from 2023-01-31 21-08-32](https://user-images.githubusercontent.com/100320/215956937-cf608e99-0c37-454f-acc2-ae6c65e1a49a.png)

nymkappa:
Not sure what we can do about the rendering since we don't have much data points. The same issue exists for all charts.
The legend positioning can be fixed tho, right now it's hardcoded.

PR #3115, opened 2023-02-23
Fix lightning widgets layout

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3004
Y-label cannot be on top of the charts anymore

image

image


wiz (approved):
Tested ACK @ v2.5.0-dev [[b1478988](https://github.com/mempool/mempool/commit/b1478988)]

PR #3119, opened 2023-02-24
Only query historical price if we're running mempool BASE_MODULE

nymkappa:


wiz (approved):
Tested ACK @ v2.5.0-dev [[20404cf6](https://github.com/mempool/mempool/commit/20404cf6)]

PR #3122, opened 2023-02-24
Remove mining db stats - replaced by runtime state variable

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3114

There is no simple way to test this aside waiting for one week to see if the network hashrate and mining pool weekly hashrate indexing is properly triggered. However the code change is trivial so I don't think testing it over a full week is necessary.

Basically we replaced a database variable with a local one. Since we were always resetting those db variable to 0 upon restart, there was no point in having them in the database anymore, in memory variable is good enough.

I've added the following new debug logs to track indexing attempt.
```
Feb 25 11:29:45 [42836] DEBUG: [Mining] Network hashrate history indexing is up to date, nothing to do
Feb 25 11:29:45 [42836] DEBUG: [Mining] Pool hashrate history indexing is up to date, nothing to do
```

PR #3125, opened 2023-02-25
Update some mining indexer log

nymkappa:
Depends on https://github.com/mempool/mempool/pull/3122
Just updated some logs I've noticed while working on #3122. No functionality change


softsimon (approved):
tested ACK @ 001be82f5

Looks safe enough.

Issue #2975, opened 2023-01-26
[Lightning] Improve error handling for missing channel data


knorrium:
When indexing lightning for the first time, we can end up on channel pages that do not yet exist.

Easy to repro at https://mempool.ninja/lightning/channel/garbage

At minimum the skeleton loader and the infinite spinner should be removed. Potentially replace the 404 error message to take into account indexing and inform the user to come back later.

![channel](https://user-images.githubusercontent.com/100320/214934752-ec30d9bd-2bfa-448d-9dea-363509627fad.png)

PR #3130, opened 2023-02-25
Improve error handling on channel component

nymkappa:
Fixes https://github.com/mempool/mempool/issues/2975

Applied the same logic as for non existing node

image


softsimon (approved):
tested ACK @ v2.5.0-dev [[9a246c68d](https://github.com/mempool/mempool/commit/9a246c68d)]

PR #2687, opened 2022-11-14
Add general-purpose disclaimer to top of faq

hunicus:
To make clear that mempool.space merely provides data...linkable at `/docs/faq#disclaimer`.

The SVG is free for commercial use with no attribution required ([source](https://pixabay.com/vectors/exclamation-mark-warning-danger-98739)).

Screenshot from 2022-11-14 00-56-57

PR #3134, opened 2023-02-25
Add i18n tag for big faq disclaimer

hunicus:
Lingering commit sitting on my machine...rebased onto current master.

re: https://github.com/mempool/mempool/pull/2687#pullrequestreview-1189222710


softsimon (approved):
tested ACK

PR #3148, opened 2023-02-26
When we re-index blocks due to mining pools change, wipe the nodejs backend cache

nymkappa:
When we re-index some blocks due to a mining pool definition change, we also need to wipe the backend cache to make sure we don't serve blocks with the wrong mining pool (usually happen with unknown blocks).

Now the only thing that is not automated and may lead to inconsistency is the nginx cache. So if you update mining pools using the `--update-pools` command, you also need to manually wipe nginx cache.

Linked to https://github.com/mempool/mempool/pull/2869

### Testing

- Confirm you have some cached data in your nodejs backend cache (`cache/cache*.json`)
- Update your nodejs backend configuration like so (you can see the expected change here https://github.com/mempool/mining-pools/compare/master...nymkappa/test)
```
    "POOLS_JSON_URL": "https://raw.githubusercontent.com/mempool/mining-pools/nymkappa/test/pools-v2.json",
    "POOLS_JSON_TREE_URL": "https://api.github.com/repos/mempool/mining-pools/git/trees/nymkappa/test"
```
- Restart your nodejs backend using `npm run start --update-pools` command line option
- Confirm your see the following `notice` in the log `Wipping nodejs backend cache/cache*.json files`
- Confirm your nodejs backend `cache/cache*.json` have been deleted
- Test ACK this PR
- Refill @nymkappa 's coffee cup
- Don't forget to remove the custom nodejs backend config shared above when you're done


softsimon (approved):
tested ACK @ c404895

PR #3157, opened 2023-02-26
Add api documentation for blocks-bulk

hunicus:
Add documentation for the new `blocks-bulk` endpoint (#3070).

Since it will only be available to enterprise sponsors, this PR also adds a mechanism to label endpoints with availability. Currently there are 2 options: available only to enterprise sponsors, and available only on mempool.space (which may be useful for block audit endpoints in the near-future).

EDIT: I have removed the code for including an availability note from this PR since `blocks-bulk` will be available publicly. The mechanism can be added in a future PR if/when it is needed.

This is how the note looks:

Screenshot from 2023-02-28 04-14-53


nymkappa:
The new note looks good.

One thing that is incorrect in the description is `up to 100 blocks`. This value actually depends on the nodejs backend configuration called `config.MEMPOOL.MAX_BLOCKS_BULK_QUERY`, which by default is set to `0` (this disable the endpoint).

I think what would be nice is to show our default production setting when `*ngIf="officialMempoolInstance"`, and `config.MEMPOOL.MAX_BLOCKS_BULK_QUERY` (literally that string) when it's a self hosted instance.

hunicus:
Thanks @nymkappa for pointing that out. Spoke to @wiz just now and the results will be limited to 10 for general public usage, and then folks who want more can sign up for an Enterprise sponsorship.

I think that's sufficient for now. Including the effect of configuration settings on API results is something that will be easier said than done (I know I said otherwise when we spoke in real-life...but I forgot that the text for descriptions is inserted with `innerHTML` so (as far as I know) it's not easily possible to get Angular logic to work that way). I am aiming to make this more feasible with my docs rewrite.

Updated the screenshot in the opening post to reflect the latest commit.

nymkappa (approved):
Tested ACK @ 7b0aa8616994db83a8fa8b8665db23ebb7dfc980

hunicus:
Rebased on current master.

PR #3158, opened 2023-02-26
Truncate `coinbase_signature` and `coinbase_signature_ascii` before insertion if needed

nymkappa:
Testnet block 000000007652b10d91fcf118ed315a57e5491a2514e6e0101950185e0b42b679 contains a coinbase transaction script of nearly 1MB.

This breaks the blocks indexing because the data does not fit the `varchar(500)`.

Since we probably don't want to store up to 1MB of that junk in the `coinbase_signature` and `coinbase_signature_ascii` fields so we just truncate them before insertion if needed.

This does not make sense economically for a miner to do those transactions on mainnet so IMO it's fine to have this hardcoded fix for now to testnet/signet indexing won't break.


wiz (approved):
Tested ACK @ v2.5.0-dev [[5fba448d](https://github.com/mempool/mempool/commit/5fba448d)]

PR #3160, opened 2023-02-26
Run Cypress tests on master after merging

knorrium:
When we updated the triggers to skip the `ops` PRs on #1923, the push trigger was removed completely, which caused the tests not to run once PRs are merged.

This adds the trigger back, and should fix the test status badge on the README.


wiz (approved):
Yolo ACK

Issue #3131, opened 2023-02-25
Fix `pool.id` object in block API response


nymkappa:
Works well, however `extras.pool.id` returned in the API does not match the `id` from `pools-v2.json` so please fix that in a separate PR

_Originally posted by @wiz in https://github.com/mempool/mempool/pull/2869#pullrequestreview-1314472328_

Issue #3121, opened 2023-02-24
Block(s) API(s) do not return the same data fields consistently


nymkappa:
### Description

- If a block is stored in the nodejs backend cache (recent blocks), then the data returned comes from the `getBlockExtended()` function
- If a block is not indexed and is queried (therefore indexed on the fly), then the data returned also comes from the `getBlockExtended()`
- If a block is already indexed, we return the data formatted using the `prepareBlock()` function

We should always `pipe` through the `prepareBlock()` function, any block we want to serve through those API so that we have a consistent returned result.

I've also noticed that the `avgFeeRate` field is missing when the block is indexed on the fly.

### Todo

- Get rid of `SELECT *` in mysql queries when querying blocks from the database (getBlockByHash, getBlockHeight)
- Cleanup the `prepareBlock()` function so that it does not "bruteforcely" try to find the available fields
- Make sure we use the `prepareBlock()` function before sending the response
- Make sure it does not involve breaking changes or break the frontend

PR #3166, opened 2023-02-27
Always return the same block structure for endpoints which return blocks data

nymkappa:
Includes https://github.com/mempool/mempool/pull/3161
Fixes https://github.com/mempool/mempool/issues/3131
Fixes https://github.com/mempool/mempool/issues/3121
Fixes https://github.com/mempool/mempool/issues/3182
Redo of #3086

This PR provides the following refactoring/changes:
- Fix the `$getBlockExtended` so that we don't store temporary data in the cache which is then serve in APIs and cached by nginx for a month or so. We also make a better use of our existing interfaces https://github.com/mempool/mempool/blob/01d699e454648a97a4f374e120f3276666691461/backend/src/api/blocks.ts#L167
- Delete the disgusting `prepareBlock` function which was basically brute-forcing objects to find available fields 
- Remove the `SELECT blocks.*` with proper field selection https://github.com/mempool/mempool/blob/01d699e454648a97a4f374e120f3276666691461/backend/src/repositories/BlocksRepository.ts#L16
- Write a function that converts an indexed blocks fetched from the database to a `BlockExtended`  https://github.com/mempool/mempool/blob/01d699e454648a97a4f374e120f3276666691461/backend/src/repositories/BlocksRepository.ts#L860
- Fix a JSON.parse bug when we tag blocks to a mining pool when database is disabled

Note: Bascically the PR should only ADD new fields to the following APIs, the current schema should not be modified. This PR is not a breaking change in our APIs. If you notice any differences let me know (you can double check using the REST API documentation, or if your run your own instance you can also compare).

Of course this is a very sensitive PR because it impacts the following APIs. Note that it should not impact Liquid and Bisq since we use the old source code for those endpoints.
- `/api/v1/block/:hash`
- `/api/v1/blocks/:height`
- `/api/v1/blocks-bulk/:minHeight/:maxHeight` (new endpoint from https://github.com/mempool/mempool/pull/3070). Also note that the response for this API is different from older ones. So response format will be different but we still need to make sure this PR did not break it.

### Testing

- Check if indexing still works properly. This is quite simple, delete a bunch of blocks from the db and query them through the API or the UI. They will be re-indexed on the fly. You can also simply test a full block re-indexing.
- Check if all UI component using blocks data works fine and do not print javascript errors in the console (block, block list, pool block list, infinite scroll...)
- Make sure this does not have side effects on other features which relies on blocks (for example, block audit)
- Make sure liquid and bisq block list and block component are working fine. They should not be impacted by this PR but it's better to double check.
- Test with no database
- Test with with `config.MEMPOOL.BACKEND: 'none'` and `config.MEMPOOL.BACKEND: 'esplora'`
- Check that querying recent blocks, blocks indexed on the fly and already indexed blocks returns consistent response format (https://github.com/mempool/mempool/issues/3182)

@mononaut Can you please look for possible breaking changes related to the `match` field. You probably used it for the audit stuff so if you can see somewhere it could be missing please let me know. Also I don't know if you use those API for rendering previews so that may also be something you wanna test.
@wiz Please wipe nginx cache for testing this


nymkappa:
Tests pass so that's at least something 😄

mononaut:
Audit stuff and unfurler previews all look good to me, but reorg detection is getting triggered on every new block:
```
WARN: <lightning> Chain divergence detected at block 778744, re-indexing most recent data
```
I think because `formatDbBlockIntoExtendedBlock` doesn't set a `block.hash` field?

nymkappa:
> Audit stuff and unfurler previews all look good to me, but reorg detection is getting triggered on every new block:
> 
> ```
> WARN: <lightning> Chain divergence detected at block 778744, re-indexing most recent data
> ```
> 
> I think because `formatDbBlockIntoExtendedBlock` doesn't set a `block.hash` field?

Thanks! I did not even saw it in the logs. Fixed in https://github.com/mempool/mempool/pull/3166/commits/a67656389ea75f591b90f9164b717d7629038a66

softsimon (approved):
tested ACK @ v2.5.0-dev [[a67656389](https://github.com/mempool/mempool/commit/a67656389)]

wiz (approved):
Tested ACK @ v2.5.0-dev [5e484d3e]

PR #3169, opened 2023-02-27
dynamically render search crawler requests

mononaut:
This PR extends the unfurler module to serve search crawler requests with dynamically server-side rendered pages.

This allows us to provide full page content to crawlers which lack support for websockets, long-running javascript, and other advanced features.

working demo running at [mempool.monospace.live](https://mempool.monospace.live) - visit with user-agent set to `googlebot` to access in SSR mode.

Requests from search crawlers (only Googlebot, so far) are rendered in an optimized Puppeteer instance to produce a hydrated HTML snapshot, which is then stripped of any `<script>` tags before being returned.

#### Ops
- Need to add Googlebot to the nginx `$unfurlbot` user-agent condition.
- The unfurler module now differentiates between unfurl crawlers and search crawlers internally, so the `user-agent` request header needs to be preserved in the proxied request.
- The unfurler module will proxy static resource requests if necessary, but it's probably better to handle this at the nginx level.


frontend/src/app/services/seo.service.ts
```diff
@@ -7,7 +7,7 @@ import { StateService } from './state.service';
 })
 export class SeoService {
   network = '';
-  baseTitle = 'mempool';
+  baseTitle = 'Mempool';
```
wiz, line 10:
```suggestion
  baseTitle = 'mempool';
```

wiz (changes requested):

mononaut:
- Added/fixed `rel="canonical"` tags in the unfurler and search ssr responses.
- Fixed bad hostnames in meta tags.
- Hard 404s for non-existent transaction/block/address/node/pool pages etc.

wiz:
@mononaut please rebase

wiz (approved):
Tested ACK @ v3.0.0-dev [cf4f779e]

PR #3140, opened 2023-02-25
Add block health faq link on block list

hunicus:
Shows on short block list on mining dashboard and full block list at `/blocks`.

Thought a simple dashed underline achieves its purpose while being cleaner than an info icon.

Screenshot 2023-02-25 at 7 49 40 PM

Alternative.

Screenshot 2023-02-25 at 7 59 57 PM

PR #3176, opened 2023-02-28
Add block health faq link on block list

### LINKED ISSUE #3108: Tooltip for Block Health
### Description

I have to clue what is meant with "Health" of a mined block.

### Proposed solution

Add a tooltip with an explanation

#### Additional info

[mempool.space "Blocks" subpage](https://mempool.space/blocks)

hunicus:
Shows on short block list on mining dashboard and full block list at `/blocks`.

I thought a simple dashed underline achieves its purpose while being cleaner than an info icon...feedback welcome.

Screenshot 2023-02-25 at 7 49 40 PM

Alternative (not implemented):

Screenshot 2023-02-25 at 7 59 57 PM

Same as #3140, just with commit pushed to this repository. Fixes #3108.


softsimon (changes requested):
We probably need to figure out how to make a consistent way of linking to FAQ. 

I like the underlined links better than the icon in this case due to positioning, but then we'll still use the icon in other places.

We could make an exception in this case but would like to discuss it first.

PR #3184, opened 2023-03-01
Only run migration 57 if bitcoin

nymkappa:


wiz (approved):
Tested ACK @ v2.5.0-dev [[88913b1a](https://github.com/mempool/mempool/commit/88913b1a)]

Issue #3167, opened 2023-02-27
Add frontend config to toggle BTC historical price fetching


nymkappa:
If you disable your database in the nodejs backend, the frontend will always try to fetch historical price and get a 404. So we need a flag to disable that and fallback to the websocket most recent price.

PR #3185, opened 2023-03-01
Add frontend config flag to toggle historical price fetching

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3167

When the nodejs backend database is disabled, the frontend still tries to query historical price, which of course does not exists since it requires the database. Now it does not seem to break anything but that's unnecessary api calls.

This PR adds a new frontend config variable to toggle fetching historical price from the backend. If the flag is set to false, we just use the websocket price all the time.


softsimon (approved):
tested ACK @ v2.5.0-dev [[d5342a4e9](https://github.com/mempool/mempool/commit/d5342a4e9)]

PR #3186, opened 2023-03-01
Ignore negative USD prices

nymkappa:


wiz (approved):
Tested ACK @ (v2.5.0-dev) [[2c55182c](https://github.com/mempool/mempool/commit/2c55182c)]

Issue #3188, opened 2023-03-01
Cleanup get blocks loop


nymkappa:
We can cleanup this since we don't use hash to fetch block anymore

https://github.com/mempool/mempool/blob/e2ef8721d78f19cd7c4fc5064458c3cf34a480a0/backend/src/api/blocks.ts#L760-L777

PR #3190, opened 2023-03-02
Remove useless code

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3188

We don't query block by hash anymore in this function, so we can just remove all the related code as well. `Common.indexingEnabled()` checks are already done in the `indexBlock()` function.


softsimon (approved):
tested ACK @ v2.5.0-dev [[be4bd691e](https://github.com/mempool/mempool/commit/be4bd691e)]

PR #3191, opened 2023-03-02
Don't run CI on "review_requested" event

nymkappa:
I've recently assign myself to review an open PR, and it triggered the test again (which were already all green). Does not seem necessary.


knorrium (approved):
Tested ACK

PR #3194, opened 2023-03-02
Specify manual deployment support for enterprise sponsors

hunicus:
In FAQ and README docs.

FAQ:

Screenshot from 2023-03-02 03-17-33

Backend README:

Screenshot from 2023-03-02 03-28-47

Production README:

Screenshot from 2023-03-02 03-29-10


softsimon:
Why "Linux server" We can also run on FreeBSD and other? Change to something more generic?

hunicus:
@softsimon you're right, no need to specify Linux.

I've generalized that FAQ to address any server deployment. As a result I mentioned Docker, so the Docker FAQ could probably be removed now too, but I'll do that in a separate PR.

Screenshot from 2023-03-04 04-22-14

hunicus:
Latest commit switches the link to `/enterprise` to plain href. There was another link to `/enterprise` on the FAQ page so I changed that one too.

softsimon (approved):
tested ACK @ [2fffd8b4]

PR #3195, opened 2023-03-02
Add promo video to about page

hunicus:
Silent autoplay enabled. Social icons moved to bottom of about page. 

Will add analytics event in separate pull request.

sample


wiz (changes requested):
Tested NACK @ v2.5.0-dev [[b1e7bbcc](https://github.com/mempool/mempool/commit/b1e7bbcc)]

Mobile layout needs a max-width tag, video is wider than screen size

PR #3196, opened 2023-03-02
Only show electrum tab on desktop

hunicus:


wiz (approved):
Tested ACK @ v2.5.0-dev [[89d9c1d7](https://github.com/mempool/mempool/commit/89d9c1d7)]

PR #3198, opened 2023-03-03
About page video improvements

softsimon:
This fixes 2 things:

1. Adds a fixed height so that the video no longer "jumps" when loading in Chrome
2. Adding mp4 videos to `.gitignore` in the resource folder


softsimon:
yolo ACK

PR #3202, opened 2023-03-03
Use core to fetch block because esplora/electrs still return integer difficulty

nymkappa:
### Testing

- Run `npm run start --reindex-blocks`
- One the indexing is completed, confirm difficulty adjustments are correct


softsimon (approved):
tested ACK @ 7ea2d3b

Perhaps the production servers only need to delete the latest blocks and difficulty, from the time we merged the other PR.

wiz (approved):
Tested ACK @ v2.5.0-dev [[d16bda06](https://github.com/mempool/mempool/commit/d16bda06)]

PR #3212, opened 2023-03-03
Fix initial pool update when db is empty

nymkappa:
When `pools_json_sha` is not initialized in the db, always download the mining pool even if `AUTOMATIC_BLOCK_REINDEXING: false`

### Testing

- Run `npm run start --reindex-blocks` and confirm you don't need the additional `--update-pools` for the pools table to be populated

### DIFF
--- a/backend/src/api/disk-cache.ts
+++ b/backend/src/api/disk-cache.ts
@@ -62,7 +62,7 @@ class DiskCache {
   }
 
   wipeCache() {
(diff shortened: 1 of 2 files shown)


softsimon (approved):
tested ACK @ 62ef1d44

softsimon (approved):
tested ACK @ 62ef1d44

PR #3213, opened 2023-03-03
Always show BTC amount in block tooltips

### LINKED ISSUE #3210: Always show BTC value in block preview tooltip
### Steps to reproduce
- Open https://mempool.space/block/00000000000000000002d74a1fa949befdd6327231df8b9ea5a828c5155c63ed
- Scroll to the transaction list
- Switch to fiat display mode by clicking on the output amount (in blue)
image

- Open any mempool block
- Hover the block preview
- Amount is shown in fiat. There is no way to switch back to BTC. Also if price is missing for some reason, we cannot know the tx value because it will show 0

Screenshot 2023-03-03 at 17 17 19

### Solution

Always show the BTC value of the transaction in the `Amount` field. 
We cannot display the usual `xxxx BTC    $xxx` layout because it won't fit in the tooltip.

mononaut:
Fixes #3210

### DIFF
--- a/frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.html
+++ b/frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.html
@@ -16,7 +16,7 @@
       </tr>
       <tr>
         <td class="td-width" i18n="dashboard.latest-transactions.amount">Amount</td>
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK @ v2.5.0-dev [[6a66d0620](https://github.com/mempool/mempool/commit/6a66d0620)]

PR #3220, opened 2023-03-04
drop decimal places from large fiat values

### LINKED ISSUE #3180: Round up fiat values over certain amounts
We don't need to show the 2 decimals on multiple places where values are often bigger.

A good solution could be to always display 2 decimals no matter the currency, as long as the value is below 100, everything above will just round up to whole digits.

Screenshot 2023-03-01 at 14 46 07
Screenshot 2023-03-01 at 14 46 19

mononaut:
Resolves #3180 by rounding fiat values >= 1000 to the nearest integer, while otherwise preserving localized currency formatting.


softsimon (approved):
tested ACK @ v2.5.0-dev [a37bcfec6]

PR #3223, opened 2023-03-04
Fix overlapping columns in cpfp table on small screens

mononaut:
| Before | After |
|-|-|
| Screenshot 2023-03-04 at 12 23 41 AM | Screenshot 2023-03-04 at 12 21 53 AM |


softsimon (approved):
tested ACK @ 269dcb2b

PR #3224, opened 2023-03-04
ops: Add missing /api/address-prefix nginx route for bisq

wiz:
Fixes bisq.markets address search autocomplete issue


softsimon (approved):
utACK

PR #3225, opened 2023-03-04
Make faq disclaimer responsive

hunicus:
Switch FAQ disclaimer from split 2-column layout to full single-column layout for mobile.

## Desktop

Screenshot from 2023-03-04 03-28-21

## Mobile

Screenshot from 2023-03-04 03-28-47


softsimon (approved):
tested ACK @ 7b24b12

PR #3234, opened 2023-03-05
Show ln channel world map using 100% height

nymkappa:


softsimon (approved):
tested ACK @ v2.5.0-dev [[789867a03](https://github.com/mempool/mempool/commit/789867a03)]

PR #3236, opened 2023-03-05
rewrite rpc api in typescript

antonilol:
rewrite rpc api to more readable and clean typescript using import and export keywords, default export and no more `// @ts-ignore`

was split of from #2846


softsimon:
Contains cookie stuff and a refactor of the rpc is a bit too risky to just merge without doing it carefully ourselves

antonilol:
> Contains cookie stuff

it was split off from that pr, it would disappear when rebasing this after the cookie pr is merged

> a refactor of the rpc is a bit too risky to just merge without doing it carefully ourselves

ok, sure
the goal was to make this part of the code more readable and easier to audit

PR #3237, opened 2023-03-06
unify time rendering components

mononaut:
Consolidates the `app-time-since`, `app-time-until` and `app-time-span` time localization & rendering components into a less verbose single `app-time` component with a `kind` input for specifying which format to use.


softsimon (approved):
tested ACK @ 7f78fefb

PR #3249, opened 2023-03-07
Fix typo in bulk config variable

knorrium:


softsimon (approved):
utACK

nymkappa:
Thank you for fixing that 😀

PR #3258, opened 2023-03-08
Redesign difficulty adjustment dashboard widget

mononaut:
This draft PR implements a redesign of the difficulty adjustment widget on the main dashboard to:
 - improve symmetry with the transaction fee widget.
 - visualize the relationship between number of expected vs mined blocks and the estimated difficulty adjustment.
 - make the current epoch progress more prominent.
 
Before:
Screenshot 2023-03-07 at 7 21 56 PM

After:
Screenshot 2023-03-07 at 7 43 45 PM

After (downward expected adjustment):
Screenshot 2023-03-07 at 7 46 05 PM

The progress bar consists of a 234x9 grid of pixels (scaled up/down to fit the UI), each representing one block in the difficulty epoch, color-coded as follows:
| Color | Mined | Expected |  |
|-|-|-|-|
| Purple/blue gradient | ✅ | ✅ | |
| Green | ✅ | ❌ | ahead of target schedule |
| Red | ❌ | ✅ | behind target schedule |
| Flashing white | ~ | ~ | next block |

Todo:
- [x] Mouseover info on difficulty progress bar sections.

### DIFF
--- a/frontend/src/app/components/difficulty/difficulty-tooltip.component.html
+++ b/frontend/src/app/components/difficulty/difficulty-tooltip.component.html
@@ -1,0 +1,41 @@
+<div
+  #tooltip
+  *ngIf="status"
+  class="difficulty-tooltip"
+  [style.visibility]="status ? 'visible' : 'hidden'"
+  [style.left]="tooltipPosition.x + 'px'"
+  [style.top]="tooltipPosition.y + 'px'"
+>
+<ng-container [ngSwitch]="status">
+  <ng-container *ngSwitchCase="'mined'">
+    <ng-container *ngIf="isAhead">
+      <ng-container *ngTemplateOutlet="expected === 1 ? blocksSingular : blocksPlural; context: {$implicit: expected }"></ng-container>
+      <ng-template #blocksPlural let-i i18n="difficulty-box.expected-blocks">{{ i }} blocks expected</ng-template>
+      <ng-template #blocksSingular let-i i18n="difficulty-box.expected-block">{{ i }} block expected</ng-template>
+    </ng-container>
+    <ng-container *ngIf="!isAhead">
+      <ng-container *ngTemplateOutlet="mined === 1 ? blocksSingular : blocksPlural; context: {$implicit: mined }"></ng-container>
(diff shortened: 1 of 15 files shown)


frontend/src/app/components/difficulty/difficulty-tooltip.component.html
```diff
@@ -0,0 +1,41 @@
...
+  class="difficulty-tooltip"
+  [style.visibility]="status ? 'visible' : 'hidden'"
+  [style.left]="tooltipPosition.x + 'px'"
+  [style.top]="tooltipPosition.y + 'px'"
+>
+<ng-container [ngSwitch]="status">
+  <ng-container *ngSwitchCase="'mined'">
+    <ng-container *ngIf="isAhead">
+      <ng-container *ngTemplateOutlet="expected === 1 ? blocksSingular : blocksPlural; context: {$implicit: expected }"></ng-container>
+      <ng-template #blocksPlural let-i i18n="difficulty-box.expected-blocks">{{ i }} <span>blocks expected</span></ng-template>
```
softsimon, line 13: The `<span>` tags can be removed in these i18n strings as they add no styling, just cause trouble in the i18n key

mononaut: fixed

mononaut:
should the progress bar go from right to left, to match the direction of the blockchain?

then we could also add a curved endcap on the left side.

i.e.
Screenshot 2023-03-09 at 4 04 57 AM

wiz (changes requested):
- Remove "1952 blocks"
- Move "+29 ahead" to the main text above
- Add "avg block time" where "+29 ahead" used to be

nymkappa:
Maybe while we're at it, we should increase the lag, because it always show wild adjustments for a few days @softsimon 
`const ESTIMATE_LAG_BLOCKS = 146; // For first 7.2% of epoch, don't estimate.`

mononaut:
Screenshot 2023-03-11 at 12 01 32 PM

wiz (changes requested):
maybe try something like this
Screenshot 2023-03-11 at 17 02 19

wiz (approved):
Tested ACK @ v2.5.0-dev [[ef0cc9d2](https://github.com/mempool/mempool/commit/ef0cc9d2)]

PR #3259, opened 2023-03-08
Fix memory-intensive getCPFPUnindexedBlocks mysql query

mononaut:
Fixes a bug in the `$getCPFPUnindexedBlocks` mysql query that inadvertently attempted to load 5000x more db rows than necessary, consuming a huge amount of memory every hour when the indexer runs.


wiz (approved):
Tested ACK @ 3e31e68a19927fdd35c2a27a6d3b5cd9c5843263

Memory leak is fixed!! 🎉 
Before and After memory usage:
Screenshot 2023-03-08 at 15 09 58

PR #3264, opened 2023-03-08
Fix stale mining dashboard data

mononaut:
Fixes a bug where the block count on the mining pools and hashrate chart widgets on the mining dashboard page would not update.

#### Steps to reproduce:
 1. Go the the /mining page
 2. Note the `Blocks (1w)` and `Hashrate` fields
 3. Wait for a few new blocks to arrive
 4. Observe that the `Blocks (1w)`, `Hashrate` etc values do not update.
 
This PR attempts to refresh the data when a new block arrives.

The api responses may be cached, so the data still might not update immediately if blocks arrive within <5 minutes.


softsimon (commented):
tested ACK @ v2.5.0-dev [c3c069684]

PR #3281, opened 2023-03-09
Add subtitles for mempool promo video

TechMiX:
This is a proof of concept for adding subtitles to the promo video on the about page.

Screenshot from 2023-03-09 01-36-36

Currently subtitles are in SRT format, but `video` tag caption work with WebVTT format. Used [srt-to-vtt](https://www.npmjs.com/package/srt-to-vtt) to convert them. Added a new script, `i18n-convert-srt-to-vtt` to do this. This script is called right after `i18n-pull-from-transifex` is called.

We should probably generate `track`s on the `video` tag dynamically. Also set the `default` according to the user's selected language.


hunicus:
Hey, thanks for working on this. 

I am working toward the same thing with a slightly different approach:
1. store vtt files in mempool-promo repo, along with other promo video assets
2. obtain vtt files from YouTube through `yt-dlp`

(2) removes the need to convert SRT to VTT, since `yt-dlp` gets subtitles in vtt format; in addition, we need to use YouTube subtitles because sometimes translators make changes to subtitles from Transifex to improve timing after seeing subtitles together with the video (for example, the timing for `ja` subtitles from Transifex is totally off, so the YouTube subtitles need to be used).

(1) is a consequence of (2) because these translations are static and can be stored in a GitHub repository, although I guess they could be fetched using `yt-dlp` at build time too. But fetching files from GitHub is probably more reliable long-term than using `yt-dlp`.

For reference, the command I am using is:

```
yt-dlp https://www.youtube.com/watch?v=hFIF61INmQs --skip-download --write-subs --sub-langs="<language code>"
```

Anyway what you've made is a good base. Would you be interested in adapting it to fit the approach above?

TechMiX:
Hey @hunicus !

Thank you for reviewing. Hope this boosts the effort. I really didn't know there was an effort towards this matter. I just checked the issues and PRs list and couldn't find anything related (Are there more places to look for active things?).

I will be more than happy to help with this. But I'm a bit unsure about youtube. Really don't like the idea of having youtube as our source of truth. Is it like the subtitles there are in sync with Transifex and they are fixed somehow? I mean what is the main reason behind subtitles on youtube being more correct for Japanese?

Also, regarding having the translations on a separate repo, should we then make that repo a git sub-module in the `resources` folder? Doesn't that have a side effect on the building process? Or is there another plan to link them?

hunicus:
We had just chatted about it amongst ourselves...the video just went live and it wasn't a huge effort so I didn't bother to make an issue.

>  Is it like the subtitles there are in sync with Transifex and they are fixed somehow? I mean what is the main reason behind subtitles on youtube being more correct for Japanese?

Yes see my explanation for (2) above. We have observed that timing can be off with Transifex strings. I'm not keen on having YouTube be a source of truth either, which is why I think it would be better to store the subtitles files on GitHub instead.

> should we then make that repo a git sub-module in the resources folder? Doesn't that have a side effect on the building process? Or is there another plan to link them?

No need since we're just downloading static files...see how we're getting the video in `sync-assets.js` for reference (#3195).

Issue #858, opened 2021-10-06
Focus search field on first page load


softsimon:
Focus the input field as the page is loaded. Make sure this doesn't steal focus from current key navigation features.

softsimon:
fixed in https://github.com/mempool/mempool/pull/3293

PR #3293, opened 2023-03-09
Autofocus search input when we load the app for the first time

nymkappa:
When you load the page, the search input is autofocused

image

Then you go on with browsing and the autofocus is not stolen as expected


softsimon (approved):
tested ACK

softsimon:
fixes https://github.com/mempool/mempool/issues/858

Issue #3263, opened 2023-03-08
need to sort ASNs for ISPs with more than one ASN


wiz:
Currently we are warm caching the following URL:
```
'/api/v1/lightning/nodes/isp/14618,16509' `# Amazon` \
```

However if you click on Amazon sometimes it's this URL and the cache misses:
```
/lightning/nodes/isp/16509,14618
```

Please always numerically sort the list of ASNs (as opposed to lexicographically sorting them)

PR #3296, opened 2023-03-10
Sort asn numerically - add few more top 10 isp in warm cache

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3263


softsimon (approved):
code ACK

Issue #304, opened 2021-01-26
Backend fails to start when mempool cache is corrupted


softsimon:
Noticed on raspberry pi which often restarts and the mempool cache dump is being corrupted upon write.

Suggested solutions:
* Don't save cache on shutdown, only on new blocks
* Gracefully handle corrupted cache files and skip the cache in that case

PR #3299, opened 2023-03-10
Save cache to disk on SIGTERM/SIGINT

### LINKED ISSUE #3292: Restore the feature of triggering disk cache write when aborting the server
We only write disk cache when a new block is found. But with the RBF Cache coming, it's better, especially when developing to write to cache on force exit (ctrl-c).

We did this before so the code should exist in the history.

mononaut:
Resolves #3292, by reverting #304 to synchronously save the cache to disk when the backend receives a SIGTERM/SIGINT signal.

Need to check if this still causes cache corruption on Raspberry Pi devices.

### DIFF
--- a/backend/src/api/disk-cache.ts
+++ b/backend/src/api/disk-cache.ts
@@ -16,14 +16,26 @@ class DiskCache {
   private static CHUNK_FILES = 25;
   private isWritingCache = false;
 
-  constructor() { }
+  constructor() {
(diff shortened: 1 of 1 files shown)


mononaut:
Updated to write data to temporary files, and then rename, so that aborted writes do not cause corrupted cache data.

softsimon (approved):
tested ACK @ 46d89ac8

PR #3300, opened 2023-03-10
Fixes changed after checked error in transaction page

softsimon:
This error is caused when navigating back and forth from a transactions page.

Fixing this may solve other unexpected issues.


nymkappa (approved):
Looks like that fixes it but it would be good to not rely on `setTimeout` because it can just create weird inconsistent behaviors.

I can see there is another one here which seem to be even more error prone to me hehe https://github.com/mempool/mempool/blob/dd01371b61a3b5b8c998585b4423c0b25a91fe7e/frontend/src/app/components/transaction/transaction.component.ts#L339

But that fixes the issue so I'll tested ACK @ dd01371b61a3b5b8c998585b4423c0b25a91fe7e

PR #3303, opened 2023-03-10
batch address outspend lookups into <50 txids per request

mononaut:
Esplora only supports batched requests for outspends for up to 50 txids at a time, but in some circumstances the address page can try to request outspends for as many as 75 transactions, resulting in a `Too many txids requested` 400 error.

This PR splits outspend requests into batches of up to 25 txids.


wiz (approved):
Tested ACK @ v2.5.0-dev [[36e1777b](https://github.com/mempool/mempool/commit/36e1777b)]

PR #3305, opened 2023-03-10
Remove search bar focus border

softsimon:
This is needed now when the Search Bar is autofocusing.

Downside is that the blinking cursor is the only indication of focus, but fine IMO.

Before
Screenshot 2023-03-10 at 20 18 14

After
Screenshot 2023-03-10 at 20 18 07


nymkappa:
Did you try on mobile? I'm wondering if auto focusing the search bar would actually trigger the os keyboard to show up automatically.

nymkappa (approved):
Tested ACK @ d3d67627f3090a36dff30a145befb3384fb803f3

PR #3306, opened 2023-03-11
Display same fee span on blocks

### LINKED ISSUE #3282: Use same Fee Span algorithm to display on Block and Block details
### Discussed in https://github.com/mempool/mempool/discussions/3278

<div type='discussions-op-text'>

<sup>Originally posted by **JoseSK999** March  8, 2023</sup>
518F2410-FEBA-4DD8-BA6C-1615F36F717D

As you can see the min fee rate of the same block is not displayed coherently, unlike the max fee rate. The upper min fee rate displayed is not true (as I have found transactions with less fee rate). On the other hand the min fee rate on the lower side is the one used in the fee rate chart, but now I’m not sure if this data is legit or not.</div>

softsimon:
fixes #3282

We were displaying the fee-span differently on the blockchain blocks and block details below.

Now displaying the same, which is removing the first fee span number in order to get rid of "1"s etc that don't really show the minimum required fee to get into the block.

### DIFF
--- a/frontend/src/app/components/block/block.component.html
+++ b/frontend/src/app/components/block/block.component.html
@@ -121,7 +121,7 @@
     <ng-container *ngIf="!isLoadingBlock; else loadingRest">
       <tr *ngIf="network !== 'liquid' && network !== 'liquidtestnet'">
         <td i18n="mempool-block.fee-span">Fee span</td>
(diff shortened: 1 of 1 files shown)


wiz (approved):
Tested ACK @ v2.5.0-dev [[627c913d](https://github.com/mempool/mempool/commit/627c913d)]

PR #3304, opened 2023-03-10
Fixes arrow position on confirmed blocks

softsimon:
fixes #3294

**How to test**

1. Start tracking a tx that will most likely be included in the next block
2. Wait for two confirmations
3. Press the back button of the web browser
4. Press the forward button of the web browser
5. There should no longer play a confirmation sound, and the arrow should not wrongly move to the first block. This should only happen at the confirmation.

PR #3307, opened 2023-03-11
Updated regtest example

nymkappa:
No code change, I've updated the regtest examples since I had to use it for testing https://github.com/mempool/mempool/pull/3304

The updated script example allows to use regtest like signet/testnet but with much more activity and short block time

image


softsimon (approved):
utACK

PR #3312, opened 2023-03-11
Fix epoch length in new difficulty widget

mononaut:
Difficulty adjusts every 2016 blocks, not every 2106 🙈


wiz (approved):
Tested ACK @ v2.5.0-dev [[6b544ac1](https://github.com/mempool/mempool/commit/6b544ac1)]

Before:
Screenshot 2023-03-11 at 19 48 02
After:
Screenshot 2023-03-11 at 19 47 48

PR #3314, opened 2023-03-12
Fix some responsive issue on the node component

nymkappa:
Some elements of the node page are not very responsive.
This PR shorten the average channel distance layout and add a dynamic show/hide of columns in the channel list component.

image

https://user-images.githubusercontent.com/9780671/224518976-9f90c73c-7e63-44d5-8566-734cf270a897.mov


softsimon (approved):
tested ACK @ 31cfbf662

PR #3315, opened 2023-03-12
Use effective fee rate heuristics for block fee span

### LINKED ISSUE #3302: Block fee span should use effective fees
Mempool blocks and mined blocks should display and store the fee range using the effective fees instead. Otherwise we don't get an informative fee information of what was actually the minimum fee to get into the block.

Screenshot 2023-03-10 at 15 26 25

mononaut:
Resolves #3302 

This PR takes advantage of our CPFP-based effective fee rate data to improve the fee span and median fee information for mined blocks.

It uses some heuristics to skip outliers and identify the minimum effective fee required and the maximum effective rate paid to get into each block.

The “median fee” now represents the approximate fee rate paid for the *median vbyte* of each block (specifically, the average rate paid for the middle ~10000 weight units). This should be a more consistent and informative metric than the fee rate of the median *transaction*, which can be skewed significantly one way or the other by the composition of the block.

Fee spans will be calculated & indexed using the new method only for new blocks. Updating the fee info for existing blocks currently requires a full CPFP re-index. This will be improved in a future PR.

Before:
_fee span shows the raw minimum and maximum fees, ignoring CPFP relationships_
Screenshot 2023-03-12 at 11 36 10 AM

After:
_fee span shows the minimum and maximum effective rates_
Screenshot 2023-03-12 at 1 39 15 PM

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -2,7 +2,7 @@ import config from '../config';
 import bitcoinApi, { bitcoinCoreApi } from './bitcoin/bitcoin-api-factory';
 import logger from '../logger';
 import memPool from './mempool';
-import { BlockExtended, BlockExtension, BlockSummary, PoolTag, TransactionExtended, TransactionStripped, TransactionMinerInfo, EffectiveFeeStats, CpfpSummary } from '../mempool.interfaces';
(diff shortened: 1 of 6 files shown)


softsimon (approved):
tested ACK @ [2fc404a]

PR #3316, opened 2023-03-12
New median fee calculation for mempool blocks

mononaut:
_builds on #3315_ 

This PR applies the new median fee calculation from #3315 to projected mempool blocks.

This will also affect our fee recommendations.


softsimon (approved):
tested ACK @ [a4c027d]

PR #3317, opened 2023-03-12
Claim your node

nymkappa:
Built on top of https://github.com/mempool/mempool/pull/3067

Demo:

https://user-images.githubusercontent.com/9780671/224529218-045127c1-556e-448c-bb7f-4a90630d0797.mov


wiz:
@softsimon this PR is ready for review. we've tested it a lot already but would appreciate your code review

softsimon (approved):
tested ACK @ [6fa8b960]

Looks good.

I hope I can find a flexboxy way of only using one instance of `<app-node-owner`  in the template, so that we can move the getNodeOwner$ call into the component itself. Also we could add the OFFICIAL_MEMPOOL_SPACE *ngIf check onto the` <app-node-owner `component instead of inside so it doesn't have to be instantiated.

nymkappa:
Integrated in https://github.com/mempool/mempool/pull/3795. Closing for now since we won't offer the feature yet.

PR #3319, opened 2023-03-12
Show tx fee ratings for older blocks

mononaut:
This PR takes advantage of the block data that is loaded & cached for the infinitely scrollable blockchain to calculate and show fee ratings badges for all confirmed transactions (instead of only transactions in the latest 8 blocks).

E.g:
Screenshot 2023-03-12 at 3 55 49 PM


wiz (approved):
Tested ACK @ v2.5.0-dev [[cb04d67d](https://github.com/mempool/mempool/commit/cb04d67d)]

softsimon (approved):
tested ACK @ v2.5.0-dev [cb04d67d]

PR #3325, opened 2023-03-12
Fix units in flow diagram tooltips for liquid assets

mononaut:
Use the proper ticker symbol for non-LBTC asset input/output values in the flow diagram tooltips on Liquid:

Before :
Screenshot 2023-03-12 at 5 42 47 PM

After:
Screenshot 2023-03-12 at 6 07 17 PM


wiz (approved):
Tested ACK @ v2.5.0-dev [72663b30]

Issue #1059, opened 2021-12-29
Add a warning on all testnet and signet networks


softsimon:
A warning message telling users that the coins have no value should be present on all the testnet networks in order to help prevent scams

Example:
<img width="390" alt="Screen Shot 2021-12-30 at 02 48 16" src="https://user-images.githubusercontent.com/8561090/147708860-3ebc081c-597a-465a-8cf1-15e31ed12d0e.png">

PR #3326, opened 2023-03-12
Show warning on testnet/signet

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1059

Concept. The warning bar can be closed and the preference is save in local storage

Screenshot 2023-03-12 at 18 37 20

Screenshot 2023-03-12 at 18 37 31

### DIFF
--- a/frontend/src/app/components/master-page/master-page.component.html
+++ b/frontend/src/app/components/master-page/master-page.component.html
@@ -62,6 +62,21 @@
 </nav>
 </header>
 
+<div *ngIf="network.val === 'testnet' || network.val === 'signet' || network.val === 'liquidtestnet'">
+  <div class="container p-lg-0 pb-0" style="max-width: 100%; margin-top: 7px" *ngIf="storageService.getValue('hideWarning') !== 'hidden'">
+    <div class="alert alert-danger mb-0 text-center">
+      <div class="d-flex justify-content-center align-items-center">
+        <fa-icon class="between-arrow mr-1" [icon]="['fas', 'exclamation-triangle']" [fixedWidth]="true"></fa-icon>
+        <span i18n="warning-testnet">This is a test network. Coins have no value</span>
+        <fa-icon class="between-arrow ml-1" [icon]="['fas', 'exclamation-triangle']" [fixedWidth]="true"></fa-icon>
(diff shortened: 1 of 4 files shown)


frontend/src/app/components/master-page/master-page.component.html
```diff
@@ -62,6 +62,21 @@ <h6 *ngIf="env.LIQUID_ENABLED || env.BISQ_ENABLED" class="dropdown-header" i18n=
 </nav>
 </header>
 
+<div *ngIf="network.val === 'testnet' || network.val === 'signet'">
+  <div class="container p-lg-0 pb-0" style="max-width: 100%; margin-top: 7px" *ngIf="storageService.getValue('hideWarning') !== 'hidden'">
+    <div class="alert alert-danger mb-0 text-center">
+      <div class="d-flex justify-content-center align-items-center">
+        <fa-icon class="between-arrow mr-1" [icon]="['fas', 'exclamation-triangle']" [fixedWidth]="true"></fa-icon>
+        Bitcoin {{ network.val | titlecase }} is used for testing - Coins have no value
```
softsimon, line 70: We should make this an i18n key so it can be translated. But we might need to find a way that makes sense to add the network as a variable. Or change to a generic title that says "this is a Testnet" even if it's Signet or Liquid testnet etc.

nymkappa: image

frontend/src/app/components/master-page/master-page.component.html
```diff
@@ -62,6 +62,21 @@ <h6 *ngIf="env.LIQUID_ENABLED || env.BISQ_ENABLED" class="dropdown-header" i18n=
 </nav>
 </header>
 
+<div *ngIf="true || network.val === 'testnet' || network.val === 'signet'">
```
softsimon, line 65: add liquidtestnet

nymkappa: Fixed in https://github.com/mempool/mempool/commit/d54aed376d0fbcee770bc4aa1e5c00c65c658ccb

softsimon (approved):
tested ACK @ [7562407]

PR #3327, opened 2023-03-12
Show cumulated fee on last mempool block

nymkappa:
### Currently

Screenshot 2023-03-12 at 19 15 05

### Fixed

Screenshot 2023-03-12 at 19 14 00


softsimon (approved):
tested ACK @ eba0e1c25

PR #3329, opened 2023-03-12
Add mutiny as community integration

hunicus:
Also juggle CSS to make rows symmetric and full.

Screenshot from 2023-03-12 06-19-30


softsimon (approved):
tested ACK @ a5e2817

PR #3330, opened 2023-03-12
Save cache to disk every 6 blocks

mononaut:
Reduce disk cache save frequency from every 1 block to every 6 blocks


softsimon (approved):
tested ACK @ v2.5.0-dev [[8bd05987e](https://github.com/mempool/mempool/commit/8bd05987e)]

PR #3331, opened 2023-03-12
Add network versioning to disk cache

mononaut:
Save the current network version to the disk cache file, so we can avoid loading an incompatible cache after switching networks.


softsimon (approved):
tested ACK @ 1b2810ec

Princeprince559:
Active

PR #3332, opened 2023-03-12
Ctdl game special

Czino:
As discussed with @wiz, this is a special easter egg to be shown only on April 1st.

It's an excerpt from the CTDL-GAME, a game I have put on ice for now but the mempool-spa stage fit's in the context of mempool.space. The stage features @wiz and @softsimon and a few other more or lesser known characters int he bitcoin space.

Yes, those are JS files and there is some hacks to make it work in an angular context.

Let me know if the temporary integration is apt for a nice little April fools day joke.


TechMiX:
Tested d93e14b679aae1f7258869e2fd89e16ecbf0bdef

Great work! :grin:  I really enjoyed it.

Here's my two cents on it:
- Adjust the border background to make it consistent with the page background. It's gray currently.
- Control buttons on mobile are invisible. Hard to find them.
- Sometimes @wiz gets stuck and don't move at all.
- Sometimes I fell down to the first platform for no apparent reason.
- :heart: the character switch feature! But couldn't figure out what to do with it.

TechMiX:
@Czino Any plans to revive it this year? :-)

PR #3345, opened 2023-03-13
Hide features row if tx has no features

### LINKED ISSUE #3323: Hide feature row when there are no features
Screenshot 2023-03-12 at 17 20 51

mononaut:
Fixes #3323 

| Before | After |
|-|-|
| Screenshot 2023-03-13 at 12 50 29 PM | Screenshot 2023-03-13 at 12 50 18 PM |

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.ts
+++ b/frontend/src/app/components/transaction/transaction.component.ts
@@ -74,6 +74,12 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
   flowEnabled: boolean;
   blockConversion: Price;
   tooltipPosition: { x: number, y: number };
+  isMobile: boolean;
+
+  featuresEnabled: boolean;
+  segwitEnabled: boolean;
+  rbfEnabled: boolean;
+  taprootEnabled: boolean;
 
   @ViewChild('graphContainer')
   graphContainer: ElementRef;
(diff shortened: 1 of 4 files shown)


frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -495,6 +517,7 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
 
   @HostListener('window:resize', ['$event'])
   setGraphSize(): void {
+    this.isMobile = window.innerWidth < 850;
```
nymkappa, line 520: Usually better to use media query for showing/hiding html elements

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -428,9 +436,23 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
       });
   }
 
+  setFeatures(): void {
+    if (this.tx) {
+      this.segwitEnabled = !this.tx.status.confirmed || this.tx.status.block_height >= 477120;
```
nymkappa, line 441: Isn't this going to break on testnet/signet?

frontend/src/app/components/transaction/transaction.component.ts
mononaut: We need to add/remove the actual table row markup element to get the correct alternating table stripes.

`display: none` via CSS media query isn't enough.

softsimon (approved):
tested ACK @ v2.5.0-dev [[130ae8c3a](https://github.com/mempool/mempool/commit/130ae8c3a)]

Issue #3348, opened 2023-03-13
Disable pools.json data checks if mempool or blocks indexing is disabled


wiz:
If config.MEMPOOL.ENABLED === false or if block indexing is disabled then we don't care about mining pools data, so could you please disable this code:

```
Mar 13 03:58:55 node206.fra  mempool[55117]: WARN <lightning> Updated mining pools data is available (71cd80eae5ce6de28f227ddec28bef18cc62b62d) but AUTOMATIC_BLOCK_REINDEXING is disabled
Mar 13 03:58:55 node206.fra  mempool[58579]: WARN <testnet-lightning> Updated mining pools data is available (71cd80eae5ce6de28f227ddec28bef18cc62b62d) but AUTOMATIC_BLOCK_REINDEXING is disabled
Mar 13 03:58:56 node206.fra  mempool[70013]: WARN <signet-lightning> Updated mining pools data is available (71cd80eae5ce6de28f227ddec28bef18cc62b62d) but AUTOMATIC_BLOCK_REINDEXING is disabled
```

PR #3351, opened 2023-03-13
Disable pool update when running lightning only

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3348

We still need the pools.json when indexing is disabled because we index on the fly. Block(s) API return the same data regardless of indexing status 👍🏻


softsimon (approved):
Concept ACK

wiz (approved):
Tested ACK @ 4d28127

PR #3353, opened 2023-03-13
Improve mempool block animations

### LINKED ISSUE #672: UX: Block animation and display when only 1 mempool block is not great
When the mempool is almost empty and there is just 1 block waiting. When a new block is found it doesn't look like it is the mempool block that is animating over, because it appears on top of the new empty mempool block.

This is especially visible on the Liquid side (because the mempool is empty most of the time)

Possible solutions:
* Empty mempool blocks should not be visible at all.
* The "empty block" should be moved to the left so that it can be animated over together with the new mined block.

mononaut:
Fixes #672, and slightly refactors the mempool animation logic - mempool blocks now ~~always~~ slide to the right when a new block is mined _that is sufficiently similar to the expected block_, except for 'stacks' of multiple mempool blocks which remain stationary on the far left side of the screen.

After:

https://user-images.githubusercontent.com/83316221/224657677-6fd4c4ca-53b4-451c-8474-820d1676e7c2.mov

https://user-images.githubusercontent.com/83316221/224657628-1c2bef49-a9b4-4e84-8a24-28347f1bbdcd.mov

### DIFF
--- a/backend/src/api/audit.ts
+++ b/backend/src/api/audit.ts
@@ -5,9 +5,9 @@ const PROPAGATION_MARGIN = 180; // in seconds, time since a transaction is first
 
 class Audit {
   auditBlock(transactions: TransactionExtended[], projectedBlocks: MempoolBlockWithTransactions[], mempool: { [txId: string]: TransactionExtended })
-   : { censored: string[], added: string[], fresh: string[], score: number } {
+   : { censored: string[], added: string[], fresh: string[], score: number, similarity: number } {
     if (!projectedBlocks?.[0]?.transactionIds || !mempool) {
-      return { censored: [], added: [], fresh: [], score: 0 };
+      return { censored: [], added: [], fresh: [], score: 0, similarity: 1 };
     }
 
     const matches: string[] = []; // present in both mined block and template
@@ -16,6 +16,8 @@ class Audit {
     const isCensored = {}; // missing, without excuse
     const isDisplaced = {};
     let displacedWeight = 0;
+    let matchedWeight = 0;
+    let projectedWeight = 0;
 
     const inBlock = {};
     const inTemplate = {};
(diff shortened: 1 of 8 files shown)


softsimon (changes requested):
Notes for the record:

* Don't animate the block if the new block happens to be empty, or mostly contain transactions that wasn't expected, eg. was manually added by the miner.

mononaut:
Updated to not animate if the new block shares <= 50% of transactions by weight with the projected block, using a new "similarity" metric calculated during the block audit.

Liquid, and backends with audits and/or block indexing disabled will continue to animate every block as before.

mononaut:
Updated to support conditional animation based on mined block similarity even when audits or indexing are disabled.

wiz (changes requested):
Tested NACK @ 777b30e

```
Mar 15 07:35:39 [98804] DEBUG: <liquid> Exception in runMainUpdateLoop(). Retrying in 5 sec. Reason: Cannot read p
roperties of undefined (reading 'transactions'). Stack trace: TypeError: Cannot read properties of undefined (read
ing 'transactions')
    at Function.getSimilarity (/mempool/liquid/backend/dist/api/common.js:161:41)
    at WebsocketHandler.handleNewBlock (/mempool/liquid/backend/dist/api/websocket-handler.js:460:55)
    at /mempool/liquid/backend/dist/api/blocks.js:540:78
    at Array.map (<anonymous>)
    at Blocks.$updateBlocks (/mempool/liquid/backend/dist/api/blocks.js:540:66)
    at runMicrotasks (<anonymous>)
    at async Server.runMainUpdateLoop (/mempool/liquid/backend/dist/index.js:189:13)
```

mononaut:
oops, I forgot to handle the case of a completely empty mempool.

It should be fixed now, and rebased on master.

softsimon (approved):
tested ACK @ v2.5.0-dev [b5c207341]

PR #3373, opened 2023-03-14
Network-specific activation heights for transaction feature badges

mononaut:
_(builds on PR #3345)_

Previously, we incorrectly reused mainnet feature activation heights to decide when to display transaction feature badges.

This PR switches to network-specific activation height logic, set to the following heights:

Mainnet
```
rbf: 399701,
segwit: 477120,
taproot: 709632,
```

Testnet
```
rbf: 720255,
segwit: 872730,
taproot: 2032291,
```

Signet
```
rbf: 0,
segwit: 0,
taproot: 0,
```

Before:
(all feature badges hidden because signet block height is much lower than the mainnet activation height of RBF, taproot and segwit features.)
Screenshot 2023-03-14 at 1 07 26 PM

After:
Signet transactions show feature badges
Screenshot 2023-03-14 at 1 06 57 PM

### DIFF
--- a/frontend/src/app/bitcoin.utils.ts
+++ b/frontend/src/app/bitcoin.utils.ts
@@ -254,3 +254,30 @@ export function selectPowerOfTen(val: number): { divider: number, unit: string }
 
   return selectedPowerOfTen;
 }
+
+const featureActivation = {
+  mainnet: {
+    rbf: 399701,
+    segwit: 477120,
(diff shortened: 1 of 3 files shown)


softsimon (approved):
tested ACK @ v2.5.0-dev [[3458c5af7](https://github.com/mempool/mempool/commit/3458c5af7)]

Issue #3360, opened 2023-03-13
Testnet and signet should always show fiat value of 0


nymkappa:
![3A5EDBEF-063A-4F1F-8F49-34619BE763B2](https://user-images.githubusercontent.com/9780671/224710979-73dcbaaa-3b7e-44ea-9851-2a369c56a994.jpeg)

ajtowns:
Rather than showing a flat "$0" for 0.0001 sBTC and also 100,000 sBTC, how about just using the BTC/USD exchange rate against a fake currency, eg "98,77 ₰FAKE" instead of "98,77 $US" and "₰98.77" instead of "$98.77" ?

I find having an exchange rate  helpful, personally, so that it feels like "50 sBTC" is a lot and "0.00005 sBTC" isn't, even though neither are actually worth anything. I also managed to the bright "$0.00" value in the balance of an address and thought that meant it had been cleaned out for a few minutes before noticing the sBTC value wasn't also 0...

(The "₰" symbol (unicode 20B0) was used for the german penny/pfennig prior to the 1950s so seems unlikely to be mistaken for real currency, while still being a widely available symbol)

PR #3379, opened 2023-03-14
Don't fetch prices on signet/testnet, always show 0

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3360


softsimon (approved):
tested ACK @ v2.5.0-dev [a5dd14193]

PR #3380, opened 2023-03-14
Use effective fee rates in mempool block visualizations & tooltips

mononaut:
This PR adds effective fee rate information to the projected mempool block visualizations:
 1. Shows effective fee rate in the mouseover tooltips where relevant
 2. Sorts transactions in the mempool block visualization by effective fee rate, rather than raw single-tx fee rate.
 
Transaction squares remain colored by raw fee rate for consistency with the mined block visualizations.

Where the effective fee rate is identical to the raw fee rate, the tooltip field is hidden.
 
The change to sorting criteria means that transaction priority will be more accurately reflected (e.g. high priority transactions towards the 'start' of the block), and CPFP transactions will be easier to identify by their contrasting raw fee colors.

The sort is still only _approximate_, so transactions with a CPFP relationship may not be placed close together in the visualization.

Screenshot 2023-03-14 at 3 40 50 PM


mononaut:
I've just realized we also need to update the mempool diff logic to account for changes to effective fee rates over time.

Converting this PR to draft until I figure that out.

feelancer21:
Is it also fixing the shown feeranges of a mined lock?

e.g.
https://mempool.space/de/block/00000000000000000001791e55f69e9b3c64750138df52ac76bdbef34e6455cc

The lowest shown fee was 3sats/vb. But these tx are all CPFPs. Hence the lower bound of the effective feerange was higher than 3 sats.

softsimon:
> Is it also fixing the shown feeranges of a mined lock?
> 
> e.g. https://mempool.space/de/block/00000000000000000001791e55f69e9b3c64750138df52ac76bdbef34e6455cc
> 
> The lowest shown fee was 3sats/vb. But these tx are all CPFPs. Hence the lower bound of the effective feerange was higher than 3 sats.

It is fixed in this open PR

https://github.com/mempool/mempool/pull/3315

mononaut:
> I've just realized we also need to update the mempool diff logic to account for changes to effective fee rates over time.

Updated to now send effective fee rate changes as part of the mempoolBlockDelta websocket updates, and apply those in the frontend to the mempool visualization & tooltips.

Tooltip values are updated immediately.

Changes in priority are not reflected right away in the visualization positions, but txs will float to their proper positions over time as the tx squares continue to be gradually sorted.

softsimon (approved):
tested ACK @ [e6e90799]

PR #3381, opened 2023-03-14
Limit special blocks by network, add future halvings

mononaut:
This PR makes special blocks events configurable by network, instead of showing the blinking yellow block animation on every network.

Now, genesis and halving events show only on Bitcoin mainnet and testnet, and the taproot activation event only shows on Bitcoin mainnet.

There are currently no special blocks configured for signet, Liquid, or Liquid testnet.

This PR also adds new special block events up to the 15th halving.


softsimon (approved):
tested ACK @ v2.5.0-dev [[7bf8fea9f](https://github.com/mempool/mempool/commit/7bf8fea9f)]

Issue #3214, opened 2023-03-04
Fix price overflow


nymkappa:
<img width="204" alt="image" src="https://user-images.githubusercontent.com/9780671/222868093-15238bfd-3368-4e5a-85ef-1dfc06c19b05.png">

nymkappa:
Related https://github.com/mempool/mempool/issues/3180

I think in some widget like the "Latest transactions" widget, we could use power of ten display when amount is getting too big.

For example, if a transaction value is `1,234,567` fiat unit, we could display `1,234M` instead, something like that. We already use the same idea for difficulty adjustments, or avg tx fee for example. So that would be easy to implement.

<img width="139" alt="image" src="https://user-images.githubusercontent.com/9780671/222938873-1ba3ce88-abcb-40df-9aba-1abb9364198a.png">

algodondev:
Good day, @nymkappa do you think it would be proper to modify the raw fiat-component so it gets the fixing on every place? I ask this because some people may want to know the entire transaction details even when it doesn't fit in the screen.

Also, was related #3180 fixed? If not I can go with both inside this issue's fix.

nymkappa:
> Good day, @nymkappa do you think it would be proper to modify the raw fiat-component so it gets the fixing on every place? I ask this because some people may want to know the entire transaction details even when it doesn't fit in the screen.

Yes, but it needs to be configurable. Maybe take my PR as it was before it got closed and fix the issue raised by wiz (support different locale).


> Also, was related #3180 fixed? If not I can go with both inside this issue's fix.

Looks like it was already fixed yes.

algodondev:
Understood and good to know, thank you!

algodondev:
I'm about to start on this @nymkappa . Doing some research (Google Gemini) I think it would be better to use only K and M shorthands since the next ones may be hard to understand wolrdwide (for example, for latin-america, billions are not a thousand millions but a million millions). What do you think about it?

nymkappa:
@algodondev does not javascript have native support for power of ten notation based on locale?

PR #3400, opened 2023-03-15
Update fiat shortener and use in the fiat component

nymkappa:
Also fixes https://github.com/mempool/mempool/issues/3214

# Update desc

This PR updates the fiat amount component so that it can either show the full fiat number (as of today), or use a shorten version.

This PR only updates the latest transaction component to avoid overflow on the main dashboard when the fiat number is too wide.

image

~This changes the way we display fiat amount everywhere we use the fiat component. Fiat amount are not shown in a much shorter fashion, showing only the relevant part and using power of ten symbols.~

~I think this is nice because fiat value are only provided for UI convenience and are not relevant to bitcoin itself, so they should not take too much UI space either.~

~This will also greatly reduce UI overflow related issues.~

~Maybe we want to scope this change so let me know what you think.~

### DIFF
--- a/frontend/src/app/dashboard/dashboard.component.html
+++ b/frontend/src/app/dashboard/dashboard.component.html
@@ -133,7 +133,7 @@
                   </a>
                 </td>
                 <td class="table-cell-satoshis"><app-amount *ngIf="(network$ | async) !== 'liquid' && (network$ | async) !== 'liquidtestnet'; else liquidAmount" [satoshis]="transaction.value" digitsInfo="1.2-4" [noFiat]="true"></app-amount><ng-template #liquidAmount i18n="shared.confidential">Confidential</ng-template></td>
-                <td class="table-cell-fiat" *ngIf="(network$ | async) === ''" ><app-fiat [value]="transaction.value" digitsInfo="1.0-0"></app-fiat></td>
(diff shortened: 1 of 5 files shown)


softsimon:
I didn't get consensus among @hunicus and @wiz about this change, so will hold off this for now.

nymkappa:
> I didn't get consensus among @hunicus and @wiz about this change, so will hold off this for now.

Yes that's a big change I would maybe scope it only to the "lastest transaction" component since that's where I noticed the overflow.

nymkappa:
@softsimon updated the PR and scoped the change to the latest component widget only

softsimon:
Looks too weird with forced 1 digit likt this.
I would expect
$487
$44.30 or $44 (rounded)

Screenshot 2023-04-01 at 19 21 39

nymkappa:
> Looks too weird with forced 1 digit likt this. I would expect $487 $44.30 or $44 (rounded)

Updated so that we show the full value if less than 100,000.
More than that and we use power of 3, with maximum 3 significant digits.

block-health-24h-1680344730

wiz:
I think this is a locale specific thing, i.e. not all people in the world will use K or M etc.

softsimon:
Rejected

PR #3401, opened 2023-03-15
Save disk cache on sighup

softsimon:
* Catch SIGHUP to save to disk cache
* Don't save to disk cache before mempool is in sync


softsimon:
Replaced with https://github.com/mempool/mempool/pull/3448

PR #3402, opened 2023-03-15
Add subtitles for promo video

hunicus:
This is a re-do of #3281 (h/t to @TechMiX) based on my comments there.

Currently draft because I'm having issues testing locale switching on master.

UPDATE

Ready for review. Tested by manually setting `localize` property in `angular.json`.

- changed `track` attribute from "subtitles" to "captions" (see `kind` docs [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/track))
- show captions only when user is in non-English locale, otherwise, do not show captions

If this works, there are more captions ready to merge at https://github.com/mempool/mempool-promo/pull/2.


softsimon (approved):
tested ACK @ v2.5.0-dev [72f25b87]

wiz (approved):
Tested ACK @ v2.5.0-dev [[72f25b87](https://github.com/mempool/mempool/commit/72f25b87)]

PR #3404, opened 2023-03-16
Fix % on heap limit warn

nymkappa:
PR

`Mar 16 14:34:44 [49088] WARN: <lightning> Used 84.77% of heap limit (1.74 / 2.05 GB)!`

Previously

`Mar 16 14:35:57 [49862] WARN: <lightning> Used 0.84% of heap limit (1.72 / 2.05 GB)!`


softsimon (approved):
tested ACK @ [d9b4ad6]

PR #3406, opened 2023-03-16
Rename block prediction to block health

### LINKED ISSUE #3372: Rename "Block Prediction Accuracy"
This graph still uses the old labels. We should use the "Block Heath" labels now.

image

nymkappa:
Fixes #3372 

image


softsimon (changes requested):
Rename Blocks health to block health (singular) everywhere to be consistent with other stuff. Block- and Blocks Health is currently mixed.

softsimon (approved):
tested ACK @ [4dec152df0]

PR #3407, opened 2023-03-16
Add faqs on mempool size and memory usage

hunicus:
Screenshot from 2023-03-17 08-35-57


softsimon (approved):
tested ACK @ v2.5.0-dev [b0a0ad11] 

Although I would skip the whole "Small node" and "Big Node" stuff and just call it 300 MB default sized node and nodes with high memory limit.

PR #3408, opened 2023-03-16
Docker digest GHA

knorrium:
This PR adds a new manual workflow that can be triggered to print the Docker images digests to be used on Docker compose. It should add a new option under the Actions section where someone on the team can specify a build or fetch the latest.

In the future we could use the output of this action to automate the Umbrel docker-compose updates.

Example run at https://github.com/knorrium/mempool/actions/runs/4433971473/jobs/7779539768 using the push trigger because forked repos can't run `dispatch_workflow` actions until they are merged into the main branch.

Screenshot from 2023-03-16 11-06-36

### DIFF
--- a/.github/workflows/get_image_digest.yml
+++ b/.github/workflows/get_image_digest.yml
@@ -1,0 +1,27 @@
+name: 'Print images digest'
+
+on:
+  push:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


wiz (approved):
utACK

PR #3438, opened 2023-03-18
Use nodejs native https client for mining pool and price updater

nymkappa:
This PR is for testing purpose. I want to see if the "networking" issues are related to axios or something else.

This PR uses the native https client for fetching mining pools and updating prices.


nymkappa:
Considering it's not an issue anymore

PR #3442, opened 2023-03-19
When a re-org happens, keep the block templates for audit

nymkappa:
Currently, we delete rows from `blocks_summaries` table when a re-org happens. However we do want to keep the block templates, so this PR changes the way we re-index block summaries in that regard.

@mononaut I noticed that `blocks_audits` are not re-indexed upon re-org. Since the mined blocks are different, we also need to re-compare them with our template to generate the new audit data (tx diff and health score).

### DIFF
--- a/backend/src/repositories/BlocksRepository.ts
+++ b/backend/src/repositories/BlocksRepository.ts
@@ -467,30 +467,6 @@ class BlocksRepository {
   }
 
   /**
-   * Get one block by hash
-   */
-  public async $getBlockByHash(hash: string): Promise<object | null> {
-    try {
-      const query = `
-        SELECT ${BLOCK_DB_FIELDS}
-        FROM blocks
-        JOIN pools ON blocks.pool_id = pools.id
-        WHERE hash = ?;
-      `;
-      const [rows]: any[] = await DB.query(query, [hash]);
-
-      if (rows.length <= 0) {
-        return null;
-      }
- 
-      return await this.formatDbBlockIntoExtendedBlock(rows[0]);
-    } catch (e) {
-      logger.err(`Cannot get indexed block ${hash}. Reason: ` + (e instanceof Error ? e.message : e));
-      throw e;
-    }
-  }
-
-  /**
    * Return blocks difficulty
    */
   public async $getBlocksDifficulty(): Promise<object[]> {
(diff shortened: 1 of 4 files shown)


backend/src/repositories/BlocksRepository.ts
```diff
@@ -466,30 +466,6 @@ class BlocksRepository {
     }
   }
 
-  /**
-   * Get one block by hash
-   */
-  public async $getBlockByHash(hash: string): Promise<object | null> {
```
nymkappa, line 472: Dead code

mononaut:
Tested ACK @ v2.6.0-dev [[9e4fe40c](https://github.com/mempool/mempool/commit/9e4fe40c)]

I performed various forks and reorgs on regtest, and verified that:
 - block templates and audit scores are preserved when we re-index the last 10 blocks after a chain divergence is detected.
 - we do not show any audit data for reorged blocks (see notes below).
 - known templates and audits become available again if an audited block is disconnected and then later reconnected to the active chain.
 - other block data is still correctly deleted and reindexed.

---

There's no need to re-index `block_audits`.

The audits are deterministic for a given combination of template & mined block, and the saved templates are only valid for the specific mempool and blockchain configuration on our node at a certain instant.

They aren't transferrable to new blocks added later, because we don't know exactly when such blocks were mined, or what the state of the miner's mempool would have been, especially in a deeper reorg.

In that case, I think we're correct not to show any audit data at all. It might be nice to record & display that a block *was* reorged, and explain why the audit data is missing, but that's probably for a separate PR.

Both the templates and the audit data are indexed & accessed by block hash, not height, so I think it's fine to just leave them in the DB as we do now. Then the data is still available in the event that an audited block gets reconnected by a subsequent re-reorg.

By the same logic, maybe we could skip deleting transactions from the `block_summaries` table too?

nymkappa:
> By the same logic, maybe we could skip deleting transactions from the block_summaries table too?

Thanks for the clear explanation. I think you're right.

Also while looking at the `block_summaries` table I noticed I do not `INDEX` the `hash` field which we use for querying the data, so maybe I should add it in a separate PR.

mononaut:
> Also while looking at the block_summaries table I noticed I do not INDEX the hash field

Isn't that the primary key? (so indexed automatically)

nymkappa:
> > Also while looking at the block_summaries table I noticed I do not INDEX the hash field
> 
> Isn't that the primary key? (so indexed automatically)

Oh yeah. Nevermind my comment lol.

nymkappa:
Pushed an additional [commit](https://github.com/mempool/mempool/pull/3442/commits/44bbb472d3683bd83efba7d50126d08ced718720) to keep block summaries in the database upon re-org.

### Testing

- Change the hash of the 5 most recent `blocks_summaries` entries with some random hash
- Change the hash of the most recent `blocks` entry with a random hash as well
- Wait for a new block
- Confirm the re-org is detected: `Chain divergence detected`
- Confirm you now have both old and new `blocks_summaries` (you will have duplicate `height` for the entries you've manually modified in the first step

For example using regtest this is what you should see after a re-org (duplicated `height`)

image

mononaut (approved):
Tested ACK @ v2.6.0-dev [[816fb3bf](https://github.com/mempool/mempool/commit/816fb3bf)]

Retested using both the db editing method and with actual reorgs on regtest.

PR #3448, opened 2023-03-20
Fix missing temp cache in disk cache

softsimon:
Removing duplicated code. Now writing temp files also during Async writing on new blocks which can cause corruption.


wiz (approved):
Tested ACK @ v2.5.0-dev [870e8951]

Tested on both server using prod configuration and RPI device using docker

PR #3490, opened 2023-03-21
Auto disable LN on macaroon fail

softsimon:
Upon startup and LND is enabled, automatically disable LIGHTNING if no macaroon file was found.

It will look like this.

```
Mar 21 15:47:50 [60293] ERR: <lightning> Could not initialize LND Macaroon and TLS Cert. Disabling LIGHTNING. ENOENT: no such file or directory, open
Mar 21 15:47:50 [60293] NOTICE: <lightning> Starting Mempool Server... (cd2bda4)
```


wiz (approved):
Tested ACK @ v2.5.0-dev [cb8fdb5e]

PR #3491, opened 2023-03-21
Audit disabled related UX fixes

softsimon:
When Audit is disabled, these pages should now properly hide the Health status
* Individual Pool pages
* Pools Ranking

Screenshot 2023-03-21 at 16 01 36
Screenshot 2023-03-21 at 16 01 04


wiz (approved):
Tested ACK @ v2.5.0-dev [f570b276]

PR #3494, opened 2023-03-21
Difficulty mining css updates

softsimon:
Reduced font-size from 20 to 18.

Screenshot 2023-03-21 at 21 25 10

Preventing word wrapping at the "Previous" as shown in some languages
Screenshot 2023-03-21 at 21 25 03


wiz (approved):
Tested ACK @ v2.5.0-dev [2c16bbb0]

PR #3547, opened 2023-03-21
Removing "d" from node fee chart loading

softsimon:


wiz (approved):
Tested ACK @ v2.5.0-dev [a34d8714]

PR #3548, opened 2023-03-21
Update channels map indexing indicator

softsimon:
New look with empty Geo channels and Node has Geo data which can happen when
* We just started up so no channels indexed
* Node channels don't have any geo data

Screenshot 2023-03-21 at 23 14 42

### DIFF
--- a/frontend/src/app/lightning/nodes-channels-map/nodes-channels-map.component.ts
+++ b/frontend/src/app/lightning/nodes-channels-map/nodes-channels-map.component.ts
@@ -206,20 +206,13 @@ export class NodesChannelsMap implements OnInit {
   prepareChartOptions(nodes, channels) {
     let title: object;
     if (channels.length === 0 && !this.placeholder) {
-      this.chartOptions = null;
-      this.showIndexingInProgress = true;
       this.isLoading = false;
-      return;
-    }
-
-    // empty map fallback
-    if (channels.length === 0 && this.placeholder) {
       title = {
         textStyle: {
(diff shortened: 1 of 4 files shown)


frontend/src/app/lightning/nodes-channels-map/nodes-channels-map.component.ts
```diff
@@ -206,20 +205,13 @@ export class NodesChannelsMap implements OnInit {
...
     let title: object;
     if (channels.length === 0 && !this.placeholder) {
-      this.chartOptions = null;
-      this.showIndexingInProgress = true;
       this.isLoading = false;
-      return;
-    }
-
-    // empty map fallback
-    if (channels.length === 0 && this.placeholder) {
```
nymkappa, line 216: I think this condition is for the node preview component used by unfurler so it looks your change would change its behavior. So need to make sure it does not break it.

softsimon: OK I believe I restored the preview "preview component" behavior

nymkappa:
I also show the generic message if we only have one data point for the node stats historical chart.
And I've added the same fix on the `Fee distribution` chart

image

nymkappa (approved):
Tested ACK @ v2.5.0-dev [[f1f6c481](https://github.com/mempool/mempool/commit/f1f6c481)]

Screenshot 2023-03-22 at 14 40 19

image

PR #3581, opened 2023-03-27
Enabling mempool in lightning prod

### LINKED ISSUE #3579: Closed Lightning Channel not caught by closed transaction scanner
This particular channel has been closed since 1 week, but yet is just marked as Inactive and not closed.

runTask almost doesn't seem to trigger in the `network-sync.service.ts` after first launch since last scan was when the servers restarted?

Examples
https://node201.fra.mempool.space/lightning/channel/850283128250630144
https://node201.fra.mempool.space/lightning/channel/852861482918477825

Ok, the bug is most likely due to our production Lightning backend not having a mempool and not checking new blocks, so it will always skip the closed scan.

softsimon:
fixes #3579

### DIFF
--- a/production/mempool-config.mainnet-lightning.json
+++ b/production/mempool-config.mainnet-lightning.json
@@ -1,6 +1,5 @@
 {
   "MEMPOOL": {
-    "ENABLED": false,
     "NETWORK": "mainnet",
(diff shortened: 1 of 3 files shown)


wiz (approved):
utACK

PR #3586, opened 2023-03-27
Add missing sanity check when fetching single price datapoint

nymkappa:
Fixes

```
Mar 27 10:53:51 [25180] ERR: Cannot fetch single historical prices from the db. Reason Cannot read properties of undefined (reading 'USD')
```


softsimon (approved):
utACK

PR #3587, opened 2023-03-27
Update mynode profile on about page

hunicus:
Logo is updated, and the name is stylized as "myNode".

## Before

Screenshot from 2023-03-27 07-38-12

## After

Screenshot from 2023-03-27 07-27-56


wiz (approved):
Tested ACK @ v2.6.0-dev [f74d651b]

Issue #3549, opened 2023-03-22
Latest block component still shows wrong latest block


nymkappa:
There is still a bug where the `Latest blocks` component is not showing the latest mined block (delayed by one block). This happen only during navigation so it's probably so rxjs stuff again.

### Step to reproduce

- Open a new page with the mining dashboard
- Confirm the latest block component shows the correct blocks

<img width="759" alt="Screenshot 2023-03-28 at 22 32 16" src="https://user-images.githubusercontent.com/9780671/228254303-6880fbf7-8be5-428c-830f-10cef8683b50.png">

- Wait for a new block
- Click on the main dashboard icon
- Click on the back button of the web browser
- Confirm that there is a missing block in the latest block component

<img width="763" alt="Screenshot 2023-03-28 at 22 33 02" src="https://user-images.githubusercontent.com/9780671/228254335-026667c5-2908-40f6-9d1d-df40d37460d2.png">

nymkappa:
Cannot reproduce anymore, closing

PR #3597, opened 2023-03-28
Fix latest block widget mining missing block sometimes

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3549 (issue includes steps to reproduce)


softsimon:
I believe fixed in https://github.com/mempool/mempool/pull/4014 because I can't reproduce it anymore

PR #3599, opened 2023-03-28
Fix node unfurl row overflow

mononaut:
Fixes a regression in the node preview component which breaks the layout of lightning node unfurls:

| Before | After |
|-|-|
| before | after |


wiz (approved):
Tested ACK @ v2.6.0-dev [ff8fecbd]

PR #3601, opened 2023-03-28
Fix broken tx diagram for non-LBTC liquid assets

### LINKED ISSUE #3441: bowtie is weird on liquid asset transactions
https://liquid.network/tx/27de9e9a8ba91c6c94f8a317f5557b03a2a7c4dac35482fbb1034202691c8d86

Screenshot 2023-03-18 at 18 29 04

mononaut:
Fixes #3441 

Flow diagrams don't really make sense for multi-asset Liquid transactions, and the previous approach was broken for asset issuance/burn transactions in particular, because we didn't differentiate between different assets when allocating values/weights in the diagram.

This PR treats non-LBTC assets on Liquid as having unknown/confidential value for the purposes of the flow diagram (although the correct asset-denominated value is still displayed in the mouseover tooltip). In most cases this means that non-LBTC inputs and outputs will be assigned zero weight and drawn with minimum width lines.

This allows the LBTC-denominated flows to be visualized correctly without breaking the diagram. The tradeoff is that flows of other assets are less clear in transactions where LBTC is minimally involved.

| Before | After |
|-|-|
| Broken diagram for an alt asset issuance tx | |
| Screenshot 2023-03-29 at 7 33 42 AM | Screenshot 2023-03-29 at 7 34 04 AM |
| | Less clear alt asset transfer diagram |
| Screenshot 2023-03-29 at 7 33 11 AM | Screenshot 2023-03-29 at 7 32 26 AM |

In future, perhaps we could implement separate flow diagrams for each asset represented in the transaction, or normalize the weights of different asset values in a single diagram instead.


softsimon (approved):
LGTM

tested ACK @ v2.6.0-dev [e5b1615c]

Issue #3605, opened 2023-03-29
Graph tooltip does not show the correct datetime format when loading the page for the first time


nymkappa:
- Open an incognito window (**make sure the local storage is empty**)
- Open https://node204.fra.mempool.space/graphs/mempool#3y
- The tooltip does not show the proper datetime formating
<img width="2009" alt="image" src="https://user-images.githubusercontent.com/9780671/228416811-e9867a02-243f-423a-9378-314ae7cc9e3c.png">

- Going back and forth between different timespan fixes the issue. So it's an initialization bug
<img width="1997" alt="image" src="https://user-images.githubusercontent.com/9780671/228417035-37c9ebc6-cae2-413b-a355-e2f26bfe75ae.png">

PR #3608, opened 2023-03-29
Use window.location object instead of angular router for default graph window preference setting

nymkappa:
Angular router does not seem to be ready when we run this code (_router.url is `/` instead of `/graphs/mempool#3y`_), so we can just use window.location object directly for setting default window preference timespan in local storage.

Fixes https://github.com/mempool/mempool/issues/3605


softsimon (approved):
tested ACK @ [0bc244b]

PR #3610, opened 2023-03-29
Fix search 1wizS test

nymkappa:
The test expects 5 addresses. @wiz do you have a new address? 😄 

image


wiz:
expand it to wiz says but xbt

softsimon (approved):
utACK. the test explains itself

Issue #3556, opened 2023-03-23
loadMore onScroll not working in all circumstances


softsimon:
When we have 10 of 15 transactions to load.. the load more doesn't trigger
Also with big enough monitor, loadmore isn't triggered.

http://umbrel.local:3006/address/bc1qhzz6jgded2uyk68ae7kuunp9qn0p23qhan82dk

Check onScroll() in `transactions-list.component.ts`

PR #3611, opened 2023-03-29
Fix infinite scroll transaction list component

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3556

I think it was never actually working (and we got around related issues with some hacks) because the infinite scroll [was added inside the transaction list loop](https://github.com/mempool/mempool/commit/90c05ccb51e3b58296e410b234593c9be9b2c43e#diff-be42597e4ce09d3849630e9ce81069487e1b6f6e4fd4a52e15788bd4d60d8bce), basically triggering one event per transaction.


softsimon (approved):
tested ACK @ [aba4989]

PR #3620, opened 2023-04-01
Fix RTL locale unfurls

mononaut:
Unfurls for RTL locales are rendered with a weird viewport offset in production, e.g:

I haven't been able to reproduce the issue locally, but I believe it may be caused by Puppeteer screenshot settings, which default to `captureBeyondViewport = true`.

This PR sets that to false and also explicitly sets the screenshot clipping coordinates, which hopefully resolves the issue.


wiz (changes requested):
Tested NACK @ v2.6.0-dev [2750661e]

Expected: 1200x600
Actual: 1200x1200

00000000000000000000846fad74aece70f6b42f2e821af8379c1b965fb560b5

mononaut:
oops, sorry, should be fixed now

wiz (approved):
Tested ACK @ v2.6.0-dev [d01fa8592]

PR #3632, opened 2023-04-03
WIP: Support spending transactions with Core txospenderindex

Pantamis:
Support spending transactions display with Core txospenderindex bitcoin/bitcoin#24539

This PR also allows the forensic service to get closure method of all channels

Since bitcoin/bitcoin#24539 is not merged, I keep this PR as a draft until it is eventually merged.


softsimon:
After it's merged to Core it would require a new release for it to be accessible though.

But it's a very helpful Core API for us!

PR #3634, opened 2023-04-04
Make fiat component color class customizable

nymkappa:
Should not change current behavior


frontend/src/app/fiat/fiat.component.scss
nymkappa, line 1: This file is not required. The `green-color` class is already defined in https://github.com/mempool/mempool/blob/master/frontend/src/styles.scss#L291

wiz (approved):
Tested ACK @ v2.6.0-dev [90fa4a8f]

PR #3636, opened 2023-04-04
Add note about enterprise support to readmes

hunicus:
Added notes on the main README as well as Docker, manual, and production install READMEs.

All the notes go something like this:

> **We only provide support for advanced installation methods to <a href="https://mempool.space/enterprise">Enterprise sponsors</a>.**


softsimon:
@hunicus is this still relevant. It has conflicts.

hunicus:
Already done with #4143.

PR #3643, opened 2023-04-05
[esplora] fallback to tcp socket if unix socket fails

nymkappa:


wiz (approved):
Tested ACK @ v2.6.0-dev [c23e529f]

would appreciate if @softsimon can review c23e529f since it touches some old retry code

PR #3644, opened 2023-04-05
Perform full cpfp calculations for the entire mempool

mononaut:
For performance reasons, the full advanced GBT algorithm is currently applied to produce only the first 8 projected blocks.

Any remaining transactions are processed without fully updating ancestor scores and dependency graphs. This is sufficient in most cases, but means some transactions with more complex CPFP relationships can end up with inaccurate CPFP data and incorrect effective fee rates.

This PR updates the `tx-selection-worker` to apply the full advanced GBT algorithm to the entire mempool, to ensure that CPFP data and effective fee rates are correctly calculated for every transaction.

Testing in current mempool conditions (~500MB / ~40k transactions), the performance impact actually doesn't seem too significant.

### DIFF
--- a/backend/src/api/tx-selection-worker.ts
+++ b/backend/src/api/tx-selection-worker.ts
@@ -107,7 +107,7 @@ function makeBlockTemplates(mempool: { [txid: string]: ThreadTransaction })
 
     if (nextTx && !nextTx?.used) {
       // Check if the package fits into this block
-      if (blockWeight + nextTx.ancestorWeight < config.MEMPOOL.BLOCK_WEIGHT_UNITS) {
+      if (blocks.length >= 7 || (blockWeight + nextTx.ancestorWeight < config.MEMPOOL.BLOCK_WEIGHT_UNITS)) {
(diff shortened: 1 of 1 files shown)


backend/src/api/tx-selection-worker.ts
```diff
@@ -175,34 +175,14 @@ function makeBlockTemplates(mempool: { [txid: string]: ThreadTransaction })
...
-    if (tx.ancestorMap.size > 0) {
-      cpfpClusters[tx.txid] = Array.from(tx.ancestorMap?.values()).map(a => a.txid);
-      mempoolTx.cpfpRoot = tx.txid;
-    }
-    mempoolTx.cpfpChecked = true;
-    transactions.push(tx);
-    tx.used = true;
+
+  if (overflow.length > 0) {
+    console.log('GBT overflow list unexpectedly non-empty after final block constructed');
```
softsimon, line 180: use logger for debug outputs

softsimon (approved):
tested ACK @ v2.6.0-dev [5950034f5]

Seems to work!

PR #3651, opened 2023-04-06
Fix block animations while syncing

mononaut:
This PR fixes some issues with the blockchain animations while the backend is syncing (e.g. after a fresh install, or being offline for a time), and also during bitcoin core initial block download.

### IBD

Before:

https://user-images.githubusercontent.com/83316221/230461099-9880403d-5646-45b4-bf39-100eb35d1fa1.mp4
 
After:

https://user-images.githubusercontent.com/83316221/230461120-c619a430-be24-48c5-9e0d-75d4c98a3dd4.mp4

### Backend syncing

Before:

https://user-images.githubusercontent.com/83316221/230458704-fa5616a7-9688-4e1c-a737-7609d545e803.mp4

After:

https://user-images.githubusercontent.com/83316221/230458729-91ac7c5e-aee4-4892-92a1-c4152c6a92c0.mp4

### DIFF
--- a/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.ts
+++ b/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.ts
@@ -22,6 +22,7 @@ export class BlockchainBlocksComponent implements OnInit, OnChanges, OnDestroy {
   @Input() offset: number = 0;
   @Input() height: number = 0; // max height of blocks in chunk (dynamic blocks only)
   @Input() count: number = 8; // number of blocks in this chunk (dynamic blocks only)
+  @Input() dynamicBlockCount: number = 8; // number of blocks in the dynamic block chunk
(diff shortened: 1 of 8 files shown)


softsimon:
A question regarding the "Backend syncing" part. It seems the blocks will now flow from left to right even though the mempool is not matching with null as match rate? Before we just popped them up to the right side and only animate over when there is a match in the mempool and mined blocks.

Here I started the backend after being offline for a few hours and blocks were animating, despite the mempool basically not changing.

Screenshot 2023-04-08 at 19 40 44

Screenshot 2023-04-08 at 19 40 47

mononaut:
Good point, we shouldn't animate the latest mined block in that case. I'll fix that.

softsimon (changes requested):
Just changing state to changes requested.

wiz:
@knorrium isn't this old PR fixing your thing

PR #3652, opened 2023-04-06
Disable mempool block animations except when new block is mined

mononaut:
Fixes #3648 

Before:

https://user-images.githubusercontent.com/83316221/230461831-7b8ef8e4-532a-4bc5-8f01-f58c2374956e.mp4

After:

https://user-images.githubusercontent.com/83316221/230461869-cd090a34-14a1-4ad4-b0c5-2432cce9effe.mp4

PR #3653, opened 2023-04-06
Fix mempool block skeleton loaders

### LINKED ISSUE #3646: Show 8 skeleton loader mempool blocks
Since the mempool is and will be full usually. Always use 8 mempool blocks in the skeleton loader.
Perhaps handle/hide, reduce to fewer blocks if screen is smaller.

bild

mononaut:
_(builds on #3652)_

Fixes #3646

Previously, we truncated the mempool block skeleton loaders to fit the window when the page is first loaded, and never updated the number of blocks shown afterwards.

This PR ensures the right number of skeleton loader blocks are always shown (up to a maximum of 8) depending on the window size.

To reproduce the issue for testing:
 - refresh the dashboard with a reduced window width.
 - increase the width of the window.
 - disconnect the backend to trigger the skeleton loader.
 - observe that fewer skeleton loader blocks are displayed than expected.
 
Before:

https://user-images.githubusercontent.com/83316221/230464892-2842b8f1-b398-4ed7-8d4c-b942887fb221.mp4

After:

https://user-images.githubusercontent.com/83316221/230464923-fae1489d-060c-412c-911e-25bda5bd86a9.mp4


softsimon (approved):
tested ACK @ [585309f9]

Issue #400, opened 2021-03-19
Occational zero matchRate result in missing new block animation


softsimon:
Most probably, a race condition with the esplora electrs engine causes the mempool to be cleared of transactions of a new block before the new block has been detected by the mempool backend, resulting in matchRate = 0, no animation in the frontend and other possible side effects.

PR #3655, opened 2023-04-07
Fix new block mempool eviction race condition

### LINKED ISSUE #3650: Block audit not handling multiple blocks arriving shortly after one another
These two blocks arrived with just 3 (?) seconds in between each other.

This caused the second block to get a 0% health score. Perhaps we don't, and need to re-run the mempool template re-calculation after each block.

https://mempool.space/block/000000000000000000002f7ce47a66916b0abfe3dec8bd489d98ef161df1d3bc
https://mempool.space/block/00000000000000000001c20c14a52b5ae7c6ca385c58943c82ce8b0a76893b68

mononaut:
This PR attempts to fix #3650, and another more general case that can cause incorrect templates and 0% audit scores due prematurely marking transactions as deleted.

---

### Problem

I believe the underlying cause is a new block arriving after we started updating blocks, but before we update the mempool in the main update loop:
https://github.com/mempool/mempool/blob/d18ebdfc5902b47a0afd869c59e3c8d92d3c90fd/backend/src/index.ts#L183-L184

When this happens, `mempool.$updateMempool` gets the new state of the mempool *after* transactions have been removed for inclusion in the new block, marks those transactions as deleted, and applies the change to the projected mempool templates.

On the next run of the main update loop, `blocks.$updateBlocks` discovers the new block and runs the audit against the post-block mempool state.

This is essentially the same bug as #400, which was addressed by adding a 30 second delay before deleting transactions from the mempool cache. Unfortunately that solution doesn't play nicely with mempool projections and audits, which need more immediate consistency.

---

### Solution

Instead of mitigating the effects of the race condition, this PR aims to eliminate it by changing the sequence of the main update loop as follows:

https://github.com/mempool/mempool/blob/3a6205ad00c515b6ca584f35d67e0bb7295e91ab/backend/src/index.ts#L183-L191

1. Fetch the current state of the mempool.
2. Check & process new blocks.
3. If there were no new blocks, update the mempool using that state.
   - since we found no new blocks after fetching the mempool, we know that this mempool update doesn't remove any transactions due to block inclusion.
4. Otherwise skip the mempool update and run the loop again asap.
   - the new blocks could have arrived between steps 1 and 2, so skipping the mempool update here ensures don't try to apply a stale mempool state.

The PR also removes the now-obsolete lazy deletion code.

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -529,13 +529,14 @@ class Blocks {
     return await BlocksRepository.$validateChain();
   }
 
-  public async $updateBlocks() {
+  public async $updateBlocks(): Promise<number> {
     // warn if this run stalls the main loop for more than 2 minutes
     const timer = this.startTimer();
 
     diskCache.lock();
 
     let fastForwarded = false;
+    let handledBlocks = 0;
     const blockHeightTip = await bitcoinApi.$getBlockHeightTip();
     this.updateTimerProgress(timer, 'got block height tip');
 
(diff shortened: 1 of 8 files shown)


softsimon (approved):
tested ACK @ [3d0f7d68]

Been running over night and more and it seems to fix the issue and nothing breaks

PR #3670, opened 2023-04-20
Push TX: Include validation to prevent DoS

junderw:
Some simple checks before sending the tx to the esplora/electrs/bitcoind.

1. Make sure it can be parsed by bitcoinjs-lib as a transaction.
2. Make sure it fits under a weight limit (bitcoind is 400000 by default, so mempool's default will mimic that.
3. Make sure it doesn't contain unreachable script paths. There is no such policy in bitcoind, so by default we should allow it.

All except 1 can be modified by config.

This PR should not change the set of accepted transactions with default config. If it does, then it is a bug.


mononaut (approved):
Tested ACK @ v2.6.0-dev [[a8098378](https://github.com/mempool/mempool/commit/a8098378)]

I confirmed the validation handles the following test cases correctly with nice error messages:
 - Normal tx rejected above default weight limit
 - Normal tx accepted below custom weight limit
 - Normal tx rejected above custom weight limit
 - Normal tx accepted below custom weight limit
 - Unreachable tapscript tx rejected with `ALLOW_UNREACHABLE=true`
 - Unreachable tapscript tx accepted with default config
 - Unreachable tapscript tx rejected above weight limit
 
 Note that this PR won't do anything in production yet, since the `/api/tx` POST route doesn't exist in `config.MEMPOOL.BACKEND='esplora'` mode, and requests are instead routed directly to esplora bypassing this validation.

backend/src/api/common.ts
```diff
@@ -440,4 +442,87 @@ export class Common {
...
+      throw Object.assign(new Error(`Transaction too large (max ${config.MEMPOOL.MAX_PUSH_TX_SIZE_WEIGHT} weight units)`), { code: -3 });
+    }
+
+    // Check 3: Check unreachable script in taproot (if not allowed)
+    if (!config.MEMPOOL.ALLOW_UNREACHABLE) {
+      tx.ins.forEach(input => {
+        const witness = input.witness;
+        // See BIP 341: Script validation rules
+        const hasAnnex = witness.length >= 2 &&
+          witness[witness.length - 1].length > 1 &&
```
antonilol, line 492: annex can be 1 byte long, although at the time of writing taproot inputs with an annex won't be relayed, this can be used to get past this check
this check is not necessary for the check here after because `witness[witness.length - 1][0]` will be `undefined`, it won't throw an error

backend/src/api/common.ts
```diff
@@ -440,4 +442,87 @@ export class Common {
...
+          witness[witness.length - 1][0] === 0x50;
+        const scriptSpendMinLength = hasAnnex ? 3 : 2;
+        const maybeScriptSpend = witness.length >= scriptSpendMinLength;
+
+        if (maybeScriptSpend) {
+          const controlBlock = witness[witness.length - scriptSpendMinLength + 1];
```
antonilol, line 498: technically correct, but would be more readable as:
```suggestion
          const controlBlock = witness[witness.length - (hasAnnex ? 2 : 1)];
```
hasAnnex ? second to last element : last element

backend/src/api/common.ts
```diff
@@ -440,4 +442,87 @@ export class Common {
...
+          if (controlBlock.length === 0 || (controlBlock[0] & 0xfe) < 0xc0) {
```
antonilol, line 499: why `< 0xc0`?

wiz:
@junderw please rebase onto current master

backend/src/api/common.ts
junderw: I disagree. "One after the min length of script spends" (ie. the script, + 1 index === the control block) looks more readable to me.

backend/src/api/common.ts
junderw: I was unaware of the extra 9 possible leaf versions... though I highly doubt the lower ones would ever be used.

I pushed a fix to be more correct.

backend/src/api/common.ts
```diff
@@ -440,4 +442,87 @@ export class Common {
...
+    } catch(e) {
+      throw Object.assign(new Error('Invalid transaction (could not parse)'), { code: -4 });
+    }
+
+    // Check 2: Simple size check
+    if (tx.weight() > config.MEMPOOL.MAX_PUSH_TX_SIZE_WEIGHT) {
+      throw Object.assign(new Error(`Transaction too large (max ${config.MEMPOOL.MAX_PUSH_TX_SIZE_WEIGHT} weight units)`), { code: -3 });
+    }
+
+    // Check 3: Check unreachable script in taproot (if not allowed)
```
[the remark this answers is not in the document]
junderw: > Note that this PR won't do anything in production yet, since the /api/tx POST route doesn't exist in config.MEMPOOL.BACKEND='esplora' mode, and requests are instead routed directly to esplora bypassing this validation.

For now I think we shouldn't add this comment to documentation yet, since prod will not have this behavior yet.

Thanks for the review.

junderw:
1. Rebased
2. Fixed comments
3. Fixed missing config templates (some of the docker templates were missing placeholders for the new config items (testing for missing configs was a good idea, awesome!))

The actual code changes shouldn't affect the outcome of the code, but I wonder what we should do for Esplora.

Should I try and mimic this logic on the esplora end as well? Or should we route esplora requests through the TS first?

PR #3672, opened 2023-04-20
Update unchained icon on about page

hunicus:
Screenshot from 2023-04-19 22-53-47


softsimon (approved):
tested ACK @ [008ec10]

Issue #3566, opened 2023-03-24
Mempool block sort bug


softsimon:
I click on a TX in block 3, but it shows tracking in block 2.

Inspecting the Websocket response, it looks like the order of the feeRanges is not correct.

https://user-images.githubusercontent.com/8561090/227566499-c2aeec81-0e9a-4151-b981-d0c277532357.mov

<img width="882" alt="Screenshot 2023-03-25 at 00 12 52" src="https://user-images.githubusercontent.com/8561090/227566548-94684595-1305-4518-9a84-83e2b4aad588.png">

mononaut:
I think those fee ranges are actually accurate, and the problem is that it's not always possible to know which block a transaction belongs to from its fee rate alone, because transaction selection also depends on size (and to some extent randomness).

Consider an extreme example:

> Imagine the mempool consists only of many thousands of huge 80,000vb transactions paying exactly 50 sats/vb, plus thousands of small 100vb transactions each paying 1 sat/vb.
> 
> Each projected block can fit 12 of the large, high-rate transactions, leaving `1,000,000 - (12 x 80,000) = 40,000 vbytes` free.
> 
> That free space can only be filled by 400 of the small 100vb, 1 s/vb transactions.
> 
> So the first projected block's feeRange looks like [1, 1, 1, 50]. But so does the second, third, fourth block etc, and so their feeRanges all overlap.
>
> That means if we're given only the fee rate of a random transaction, there's no way to tell which block it "belongs" to.

I'm not sure what the best solution is.

We do have the actual positions of transactions in each projected block available in the backend, so perhaps we should just send that info over the websocket when you subscribe to an unconfirmed transaction?

softsimon:
Ok. So that is also the cause of this thing when the fee range doesn't look sorted by fee and you get a sudden high fee data point like here?

<img width="511" alt="Screenshot 2023-03-15 at 11 25 49" src="https://user-images.githubusercontent.com/8561090/229266350-78fe7888-2add-46c8-a1a2-6856786ebecf.png">

nymkappa:
> We do have the actual positions of transactions in each projected block available in the backend, so perhaps we should just send that info over the websocket when you subscribe to an unconfirmed transaction?

Never really worked in that code but I would have thought it was already the case. Are you saying the backend and the frontend have different code to compute the position in the mempool?

mononaut:
[...]

PR #3673, opened 2023-04-21
Improve stability of the mempool transaction marker arrow

mononaut:
_(builds on PR #3714)_

_resolves #3566_

This PR improves the accuracy and stability of the position of the marker arrow when a mempool transaction is selected.

Previously, we estimated the mempool position based on the effective fee rate of the transaction and the reported fee bands of the mempool blocks. This could produce ambiguous and unstable results when mempool blocks had overlapping fee bands (see https://github.com/mempool/mempool/issues/3566#issuecomment-1493050077 for more discussion).

The PR makes two main changes:
 - stores the actual block index and position within the block for each mempool transaction, sends it to any client subscribed to websocket updates for that tx, and uses that to place the marker arrow.
   - this position is now based on cumulative vsize, so it should also more closely match positions in the mempool block visualizations (or at least it will after #3380 is merged).
 - adds a stable 'tie-breaker' to the GBT algorithms, so that transactions with identical effective fee rates are always sorted into the same order.

If the backend calculated position is unavailable for some reason (e.g. because the websocket is disconnected or we haven't received an event yet), marker positioning falls back to the old fee-rate-based estimate.

### DIFF
--- a/backend/src/api/mempool-blocks.ts
+++ b/backend/src/api/mempool-blocks.ts
@@ -252,9 +252,17 @@ class MempoolBlocks {
 
   private processBlockTemplates(mempool, blocks, clusters, saveResults): MempoolBlockWithTransactions[] {
     // update this thread's mempool with the results
-    blocks.forEach(block => {
+    blocks.forEach((block, blockIndex) => {
+      let runningVsize = 0;
       block.forEach(tx => {
         if (tx.txid && tx.txid in mempool) {
+          // save position in projected blocks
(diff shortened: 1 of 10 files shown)


mononaut:
rebased on #3714

softsimon (changes requested):
Something seems not working here right now. On Master it's fine, but with the PR it never finds the arrow position.

image

softsimon (approved):
tested ACK @ [4894706]

PR #3689, opened 2023-04-27
detect and log stalls in the main loop

mononaut:
~~This PR adds a temporary extra layer of explicit timeout to the esplora api, to help debug possible failures relating to unix socket requests.~~

~~If an esplora request has not returned, failed, or triggered an internal timeout within 30 seconds, the request throws an error.~~

The PR ~~also~~ implements timed logging functions in the `$updateBlocks` and `$updateMempool` functions to detect stalls in the main loop for any other reason, and log an error indicating where in the loop the stall occurred.


softsimon (approved):
tested ACK @ [58b08f2]

Lowered the timeout to get various mempool and block debug outputs so it works, and it's safe because the only thing it does is logging.

```
May  1 00:15:39 [38152] ERR: $updateMempool stalled at "running async mempool callback"
```

PR #3693, opened 2023-04-27
prevent table overflow in unfurl previews

mononaut:
Table-based unfurl previews can overflow the intended 1200x600 bounds in some locales due to long translation strings.

This PR limits the layout width of tables in those preview components to prevent overflow.

e.g:

| Before | After |
|-|-|
| Screenshot 2023-04-26 at 4 36 33 AM | Screenshot 2023-04-26 at 4 35 12 AM |
| Screenshot 2023-04-26 at 4 38 03 AM | Screenshot 2023-04-26 at 4 41 21 AM |

Affected previews are:
 - `/preview/address/...` (overflows in korean)
 - `/preview/block/...` (can overflow in georgian)
 - `/preview/lightning/channel/...` (georgian)
 - `/preview/lightning/node/...` (georgian)

This PR is an improvement, but not a complete fix - in some cases the content of the tables is still simply too wide, causing the right hand column to be truncated or hidden. Long-term solutions might be to reduce the length of the translated strings, or reduce font size when necessary.


wiz (approved):
Tested ACK @ v2.6.0-dev [93b1c6448]

PR #3696, opened 2023-04-29
Add explicit timeout to mysql DB queries

mononaut:
When the backend is configured to connect to the database via unix socket, it appears to be possible for database query requests to never resolve or reject, stalling the calling context indefinitely.

This PR adds an extra timeout wrapper to the `DB.query()` method to ensure it always eventually settles.

The timeout is configurable via a new config parameter `config.DATABASE.TIMEOUT` (in milliseconds), defaulting to 180000ms (3 minutes).

If `DB.query()` is called with an explicit `timeout` parameter, that value takes precedence. For example the database migration queries still run with a one hour timeout.

If the timeout is set to zero, the new wrapper is not used.

The wrapper can't cancel or rollback a timed-out DB query, so it's important to set the new timeout config to a high value that won't be triggered during normal operation.


softsimon (approved):
tested ACK @ [e05f2198]

With very low timeout limit it rejects properly and continue the main loop.

```
Apr 30 22:01:58 [28178] DEBUG: Updating mempool...
Apr 30 22:01:59 [28178] DEBUG: Mempool updated in 0.691 seconds. New size: 178527 (0)
Apr 30 22:01:59 [28178] ERR: Cannot generate pools stats. Reason: DB query failed to return, reject or time out within 3s - 
      SELECT
        COUNT(blocks.height) As blockCount,
          pool_id AS poolId,
          pools.name AS name,
          pools.link AS link,
          slu
Apr 30 22:02:00 [28178] DEBUG: Running statistics
Apr 30 22:02:01 [28178] DEBUG: Updating mempool...
Apr 30 22:02:02 [28178] DEBUG: Mempool updated in 1.427 seconds. New size: 178544 (+17)
Apr 30 22:02:02 [28178] DEBUG: Mempool templates calculated in 0.999 seconds
```

PR #3700, opened 2023-04-30
await for mempool change handler after loading disk cache

mononaut:
Fixes a bug caused by failing to `await` the asynchronous `handleMempoolChange` callback after restoring the mempool from disk cache.

If the timing is right, the callback might not resolve by the time the main loop starts, resulting in two separate `updateMempoolBlocks` calls requesting templates from the thread worker at the same time.

The first thread worker response would satisfy *both* waiting tasks, leaving a spare result in the worker message queue. Every subsequent run of `updateMempoolBlocks` would then receive the thread worker results of the *previous* task.

This PR adds the missing `await`s, to ensure that thread worker tasks are always invoked sequentially.

It may also be worth adding some sort of mutex lock and/or message sequence numbers to the thread worker to prevent similar issues in future.

### DIFF
--- a/backend/src/api/disk-cache.ts
+++ b/backend/src/api/disk-cache.ts
@@ -124,7 +124,7 @@ class DiskCache {
     }
   }
 
-  async loadMempoolCache(): Promise<void> {
+  async $loadMempoolCache(): Promise<void> {
     if (!fs.existsSync(DiskCache.FILE_NAME)) {
       return;
     }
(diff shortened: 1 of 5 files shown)


softsimon (approved):
tested ACK @ d72bee25

No more out of sync errors `WARN: projected transaction ancestor missing from mempool cache `

softsimon (approved):
tested ACK @ [4597bfa5]

mononaut:
rebased

PR #3703, opened 2023-05-01
delay writing disk cache until block handler completes

mononaut:
This PR adds a semaphore-style lock to the disk cache task to prevent it from writing cache files while new blocks are being processed.

If a new block arrives while the `$saveCacheToDisk(true)` is running, it will pause and wait for the block handler to complete before starting the next `writeFile` operation.


mononaut:
rebased on current master [e807b3c]

softsimon (approved):
tested ACK @ [e5f97ace]

PR #3705, opened 2023-05-02
Increase client websocket timeout

mononaut:
This PR boosts the frontend websocket timeout from 10 seconds to 30 seconds.

It also increases the minimum delay between reconnection attempts from 1 second to 2 seconds, and adds up to 2 seconds of random extra delay to spread the impact of mass reconnections over a longer period of time.


softsimon (approved):
tested ACK @ [3691ba82]

PR #3707, opened 2023-05-02
Removing dead code causing slowdown

softsimon:


mononaut (approved):
Tested ACK @ [2fbe2b2f]

Big speed up, and audits still work.

PR #3709, opened 2023-05-03
skip unnecessary makeBlockTemplates call

mononaut:
This PR significantly speeds up `handleNewBlock` by skipping the expensive pre-audit run of `makeBlockTemplates` when using the same GBT algorithm for both mempool updates and block audits.

This is possible because in that case the projected templates from the last run should still  be valid (since our internal copy of the mempool shouldn't have changed in that time).

---

The PR then replaces the later call to `updateBlockTemplates` in that function with `makeBlockTemplates`, so that the thread worker state is still reset every block.

This might not be strictly necessary, and adds a small amount of extra thread I/O, but it's a nice sanity check in case the thread worker ever drifts out of sync for any reason.

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -439,13 +439,19 @@ class WebsocketHandler {
 
     if (config.MEMPOOL.AUDIT) {
       let projectedBlocks;
+      let auditMempool = _memPool;
       // template calculation functions have mempool side effects, so calculate audits using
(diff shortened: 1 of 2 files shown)


softsimon:
Do you think this might affect the case when 2 blocks are found at the same time? I mean it might not run the `$updateBlockTemplates` between the 2 blocks, or is that handled?

softsimon (approved):
tested ACK @ [03ee5c7c]

PR #3710, opened 2023-05-03
Mononaut/optimize gbt data

mononaut:
_(builds on PR #3709)_

This PR optimizes some of the data structures used in the advanced GBT implementation with the aim of speeding up mempool updates.

The main change is switching from txid strings to short numerical ids to identify transactions in the GBT thread worker.

ids are assigned sequentially, and tracked by a Map in the `MempoolBlocks` module. Some helper functions convert between numeric ids and txids when communicating with the worker thread.

This significantly reduces the amount of data copied back and forth from the thread worker, since now we only need a single javascript number to identify each transaction instead of a 128-byte txid string, and the majority of the GBT data consists of transaction identifiers.

In order to work with numerical ids, a few structures in the GBT implementation have been converted from simple Objects to ES6 Maps. Setting and accessing Map elements by numeric key seems to be more efficient than the equivalent operations with txid-keyed Objects, and the GBT algorithm runs noticeably faster.

The whole concept is a bit gross, but it does speed up mempool updates by 20-40% on my machine in current conditions, and significantly reduces the time spent on blocking cross-thread communication.


mononaut:
rebased on current master (07d9315)

softsimon (approved):
tested ACK @ [428d4fc6]

Issue #3695, opened 2023-04-28
Transaction ETA shows the wrong value most of the time


nymkappa:
Related:
- https://github.com/mempool/mempool/issues/2871
- https://github.com/mempool/mempool/issues/65

Transaction ETA is completely broken. Consider fixing it or removing it. Here is some examples while randomly browsing pending tx on prod, very easy to reproduce, just click on random pending txs.

### An example of a correct value

<img width="708" alt="Screenshot 2023-04-28 at 17 12 56" src="https://user-images.githubusercontent.com/9780671/235186765-b9d02685-31e8-4476-be0e-0b988c2dc9eb.png">

### Some examples of incorrect value

<img width="750" alt="image" src="https://user-images.githubusercontent.com/9780671/235187485-c3bd88b0-c3e8-4a61-ae80-4ab9ba4ee70e.png">
<img width="727" alt="Screenshot 2023-04-28 at 17 13 08" src="https://user-images.githubusercontent.com/9780671/235186760-b0404a92-bcbb-4352-8eb6-0f360ae2cf1f.png">
<img width="684" alt="Screenshot 2023-04-28 at 17 12 42" src="https://user-images.githubusercontent.com/9780671/235186772-3cf262bb-d095-48bb-9ff7-e4c51f29ba99.png">

### Monitoring a single pending tx

<img width="764" alt="Screenshot 2023-04-28 at 17 15 05" src="https://user-images.githubusercontent.com/9780671/235186718-99bdda2c-5b27-4b05-97d9-a39c2569fe9a.png">
<img width="748" alt="image" src="https://user-images.githubusercontent.com/9780671/235187783-8f2cce77-0ce4-4be4-9c04-0e1312e36476.png">
<img width="743" alt="Screenshot 2023-04-28 at 17 21 22" src="https://user-images.githubusercontent.com/9780671/235188277-0d935b3d-4fda-4bff-8389-a425cfb32f8a.png">
<img width="742" alt="image" src="https://user-images.githubusercontent.com/9780671/235188320-788df08c-2533-431b-8f6d-91c981f4172b.png">

PR #3714, opened 2023-05-03
Fix transaction ETA calculation

mononaut:
_resolves #3695 and resolves #2871_

The current ETA calculation chooses the *last* projected block in which the fee span includes the transaction's effective rate.

Because of the issues discussed in #3566, when the mempool is extremely large that means it almost always chooses the final "overflow" mempool block (since the very high number of transactions ensures the fee span is very wide):

Screenshot 2023-05-03 at 10 08 13 AM

This PR switches the ETA calculation to choose the *first* matching projected block, which should be correct for the majority of transactions.

However there are still a few edge cases from #3566 where this will underestimate the ETA, so we still need PR #3673 to completely fix the calculation.

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.ts
+++ b/frontend/src/app/components/transaction/transaction.component.ts
@@ -59,7 +59,7 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
   fetchRbfHistory$ = new Subject<string>();
   fetchCachedTx$ = new Subject<string>();
   isCached: boolean = false;
-  now = new Date().getTime();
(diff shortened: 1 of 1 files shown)


mononaut:
Good catch.

Apparently we never updated the `now` variable to which we add the estimated delay, so when the block index changed the ETA would slip backwards (presumably the block got mined about 8 minutes after you loaded the tx page?).

I've just pushed a fix for that, and also backported a fix from #3673 for a subscription leak which could also cause some issues.

softsimon (approved):
tested ACK @ [9bd968f6]

PR #3716, opened 2023-05-03
display fee box skeleton while mempool not in sync

mononaut:
Displays the fee box skeleton loader when either the websocket is disconnected or the mempool is not in sync, to avoid showing bad fee estimates.

Before:
Screenshot 2023-05-03 at 2 45 51 PM

After:
Screenshot 2023-05-03 at 2 46 15 PM


softsimon (approved):
tested ACK @ [65189d6f]

PR #3719, opened 2023-05-05
fix rbf cache eviction logic

mononaut:
This PR fixes a bug in the RBF cache eviction logic which caused mined RBF transactions not to be removed from the in-memory RBF cache (although they would eventually be removed after restarting the backend).

The PR also fixes a separate issue where unnecessary data was being tracked.

These two bugs combined to make the RBF cache file much larger than it needed to be.

Finally, the PR increases the frequency at which the RBF cache is cleared of expired transactions, and periodically logs the size of the cache.


softsimon (approved):
tested ACK @ [f4569126]

PR #3722, opened 2023-05-05
Filter GBT input

mononaut:
### Summary

This draft PR implements a possible optimization to reduce the amount of work for the advanced GBT algorithm.

Generally, the algorithm needs to consider the entire mempool in order to accurately assemble CPFP packages, calculate effective fee rates and prioritize transactions.

However, transactions with no in-mempool relatives are guaranteed to not be part of any CPFP package, or to affect other CPFP packages, so we can safely filter these transactions out below a certain fee rate threshold without affecting the resulting GBT templates.

### Implementation

This PR implements that filtering, including tracking and accommodating changes to the set of filtered transactions as the threshold changes. After templates are calculated from the filtered set, any omitted transactions are reinserted in the last block.

The threshold is currently set as `max(purging rate, avg effective fee rate of projected blocks 8-24)`. More investigation is needed to see how aggressively the threshold can be raised without excluding any transactions which ought to be in the first 8 projected blocks.

### Results

At this threshold, in current mempool conditions, applying the filter reduces the total number of transactions to consider for the GBT algorithm from ~277k to ~213k, which results in about a **20% overall speed up** for the `$updateMempool` function on average (on my machine).

Using a more conservative threshold of just the current default mempool purging rate, the filter reduces the number transactions to consider to ~250k, for an average **10% speed up**.

In theory, I think the impact should be larger as the mempool grows, since the proportion of transactions falling below the threshold will also increase.

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -529,13 +529,14 @@ class Blocks {
     return await BlocksRepository.$validateChain();
   }
 
-  public async $updateBlocks() {
+  public async $updateBlocks(): Promise<number> {
     // warn if this run stalls the main loop for more than 2 minutes
     const timer = this.startTimer();
 
     diskCache.lock();
 
     let fastForwarded = false;
+    let handledBlocks = 0;
     const blockHeightTip = await bitcoinApi.$getBlockHeightTip();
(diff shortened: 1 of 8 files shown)


softsimon:
Closing because we moved on to Rust GBT

PR #3723, opened 2023-05-06
Added the total number of inputs and outputs in tx details.

bholm:
I often find myself wanting to see at a glance the total number of inputs or outputs to a transaction but I'm currently forced to either manually count them (impractical for large transactions) or just copy the txid to another block explorer that does. I just made a front end change that adds an extra row to the Details table of the transaction component with the total inputs and outputs for that transaction.

### DIFF
--- a/contributors/bholm.txt
+++ b/contributors/bholm.txt
@@ -1,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of May 5, 2023.
+
+Signed: bholm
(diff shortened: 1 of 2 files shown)


softsimon:
Can you provide a screen shot in the PR?

bholm:
@softsimon sure, here are some screenshots of how it looks for Desktop and Mobile for the example transaction of `16f11bc3d4eeed5759f4fa40ac601e2b403b8eba881f21368b8e51885b280700`.
Screenshot-Mempool-Desktop
Screenshot-Mempool-Mobile

antonilol:
why not display "3 Inputs & 7 Outputs" instead of "Inputs & Outputs"? looks cleaner to me

softsimon:
We need to figure out a better place to put it. I don't think the sum of the input/outputs belong at the details section that currently only contain transaction properties.

softsimon:
Closing due to inactivity

PR #3727, opened 2023-05-07
Rapid mempool sync

softsimon:
Instead of pausing all the time to calculate block templates and other, we now let the mempool sync run until finished, which will speed up first runs and restarts a lot. Only downside is that we might not see something visible on the mempool blocks growing, other than the progress indicator moving up.

```
May  7 18:47:00 [85265] NOTICE: Mempool Server is running on port 8999
May  7 18:47:00 [85265] DEBUG: Updating mempool...
May  7 18:47:14 [85265] DEBUG: Mempool is synchronizing. Processed 10227/16555 txs (98%)
May  7 18:47:18 [85265] DEBUG: Mempool is synchronizing. Processed 13284/16555 txs (99%)
May  7 18:47:25 [85265] DEBUG: Mempool templates calculated in 2.473 seconds
May  7 18:47:29 [85265] DEBUG: Mempool updated in 28.971 seconds. New size: 407203 (+16555)
```

### DIFF
--- a/backend/src/api/mempool.ts
+++ b/backend/src/api/mempool.ts
@@ -11,7 +11,6 @@ import bitcoinSecondClient from './bitcoin/bitcoin-second-client';
 import rbfCache from './rbf-cache';
 
 class Mempool {
-  private static WEBSOCKET_REFRESH_RATE_MS = 10000;
   private inSync: boolean = false;
   private mempoolCacheDelta: number = -1;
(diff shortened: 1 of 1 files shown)


mononaut (approved):
Tested ACK @ v2.6.0-dev [[52fe59ff](https://github.com/mempool/mempool/commit/52fe59ff)]

noticeably reduces cold start-up time, and avoids showing an incomplete mempool while syncing.

the progress updates are nicer too.

PR #3733, opened 2023-05-08
Fee widget gradient

mononaut:
This PR fixes the colors & gradient in the recommended fee widget.

Before, we used a single linear gradient from the `minimumFee` color to the `fastestFee` color (even though the gradient runs from "Low Priority" to "High Priority".

After, we start from the correct color for "Low Priority", and add a halfway stop in the gradient at the "Medium Priority" color. 

It also correctly uses the `economyFee` color for the "No Priority" section (instead of `minimumFee`)

| Before | After |
|-|-|
| Screenshot 2023-05-08 at 1 06 38 PM | Screenshot 2023-05-08 at 1 06 24 PM |
| Screenshot 2023-05-08 at 1 00 47 PM | Screenshot 2023-05-08 at 1 00 56 PM |

### DIFF
--- a/frontend/src/app/app.constants.ts
+++ b/frontend/src/app/app.constants.ts
@@ -29,6 +29,14 @@ export const mempoolFeeColors = [
   'ba3243',
   'b92b48',
   'b9254b',
+  'b8214d',
+  'b71d4f',
+  'b61951',
+  'b41453',
+  'b30e55',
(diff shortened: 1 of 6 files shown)


wiz (changes requested):
Tested Concept NACK - sorry but simon and I prefer the current behavior

PR #3735, opened 2023-05-09
Optimize main thread processing of GBT updates

mononaut:
### Summary

This PR refactors the data structures passed back from the GBT worker thread (again) to significantly reduce the total amount of data and the amount of work required to process the results in the main thread.

The main thread processing is then further optimized to eliminate the most expensive operations identified via CPU profiling.

### Results

The result is a large performance increase in the `$updateBlockTemplates` function, and a more modest gain in `$makeBlockTemplates`.

On my machine in current mempool conditions (~430k transactions) the total duration of main thread work in `$updateBlockTemplates` is reduced from ~2 seconds to ~400ms.

### Other changes

One of the expensive operations removed by this PR is calculating the fee statistics for the entire rest-of-the-mempool. This means that the last mempool block now shows the fee range and fee distribution graph for the block itself, instead of for the whole stack of N blocks.

It should be possible to calculate statistics for the full stack of blocks more efficiently, but I haven't spent the time to implement this yet.

The PR also adds new debug statements to log the total time taken by the `$updateBlockTemplates` and `$makeBlockTemplates` functions.


wiz (approved):
Tested ACK @ v2.6.0-dev [033e78c0]

Would really prefer to have @softsimon review this before merging but it seems like we need it ASAP so will merge for now

mononaut:
yes

> One of the expensive operations removed by this PR is calculating the fee statistics for the entire rest-of-the-mempool. This means that the last mempool block now shows the fee range and fee distribution graph for the block itself, instead of for the whole stack of N blocks.
> 
> It should be possible to calculate statistics for the full stack of blocks more efficiently, but I haven't spent the time to implement this yet.

I haven't figured out a more efficient way to calculate the full stack fee statistics yet, but in the meantime I could open a PR to revert to the old behavior (at the expense of some of the performance gains from this PR)?

PR #3743, opened 2023-05-11
stack-of-n-blocks fee statistics

mononaut:
Implements an efficient online fee range calculation, in order to restore the fee statistics for the full stack of N mempool blocks (which was removed in PR #3735).

Before
Screenshot 2023-05-10 at 5 56 13 PM

After
Screenshot 2023-05-10 at 5 54 52 PM

Before
Screenshot 2023-05-10 at 5 56 03 PM

After
Screenshot 2023-05-10 at 5 55 21 PM


softsimon (approved):
tested ACK @ [3d1cd319]

PR #974, opened 2021-12-05
Display missing taproot inner script

softsimon:
fixes #973

This PR reveals the inner spending script on a taproot transaction.
This does _only_ fix the missing spend script for the backend when running `bitcoind` and not when using `esplora` as blockstream/electrs needs to implement this natively just like with other output types.

Some open questions remains to be tested/answered:
* Will this work on all types of Taproot-transactions
* Should it be called `P2WSH witness script`?

**Before:**
Screen Shot 2021-12-05 at 23 54 10

**After:**

Screen Shot 2021-12-05 at 23 54 34

Issue #2380, opened 2022-08-23
Esplora doesn't support Taproot tapscripts


softsimon:
This could either be a `blockstream/electrs` or a `rust-bitcoin` deficiency, but it's still a problem that we still don't support showing Tapleaf scripts in production.

When running your local Bitcoin Core (or Elements Core) we support it simply by doing the same thing as with Segwit: https://github.com/mempool/mempool/blob/master/backend/src/api/bitcoin/bitcoin-api.ts#L412

So it should be an easy fix / addition for Esplora as well. But requires some investigation where the problem lies. 

Here is a Liquid Testnet Taproot transaction for example: a0dc0d0128aff1c35cd6a911153abb57b1baa854e4037104f04cb36394ab1fc1

Elements Core:
<img width="552" alt="Screen Shot 2022-08-23 at 23 07 21" src="https://user-images.githubusercontent.com/8561090/186243486-65b843ca-9793-477b-8cf8-4ac8933a6f9d.png">

Blockstream/electrs
<img width="552" alt="Screen Shot 2022-08-23 at 23 07 25" src="https://user-images.githubusercontent.com/8561090/186243594-dead27ef-dd60-4ede-b5ad-6aead1fce411.png">

wiz:
This is an ops issue, I probably just need to upgrade the electrs and re-index it. I'm sure it works on blockstream.info, right?

softsimon:
> This is an ops issue, I probably just need to upgrade the electrs and re-index it. I'm sure it works on blockstream.info, right?

Esplora does not support it. Another example:

https://blockstream.info/tx/977b937e7ab7ca52d87ec25a50e36ff164dacb00165ef30dbb13f63c0e7bf7eb

With Bitcoin Core only:

<img width="548" alt="Screen Shot 2022-08-24 at 12 52 13" src="https://user-images.githubusercontent.com/8561090/186375020-c8d83748-fb95-4103-80b0-a07f0d43e158.png">

wiz:
If so then I would just report the issue to Blockstream, it's not a bug in this repository

ProgramComputer:
I checked the version of electron.blockstream.info which blockstream.info is based on 
```
var ElectrumClient = require("electrum-client-js")

const config = {
  host: 'electrum.blockstream.info',
  port: 50002,
  protocol: 'ssl',
}

const main = async () => {
  console.log('Connecting...')
  const client = new ElectrumClient(config.port, config.host, config.protocol)

  await client.connect()

[...]

PR #3744, opened 2023-05-11
Add InnerScripts to bitcoin-api

ProgramComputer:
[Esplora API](https://github.com/Blockstream/esplora/blame/947e4b7bbc79fe359c6e5eb0661e0a7be75bc6a9/API.md#L366) has inner scripts. Ref #974


softsimon (commented):
nACK

We are not adding `bitcoinjs-lib.js` to the frontend. The correct solution is fixing it directly in Elects.

ProgramComputer:
> nACK
> 
> We are not adding `bitcoinjs-lib.js` to the frontend. The correct solution is fixing it directly in Elects.

Is it [blockstream/electrs](https://github.com/Blockstream/electrs) or [romanz/electrs](https://github.com/romanz/electrs)? 

And the change has to happen upstream at that repository?

softsimon:
> > nACK
> > We are not adding `bitcoinjs-lib.js` to the frontend. The correct solution is fixing it directly in Elects.
> 
> Is it [blockstream/electrs](https://github.com/Blockstream/electrs) or [romanz/electrs](https://github.com/romanz/electrs)?
> 
> And the change has to happen upstream at that repository?

When you self host with just Bitcoin Core and Electrs, we already get the InnerScripts. The problem is `blockstream/electrs `that is used in production that needs to be patched. Here is the open issue:  https://github.com/mempool/mempool/issues/2380

PR #3751, opened 2023-05-11
Fix RBF timestamps to always use seconds

mononaut:
Fixes a bug where RBF timestamps fell back to a timestamp in milliseconds when `tx.firstSeen` was not set, when all RBF times should be in seconds.


softsimon (approved):
utACK

Couldn't reproduce right now

PR #3753, opened 2023-05-11
Fix difficulty estimate

mononaut:
The current difficulty adjustment estimation is subtly wrong at the very end of each epoch.

### Problem

Actual difficulty adjustments are determined by
```
difficultyChange = ((targetEpochDuration / actualEpochDuration) - 1) * 100
```
where the `targetEpochDuration` is 20160 minutes, and the `actualEpochDuration` is the time taken to mine the previous 2015 blocks (not 2016, due to a well documented off-by-one error in Bitcoin Core).

This means the next difficulty adjustment is "locked in" as soon as the last block in an epoch is mined, since by that point the `actualEpochDuration` is known (equal to the difference in timestamps between the 1st and 2016th blocks in the epoch).

---

Our estimates were made with
```
difficultyChange = ((BLOCK_SECONDS_TARGET / (diffSeconds / blocksInEpoch)) - 1) * 100;
```
where `diffSeconds` is the time since the 1st block of the epoch, and `blocksInEpoch` is the number of blocks mined so far.

`diffSeconds` continues to increase after the last block in the epoch is mined, which has the effect of gradually lowering the estimate below the correct value while we wait for the next epoch to begin.

---

For example, in the most recent epoch, the next difficulty adjustment was guaranteed to be -1.45% as soon as block 788255 was received. However, we continued to lower the estimate while waiting for block 788256 (the first of the next epoch):

Screenshot 2023-05-11 at 11 16 05 AM

Screenshot 2023-05-11 at 11 06 42 AM

### Fix

This PR freezes the predicted adjustment at the known correct value as soon as the last block in an epoch is mined:

Screenshot 2023-05-11 at 12 44 48 PM

It also adjusts the estimate calculation to include the expected next block in the `blocksInEpoch` count. i.e. the estimate asks "what would the average block time be if the next block arrived right now?".

That doesn't make a huge difference to ongoing estimates, but produces a nicer transition from estimate to confirmed adjustment on the arrival of the last block.

### DIFF
--- a/backend/src/__tests__/api/difficulty-adjustment.test.ts
+++ b/backend/src/__tests__/api/difficulty-adjustment.test.ts
@@ -14,11 +14,11 @@ describe('Mempool Difficulty Adjustment', () => {
           750134,                         // Current block height
           0.6280047707459726,             // Previous retarget % (Passed through)
           'mainnet',                      // Network (if testnet, next value is non-zero)
-          0,                              // If not testnet, not used
+          0,                              // Latest block timestamp in seconds (only used if difficulty already locked in)
         ],
         { // Expected Result
           progressPercent: 9.027777777777777,
-          difficultyChange: 12.562233927411782,
+          difficultyChange: 13.180707740199772,
           estimatedRetargetDate: 1661895424692,
           remainingBlocks: 1834,
           remainingTime: 977591692,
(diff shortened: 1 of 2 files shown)


softsimon (approved):
tested ACK @ [d3226657]

PR #3754, opened 2023-05-11
Difficulty adjustment lock in

mononaut:
_(builds on PR #3753)_

Display a message to show that the difficulty adjustment is confirmed after the last block in an epoch.

Screenshot 2023-05-11 at 1 17 31 PM


softsimon:
I'm always hesitant to adding new i18n strings. Do you think it could be enough with just showing the full bar and no text?

PR #3756, opened 2023-05-11
Adds SSL to config

ProgramComputer:
Adds SSL to config. #2840


knorrium:
@ProgramComputer heads up that I'm adding tests for the Docker config and you'll have to rebase once #3776 gets merged. Sorry for the inconvenience.

knorrium:
@ProgramComputer we merged the other PR, can you please fix the merge conflicts so the new tests can run?

ProgramComputer:
> @ProgramComputer we merged the other PR, can you please fix the merge conflicts so the new tests can run?

Merge conflicts fixed

wiz:
Thanks for the PR, but could you default to SSL = false?
Since this is something new it should not be enabled by default

ProgramComputer:
> Thanks for the PR, but could you default to SSL = false? Since this is something new it should not be enabled by default

Changed the SSL default to false

production/mempool-config.testnet.json
```diff
@@ -22,7 +22,8 @@
   "CORE_RPC": {
     "PORT": 18332,
     "USERNAME": "__BITCOIN_RPC_USER__",
-    "PASSWORD": "__BITCOIN_RPC_PASS__"
+    "PASSWORD": "__BITCOIN_RPC_PASS__",
+    "SSL":false
```
softsimon, line 26: You can revert all production config additions, as we don't need to add when the default config is the same. Keeps them cleaner.

softsimon (changes requested):

PR #3770, opened 2023-05-12
Redis cache prototype

mononaut:
This draft PR prototypes integrating Redis as a replacement for the JSON file cache.

So far, mempool, block and RBF data is implemented.

Loading data at startup seems to be a little slower than with JSON files, but being able to update specific cache entries in real time completely eliminates the huge overhead from periodically saving cache data.

### Config

New config variables:
```
MEMPOOL: {
  ...
  CACHE_ENABLED: boolean // toggles JSON file cache (default true)
  ...
},
REDIS: {
  ENABLED: boolean // toggles Redis cache (default false)
  UNIX_SOCKET_PATH: '"/tmp/redis.sock"
}
```

### Prerequisites

The new functionality requires a Redis instance available at `config.REDIS.UNIX_SOCKET_PATH`, with the RedisJSON module loaded.

The easiest way to get that running is with [Redis Stack](https://redis.io/docs/stack/get-started/install/).

Alternatively, you can compile the RedisJSON module separately and run with plain Redis.

Make sure Redis RDB persistence is enabled to avoid losing the cache when Redis restarts.

### Running

The backend can be run with either file caching enabled, or Redis caching, or both.

If both are enabled, the backend will only load from disk but save to both caches. You can use this to migrate existing cache data from JSON files to Redis.

### Todo
- [x] network/schema versioning
- [ ] implement LN funding tx cache
- [ ] handle Redis errors / downtime / connection issues

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -19,6 +19,7 @@
         "maxmind": "~4.3.11",
         "mysql2": "~3.5.2",
         "rust-gbt": "file:./rust-gbt",
+        "redis": "^4.6.6",
         "socks-proxy-agent": "~7.0.0",
         "typescript": "~4.9.3",
         "ws": "~8.13.0"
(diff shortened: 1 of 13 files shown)


wiz:
fixes #338 ?

softsimon:
> fixes #338 ?

This is to replace the JSON cache disk write/load and issues that comes with it.

I see no point with #338 because how would we perform constant calculations on the current mempool if we just store it in Redis, it would have to be loaded back and forward into nodejs memory every second which apparently takes multiple seconds.

mononaut:
Rebased on current master (0fb92e6), added network and schema versioning, and improved handling of large batches of transactions.

mononaut:
Rebased on current master [b58abe47] and optimized cached data to speed up load times.

on an M1 mac this now seems to be faster than loading from disk cache.

wiz:
Replaced by #4063

PR #3771, opened 2023-05-13
Add missing clock websocket subscriptions

mononaut:
Adds missing websocket subscriptions needed to keep the clock interfaces updated.


softsimon (approved):
tested ACK @ [6a5afd7f9]

Issue #3763, opened 2023-05-12
Possible backend crash


nymkappa:
```
May 12 07:32:05 [84333] DEBUG: Cannot save block template for 00000000000000000002ce5109621be135c55557551a11e0e74eacf6ed19dd97. Reason: DB query failed to return, reject or time out within 180s -
        INSERT INTO blocks_summaries (height, id, transactions, template)
        VALUE (?, ?, ?, ?)
        ON DUPLICATE KEY UPDATE
          template = ?

/Users/nymkappa/Documents/mempool/backend/dist/database.js:46
                    reject(new Error(`DB query failed to return, reject or time out within ${hardTimeout / 1000}s - ${query?.sql?.slice(0, 160) || (typeof (query) === 'string' || query instanceof String ? query?.slice(0, 160) : 'unknown query')}`));
                           ^

Error: DB query failed to return, reject or time out within 180s -
        INSERT INTO blocks_summaries (height, id, transactions, template)
        VALUE (?, ?, ?, ?)
        ON DUPLICATE KEY UPDATE
          template = ?

    at Timeout._onTimeout (/Users/nymkappa/Documents/mempool/backend/dist/database.js:46:28)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
```

This is not caught because we

```
        BlocksSummariesRepository.$saveTemplate({
          height: block.height,
          template: {
            id: block.id,
            transactions: stripped
          }
        });
```

without an await, and this function can throw an exception

```
      if (e.errno === 1062) { // ER_DUP_ENTRY - This scenario is possible upon node backend restart
        logger.debug(`Cannot save block template for ${blockId} because it has already been indexed, ignoring`);
      } else {
        logger.debug(`Cannot save block template for ${blockId}. Reason: ${e instanceof Error ? e.message : e}`);
        throw e;
      }
```

nymkappa:
So we either add an await + try/catch in the upper level or we do not rethrow the exception once caught and we log a warn instead (not sure why it's a debug right now, probably a copy paste mistake I've made in the past)

nymkappa:
So we either add an await + try/catch in the upper level or we do not rethrow the exception once caught and we log a warn instead (not sure why it's a debug right now, probably a copy paste mistake I've made in the past)

mononaut:
I'd opt for not rethrowing the exception inside the `$saveTemplate` function, unless we want to retry-on-failure in the upper level?

[...]

PR #3772, opened 2023-05-13
[audit] warn if we cannot save templates and remove exception re-throw

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3763


mononaut (approved):
Tested ACK @ v2.6.0-dev [f95c16a7]

It might be worth making the same change for the `BlocksAuditsRepository.$saveAudit` function, since we also call that without an await in `$handleNewBlock`.

softsimon:
> It might be worth making the same change for the `BlocksAuditsRepository.$saveAudit` function, since we also call that without an await in `$handleNewBlock`.

@nymkappa is this something you want to do in this same PR before we merge?

nymkappa:
> > It might be worth making the same change for the `BlocksAuditsRepository.$saveAudit` function, since we also call that without an await in `$handleNewBlock`.
> 
> @nymkappa is this something you want to do in this same PR before we merge?

Done.

There are few other instances of this issue here and there in the code. So another PR will be needed to clean up all the uncaught re-throw.

softsimon (approved):
ACK

PR #3776, opened 2023-05-14
Docker vars tests

knorrium:
Adding a test that will check if all the config vars are set in the Docker backend start script.

It checks if the value is read from the environment and if it's replaced later.

The first commit will fail, I'll fix the missing configs to validate that the test works.


knorrium:
The first run found a missing entry:

```
  ● Mempool Backend Config › should ensure the docker start.sh script has default values

    expect(received).toContain(expected) // indexOf

    Expected substring: "sed -i \"s!__DATABASE_TIMEOUT__!${__DATABASE_TIMEOUT__}!g\" mempool-config.json"
```

knorrium:
The part that reads the environment variable being detected:

```
  ● Mempool Backend Config › should ensure the docker start.sh script has default values

    expect(received).toMatch(expected)

    Expected pattern: /__LIGHTNING_FORENSICS_INTERVAL__=\${LIGHTNING_FORENSICS_INTERVAL:=(.*)}/
```

knorrium:
Added another test that checks if all the keys set in the unit test fixture (`mempool-config.template.json`) are included in the `mempool-config.json` file used by Docker.

The fixture now has to use proper variable types (array, string, boolean or string) to be checked against the Docker config.

```
  ● Mempool Backend Config › should ensure that the mempool-config.json Docker template has all the keys

    expect(received).toContain(expected) // indexOf

    Expected substring: "\"POOLS_JSON_TREE_URL\": \"__MEMPOOL_POOLS_JSON_TREE_URL__\""
```

knorrium:
@softsimon @mononaut @nymkappa these tests should help us keep all the configs in sync whenever they are updated. Please note that the fixture itself needs to use proper typing or else the test will suggest doing the wrong thing.

wiz (approved):
Yolo ACK

Issue #3658, opened 2023-04-08
Mempool templates and audits should account for high-sigop transactions


mononaut:
Mempool transactions with a high sigop count are treated by Bitcoin Core as having an inflated vsize for fee purposes and mempool sorting, to account for their disproportionate consumption of the 80,000-per-block sigop limit.

```
int64_t GetVirtualTransactionSize(int64_t nWeight, int64_t nSigOpCost, unsigned int bytes_per_sigop)
{
    return (std::max(nWeight, nSigOpCost * bytes_per_sigop) + WITNESS_SCALE_FACTOR - 1) / WITNESS_SCALE_FACTOR;
}
```
(https://github.com/bitcoin/bitcoin/blob/e9262ea32a6e1d364fb7974844fadc36f931f8c6/src/policy/policy.cpp#L295-L298)

This reduces the effective fee rate and relative priority of high-sigop transactions in the mempool.

We do not account for this policy while generating block templates, which can significantly affect the quality of projected blocks and block audits, especially with the recent surge in these types of transactions.

For example, almost all of the bright pink transactions in the template for the block below are high-sigop transactions. We calculate their effective fee rates as 6-12 sats/vb, while the proper sigop-adjusted rates would only be 2-3 sats/vb.

Since the actual minimum fee rate for this block was 5 sats/vb, applying the sigop penalty rule would have removed these transactions from the template and boosted the block health score to 100%. 

<img width="1131" alt="Screenshot 2023-04-08 at 10 04 33 AM" src="https://user-images.githubusercontent.com/83316221/230696731-36017c7b-bde1-4d39-853b-a9eb3d946d92.png">

Unfortunately the necessary data doesn't seem to be exposed by any esplora api methods. The adjusted vsizes are available in responses from mempool-related Bitcoin Core RPC commands (`getRawMempool`, `getMempoolEntry` etc), but these calls are very expensive, so we'll probably need to handle sigop counting internally.

Sjors:
Thanks @mononaut for your example on [Twitter](https://twitter.com/mononautical/status/1646166180145577990). On a recent master-ish node:

```sh
$ bitcoin-cli getrawtransaction 0799c29dfbdec71920b67fe1bd53f8ea89d06fae82d58f31f19da6b1ca023930 true | jq .vsize
1707

$ bitcoin-cli getmempoolentry 0799c29dfbdec71920b67fe1bd53f8ea89d06fae82d58f31f19da6b1ca023930 | jq .vsize
4820
```

[...]

PR #3783, opened 2023-05-17
Omit possible high-sigop txs from block health score

mononaut:
This PR temporarily mitigates the impact of #3658 on block audits/health scores by identifying and excluding transactions which are likely to have an adjusted mempool vsize due to high sigop count.

So far, only transactions with many `OP_CHECKMULTISIG` outputs are excluded (based on a simple formula comparing the adjusted weight of the multisig outputs to the normal vsize of the whole transaction).

A new column is added to the `blocks_audits` table to track these transactions, and they are labeled "High sigop count" in the frontend audit visualization with the same coloring as "marginal fee rate" transactions.

Screenshot 2023-05-17 at 11 45 25 AM


softsimon (approved):
tested ACK @ [81ec54fcb]

Production
Screenshot 2023-05-20 at 23 26 21

With this PR:
Screenshot 2023-05-20 at 23 27 08

PR #3785, opened 2023-05-18
Detect RBF by mined transactions

### LINKED ISSUE #3758: Consider adding confirmed transactions to the RBF chain
If we're not doing this. We should consider adding new block transactions to the RBF chain, since we not always get transactions before they are mined.

mononaut:
Resolves #3758.

This PR tracks the utxos spent by each transaction in our mempool, which allows us to efficiently detect when those are replaced by previously unseen transactions in newly mined blocks.

This means we can show the full RBF chain even when the final version of the transaction doesn't reach us before a block is mined (e.g. due to propagation delays, out-of-band submission, or use of full RBF).

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -77,6 +77,24 @@ export class Common {
     return matches;
   }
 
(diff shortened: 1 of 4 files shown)


softsimon (approved):
tested ACK @ [369db7a6]

PR #3786, opened 2023-05-18
Fix effective fee rates for non-cpfp dependents

### LINKED ISSUE #3674: Improve handling of CPFP special cases (mempool fee gradient bug)
### Description

Quirks of the CPFP rules mean there are some special cases where CPFP-based effective fee rates are poorly defined.

This can produce confusing results in our mempool projections, including unexpected mempool block fee ranges and gradients:

---

### Problem to be solved

In the current GBT algorithm, each CPFP package can only be "sponsored" by a single root transaction.

If an in-mempool transaction dependency graph has multiple higher-fee-rate terminal nodes, only the best candidate is included in the CPFP package.

Any other transactions which are not direct ancestors of that selected candidate are not included in the CPFP package, and their effective fee rates and CPFP relatives are set accordingly.

This can produce confusing results where the effective fee rates displayed for these "orphaned" transactions are much higher than implied by their actual mempool priority (which is strictly lower than the CPFP package they were excluded from, since a transaction can never be included in a block before its ancestors).

---

### Simple worked example

Consider an unconfirmed transaction A, with two children B and C, all having the same vsize.
- A pays 1 s/vb
- B pays 4 s/vb
- C pays 5 s/vb

Possible CPFP packages are:
 - `{A, B}`, paying 2.5 s/vb
 - `{A, C}`, paying 3 s/vb

The advanced GBT algorithm selects the best package first - `{A, C}` with a combined effective fee rate of 3 s/vb.

Each transaction can only belong to a single CPFP package, so the algorithm removes A as a valid ancestor, leaving a single package `{B}` paying an effective rate of 4 s/vb.

Next the algorithm selects `{B}`, as that's now the highest-paying package available.

The result is that B is included _after_ A and C in the projected blocks, but is assigned a _higher_ effective fee rate (since technically C, not B, is "paying for" their parent A).

---

### Proposed solution

The "effective fee rate" we actually want to display to users is the equivalent rate a single simple transaction would have to pay to achieve the same mining priority.

In the worked example above, this _true_ effective fee rate for transaction B would be 3 s/vb.

The advanced GBT algorithm should keep track of the minimum effective rate of CPFP packages containing each transaction's ancestors, and use that as an upper bound on the transaction's final effective fee rate.

Since each transaction must have a strictly lower priority than all of its ancestors, this should be a good approximation of the maximum true effective fee rate.

mononaut:
Resolves #3674

This PR implements the solution proposed in #3674, by adjusting the effective fee rate calculation to consider non-CPFP dependencies.

This new dependency-based effective fee rate is the minimum of the calculated CPFP rate, and the lowest effective fee rate of any unconfirmed ancestors.

This brings the effective fee rate variable in line with the actual mining priority of the transaction.

For example, this transaction depends on an unconfirmed ancestor with an effective fee rate of 78.6 s/vb:
Screenshot 2023-05-18 at 2 01 23 PM
Previously we would show only the basic fee rate of 102 s/vb, since this transaction doesn't belong to any CPFP clusters. This was misleading, because the transaction can't confirm before its lower effective fee rate ancestors which only pay 78.6 s/vb.

---

Fixing these effective fee rate outliers also improves the fee ranges of the projected blocks:

Before:
Screenshot 2023-05-18 at 2 28 51 PM

After:
Screenshot 2023-05-18 at 2 28 26 PM

---

We probably also need to add specific handling in the UI for non-CPFP effective fee rates, to explain how the rate was obtained.

### DIFF
--- a/backend/src/api/tx-selection-worker.ts
+++ b/backend/src/api/tx-selection-worker.ts
@@ -127,7 +127,7 @@ function makeBlockTemplates(mempool: Map<number, CompactThreadTransaction>)
           cpfpClusters.set(nextTx.uid, sortedTxSet.map(tx => tx.uid));
           isCluster = true;
         }
-        const effectiveFeeRate = nextTx.ancestorFee / (nextTx.ancestorWeight / 4);
+        const effectiveFeeRate = Math.min(nextTx.dependencyRate || Infinity, nextTx.ancestorFee / (nextTx.ancestorWeight / 4));
         const used: AuditTransaction[] = [];
         while (sortedTxSet.length) {
           const ancestor = sortedTxSet.pop();
@@ -155,7 +155,7 @@ function makeBlockTemplates(mempool: Map<number, CompactThreadTransaction>)
         // remove these as valid package ancestors for any descendants remaining in the mempool
         if (used.length) {
           used.forEach(tx => {
-            updateDescendants(tx, auditPool, modified);
+            updateDescendants(tx, auditPool, modified, effectiveFeeRate);
           });
         }
 
@@ -251,6 +251,7 @@ function updateDescendants(
   rootTx: AuditTransaction,
   mempool: Map<number, AuditTransaction>,
   modified: PairingHeap<AuditTransaction>,
+  clusterRate: number,
 ): void {
   const descendantSet: Set<AuditTransaction> = new Set();
   // stack of nodes left to visit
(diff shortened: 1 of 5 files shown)


softsimon (changes requested):
Not surewe should display an Effective fee rate, if the fee is practically the same. Perhaps we need in order to reveal the CPFP button when there are related TXIDs to list.

Screenshot 2023-05-31 at 16 23 24

Only the Effective fee rate should display the "fee score" badge when we show both.

Screenshot 2023-05-30 at 16 15 59

PR #3791, opened 2023-05-23
Fix for pool output address matching

### LINKED ISSUE #3782: Unknown block should be tagged to F2Pool?
https://mempool.space/block/00000000000000000001aadcd2ac42e0fabcc46b9a2d2150963d4c594f22e294

softsimon:
fixes #3782

Testing:

Remove block (791074) [00000000000000000001aadcd2ac42e0fabcc46b9a2d2150963d4c594f22e294] from your database and reload it, it should now match it as F2Pool because we match with all output addresses.

This PR doesn't trigger a rescan of unknown pools that we might have missed because this will be fixed after this is merged and we update the miningpool repo next time? (we should add the /F2pool/ tag to the repo to not just rely on output addresses in this case)

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -306,7 +306,7 @@ class Blocks {
     }
 
     const asciiScriptSig = transactionUtils.hex2ascii(txMinerInfo.vin[0].scriptsig);
-    const address = txMinerInfo.vout[0].scriptpubkey_address;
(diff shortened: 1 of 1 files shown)


nymkappa (approved):
Tested ACK @ 31ea2e1d4bef21d4aa037599e4a81b80c0444ffe

### local vs prod
Screenshot 2023-05-24 at 08 40 20


Screenshot 2023-05-24 at 08 40 54

PR #3794, opened 2023-05-24
Restore liquid max block weight to defaults

mononaut:
Removes the artificially low 300kWU block weight cap in the production liquid frontend, which fixes some UI issues when liquid blocks are full.

Before:

Screenshot 2023-05-24 at 3 52 40 PM

After:

Screenshot 2023-05-24 at 3 53 55 PM


softsimon (approved):
utACK

PR #3800, opened 2023-05-25
Don't wipe mempool cache in pools updater

softsimon:
```
May 25 19:17:47 [96012] INFO: Restoring mempool and blocks data from disk cache
May 25 19:18:01 [96012] INFO: Re-saving cache with empty recent blocks data
May 25 19:18:01 [96012] DEBUG: Writing mempool and blocks data to disk cache (sync)...

```


wiz (approved):
Tested ACK  @ 038f965

PR #3801, opened 2023-05-25
Always re-indexing blocks when updating mining pools

nymkappa:


wiz (approved):
utACK - will test as part of staging upgrade

PR #3804, opened 2023-05-25
Break block templates into their own db table

mononaut:
This PR moves block template data out of the `blocks_summaries` db table and into a new dedicated `blocks_templates` table.

It never made much sense to have templates and block summaries in the same table, and separating them makes db backup and re-indexing logic much more straightforward.


backend/src/api/database-migration.ts
```diff
@@ -521,6 +521,18 @@ class DatabaseMigration {
       await this.$executeQuery('ALTER TABLE `blocks_audits` ADD sigop_txs JSON DEFAULT "[]"');
       await this.updateToSchemaVersion(60);
     }
+
+    if (databaseSchemaVersion < 61 && isBitcoin === true) {
+      // Break block templates into their own table
+      if (! await this.$checkIfTableExists('elements_pegs')) {
```
softsimon, line 527: `elements_pegs` ?

softsimon (changes requested):

backend/src/api/database-migration.ts
mononaut: oops, fixed and rebased on latest mater

PR #3808, opened 2023-05-26
Add nostr as a transport mechanism for bitcoin transactions

joostjager:
Bypass the current limitations of the bitcoin p2p network.

For more context, see https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2023-May/021700.html

A follow-up can leverage core's `submitpackage` rpc for package relay.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -19,8 +19,10 @@
         "maxmind": "~4.3.8",
         "mysql2": "~3.2.0",
         "node-worker-threads-pool": "~1.5.1",
+        "nostr-tools": "^1.10.1",
         "socks-proxy-agent": "~7.0.0",
         "typescript": "~4.7.4",
+        "websocket-polyfill": "^0.0.3",
         "ws": "~8.13.0"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


joostjager:
Seems the cla-bot isn't pleased with my contributors file?

wiz:
Thanks for the PR. As much as I love nostr, this PR seems to add many new npm deps, and as a general rule we are trying to reduce the number of npm deps in the project - also from looking at the changed code it is not really clear what these deps do or how the integration works... will have to think about this one.

joostjager:
>As much as I love nostr, this PR seems to add many new npm deps, and as a general rule we are trying to reduce the number of npm deps in the project

That is understandable. There are probably much leaner ways to do it, which I am happy to look if it is decided that the concept that is shown here is viable.

>also from looking at the changed code it is not really clear what these deps do or how the integration works... will have to think about this one.

The main idea is that going forward, the bitcoin p2p network may no longer be the only source of transactions. To an extent this is already the case with certain bitcoin users having direct ties to miners for for example non-standard transactions.

Nostr is a way to make this process more transparent and accessible, improving operations such as fee estimation that rely on the state of the mempool for all users. Naturally I think that users would want those out-of-band transactions visible in mempool.space. 

As for integration, I think there are multiple ways to do it. This PR shows an integration where mempool passes on transactions acquired via nostr to bitcoin core. But perhaps it is also conceivable that the backend ingests these transactions directly?

nymkappa:
I'd probably prefer to see this as a small standalone server we can run anywhere rather than being integrated in the main project, because it seems to be a door for dos'ing our backend 👍🏻.

wiz (commented):
Concept NACK

joostjager:
Definitely a fair point. Would you consider an addition of such a standalone server to https://github.com/mempool/mempool/blob/master/docker/docker-compose.yml viable at some point in the future?

PR #3809, opened 2023-05-26
Switch tm to r for registered trademarks

hunicus:
These changes switch (tm) for (r) for the following marks, since they are now registered:

Screenshot from 2023-05-26 17-41-48

Screenshot from 2023-05-26 17-42-03


wiz (approved):
utACK

PR #3811, opened 2023-05-26
Fix missing fees in $updateBlocks without esplora

### LINKED ISSUE #3805: Median fee set to 0
image

mononaut:
Fixes #3805

Running without esplora, if we have to fetch any transactions during `$updateBlocks` we do so without adding prevouts and calculating fees. (presumably to maintain performance?)

We then attempt to calculate fee statistics using this list of transactions, and the incomplete fee data obviously produces garbage results.

Most of the time it isn't too noticeable, since we can find most transactions from new blocks in our mempool, where prevouts and fees have already been added.

But after being offline for a while, the mempool may be missing the majority of transactions in new blocks, and a higher proportion of transactions will lack fee data.

This PR fills in any missing transaction fees using data we already have available from an earlier call to `getBlock(verbosity=2)`, so that we can calculate the correct fee statistics.


wiz:
@softsimon did you want to review this since it's a non-esplora setup

nymkappa (changes requested):
It worked fine but would not it make sense to fix the `getTransactionsExtended` function instead?

We request all transactions through the 3rd param `onlyCoinbase: false`, but we do not return those transactions if they're not in the mempool and if we do not use esplora.

Maybe this test is buggy:

```
else if (config.MEMPOOL.BACKEND === 'esplora' || !memPool.hasPriority() || i === 0)
```

The function should return all transactions regardless of the backend we use?

mononaut:
My assumption was that the reason we skip fetching prevouts in that non-esplora case is because it would be prohibitively expensive (especially for low powered devices).

Considering we only need the fees here, already have them from the verbose block anyway, and this is the only instance of `getTransactionsExtended(onlyCoinbase=false)` in the codebase, it didn't seem worth changing `getTransactionsExtended` to fetch all of those prevouts and calculate fees again manually.

But maybe it would make sense to have a separate `TransactionExtendedNoFees` type and `getTransactionsExtendedNoFees()` function so that we don't make this mistake of relying on null transaction fees again in future.

nymkappa:
> But maybe it would make sense to have a separate `TransactionExtendedNoFees` type and `getTransactionsExtendedNoFees()` function so that we don't make this mistake of relying on null transaction fees again in future.

I'll open an issue for that. It's low priority. Thanks for your explanation 👍🏻.

nymkappa (approved):
Tested ACK @ 6232a92de3aebe969e43316a025be075c64de05e

PR #3815, opened 2023-05-27
Unmute about promo video on click

hunicus:
Curious if/how this works on iOS?


knorrium:
utACK

knorrium (approved):

hunicus:
Added `touchstart` so it works on my Android phone now.

softsimon (approved):
tested ACK @ [8c78e350]

Works on Android and iOS

PR #3818, opened 2023-05-29
Count sigops and use adjusted vsizes

### LINKED ISSUE #3658: Mempool templates and audits should account for high-sigop transactions

mononaut:
_(builds on PR #3786)_

Fixes #3658.

This PR examines input and output scripts in transactions loaded into the mempool cache in order to count the number of signature operations (sigops) per transaction.

We then calculate an `adjustedVsize` and `adjustedFeePerVsize` as defined in https://github.com/bitcoin/bitcoin/blob/e9262ea32a6e1d364fb7974844fadc36f931f8c6/src/policy/policy.cpp#L295-L298, to correctly reflect the reduced mining priority of high-sigop transactions.

The GBT algorithm is adapted to construct blocks using these adjusted vsizes and rates, as well as enforce the per-block total sigop limit.

This massively improves expected blocks and audit scores when there are many high-sigop transactions in the mempool.

---

A side effect of properly accounting for sigops means projected blocks with lots of high-sigop transactions may appear underfilled:

Screenshot 2023-05-29 at 5 16 33 PM

If that will be too confusing for users, perhaps we can treat the high-sigop transactions as if they have an even larger weight for the purposes of the mempool visualizations. (e.g. `max(weight, sigops × 50)`)

---

Sigop count and adjusted vsize are added in a new row in the "details" section of the transaction page:

Screenshot 2023-05-29 at 5 15 36 PM

and reduced mining priority is accounted for in the effective fee rate field where appropriate:
 
Screenshot 2023-05-29 at 5 15 54 PM

---

Example of the impact on block audits:

Before:

Screenshot 2023-05-31 at 10 45 20 AM

After:

Screenshot 2023-05-31 at 10 59 44 AM


backend/src/api/transaction-utils.ts
```diff
@@ -50,7 +59,31 @@ class TransactionUtils {
...
+    const fractionalVsize = (transaction.weight / 4);
+    const sigops = this.countSigops(transaction);
+    // https://github.com/bitcoin/bitcoin/blob/e9262ea32a6e1d364fb7974844fadc36f931f8c6/src/policy/policy.cpp#L295-L298
+    const adjustedVsize = Math.max(fractionalVsize, sigops *  5); // adjusted vsize = Max(weight, sigops * bytes_per_sigop) / witness_scale_factor
+    const feePerVbytes = Math.max(Common.isLiquid() ? 0.1 : 1,
+      (transaction.fee || 0) / fractionalVsize);
+    const adjustedFeePerVsize = Math.max(Common.isLiquid() ? 0.1 : 1,
+      (transaction.fee || 0) / adjustedVsize);
+    const transactionExtended: MempoolTransactionExtended = Object.assign(transaction, {
+      vsize: Math.round(transaction.weight / 4),
```
junderw, line 79: This should be ceil. Or you could use the vsize from a few lines up.

backend/src/api/transaction-utils.ts
```diff
@@ -50,7 +59,31 @@ class TransactionUtils {
...
+    }
+    return transactionExtended;
+  }
+
+  public extendMempoolTransaction(transaction: IEsploraApi.Transaction): MempoolTransactionExtended {
+    const vsize = Math.ceil(transaction.weight / 4);
+    const fractionalVsize = (transaction.weight / 4);
+    const sigops = this.countSigops(transaction);
+    // https://github.com/bitcoin/bitcoin/blob/e9262ea32a6e1d364fb7974844fadc36f931f8c6/src/policy/policy.cpp#L295-L298
+    const adjustedVsize = Math.max(fractionalVsize, sigops *  5); // adjusted vsize = Max(weight, sigops * bytes_per_sigop) / witness_scale_factor
```
junderw, line 73: Since sigops * 5 can only be an integer, but fractionalVsize can be a non-integer, if we imagine the case where fractionalVsize is between the sigops * 5 and (sigops * 5) - 1, the sigops * 5 will be chosen, but if you ceil fractionalVsize, it will be equal and won't matter.

So it would be better to use vsize which is an integer, so that adjustedVsize is also an integer.

junderw: Or maybe the ceil should be done inside the Object.assign below.

softsimon (approved):
tested ACK @ [0ac76e12]

backend/src/api/audit.ts
```diff
@@ -140,19 +137,11 @@ class Audit {
       censored: Object.keys(isCensored),
       added,
       fresh,
-      sigop,
+      sigop: [],
```
Pyakouto, line 140: My funds

Pyakouto (commented):
My funds

PR #3822, opened 2023-05-31
fix possible backend crash

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3819


softsimon (approved):
tested ACK @ [0b74cf1d8]

PR #3825, opened 2023-06-03
Fix negative confirmations

mononaut:
Fixes an issue where transactions could display with a zero or negative confirmations badge when confirmed in a block which has not finished processing in the backend.

This PR also consolidates confirmation badges on various pages into a single reusable `<app-confirmations>` component.

### DIFF
--- a/frontend/src/app/services/state.service.ts
+++ b/frontend/src/app/services/state.service.ts
@@ -92,7 +92,7 @@ export class StateService {
 
   networkChanged$ = new ReplaySubject<string>(1);
   lightningChanged$ = new ReplaySubject<boolean>(1);
-  blocks$: ReplaySubject<[BlockExtended, boolean]>;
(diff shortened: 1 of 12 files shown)


frontend/src/app/services/state.service.ts
```diff
@@ -92,7 +92,7 @@ export class StateService {
 
   networkChanged$ = new ReplaySubject<string>(1);
   lightningChanged$ = new ReplaySubject<boolean>(1);
-  blocks$: ReplaySubject<[BlockExtended, boolean]>;
+  blocks$: ReplaySubject<[BlockExtended, string | boolean]>;
```
softsimon, line 95: Why keep `boolean` as type?

mononaut: good point, fixed

softsimon (approved):
tested ACK @ [8b1dff6]

PR #3826, opened 2023-06-04
fix firstSeen reset migration bug

mononaut:
Fixes an issue where transaction `firstSeen` times could get reset while migrating transactions in the disk cache to use the new sigop feature (#3818)


softsimon (approved):
tested ACK @ [37dd95a4]

PR #3829, opened 2023-06-05
Add expected total fees audit

joostjager:
This PR adds a new field to the block audit view that shows the expected total fees for a block based on the locally produced block template. The difference with the actually mined block tells something about the optimality of the constructed block.

The field `expected_fees` can be added in other views as well, but first wanted to see what thoughts are on this functionality in general. Clean up to do too.

image

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -282,10 +282,12 @@ class Blocks {
       }
 
       extras.matchRate = null;
+      extras.expectedFees = null;
       if (config.MEMPOOL.AUDIT) {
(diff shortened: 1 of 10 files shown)


mononaut:
thanks for this! I've extended the implementation a bit in #3846

would you mind adding a [Contributor License Agreement file](https://github.com/mempool/mempool/blob/master/CONTRIBUTING.md) to this PR so that we can get this ready to merge?

joostjager:
@mononaut I've added the CLA, but the bot is still complaining it seems?

joostjager:
Thanks for finishing my incomplete and not as good as this one PR 🎉

PR #3830, opened 2023-06-05
Fix: Annex parsing for p2tr on bitcoind/romanz backends

### LINKED ISSUE #3828: Taproot annex interpreted as script
### Description

The taproot annex seems to be not properly recognized and is interpreted as script instead.

#### Version

v2.5.0

### Steps to reproduce

View transaction with taproot annex. This means that the last witness element starts with 0x50 followed by the annex bytes.

### Expected behaviour

Even though the annex is non-standard, it is consensus-valid. It would therefore be better to either not try to interpret the annex as script or add a new display field that shows the annex bytes.

### Actual behaviour

Annex interpreted as script.

### Screenshots

image

junderw:
Fixes #3828

The fix for esplora backends (production mempool space) is [here](https://github.com/mempool/electrs/pull/2) (Note: for esplora backends, it would just show as having no script, it would not try to parse the signature as a script like bitcoind backends do currently (this PR will fix bitcoind backends))

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api.ts
+++ b/backend/src/api/bitcoin/bitcoin-api.ts
@@ -415,10 +415,36 @@ class BitcoinApi implements AbstractBitcoinApi {
       vin.inner_witnessscript_asm = this.convertScriptSigAsm(witnessScript);
     }
 
-    if (vin.prevout.scriptpubkey_type === 'v1_p2tr' && vin.witness && vin.witness.length > 1) {
-      const witnessScript = vin.witness[vin.witness.length - 2];
-      vin.inner_witnessscript_asm = this.convertScriptSigAsm(witnessScript);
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK @ [9e1de656]

@joostjager  verified annex parsing now works. I confirmed taproot script still works.

joostjager:
Does taproot script spend still work? In https://github.com/mempool/mempool/issues/3828#issuecomment-1578944019, I am getting something different compares to mempool.space even though there is no annex

junderw:
https://github.com/mempool/mempool/issues/3828#issuecomment-1579139672

mempool production doesn't show any taproot script spends right now, because esplora is broken (PR to fix is open)

I checked your screenshot and it looks fine.

Thanks!

PR #3838, opened 2023-06-06
calculate & index ancestor-dependent effective rates

mononaut:
This PR extends the dependency-based effective fee rate logic of #3786 to CPFP calculations for mined blocks.

The effective fee rate of each transaction in a mined block is now calculated as the minimum of the effective rate of that transaction and that of all of its ancestors in the same block.

For example, this transaction has a fee rate of 93.2 sat/vb, but depends on an ancestor in the same block with an effective rate of only 90.5 sat/vb:

Screenshot 2023-06-06 at 6 24 30 PM

This makes rates more consistent between projected and mined blocks, and should also make it easier to compute accurate block fee ranges.


softsimon (approved):
tested ACK @ [ca9b4828]

PR #3839, opened 2023-06-06
Clean up legacy CPFP calculations

mononaut:
This PR removes some unnecessary legacy CPFP calculations which can cause performance issues during new block processing when the mempool is very large.

It also ensures that CPFP relatives are reset between runs of the GBT algorithm, to ensure transactions can't accumulate stale ancestors.


softsimon (approved):
tested ACK @ [c8fc416c]

PR #3842, opened 2023-06-07
Fix fee range inconsistencies

### LINKED ISSUE #3836: Different fee span based on uptime
Even after wiping my cache and deleting the last 50 blocks from my db, my local does not match prod. Running 7d92cdf015d77944c0cc3dc5030fd2df1d172a01.

If possible, use same algo to generate fee span, regardless if we're online or catching up.

### Local
Screenshot 2023-06-06 at 18 09 15

### Prod
Screenshot 2023-06-06 at 18 09 35

### LINKED ISSUE #3837: Fee span is different between block preview and block overview
Probably related to https://github.com/mempool/mempool/issues/3836

Expanding a block shows different min value between the block preview and the block overview panel. It happens both in prod and in master branch.

### Local
image

### Prod
image

mononaut:
Fixes #3837 by removing the fee range offset on the block page (previously used to reduce the effect of outliers, but not necessary with our newer fee range heuristics).

Also partially fixes #3836, by saving extended blocks to the DB during fast-forwarded block processing. We were already doing all of the work to create the extended block, including calculating fee ranges, but for some reason did not save the results to the DB.

Historical blocks will continue to be indexed using simple fee ranges direct from Bitcoin Core. Fixing that is possible, but would require a complete overhaul of the block indexing logic.

### DIFF
--- a/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
+++ b/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
@@ -32,8 +32,8 @@
             </ng-template>
             <div [attr.data-cy]="'bitcoin-block-' + offset + '-index-' + i + '-fee-span'" class="fee-span"
               *ngIf="block?.extras?.feeRange; else emptyfeespan">
-              {{ block?.extras?.feeRange?.[0] | number:feeRounding }} - {{
-              block?.extras?.feeRange[block?.extras?.feeRange?.length - 1] | number:feeRounding }} <ng-container
+              {{ getMinBlockFee(block) | number:feeRounding }} - {{
+              getMaxBlockFee(block) | number:feeRounding }} <ng-container
                 i18n="shared.sat-vbyte|sat/vB">sat/vB</ng-container>
             </div>
             <ng-template #emptyfeespan>
--- a/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
+++ b/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
@@ -31,9 +31,8 @@
               </div>
             </ng-template>
             <div [attr.data-cy]="'bitcoin-block-' + offset + '-index-' + i + '-fee-span'" class="fee-span"
-              *ngIf="block?.extras?.feeRange; else emptyfeespan">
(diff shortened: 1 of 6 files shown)


softsimon:
> but not necessary with our newer fee range heuristics).

Are you sure. The problem is still miner manually selected TXs with 0 fee for example.

mononaut:
yes, the new fee range calculation handles that fine.

e.g here's a recent block with some manually selected txs paying 7 s/vb, and with this PR the fee span correctly starts at the actual market rate of ~31 s/vb.

Screenshot 2023-06-14 at 6 15 44 PM

mononaut:
But I've just realized that obviously we still need to handle older blocks that still have simple fee ranges from Bitcoin Core.

mononaut:
I've just pushed a change that checks which kind of `feeRange` we have, and chooses the minimum rate for the fee span accordingly.

it's a bit of a hack, but avoids adding flags or forcing a block reindex.

longer term, maybe we should reserve `feeRange` for the stats from Bitcoin Core, and add new `minEffectiveFeeRate` and `maxEffectiveFeeRate` fields for the fee span using the new heuristics.

frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
```diff
@@ -32,8 +32,8 @@
             </ng-template>
             <div [attr.data-cy]="'bitcoin-block-' + offset + '-index-' + i + '-fee-span'" class="fee-span"
               *ngIf="block?.extras?.feeRange; else emptyfeespan">
-              {{ block?.extras?.feeRange?.[0] | number:feeRounding }} - {{
-              block?.extras?.feeRange[block?.extras?.feeRange?.length - 1] | number:feeRounding }} <ng-container
+              {{ getMinBlockFee(block) | number:feeRounding }} - {{
```
softsimon, line 35: Is it possible to refactor this into not using function calls from the angular template. The problem is that the function is executed 500 times on load and 160 times every second to check for changes.

mononaut: yes, I've just moved that logic out of the template and into the data pipeline.

softsimon (approved):
tested ACK @ [ee392832]

Issue #3065, opened 2023-02-16
Next block Y axis provides redundant information


MwithM:
### Description

![image](https://user-images.githubusercontent.com/50149324/219337457-11c05329-e81c-48be-88ac-cfd77da8adc2.png)

On this table, Y axis shows that there are tx in the mempool paying 224 sat/vbyte, and it appears that there are a lot. Still, the median fee is 23 sats/vbyte.

### Problem to be solved

We already know that 224 is a lot more than 23, this graphic does not provide relevant information.

### Proposed solution

Add the frecuency in the Y axis for a range of mining fees in the X axis, so there will be i.e. 500 tx paying 20-25 sats/vbyte and only 100 tx paying more than 100 sats/vbyte.
This way, the table will give very good info.

#### Additional info

I have no idea what the 1 - 7 is referring to in the X axis.

antonilol:
related: #1721 and #1950

Issue #1721, opened 2022-05-25
mempool size graph bug with numbers > 999


antonilol:
### Description

mempool block fee graph doesnt seem to handle numbers larger than 999 sat/vB 

#### Version

mempool.space v2.4.0-dev [774cd98a]

### Steps to reproduce

wait for a transaction with fee rate > 999 sat/vB and click the mempool block

### Expected behaviour

4 digits

### Actual behaviour

3 digits

### Screenshots

![image](https://user-images.githubusercontent.com/49868160/170212091-969de706-a821-4088-ad8c-3bf015ab69af.png)

#### Device or machine

chromium linux pc

antonilol:
other remark: this graph does not give any direct insight other than 'there is someone paying extreme fee', 20 sat/vB is maybe 1 pixel higher than 1 sat/vB

mononaut:
Fixed by #3843

PR #3843, opened 2023-06-07
Better projected fee graph

mononaut:
This PR leverages the detailed fee data we already have available to improve the resolution of the fee distribution graph on the projected block page.

| Before | After |
|-|-|
| Screenshot 2023-06-07 at 2 26 32 PM | Screenshot 2023-06-07 at 2 26 07 PM |


nymkappa:
Maybe fixes https://github.com/mempool/mempool/issues/3065 ?
Also see if https://github.com/mempool/mempool/issues/1721 is still a bug and maybe include it in this PR 👍🏻

nymkappa:
> The bug in #1721 seems to be resolved up to 4-digit fees now, but fees over 9999 s/vb will still be cut off. I'll open a separate PR for that because it might not be a trivial fix.

Use power of ten label to keep the length fixes 👍🏻. No one will care about the exact value when we're at 10,000+ s/vb IMO.

softsimon (approved):
tested ACK @ [eca40f94c]

PR #3846, opened 2023-06-08
Add expected vs actual audit details comparison table

mononaut:
_(builds on PR #3829)_

This PR adds some new mini tables to the audit panel on the block page, to allow quick comparison of key details between the expected and actual blocks:

Screenshot 2023-06-08 at 6 40 54 PM

The repetition of "Expected" and "Actual" in the labels might be overkill, but I want to make it abundantly clear which stats relate the projected template and which to the actual mined block. Perhaps the layout cues are sufficient.

---

~~There is some overlap with PR #3829, which also adds an "expected fees" field.~~

~~The approach in this PR is slightly different, in that we calculate the expected fees and total weight client-side by iterating over the transactions in the expected block template. The server-side approach may be preferable if we want to use this data elsewhere or expose it in the API.~~

This PR now builds on top of #3829.

### DIFF
--- a/frontend/src/app/components/blocks-list/blocks-list.component.html
+++ b/frontend/src/app/components/blocks-list/blocks-list.component.html
@@ -18,7 +18,7 @@
         <th *ngIf="indexingAvailable" class="reward text-right" i18n="latest-blocks.reward" [ngClass]="{'widget': widget, 'legacy': !indexingAvailable}"
           i18n-ngbTooltip="latest-blocks.reward" ngbTooltip="Reward" placement="bottom" #reward [disableTooltip]="!isEllipsisActive(reward)">Reward</th>
         <th *ngIf="indexingAvailable && !widget" class="fees text-right" i18n="latest-blocks.fees" [class]="indexingAvailable ? '' : 'legacy'">Fees</th>
-        <th *ngIf="auditAvailable && !widget" class="fee-delta text-right" i18n="latest-blocks.vs-expected-fees" [ngClass]="{'widget': widget, 'legacy': !indexingAvailable}">vs Expected</th>
+        <th *ngIf="auditAvailable && !widget" class="fee-delta" [ngClass]="{'widget': widget, 'legacy': !indexingAvailable}"></th>
         <th *ngIf="indexingAvailable" class="txs text-right" i18n="dashboard.txs" [ngClass]="{'widget': widget, 'legacy': !indexingAvailable}"
           i18n-ngbTooltip="dashboard.txs" ngbTooltip="TXs" placement="bottom" #txs [disableTooltip]="!isEllipsisActive(txs)">TXs</th>
         <th *ngIf="!indexingAvailable" class="txs text-right" i18n="dashboard.txs" [ngClass]="{'widget': widget, 'legacy': !indexingAvailable}">Transactions</th>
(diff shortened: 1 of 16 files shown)


joostjager:
>The server-side approach may be preferable if we want to use this data elsewhere

Yes, I think you can't use this approach to for example make a pool comparison showing a score based on expected vs actual fees?

softsimon:
Should we not add (+154) after the Actual transaction number and (-0,16 MWU) after the actual weight

mononaut:
Rebased on #3829 with recent changes from master merged in.

I've lightly refactored and extended the backend part to also save expected block weight and to backfill expected fee/weight data for previously audited blocks.

I also added the deltas between expected and actual values for weight and transaction count:

Screenshot 2023-06-09 at 2 33 42 PM

using relative % instead of absolute differences, since that looked a bit cleaner, and makes it easier to compare between different blocks.

joostjager:
While testing, I noticed that the expected weight for a block was exactly 4 MWU. Hadn't seen that before. Could there be some rounding and/or invalid counting issue?

image

frontend/src/app/components/blocks-list/blocks-list.component.html
```diff
@@ -64,6 +66,9 @@ <h1 *ngIf="!widget" class="float-left" i18n="master-page.blocks">Blocks</h1>
               <span class="skeleton-loader" style="max-width: 60px"></span>
             </ng-template>
           </td>
+          <td *ngIf="auditAvailable" class="text-right" [ngClass]="{'widget': widget, 'legacy': !indexingAvailable}">
+            <app-amount [satoshis]="block.extras.expectedFees" [noFiat]="true" digitsInfo="1.2-2"></app-amount>
```
joostjager, line 70: I wonder whether it would be better to show the percentage change here? I find myself comparing the two columns a lot in this view.

mononaut: yes, I agree

mononaut:
reworked the blocks list page to show % difference in fees instead of absolute value, and to fix some responsive width issues.

there isn't really enough space to add an extra column here, so I've sacrificed the (imo redundant) "Mined" column to free up some room:

Screenshot 2023-06-10 at 12 06 18 PM

the "vs Expected" column header isn't ideal, and probably doesn't translate at all well. I'm open to suggestions for a better approach.

joostjager:
Removing "Mined" makes sense indeed. Maybe it can also work if you make fees and expected fees one column:

0.28 BTC (-3.08%)

For naming maybe something like "relative fee (%)" could work too.

Or a double header bar with the top row spanning two columns and displaying 'fee', to avoid potential duplication.

nymkappa:
> the "vs Expected" column header isn't ideal, and probably doesn't translate at all well. I'm open to suggestions for a better approach.

Maybe keep one column `Fees (expected)` and simply show something like `0.28 BTC (+1.5%)`

softsimon:
Is there any chance we could drop the "Expected.. " and "Actual .." prefix and just assume that from the top title. This would make us re-use translated strings and not create 6+ new i18n strings to translate.

mononaut:
yeah, I think that's reasonable, as long as it isn't too confusing in the mobile layout where we don't have both columns side-by-side.

mononaut:
Updated to remove "Expected/Actual..." prefixes:
Screenshot 2023-06-12 at 12 42 40 PM


an the "vs Expected" blocks list table header:
Screenshot 2023-06-12 at 12 39 23 PM

softsimon (changes requested):
Update database migration `currentVersion` variable.

Just an UX opinion if you agree, perhaps remove the ( ) parenthesis on the blocks list.

softsimon (approved):
tested ACK @ [30f8d5cf]

PR #3848, opened 2023-06-09
Chunked worker thread messaging

mononaut:
This draft PR aims to minimize the amount of time the main thread can be blocked by inter-thread communication for the GBT worker.

Sending mempool data to the worker thread is synchronous and slow. When the mempool is very full, it can block the main thread and delay other time-sensitive backend tasks like responding to API requests.

This PR splits some particularly large worker messages into multiple chunks, and yields to the javascript event loop between sending each chunk so that other tasks can continue to make progress.

Keeping in draft for now while I test more thoroughly and figure out optimal chunk sizes etc.

### DIFF
--- a/backend/src/api/mempool-blocks.ts
+++ b/backend/src/api/mempool-blocks.ts
@@ -5,6 +5,8 @@ import config from '../config';
 import { Worker } from 'worker_threads';
 import path from 'path';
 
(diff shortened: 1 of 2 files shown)


softsimon:
Not required anymore

PR #3853, opened 2023-06-12
Replace disclaimer text regarding tx acceleration

hunicus:
This change removes the text regarding transaction acceleration. A link to a landing page / more information about the accelerator can be added when the accelerator is made available.

Screenshot from 2023-06-12 07-18-49


wiz (approved):
utACK

PR #3855, opened 2023-06-12
Fix inconsistent websocket responses

### LINKED ISSUE #3415: blockchain blocks sometimes disappear while restarting backend
Screenshot 2023-03-17 at 20 50 35

mononaut:
This PR refactors the backend websocket handler slightly to fix some inconsistencies:
 - ensure we always send the latest data immediately when a client starts a subscription.
 - fix some bugs which could cause our pre-serialized websocket data to get briefly out of date.
 - avoid sending duplicate or redundant data when a websocket connection is initialized.
 - use pre-serialized data where possible in direct websocket responses.
 
It also allows the frontend to request missed blocks when necessary.
 
Together this resolves #3415, at least in one reproducible case where the frontend receives `block` websocket updates out of sequence (because, e.g, some blocks were missed due to a backend restart, or the websocket init data was out of date).

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -730,6 +730,11 @@ class Blocks {
         this.currentDifficulty = block.difficulty;
       }
 
+      // wait for pending async callbacks to finish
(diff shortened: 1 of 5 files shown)


softsimon (approved):
tested ACK @ [eaad63a]

PR #3859, opened 2023-06-14
Broadcast transaction packages

joostjager:
This PR demonstrates a possible implementation of package broadcasting. It facilitates the addition of transactions to the mempool even when their fee rate is below the minimum, as long as the total package fee rate suffices. An example of this could be a 1 sat/b lightning commitment transaction, which is time-sensitive (the parent tx), along with a high fee rate child transaction that covers the cost for the parent tx too.

According to the FAQ, mempool.space utilizes a virtually limitless mempool. In this scenario, transactions would still be accepted even if they were broadcasted individually. However, versions hosted in other locations might operate with more confined mempools. For these versions, package broadcasting might be the only technique to push a package similar to the one described above into their mempool.

To utilize this on the mainnet, it necessitates the application of this patch: https://github.com/bitcoin/bitcoin/pull/27609. The discussions around the PR suggest its usage is safe.

https://github.com/mempool/mempool/assets/4638168/670504af-a86e-4e86-9cda-a89cb09a1347

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
+++ b/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
@@ -15,6 +15,7 @@ export interface AbstractBitcoinApi {
   $getAddressTransactions(address: string, lastSeenTxId: string): Promise<IEsploraApi.Transaction[]>;
   $getAddressPrefix(prefix: string): string[];
   $sendRawTransaction(rawTransaction: string): Promise<string>;
+  $submitPackage(rawTransactions: string[]): Promise<string>;
   $getOutspend(txId: string, vout: number): Promise<IEsploraApi.Outspend>;
   $getOutspends(txId: string): Promise<IEsploraApi.Outspend[]>;
   $getBatchedOutspends(txId: string[]): Promise<IEsploraApi.Outspend[][]>;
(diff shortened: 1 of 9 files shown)


softsimon:
Looks cool!

softsimon:
Wait. So this is not supported by the latest official Bitcoin Core 25.0?

joostjager:
That's correct. The feature unfortunately didn't make the release. As long as `submitpackage` is only useable on regtest, I guess this PR only serves as an illustration on how the api can be made use of.

softsimon:
Can't and won't be implemented.

emilcondrea:
Can this be used on mainnet now?

glozow:
I think now would be a good time to reconsider this. The RPC has been available in Bitcoin Core since v26.0. From v28.0, nodes will also do some relaying of packages (only if they're 1-parent-1-child), so hopefully we will have some propagation soon.

Assuming adoption, users can submit a 1-parent-1-child CPFPing package and, even if the parent is below mempool minimum feerate, it will be accepted. The child's fees can even be used to replace stuff that the parent conflicts with. Other >=v28.0 nodes will also download these packages from the node! (It will work in the "happy cases" but not 100% reliable in adversarial contexts). We also allow the parent to be 0-fee if it's using version 3 (see bip 431). The RPC also works for submitting any multi-parent-1-child package, but the p2p stuff is not set up yet.

If you want to give users a warning, this kind of text could be appropriate: "Warning: this feature is experimental and still a work in progress. Some features may be unavailable and transactions may not necessarily be accepted by miners and the rest of the network. See Bitcoin Core's [documentation](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/packages.md) for more details."

I know you have some great docs around "there's no such thing as The mempool" so hopefully that's enough.

glozow:
(Can close as resolved)

PR #3861, opened 2023-06-14
Add luminex as enterprise sponsor

hunicus:
It's intentionally _not_ aligned with existing icons because it is not circular and taller than it is wide...looks too large when it is the same height as the other icons.

Screenshot from 2023-06-14 19-00-02

### DIFF
--- a/frontend/src/app/components/about/about.component.html
+++ b/frontend/src/app/components/about/about.component.html
@@ -173,6 +173,21 @@
         </svg>
         <span>Exodus</span>
       </a>
+      <a href="https://www.luminex.io" target="_blank" title="Luminex">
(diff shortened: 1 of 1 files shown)


wiz (changes requested):
You cannot use the cls-1 CSS class as it's used by Spiral, and the breaking of the Spiral logo is visible in your screenshot

softsimon (approved):
tested ACK @ [816f4108]

I believe @wiz is wrong(?) because it looks good to me and no css class is conflicting.

hunicus:
@softsimon he was right...then I fixed it and updated the screenshot.

softsimon:
> @softsimon he was right...then I fixed it and updated the screenshot.

Ah I see you made another push

softsimon (approved):
tested ACK @ [816f410]

PR #3862, opened 2023-06-15
update frontend local instructions

secondl1ght:
While setting up Mempool locally I realized there is a slight missing step in the Frontend README instructions. I simply added the extra step of `cd` into the `frontend` directory so that you are able to follow the remaining steps successfully. Thanks


softsimon (approved):
utACK

PR #3863, opened 2023-06-15
scroll to see all mempool blocks

mononaut:
This PR makes the blockchain bar scrollable in both directions, so that we can display the full set of projected mempool blocks even on smaller screens and mobile devices.

https://github.com/mempool/mempool/assets/83316221/340dd669-4873-4d06-b87e-daa93df00f8e

other uses of these components are unaffected (TV view, docs/faq, clocks etc).

making this work nicely with the flip-blockchain-direction animation is tricky, and maybe depends on browser quirks, ~~so I'm leaving this in draft while I verify that the current solution works on other browsers and devices.~~

### DIFF
--- a/frontend/src/app/components/blockchain/blockchain.component.html
+++ b/frontend/src/app/components/blockchain/blockchain.component.html
@@ -1,9 +1,9 @@
 <div class="text-center" class="blockchain-wrapper" [class.time-ltr]="timeLtr" [class.ltr-transition]="ltrTransitionEnabled" #container>
-  <div class="position-container" [ngClass]="network ? network : ''">
+  <div class="position-container" [ngClass]="network ? network : ''" [style.--divider-offset]="dividerOffset + 'px'" [style.--mempool-offset]="mempoolOffset + 'px'">
     <span>
       <div class="blocks-wrapper">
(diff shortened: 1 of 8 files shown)


mononaut:
yes, very good point. I'll try to implement that.

mononaut:
the relevant block is now automatically scrolled into view when selecting a mempool block or an unconfirmed transaction:

https://github.com/mempool/mempool/assets/83316221/699d8e3c-2560-4f8b-aebb-52e30ca1f5ff

mononaut:
taking this out of draft now that I've tested across several browser/device combinations.

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3888

softsimon (approved):
tested ACK @ [58b805]

But not sure everyone in the team agree of this change.

wiz (approved):
Tested ACK @ v2.6.0-dev [efcb58a4a]

I think this is great, and I would even suggest considering to take the idea further to make the mempool scrolled over a bit by default on the mempool dashboard, and the blockchain scrolled over a bit by default on the blockchain dashboard, but this could be a stupid idea :)

Screenshot 2023-07-05 at 9 28 56
Screenshot 2023-07-05 at 9 29 13

PR #3864, opened 2023-06-15
remove redundant audit score api calls from blocks list

mononaut:
The blocks list component currently makes separate API calls to `/mining/blocks/audit/scores` and `/mining/blocks/audit/score` to fetch block audit health scores, even though all of the relevant data is already included in the extended blocks from `/blocks/:height`.

I believe this behavior is left over from before PR #3166, when blocks did not reliably include the `matchRate` field.

This PR removes those calls and related redundant code, and instead just uses the `matchRate`'s returned by `/blocks/:height`, which simplifies the component considerably.


softsimon (approved):
tested ACK @ [618ba56c]

PR #3870, opened 2023-06-16
the last two docker-compose overrides need MEMPOOL_ in front

pfoytik:
Simple correction that hung me up, in the docker README.MD the last two override variables for the MEMPOOL variables were listed as shown below without "MEMPOOL_" in front
image

This pull request simply corrects those values to be:
image

I tested on my local instance and verified that this change is reflected in the mempool-config.json file when the containers startup.


softsimon (approved):
utACK

PR #3871, opened 2023-06-16
Change RBF status badges

### LINKED ISSUE #3866: Invalid RBF-transactions mislabeled as Unconfirmed
* A RBF-transaction that becomes invalid after a previous transaction being mined should not display as Unconfirmed, but "Removed" or "Invalid", (the first one we have existing i18n tag for;) )

* And the Replaced and Removed tag should be yellow, not red.

Screenshot 2023-06-15 at 22 12 16

mononaut:
Fixes #3866 

Label cached RBF transactions as "Removed" if a prior version gets mined, and change the color classes of both the "Replaced" and "Removed" badges from `btn-danger` to `btn-warning`:

Screenshot 2023-06-16 at 5 34 40 PM
Screenshot 2023-06-16 at 5 34 32 PM


softsimon (approved):
tested ACK @ [20d7e56]

PR #3872, opened 2023-06-16
Reinstate the mempool loop time limit

softsimon:
If we get hundreds/thousands of new transactions, the mempool loop might take time to process causing the backend to fall behind on block height.

Reinstating the 10 second timeout before we break and continue next loop. Only when the mempool is in sync.


mononaut (approved):
Tested ACK @ [[176f5e1](https://github.com/mempool/mempool/pull/3872/commits/176f5e137706dcdc51a9f206b8bd710a03a7307b)]

confirmed that the loop breaks after 10 seconds, but only if the mempool is already in sync.

PR #3873, opened 2023-06-16
Sanitize node pubkey search

softsimon:
Only pubkey characters allowed. Preventing random stuff like mixed collations.


junderw (approved):
ACK

PR #3875, opened 2023-06-17
Bitcoin-Fiat calculator tool

softsimon:
This PR adds a calculator app to the` /tools/calculator `URL looking like this:

Screenshot 2023-06-18 at 00 16 18

Edit: Removed logos because they indicate the network, not the currency
Screenshot 2023-06-18 at 02 08 44

This calculator solves the following issues common to most current calculators out there:

* All conversions are done client side by only fetching the fiat prices of bitcoin, not leaking any data to any third party service or server
* You can easily paste any value into any field and it is parsed correctly even if there are spaces or a comma , instead of a dot which is common in many languages
* You can easily copy any of the results using the buttons
* The Satoshis and Fiat values are rendered clearly at the bottom making it no mistake on how much money using a clever BitcoinSatoshis hybrid number.
* To further explain the point above. Google cuts zeroes making it impossible to know the Sats amount at a glance, eg: `0.000038`

Try it out!

Partly fixes https://github.com/mempool/mempool/discussions/554

### DIFF
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -22,6 +22,7 @@ import { AssetsFeaturedComponent } from './components/assets/assets-featured/ass
 import { AssetsComponent } from './components/assets/assets.component';
 import { AssetComponent } from './components/asset/asset.component';
 import { AssetsNavComponent } from './components/assets/assets-nav/assets-nav.component';
+import { CalculatorComponent } from './components/calculator/calculator.component';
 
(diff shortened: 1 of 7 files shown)


softsimon:
Ninja ack

nymkappa:
- Some padding would be nice on mobile

image


- Maybe limit the BTC decimal count to 8
100,000,001 sats = 1,00000001 BTC

image


- In the same idea, maybe limit sats to integer only
image

PR #3878, opened 2023-06-19
Add 'all time' option for mempool graph

### LINKED ISSUE #3865: Let user see full historical mempool graph
Now our 4 year view is not enough to show all our existing data that is from late 2018.

Consider adding a [All] button next that reveal the full history no matter the years of data we got.

Screenshot 2023-06-15 at 20 49 46

Screenshot 2023-06-15 at 20 51 37

mononaut:
Resolves #3865

Screenshot 2023-06-19 at 12 48 25 PM

### DIFF
--- a/backend/src/api/statistics/statistics-api.ts
+++ b/backend/src/api/statistics/statistics-api.ts
@@ -211,7 +211,7 @@ class StatisticsApi {
       CAST(avg(vsize_1800) as DOUBLE) as vsize_1800,
       CAST(avg(vsize_2000) as DOUBLE) as vsize_2000 \
       FROM statistics \
-      ${interval === 'ALL TIME' ? '' : `WHERE added BETWEEN DATE_SUB(NOW(), INTERVAL ${interval}) AND NOW()`} \
+      ${interval === 'ALL' ? '' : `WHERE added BETWEEN DATE_SUB(NOW(), INTERVAL ${interval}) AND NOW()`} \
       GROUP BY UNIX_TIMESTAMP(added) DIV ${div} \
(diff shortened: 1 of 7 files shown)


nymkappa (changes requested):
Does not fit on mobile (`iPhone SE 2nd gen iOS 14.6` preset, Firefox). Same using my iPhone 13 mini directly.

Screenshot 2023-06-21 at 09 42 42

softsimon:
Perhaps we can remove some of the other time spans, like 3Y ?

Also wasn't there some pre-cashing script cronjob that needs to be updated?

nymkappa:
> Also wasn't there some pre-cashing script cronjob that needs to be updated?

Yes good point. @mononaut please the warm cache script for that new endpoint 👍

mononaut:
I've added the new endpoint to `production/nginx-cache-warmer`, are there any other warm cache scripts?

Also just pushed a change to wrap the toggles into two rows on screens <500px wide:
Screenshot 2023-06-21 at 9 51 19 AM
Screenshot 2023-06-21 at 9 51 37 AM
Screenshot 2023-06-21 at 9 52 01 AM

or we could just drop one of the time spans on small screens, if that doesn't look good enough.

nymkappa (approved):
Thanks @mononaut, splitting into two rows is good for me. Eventually we may just want a dropdown selector on mobile instead.

Pushed a small css fix and "code preference" fix.

Tested ACK @ v2.6.0-dev [[8f995d89](https://github.com/mempool/mempool/commit/8f995d89)]

softsimon (approved):
tested ACK @ [d939391]

PR #3881, opened 2023-06-20
Separate summary and audit-summary API endpoints

### LINKED ISSUE #3740: Missing block template due to nginx cache
The `/audit-summary` is cached empty on request, so it takes a while until we can see the block template.

We should 404 that endpoint if the data isn't ready yet.

mononaut:
This PR disentangles the `/block/:hash/audit-summary` and `/block/:hash/summary` API endpoints to keep audit data and mined block summary data separate.

`.../audit-summary` now:
 - returns *only* audit data
     - projected template, health score, expected fees, audit transaction labels etc
     - *no* mined block summary `transactions` field
 - throws a 404 if no audit data is available, instead of returning the block summary
 
This is technically a breaking change from the previous behavior, but as far as I'm aware this endpoint has never been officially documented anywhere, so nobody should be relying on it.

The `.../summary` endpoint is unchanged.
 
The block page is updated accordingly to fetch those sets of data separately from the two different endpoints.

---

This should fix #3740, since we now throw a 404 instead of sending and caching a response without audit data when the `audit-summary` API is called before audits are ready.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -394,9 +394,13 @@ class BitcoinRoutes {
 
   private async getBlockAuditSummary(req: Request, res: Response) {
     try {
-      const transactions = await blocks.$getBlockAuditSummary(req.params.hash);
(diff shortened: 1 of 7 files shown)


softsimon (approved):
tested ACK @ [4fbab085]

PR #3886, opened 2023-06-21
Hotfix for undefined cpfp cluster bug

mononaut:
The bug is triggered by a stray property access without a null check introduced in #3838:

https://github.com/mempool/mempool/blob/ca9b48283dbd07ff622b5d878e83ebc3aa937a45/backend/src/repositories/CpfpRepository.ts#L113-L129

In theory that db query should never return an empty result, since the cluster id is obtained from a corresponding `compact_cpfp_transaction` record, and we always add transaction and cluster records together (so the cluster should always exist)

In practice, saving cpfp info is not an atomic operation, and transaction records are saved before their corresponding cluster record, so there's a brief moment in which someone could make a `getCpfpInfo` request, or an unrelated crash could occur and leave the db in an inconsistent state.

This hotfix:
 - adds the missing null check.
 - wraps the function call in a try/catch.
 - changes the order of operations so that cluster records are saved before their corresponding transaction records.
 
I'll look into making the database operations in `CpfpRepository.$batchSaveClusters` actually atomic in a separate PR.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -224,7 +224,12 @@ class BitcoinRoutes {
     } else {
       let cpfpInfo;
       if (config.DATABASE.ENABLED) {
-        cpfpInfo = await transactionRepository.$getCpfpInfo(req.params.txId);
+        try {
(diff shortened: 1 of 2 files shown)


wiz (approved):
Tested ACK @ 9ff5ce0
Confirmed fixes the crash bug in production

PR #3887, opened 2023-06-22
Atomic CPFP database operations

mononaut:
_(builds on #3886)_

This PR:
 - Extends the DB class to support atomic database transactions via a new `$atomicQuery` function
 - Makes `CpfpRepository.$batchSaveClusters` atomic using that new function
 - Removes unused `CpfpRepository.$saveCluster` function
 - Adds a try/catch to `blocks.$saveCpfp` to gracefully handle any exceptions from `$batchSaveClusters`.


softsimon (approved):
tested ACK @ v2.6.0-dev [[f80257c5](https://github.com/mempool/mempool/commit/f80257c5)]

PR #3889, opened 2023-06-24
Rust GBT

mononaut:
This PR explores re-implementing the getBlockTemplate algorithm in Rust for use as a native Node addon via ~~[Neon](https://neon-bindings.com/)~~ [NAPI_RS](https://napi.rs/).

The Rust module is more or less a direct port of the typescript implementation in `tx-selection-worker`, edited for style and mangled until the borrow checker stopped complaining.

The PR adds an extra step to the npm build command to compile the Rust module first, so
`npm run build && npm run start-production` should just work if rust is installed and available.

So far, the Rust implementation seems to speed up `updateBlockTemplates` by a consistent ~~30-50%~~ 50-75%, and `makeBlockTemplates` possibly even more.


junderw:
Please add the following to `.vscode/settings.json`

```diff
diff --git a/.vscode/settings.json b/.vscode/settings.json
index 69279118..8ff18083 100644
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,8 @@
 {
   "editor.tabSize": 2,
   "typescript.preferences.importModuleSpecifier": "relative",
-  "typescript.tsdk": "./backend/node_modules/typescript/lib"
+  "typescript.tsdk": "./backend/node_modules/typescript/lib",
+  "rust-analyzer.linkedProjects": [
+    "./backend/rust-gbt/Cargo.toml"
+  ]
 }
\ No newline at end of file
```

junderw:
I created #3890 which is pointed at this feature branch. Please review and merge when possible.

backend/rust-gbt/src/gbt.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+/*
+* Build projected mempool blocks using an approximation of the transaction selection algorithm from Bitcoin Core
+* (see BlockAssembler in https://github.com/bitcoin/bitcoin/blob/master/src/node/miner.cpp)
+* Ported from https://github.com/mempool/mempool/blob/master/backend/src/api/tx-selection-worker.ts
+*/
+fn make_block_templates(
+    mempool: &mut HashMap<u32, ThreadTransaction>,
+) -> Option<(Vec<Vec<u32>>, Vec<(u32, f64)>, Vec<Vec<u32>>)> {
```
junderw, line 52: Why does gbt function exist? Just rename make_block_template to gbt.

```suggestion
/*
* Build projected mempool blocks using an approximation of the transaction selection algorithm from Bitcoin Core
* (see BlockAssembler in https://github.com/bitcoin/bitcoin/blob/master/src/node/miner.cpp)
* Ported from https://github.com/mempool/mempool/blob/master/backend/src/api/tx-selection-worker.ts
*/
pub fn gbt(
    mempool: &mut HashMap<u32, ThreadTransaction>,
) -> Option<(Vec<Vec<u32>>, Vec<(u32, f64)>, Vec<Vec<u32>>)> {
```

mononaut: yup, good point. I think there was some data initialization logic in there originally that evaporated during a refactor

junderw:
I updated my PR against this PR with a migration to napi-rs.

For now I have mimicked the callback structure you made originally, but napi-rs has a tokio integration which allows you to return a Promise directly from the exposed functions, I'll look into it later today maybe.

junderw:
OK, I think my PR against this PR is mostly complete.

Please review and "Squash & Merge" when you can.

Then I think we can review/test this PR and merge it to master.

The next step (a separate PR) should be setting up some sort of benchmark testing. Then it will be easier to tweak the Rust to see how it affects performance. (Sometimes just tweaking compiler settings can get some big gains.)

mononaut:
Added a config flag for the Rust GBT feature (default off, but enabled in production) so that we have an escape hatch in case we need to fall back to the tried-and-tested typescript implementation for any reason.

backend/src/api/mempool-blocks.ts
```diff
@@ -321,6 +327,96 @@ class MempoolBlocks {
...
+
+  public async $oneOffRustBlockTemplates(newMempool: { [txid: string]: MempoolTransactionExtended }): Promise<MempoolBlockWithTransactions[]> {
+    return this.$rustMakeBlockTemplates(newMempool, false);
+  }
+
+  public async $rustUpdateBlockTemplates(newMempool: { [txid: string]: MempoolTransactionExtended }, added: MempoolTransactionExtended[], removed: MempoolTransactionExtended[]): Promise<void> {
+    // sanity check to avoid approaching uint32 uid overflow
+    if (this.nextUid > 4_000_000_000) {
+      this.resetRustGbt();
+    }
```
junderw, line 380: Why not `4_294_967_295`? (Exactly u32::MAX)

junderw:
https://github.com/mempool/mempool/pull/3893

Forgot to use the hasher for HashSet.

backend/src/api/mempool-blocks.ts
mononaut: just to leave a margin of error for the added transactions (obviously there won't ever be 300m new transactions, but if the threshold isn't exactly u32::MAX it might as well be a round number)

I suppose

```
if (this.nextUid + added.length > u32::MAX) {
  this.resetRustGbt();
}
```
would be convey the intent more clearly

backend/rust-gbt/README.md
```diff
@@ -0,0 +1,121 @@
+# gbt
+
+**gbt:** rust implementation of the getBlockTemplate algorithm
+
+This project was bootstrapped by [create-neon](https://www.npmjs.com/package/create-neon).
+
+## Installing gbt
+
+Installing gbt requires a [supported version of Node and Rust](https://github.com/neon-bindings/neon#platform-support).
```
junderw, line 9: hehe, oops. The README is still neon-based.

backend/src/api/mempool-blocks.ts
```diff
@@ -219,15 +224,16 @@ class MempoolBlocks {
     const strippedMempool: Map<number, CompactThreadTransaction> = new Map();
     Object.values(newMempool).forEach(entry => {
       if (entry.uid != null) {
```
junderw, line 226: `!==`

backend/src/api/mempool-blocks.ts
```diff
@@ -219,15 +224,16 @@ class MempoolBlocks {
...
-        strippedMempool.set(entry.uid, {
+        const stripped = {
           uid: entry.uid,
           fee: entry.fee,
           weight: (entry.adjustedVsize * 4),
           sigops: entry.sigops,
           feePerVsize: entry.adjustedFeePerVsize || entry.feePerVsize,
           effectiveFeePerVsize: entry.effectiveFeePerVsize || entry.adjustedFeePerVsize || entry.feePerVsize,
           inputs: entry.vin.map(v => this.getUid(newMempool[v.txid])).filter(uid => uid != null) as number[],
```
junderw, line 234: `!==`

backend/src/api/mempool-blocks.ts
```diff
@@ -279,7 +287,7 @@ class MempoolBlocks {
     const start = Date.now();
 
     for (const tx of Object.values(added)) {
-      this.setUid(tx);
+      this.setUid(tx, true);
     }
     const removedUids = removed.map(tx => this.getUid(tx)).filter(uid => uid != null) as number[];
```
junderw, line 292: `!==`

backend/src/api/mempool-blocks.ts
```diff
@@ -321,6 +329,96 @@ class MempoolBlocks {
...
+      await this.$rustMakeBlockTemplates(newMempool, true);
+      return;
+    }
+
+    const start = Date.now();
+
+    for (const tx of Object.values(added)) {
+      this.setUid(tx, true);
+    }
+    const removedUids = removed.map(tx => this.getUid(tx)).filter(uid => uid != null) as number[];
```
junderw, line 394: `!==`

backend/src/api/mempool-blocks.ts
```diff
@@ -496,6 +598,96 @@ class MempoolBlocks {
...
+    }
+    return { blocks: convertedBlocks, rates: convertedRates, clusters: convertedClusters } as { blocks: string[][], rates: { [root: string]: number }, clusters: { [root: string]: string[] }};
+  }
+
+  private mempoolToArrayBuffer(txs: MempoolTransactionExtended[], mempool: { [txid: string]: MempoolTransactionExtended }): ArrayBuffer {
+    let len = 4;
+    const inputs: { [uid: number]: number[] } = {};
+    let validCount = 0;
+    for (const tx of txs) {
+      if (tx.uid != null) {
```
junderw, line 649: `!==`

backend/src/api/mempool-blocks.ts
```diff
@@ -496,6 +598,96 @@ class MempoolBlocks {
...
+        validCount++;
+        const txInputs = tx.vin.map(v => this.getUid(mempool[v.txid])).filter(uid => uid != null) as number[];
```
junderw, line 651: `!==`

backend/src/api/mempool-blocks.ts
```diff
@@ -496,6 +598,96 @@ class MempoolBlocks {
...
+        inputs[tx.uid] = txInputs;
+        len += (10 + txInputs.length) * 4;
+      }
+    }
+    const buf = new ArrayBuffer(len);
+    const view = new DataView(buf);
+    view.setUint32(0, validCount, false);
+    let offset = 4;
+    for (const tx of txs) {
+      if (tx.uid != null) {
```
junderw, line 661: `!==`

backend/src/api/websocket-handler.ts
```diff
@@ -624,7 +632,11 @@ class WebsocketHandler {
     }
 
     if (config.MEMPOOL.ADVANCED_GBT_MEMPOOL) {
-      await mempoolBlocks.$makeBlockTemplates(_memPool, true);
+      if (config.MEMPOOL.RUST_GBT) {
+        await mempoolBlocks.$rustUpdateBlockTemplates(_memPool, [], transactions);
+      } else {
+        await mempoolBlocks.$makeBlockTemplates(_memPool, true);
```
junderw, line 638: Why is this one the only one doing "make" when the others are "update"?

junderw (approved):
My comments are all nits. (1 Question though)

Also, needs a rebase.

LGTM!

utACK!

junderw:
I found a workaround for the Rust IDE plugin bugging out inside napi tagged impl blocks etc.

Here's the PR to fix it: https://github.com/mempool/mempool/pull/3897

There's a link to the issue with the workaround in the commit message.

junderw:
Rebased onto master.

mononaut:
pushed fixes for various bugs in the implementation of the gbt algorithm which could produce subtly (and not so subtly) wrong templates.

junderw:
Opened https://github.com/mempool/mempool/pull/3900 against this PR. Please merge when convenient.

mononaut:
In 9f580fa I dropped the ArrayBuffer serialization and instead pass the javascript array of transactions directly into the Rust functions.

There's no obvious difference in performance, but it does make the code much cleaner on both sides.

backend/rust-gbt/src/gbt.rs
```diff
@@ -35,10 +36,12 @@ impl PartialEq for TxPriority {
 impl Eq for TxPriority {}
 impl PartialOrd for TxPriority {
     fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        if self.score == other.score {
-            Some(self.uid.cmp(&other.uid))
-        } else {
+        if self.score != other.score {
             self.score.partial_cmp(&other.score)
+        } else if self.order != other.order {
+            Some(other.order.cmp(&self.order))
```
junderw, line 42: Double checking: order cmp value is flipped compared to others.

backend/rust-gbt/src/audit_transaction.rs
```diff
@@ -50,19 +51,24 @@ impl PartialEq for AuditTransaction {
...
-pub fn partial_cmp_uid_score(a: (u32, f64), b: (u32, f64)) -> Option<Ordering> {
+pub fn partial_cmp_uid_score(a: (u32, u32, f64), b: (u32, u32, f64)) -> Option<Ordering> {
     // If either score is NaN, this is false,
     // and partial_cmp will return None
-    if a.1 == b.1 {
-        Some(a.0.cmp(&b.0))
+    if a.2 != b.2 {
+        a.2.partial_cmp(&b.2)
+    } else if a.1 != b.1 {
+        Some(b.1.cmp(&a.1))
```
junderw, line 60: This is also flipped.

Maybe there should be a comment in both places about the flip.

(Also, the only reason I didn't use this fn in TxPriority was because the self and other were flipped. Maybe we can use this fn by flipping the inputs somehow, that way this is the only place we need to comment.)

backend/rust-gbt/src/gbt.rs
mononaut: yep, I should have added a comment - transactions are ordered by *descending* score but *ascending* txid.

backend/rust-gbt/src/audit_transaction.rs
mononaut: good point, it would be nice to consolidate that logic into a single ordering

junderw (commented):
Avoiding regex is about 2.4x faster.

That said, on my local laptop the difference for 500k runs is about 100ms, which is not too bad. (175ms vs. 75ms)

junderw:
Looks like the Vec isn't that different from the HashMap performance-wise.

softsimon (approved):
tested ACK @ [390c4a7]

Time to calculate the templates is reduced by -80%.

The mempool blocks order looks fine and the audits are in good health.

`Jul  5 15:53:30 [99330] DEBUG: RUST updateBlockTemplates completed in 0.063 seconds`

junderw:
Moved the napi cli from devDependencies to dependencies.

Otherwise the production tests (and `npm ci --omit=dev` in general) does not work due to lack of the napi command.

junderw:
TODO: Add documentation (somewhere) that explains that `rustc` and `cargo` version 1.70.x need to be installed, OR rustup (which will automatically install 1.70.x for them if they don't have it.

junderw:
TODO: Fix docker/backend to install rust in the builder step so that npm install doesn't fail.

junderw:
Should be good to merge, I think. docker packaging (and the cleanup scripts in package.json) I think was the last thing we needed to account for...

Let me know if you think of anything else.

wiz (commented):
Testing begin @ f8e910f
Had to upgrade rust to v1.70 for it to build but seems to work well so far with both configurations disabled/enabled. Will test for a few days...

wiz (approved):
Tested ACK @ f8e910f

Ran well for a day on staging server, no issues, going to merge 🚀

PR #3890, opened 2023-06-24
Fix: Various small fixes

junderw:
**Note: This PR is pointed at #3889 and not the master branch.**

Just a few of the pedantic level lints for Clippy.

A lot of the `as` conversions should also be brushed up. Maybe tomorrow.

Edit 2: Instead of just using a vscode setting, making the root of the repo into a cargo workspace means you can run all the cargo commands etc. from the root, plus it works for all editors.


mononaut (approved):
Tested ACK @ f4b0bda6

Runs smoothly, produces templates that match the prior implementation, and the interface on the typescript side is great.

Performance looks like another solid 30-50% improvement over my original proof-of-concept, which means something like 50-75% faster than the old typescript thread worker.

PR #3892, opened 2023-06-25
Liquidity > Capacity

lukechilds:
On the main Lightning view these two views are titled as "Liquidity Ranking" and "Connectivity Ranking":

Screenshot 2023-06-25 at 14 03 16

However on the individual pages they are both titled "Liquidity Ranking":

Screenshot 2023-06-25 at 14 03 28

Screenshot 2023-06-25 at 14 03 32

This PR brings the listing view title inline with the main Lightning view.


softsimon:
Thanks for finding and reporting the wrong title on the "Connectivity page".

I am not sure about the change to Capacity though because it would mismatch the dashboard widget. Instead I made this PR to correct the title: https://github.com/mempool/mempool/pull/3952

wiz:
Thanks Luke but replaced by #3952

lukechilds:
🫡

PR #3893, opened 2023-06-26
Use U32HasherState for HashSet

junderw:
Forgot to use it for HashSet too.


mononaut (approved):
Tested ACK @ 724d4b39

PR #3897, opened 2023-06-26
Fix: napi macro breaks rust-analyzer

junderw:
See issue: https://github.com/napi-rs/napi-rs/issues/944#issuecomment-1013002760

This will ignore expanding the napi macro for rust-analyzer, letting auto-complete work inside the napi impl blocks.


junderw:
Also fixed the README

PR #3906, opened 2023-06-27
Lowering mempool loop break limit

softsimon:
This PR changes 2 things:

* Lowering the break limit from 10s to 5s that the mempool loop is allowed to run before it breaks and continues
* Don't output any mempool loop progress when the mempool "in sync" status is true.

This hopefully fixes slowness in the mempool loop when a lot of new transactions comes in. But also not sending 99% progress to the frontend causing the frontend to go into "loading spinner" mode, because we are not out of sync!

### DIFF
--- a/backend/src/api/mempool.ts
+++ b/backend/src/api/mempool.ts
@@ -156,6 +156,7 @@ class Mempool {
       }
     };
 
+    let intervalTimer = Date.now();
     for (const txid of transactions) {
       if (!this.mempoolCache[txid]) {
         try {
--- a/backend/src/api/mempool.ts
+++ b/backend/src/api/mempool.ts
@@ -156,7 +156,6 @@ class Mempool {
       }
     };
 
-    let loggerTimer = new Date().getTime() / 1000;
     for (const txid of transactions) {
       if (!this.mempoolCache[txid]) {
         try {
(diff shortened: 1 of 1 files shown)


backend/src/api/mempool.ts
```diff
@@ -156,7 +156,6 @@ class Mempool {
       }
     };
```
mononaut, line 158:
```suggestion

    let intervalTimer = Date.now();
```

backend/src/api/mempool.ts
```diff
@@ -179,18 +178,19 @@ class Mempool {
           logger.debug(`Error finding transaction '${txid}' in the mempool: ` + (e instanceof Error ? e.message : e));
         }
       }
-      const elapsedSeconds = Math.round((new Date().getTime() / 1000) - loggerTimer);
-      if (elapsedSeconds > 4) {
+
+      if (new Date().getTime() - start > 5_000) {
         const progress = (currentMempoolSize + newTransactions.length) / transactions.length * 100;
         logger.debug(`Mempool is synchronizing. Processed ${newTransactions.length}/${diff} txs (${Math.round(progress)}%)`);
```
mononaut, line 184:
```suggestion
      if (Date.now() - intervalTimer > 5_000) {
```

backend/src/api/mempool.ts
```diff
@@ -179,18 +178,19 @@ class Mempool {
...
-        break;
+        
+        if (this.inSync) {
+          // Break and restart mempool loop if we spend too much time processing
+          // new transactions that may lead to falling behind on block height
+          logger.debug('Breaking mempool loop because the 5s time limit exceeded.');
+          break;
+        } else {
+          loadingIndicators.setProgress('mempool', progress);
+        }
```
mononaut, line 193:
```suggestion
        if (this.inSync) {
          // Break and restart mempool loop if we spend too much time processing
          // new transactions that may lead to falling behind on block height
          logger.debug('Breaking mempool loop because the 5s time limit exceeded.');
          break;
        } else {
          const progress = (currentMempoolSize + newTransactions.length) / transactions.length * 100;
          logger.debug(`Mempool is synchronizing. Processed ${newTransactions.length}/${diff} txs (${Math.round(progress)}%)`);
          loadingIndicators.setProgress('mempool', progress);
          intervalTimer = Date.now()
        }
```

mononaut (changes requested):
Tested @ 31ae382f

I confirmed this fixes the loading spinner issue in the frontend, but now it spams the debug logs with
> "Mempool is synchronizing. Processed x/y txs..."

for every transaction after the first 5 seconds during the initial sync.

mononaut (approved):

mononaut:
Tested ACK @ [a567a28c]

PR #3908, opened 2023-06-27
Rust GBT test

mononaut:
Adds a simple unit test for the rust gbt implementation, and tweaks the algorithm a bit so that the test passes.

The main changes are:
 - ancestor scores now represent fee rates calculated from integer adjusted vsizes.
   - previously we used exact vsize (i.e. weight / 4), but this is how Bitcoin Core handles fee rates.
   - these scores are *only* used for transaction selection, fee rates for display in the frontend etc are still calculated using exact vsize.
 - the integer adjusted vsize is now used to test for package acceptance, instead of actual weight, to match [a quirk in Core's implementation](https://github.com/bitcoin/bitcoin/blob/7ee41217b3b3fe4d8b7eb4fd1d4577b9b33d466d/src/node/miner.cpp#L198-L208).
 - weight usage is now counted using actual weights instead of `adjusted_vsize * 4`.

### Methodology

I took a mempool snapshot from an instance of Bitcoin Core running with maxmempool=30, and generated a target template using
```
bitcoin-cli getblocktemplate '{"rules": ["segwit"]}' | jq '.transactions | map({txid,weight})'
```

From the mempool snapshot, I generated:
 - An ArrayBuffer using `MempoolBlocks.mempoolToArrayBuffer` to form the input to `rustGbt.make`
 - A map of uids to txids
 
These uids were assigned in ascending order of numerical txid value, so that ties are broken in the same direction as Bitcoin Core's gbt implementation, which allows us to exactly reproduce Core's results.


backend/rust-gbt/src/audit_transaction.rs
```diff
@@ -14,6 +14,7 @@ pub struct AuditTransaction {
     pub uid: u32,
     pub fee: u64,
     pub weight: u32,
+    pub vsize: u32,
```
junderw, line 17: Perhaps this should be a different name...

Later on we do `(4 * next_tx.ancestor_vsize())` which to the untrained eye would look like an "oh, I'll just replace it with weight, it's the same thing... when it's not.

It's the MAX of vsize or sigops * 5.

`sigop_adjusted_vsize` maybe?

mononaut: yes, very good point

junderw:
I made a PR against this (but it can be merged into the main PR after this PR is merged... whatever you feel like)

#3909

PR #3909, opened 2023-06-28
Increased performance

junderw:
This almost halves the time of the gbt function. (Though the gbt function itself is surprisingly a small portion of the total time including TypeScript.)

The peak memory usage was 46MB, even with this really large initial capacity for a bunch of Vecs and Queues.

I was also shocked that converting the Vec for mempool_stack twice (collecting both times and re-allocating) actually was faster.

I guess it makes sense because we moved the hashmap lookups to an O(n) operation instead of an O(n log n) (worst case) operation, and the reallocation only happens 2 times (O(1))

I was able to measure this thanks to the test case. :+1:

### DIFF
--- a/backend/rust-gbt/src/audit_transaction.rs
+++ b/backend/rust-gbt/src/audit_transaction.rs
@@ -49,15 +49,20 @@ impl PartialEq for AuditTransaction {
 
 impl Eq for AuditTransaction {}
 
+#[inline]
+pub fn partial_cmp_uid_score(a: (u32, f64), b: (u32, f64)) -> Option<Ordering> {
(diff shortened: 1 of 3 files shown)


mononaut (approved):
Tested ACK @ 649d9e17

I tried with both the jest test case (2x speedup) and a larger ~280k transaction mempool (1.3x speedup).

I wonder if it's worth differentiating the starting capacities of the different Vecs/Queues/HashMaps?

e.g. we know exactly how large the `mempool_stack` and `audit_pool` need to be at allocation time, and the `transactions` Vec only needs to fit 4MWU of transactions at a time until the final unbounded block.

PR #3916, opened 2023-06-29
Fix transaction component ETA

### LINKED ISSUE #3899: ETA might go out of sync
I can reproduce this on master, by just jumping between different transactions by using the search bar or browser navigation back and forth. Suddenly the ETA is 1 minute out of sync.

Screenshot 2023-06-26 at 21 08 49

This happens sometimes, I believe it's related but I don't know how to reproduce.

Screenshot 2023-06-26 at 19 43 41

mononaut:
This PR should fix #3899, which seems consist of two related issues:

 - The relative time components for the mempool blocks ETAs would not update when the difficulty adjustment changed, while the transaction ETA did, causing the two value to drift out of sync.
 - A regression of https://github.com/mempool/mempool/pull/3714#issuecomment-1533660715 introduced in #3673, where the "now" variable which forms the baseline of the relative time calculation did not always get updated properly, causing the ETA to slip backwards.


wiz:
@softsimon any reason not to merge this? it seems quite high priority

wiz (approved):
Tested ACK @ v2.6.0-dev [d9ed02a03]

Issue #2269, opened 2022-08-10
Smoothed graphs


ajtowns:
Would be nice if the vbytes/second graphs at https://mempool.space/graphs/mempool#3y had a moving average to smooth the daily (or weekly?) fluctuations out.

pfoytik:
I took at stab at this and submitted a pull request: 
smoothed graphs #3920

pfoytik:
I took another stab, this time for the vbytes/second chart :) see this pull request: #3923

mononaut:
resolved by #4371

PR #3921, opened 2023-07-03
smoothed graph, added moving average to vbytes chart

pfoytik:
This pull request takes a stab at #2269 , it creates a moving average line over top of the Mempool by vBytes chart to show a smoothed curve in addition to the data plots.

image

The moving average line is set to be white as it stands out best among the default colors of the chart and dark theme. The moving average window size is set to 10 as a default, but is a variable MAWindowsSize at the top of mempool-graph.components.ts. The MA is currently displayed over all time frames, but this could be adjusted to only show for the larger time frames if needed. We could also add the ability for users to change the windowsize for the moving average (smaller windows size will reflect more recent movement and larger windowsize will smooth out the MA more).


pfoytik:
@nymkappa pointed me to the correct chart (vbytes/sec) that the MA was supposed to be for. I'm looking into it and will make it a separate pull request. Feel free to close this one if needed as its the wrong chart for the issue.
Thanks!

PR #3923, opened 2023-07-03
added moving average to Transactions vBytes per second (vB/s)

pfoytik:
This pull request takes _another_ stab at #2269 it creates a moving average line over top of the vBytes/sec chart to show a smoothed curve in addition to the data plots.

image

The moving average line is set to be white as it stands out best among the default colors of the chart and dark theme. The moving average window size is set to 100 as a default, but is a variable MAWindowsSize at the top of incoming-transactions-graph.components.ts. The MA is currently displayed over all time frames, but this could be adjusted to only show for the larger time frames if needed. We could also add the ability for users to change the windowsize for the moving average (smaller windows size will reflect more recent movement and larger windowsize will smooth out the MA more).

### DIFF
--- a/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
+++ b/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
@@ -37,6 +37,7 @@ export class IncomingTransactionsGraphComponent implements OnInit, OnChanges, On
   };
   windowPreference: string;
   chartInstance: any = undefined;
+  MA: number[][] = [];
   weightMode: boolean = false;
   rateUnitSub: Subscription;
 
@@ -62,6 +63,7 @@ export class IncomingTransactionsGraphComponent implements OnInit, OnChanges, On
       return;
     }
     this.windowPreference = this.windowPreferenceOverride ? this.windowPreferenceOverride : this.storageService.getValue('graphWindowPreference');
+    this.MA = this.calculateMA(this.data.series[0]);
     this.mountChart();
   }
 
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
```diff
@@ -72,7 +75,96 @@ export class IncomingTransactionsGraphComponent implements OnInit, OnChanges, On
...
+    //calculate the moving average
+    for (i = 0; i < len; i++) {
+      sum += this.data.series[0][i][1];
+      if (i >= this.MAWindowSize) {
+        sum -= this.data.series[0][i - this.MAWindowSize][1];
+        ma[i][1] = sum / this.MAWindowSize;
+      } else {
+        ma[i][1] = sum / (i + 1);
+      }
+      ma[i][0] = this.data.series[0][i][0];
```
mononaut, line 99: this section might be easier to read if we assign `this.data.series[0]` to a convenience variable at the top of the function (or even pass it in as an argument like `calculateMA(series: number[][])`)

mononaut (changes requested):
Hey, sorry for the delay in reviewing this.

The code looks good, apart from a few nits I've commented on.

Some other more general suggestions:

1) The fixed moving window size works better for some graph timespans than others, for example the 2hr chart only has ~120 data points in total. Can we set the size of the window dynamically depending on the timespan?

2) The line gets less smooth towards the beginning of the chart. I'd prefer to plot the line only where we have a full window of data to average over. (See the Hashrate & Difficulty chart for an example).

3) Would the chart be easier to read as a centered moving average, instead of a trailing average?

frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
pfoytik: Love the comments, thanks mononaut! I agree with the feedback, I'll make the changes and update! 

I was even thinking of adding a window selection option for the user so they could choose how much of a smooth MA they would like to see, but still have some presets for the different time frames like you suggest. 

The center moving average definitely makes more sense for visualization (trailing moving average is more for forecasting). I'll make that change too!

pfoytik:
> Hey, sorry for the delay in reviewing this.
> 
> The code looks good, apart from a few nits I've commented on.
> 
> Some other more general suggestions:
> 
> 1. The fixed moving window size works better for some graph timespans than others, for example the 2hr chart only has ~120 data points in total. Can we set the size of the window dynamically depending on the timespan?
> 2. The line gets less smooth towards the beginning of the chart. I'd prefer to plot the line only where we have a full window of data to average over. (See the Hashrate & Difficulty chart for an example).
> 3. Would the chart be easier to read as a centered moving average, instead of a trailing average?

Updated code based on code recommendations, also made a couple of other corrections. 

I modified the moving average window to be 5% of the points, based on testing this looked pretty good. I'll look into a toggle or text box UI selection for the user in the future. For now I think the MA window as 5% is a pretty solid default that we can use for now.

I modified the moving average to be a center moving average instead of trailing. It now plots the line only when there is a full window of data to calculate the average. Here are a few images to show the MA window size based on data length and the centered MA:
2 hour window
image

24 hour window
image

1 week window
image

mononaut:
update looks good, thanks!

can you revert the changes to `.gitignore` and `.vscode`, and then squash the rest of the PR down to one commit please?

mononaut (approved):
Tested ACK @ 76621464

Issue #1515, opened 2022-04-10
Handle block reorgs


softsimon:
* The internal backend latest block array need to handle block reorgs
* The frontend needs to receive a new list of blocks or correct after a reorg
* Transaction tracking should handle the possibility of going from confirmed back to unconfirmed

antonilol:
A transaction that was put in a block (and was valid according to _consensus_ rules) won't be put back in the mempool if it doesn't conform to _relay_ rules. (e.g. TXs below relay fee, non standard TXs, coinbase)

Transactions can also be removed by double spending in a fork

dambaev:
about frontend: maybe, we can just send a "reload" websocket message to all the clients to force frontend to refresh itself?
can you outlook the downsides of such way of dealing with reorgs for frontend?

antonilol:
why not display the fork, add the orphan block(s) above or below the blockchain view to browse them and a diff of transactions would also be cool between two blocks with the same height

example with bitcoin-cli 
get fork height and hash
```
$ bitcoin-cli getchaintips
[
  {
    "height": 750574,
    "hash": "000000000000000000084841119d77a47aafc1765feedfa32f44016cdb9cd18e",
    "branchlen": 0,
    "status": "active"
  },
  {                                                                                   most recent fork
    "height": 733430,     <---------------------------------------------------------- height
    "hash": "00000000000000000006ead1cff09f279f7beb31a7290c2a603b0776d98dc334",   <-- hash
    "branchlen": 1,
    "status": "valid-headers"
  },
...
```
get hash of the block with same height in the active chain
```
$ bitcoin-cli getblockhash 733430
000000000000000000009b84ee4ca1b965618de1a2e056a51d994ff070791d28
```

write blocks to files
```
$ bitcoin-cli getblock 00000000000000000006ead1cff09f279f7beb31a7290c2a603b0776d98dc334 > /tmp/fork.json
$ bitcoin-cli getblock 000000000000000000009b84ee4ca1b965618de1a2e056a51d994ff070791d28 > /tmp/active.json
```

diff
```diff
$ diff -u --color /tmp/{active,fork}.json
--- /tmp/active.json	2022-08-22 12:41:13.529726856 +0200
+++ /tmp/fork.json	2022-08-22 12:41:00.919346260 +0200
@@ -1,24 +1,23 @@
 {
-  "hash": "000000000000000000009b84ee4ca1b965618de1a2e056a51d994ff070791d28",
-  "confirmations": 17145,
[...]

PR #3932, opened 2023-07-08
Stale blocks

### LINKED ISSUE #3559: Don't automatically redirect from reorgd block page
There are two blocks at height 782129:

000000000000000000036807803e4ccbbfac2191d6b42c519decca948ceffb2e
000000000000000000036f461ab63c78f08401d3907a67fd2237166d8a373193\

Searching for either causes a redirect to https://mempool.space/en/block/000000000000000000036f461ab63c78f08401d3907a67fd2237166d8a373193 (i.e. the longest chain)

It would be better if the original block page was displayed, with a message that it was reorged out and a link to the longest chain. It's nice to be able to visually inspect any differences.

mononaut:
Fixes #3559. Partially addresses #1515.

This PR fixes some issues related to stale blocks and reorgs:
  1. After a reorg, the backend block cache is now updated to the canonical chain, so that we serve the correct list of recent blocks to new clients.
  2. Serve the correct block
      - Previously, requests for stale blocks would silently return the canonical block at the same height without changing the url etc.
      - Now, we always return the actual requested block for a given block hash.
  3. Now that it's possible to view stale blocks, we display an alert banner to make it clear that the block is not part of the main chain:
  
Screenshot 2023-07-08 at 12 36 20 AM

---

For future PRs:
 - [x] Replace client-side cached blocks after a reorg
 - [ ] Link to known stale blocks from main chain
 - [ ] Visualize difference between stale + canonical blocks

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api.ts
+++ b/backend/src/api/bitcoin/bitcoin-api.ts
@@ -29,6 +29,7 @@ class BitcoinApi implements AbstractBitcoinApi {
       weight: block.weight,
       previousblockhash: block.previousblockhash,
       mediantime: block.mediantime,
+      stale: block.confirmations === -1,
     };
   }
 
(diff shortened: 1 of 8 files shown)


softsimon (commented):
tested ACK @ [6942a6fd]

softsimon (approved):

PR #3936, opened 2023-07-09
Fix random lint issues

knorrium:
These should silence some of the warnings we see in PRs

### DIFF
--- a/backend/src/api/audit.ts
+++ b/backend/src/api/audit.ts
@@ -6,7 +6,7 @@ const PROPAGATION_MARGIN = 180; // in seconds, time since a transaction is first
 
 class Audit {
   auditBlock(transactions: TransactionExtended[], projectedBlocks: MempoolBlockWithTransactions[], mempool: { [txId: string]: TransactionExtended })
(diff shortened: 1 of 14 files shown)


backend/src/api/audit.ts
```diff
@@ -71,7 +71,7 @@ class Audit {
           if (fits) {
             lastFeeRate = Math.min(lastFeeRate, tx.effectiveFeePerVsize);
           }
-          if (tx.firstSeen == null || (now - (tx?.firstSeen || 0)) > PROPAGATION_MARGIN) {
+          if (tx.firstSeen === null || (now - (tx?.firstSeen || 0)) > PROPAGATION_MARGIN) {
```
softsimon, line 74: Afraid this and that other case might be `undefined` here, changing to `!==` will break it .. maybe  @mononaut  can confirm.

mononaut: yeah, non-strict null equality checks are usually deliberate.

can we change the `eqeqeq` linting rule to `["error", "always", {"null": "ignore"}]`?

softsimon (changes requested):
Comments

softsimon:
Closing due to low activity

PR #3945, opened 2023-07-11
Fix mined rbf conflict prevention

mononaut:
Fixes a bug where a mempool transaction could be identified as replaced by multiple different mined transactions, producing nonsensical RBF trees, e.g:

Screenshot 2023-07-11 at 11 45 13 AM

This PR properly deletes all txos of replaced transactions from the `spendMap`, to guarantee that each transaction can only be replaced once.


softsimon (approved):
partly tested ACK @ [f871300bf]

Didn't encounter a similar structure to compare the result.

PR #3947, opened 2023-07-11
fix price updater loop on testnet/signet

mononaut:


wiz (approved):
Tested ACK @ v2.6.0-dev [[3ece5acd5](https://github.com/mempool/mempool/commit/3ece5acd5)]

PR #3948, opened 2023-07-11
show "loading" message while checking for cached txs

mononaut:
On slow network connections, or when the servers are running slowly, it takes a while to check for replaced transactions in the backend cache.

Currently we show the "Transaction not found. Waiting for it to appear in the mempool..." message during this time, which can be confusing.

This PR displays "Loading transaction" while checking the backend cache, and then reverts to "Transaction not found..." only if the transaction is missing from both the mempool and the cache:

Screenshot 2023-07-11 at 4 08 01 PM
Screenshot 2023-07-11 at 4 08 33 PM

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -454,7 +454,7 @@
   <ng-template [ngIf]="error">
 
     <div class="text-center" *ngIf="loadingCachedTx; else waitingTemplate">
-      <h3 i18n="transaction.error.transaction-not-found">Loading transaction</h3>
+      <h3 i18n="transaction.error.loading-transaction">Loading transaction</h3>
       <div class="spinner-border text-light mt-2"></div>
     </div>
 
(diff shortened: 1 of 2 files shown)


frontend/src/app/components/transaction/transaction.component.html
```diff
@@ -453,12 +453,19 @@ <h2 i18n="transaction.details">Details</h2>
 
   <ng-template [ngIf]="error">
 
-    <div class="text-center" *ngIf="waitingForTransaction; else errorTemplate">
-      <h3 i18n="transaction.error.transaction-not-found">Transaction not found.</h3>
-      <h5 i18n="transaction.error.waiting-for-it-to-appear">Waiting for it to appear in the mempool...</h5>
+    <div class="text-center" *ngIf="loadingCachedTx; else waitingTemplate">
+      <h3 i18n="transaction.error.transaction-not-found">Loading transaction</h3>
```
softsimon, line 457: Bad i18n tag

softsimon (changes requested):
Just do the skeleton instead of the new loading text

softsimon (approved):
tested ACK @ [132923e7]

PR #3952, opened 2023-07-12
Connectivity ranking title fix

softsimon:
Fixed the bad header title and the browser title.

This is an alternative fix to https://github.com/mempool/mempool/pull/3892

Screenshot 2023-07-12 at 12 25 07


wiz (approved):
Tested ACK @ v2.6.0-dev [a6dc4fa38]

PR #3953, opened 2023-07-12
always show latest difficulty on hashrate chart

mononaut:
We currently skip difficulty datapoints on the hashrate and difficulty chart when there isn't a corresponding hashrate datapoint.

This means the chart doesn't show the latest difficulty until the hashrate is next indexed.

This PR includes trailing difficulty points so that the latest difficulty is always visible.

Before:
Screenshot 2023-07-12 at 12 59 41 PM

After:
Screenshot 2023-07-12 at 12 59 18 PM


wiz (approved):
Tested ACK @ v2.6.0-dev [4144746e1]

PR #3957, opened 2023-07-13
highlight & tag fullrbf replacements in RBF timeline

mononaut:
This PR extends the RBF Timelines component to more explicitly highlight full-rbf replacements:

Screenshot 2023-07-13 at 12 30 16 PM

Now the dark blue lines show replacements that all nodes accept, while the cyan lines indicate replacements that are only valid on nodes running `mempoolfullrbf=1`


softsimon (approved):
tested ACK @ v2.6.0-dev [[6a2a9bda0](https://github.com/mempool/mempool/commit/6a2a9bda0)]

PR #3959, opened 2023-07-13
Load more mempool transactions

softsimon:
This PR requires this patch https://github.com/mempool/electrs/pull/8 to `mempool/electrs` and make the address loading backwards incompatible with `blockstream/electrs` !

It allows us to load more unconfirmed mempool transactions and not just limit us to 25 before start loading onchain transactions.

TODO:
- [x] Make sure it works with `romanz/electrs` backend


mononaut (approved):
Tested ACK @ [a59a7fe2]

Successfully loads hundreds of unconfirmed address transactions:


https://github.com/mempool/mempool/assets/83316221/fadf5f3d-f502-4c2f-8e37-7d8707356b62

PR #3961, opened 2023-07-14
Fix difficulty chart off-by-one bug

mononaut:
Fixes a bug in #3953 which could cause the last datapoint in the chart to use the wrong timestamp.

Before:
Screenshot 2023-07-14 at 11 05 47 AM

After:

Screenshot 2023-07-14 at 11 05 37 AM


softsimon (approved):
tested ACK @ [8477600]

PR #3962, opened 2023-07-14
Switch "Latest blocks" to "Latest replacements"

### LINKED ISSUE #3954: Latest RBF transactions list
Replace Latest blocks with Latest RBF transactions (or if there's a better title) on the Mempool main dashboard.

Consider re-using the same FullRBF label tags to differentiate between RBFs and Full-RBF transactions.

**Rationale:**
The Mempool dashboard should only contain mempool related data and the mining dashboard should display blockchain/mining related so currently the Latest blocks on the mempool dashboard is out of place and also redundant.

mononaut:
Resolves #3954 by changing the "latest blocks" widget to "latest replacements", which shows the last 6 RBF replacements:

Screenshot 2023-07-14 at 12 24 58 PM

The title of the widget links to the (formerly secret) `/rbf` page.

### DIFF
--- a/backend/src/api/rbf-cache.ts
+++ b/backend/src/api/rbf-cache.ts
@@ -6,6 +6,7 @@ import { Common } from "./common";
 interface RbfTransaction extends TransactionStripped {
   rbf?: boolean;
   mined?: boolean;
+  fullRbf?: boolean;
 }
 
 interface RbfTree {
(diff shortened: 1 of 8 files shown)


softsimon (changes requested):
Customize the websocket data for the dashboard with minimal RBF data.

mononaut:
I've pushed a change to reduce the data sent for the latest replacements widget to the absolute minimum.

Now the server only sends the specific data displayed in the widget (i.e. the txid, status, and fee info for the last 6 replacements, instead of the last 25 full rbf trees).

It also only sends updates when clients first subscribe or when the underlying data changes, not on every mempool loop.

softsimon (approved):
tested ACK @ [fa48791c59]

Issue #3966, opened 2023-07-15
Price timestamp is not up todate


softsimon:
The price sent from the backend to frontend seems to be old or cached.

Current latest price from the websocket claims the timestamp to be very old:

1687514400
Fri Jun 23 2023 10:00:00 GMT+0000
( 1 Month old )

Locally my timestamp is 6 hours ago.

This affects the current value of bitcoin displayed.

Make sure the timestamp and price is updated more frequently to display current bitcoin value more correctly.

Perhaps the price is correct but the timestamp is wrong.

PR #3967, opened 2023-07-15
[price updater] update latestPrices timestamp before pushing to websocket

nymkappa:
Fixes https://github.com/mempool/mempool/issues/3966


softsimon (approved):
tested ACK @ [73d9b4ef]

PR #3974, opened 2023-07-16
Set missing websocket init data

mononaut:
Fixes a bug where the websocket init data could sometimes be set incompletely, leading to missing prices and backend version info.


wiz (approved):
Tested ACK @ v3.0.0-dev [565336df2]

PR #3978, opened 2023-07-17
unbork mining pool blocks list

mononaut:
Fixes an issue in the blocks list on the `/pool/:slug` page, where we were using the wrong block height to load more blocks.

| Before | After |
|-|-|
| Screenshot 2023-07-17 at 11 21 13 AM | Screenshot 2023-07-17 at 11 20 43 AM


softsimon (approved):
tested ACK @ [2c39e1e20]

PR #3979, opened 2023-07-17
get chain tip direct from Bitcoin Core to avoid race conditions

mononaut:
PR #3655 mitigated but did not fully solve a race condition where newly confirmed transactions could get removed from the mempool before the corresponding block is detected (leading to zero-score audits).

It seems that the same race condition also occurs within `mempool/electrs`, so this PR attempts to eliminate the issue by checking for new blocks directly from Bitcoin Core.


softsimon (approved):
tested ACK @ [7a059ba2]

PR #3981, opened 2023-07-17
Fix websocket null data for undefined rbfSummary

mononaut:
This PR fixes a bug where which occurs if `rbfSummary` hasn't been set in the websocket init data when a client starts a `track-rbf-summary` subscription, we would set an undefined value in the response and send the client invalid json.


softsimon (approved):
tested ACK @ [4309bfd51]

PR #3982, opened 2023-07-17
Fix mempool update poll delay

mononaut:
Run the main update loop every `config.MEMPOOL.POLL_RATE_MS` instead of delaying the loop for an additional `config.MEMPOOL.POLL_RATE_MS` every iteration.


softsimon (approved):
tested ACK @ [2c1b9b90]

PR #3987, opened 2023-07-18
Enable fullrbf in production frontend config

mononaut:
Adds `FULL_RBF_ENABLED` to the production `mempool-frontend-config.mainnet.json` config file, which enables this toggle on the RBF Replacements page:

Screenshot 2023-07-18 at 10 47 39 AM


softsimon (changes requested):
Just make the toggle always visible since we still can show fullrbf (when mined) on non-fullrbf.

softsimon:
Using https://github.com/mempool/mempool/pull/3989 instead

PR #3988, opened 2023-07-18
Fix clock horizontal scroll bug

mononaut:
This PR fixes a css bug which lets the clock scroll off the page horizontally:

https://github.com/mempool/mempool/assets/83316221/cb8cf72c-8d9d-4b38-bf42-d06a4893b4c2


softsimon (approved):
tested ACK @ [5e46176c]

PR #3989, opened 2023-07-18
Remove frontend FULL_RBF_ENABLED flag

mononaut:
An alternative to #3987, this removes the `FULL_RBF_ENABLED` frontend config setting entirely, and just always shows the toggle.


softsimon (approved):
tested ACK @ [17866f80]

PR #3996, opened 2023-07-19
Fix for 0 sat minimum fee

softsimon:
Alternative version of https://github.com/mempool/mempool/pull/1951


softsimon:
Tested and ACKEd

PR #3999, opened 2023-07-19
Fix fee handling on Liquid

mononaut:
Fixes a number of related bugs where, e.g. fees could sometimes show as ∞ sat/vb or throw errors on Liquid.


mononaut:
possibly fixed

wiz (approved):
Tested ACK @ v3.0.0-dev [7a6da07a6]

PR #4013, opened 2023-07-20
Fix some icon css color changes

softsimon:
After upgrading to Angular 16 some CSS bug revealed itself.

Before

Screenshot 2023-07-20 at 16 02 15

After this PR
Screenshot 2023-07-20 at 16 02 23


nymkappa (approved):
Tested ACK @ 41928695934878ee0eab0d13294ea672efc483fe

PR #4025, opened 2023-07-22
load mempool txs in bulk from esplora

mononaut:
This PR leverages a new bulk [`/mempool/txs` endpoint in mempool/electrs](https://github.com/mempool/electrs/pull/18) to synchronize the mempool faster on startup when the disk cache is unavailable.

Handling the extremely large JSON response requires a new dependency, JSONStream, which processes the response transaction-by-transaction instead of attempting to parse the entire array in one shot.

On an M1 mac this cuts the time to synchronize a 260k transaction mempool down from over 11 minutes to ~30 seconds.

Presumably there's some hard limit on how much data we can fetch and parse in a single request, so we may want to figure out how to break this into smaller batches.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
+++ b/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
@@ -3,6 +3,7 @@ import { IEsploraApi } from './esplora-api.interface';
 export interface AbstractBitcoinApi {
   $getRawMempool(): Promise<IEsploraApi.Transaction['txid'][]>;
   $getRawTransaction(txId: string, skipConversion?: boolean, addPrevout?: boolean, lazyPrevouts?: boolean): Promise<IEsploraApi.Transaction>;
+  $getMempoolTransactions(expectedCount: number);
   $getTransactionHex(txId: string): Promise<string>;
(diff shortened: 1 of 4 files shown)


mononaut:
Updated to use the new paginated `mempool/txs/:txid` endpoint from https://github.com/mempool/electrs/pull/18.

This eliminates the need for the JSONStream parser, should scale better, and actually speeds up synchronization times by another 20-30%.

softsimon (approved):
tested ACK @ [7091a67e]

Loaded the whole mempool into cache in 7 seconds.

wiz (changes requested):
Tested NACK @ 2cc53749a5110b92a0f6461c47326f5f9e9904a5

Works very well, but one small issue is that the mempool is set "in sync" too early and gives incorrect fee estimates and displays empty mempool blocks for a few seconds before displaying the correct data:

Screenshot 2023-07-24 at 14 49 54

Screenshot 2023-07-24 at 14 50 58

mononaut:
Rebased, fixed the in-sync status and added progress updates:


https://github.com/mempool/mempool/assets/83316221/8323466a-39d5-43e0-ab37-50c01b168a7f

PR #4027, opened 2023-07-22
Support P2PK address types

### LINKED ISSUE #4019: Support P2PK on the address page
Instead of displaying P2PK, actually print out the Pub Key as that was the "address" at that time.

Create a new URL: `/pubkey/<pubkey>` as opposed to the current `/address/<address>` but redirect the URL to the same Address Component.

In the Address Component, detect that we are coming from the '/pubkey/' url and now convert the pubkey to a script hash in order to look the address upp against the esplora `/api/scripthash/:hash` and `/api/scripthash/:hash/txs` 

The conversion algorithm is like the following in NodeJS, but I guess we can do it in the frontend as well(?), but if it requires an extra 3rd party library, it make sense to instead implement it straight into `mempool/electrs` as a new /pubkey/ endpoint.

```
Buffer.from('pubkey, 'hex');
crypto.createHash('sha256').update(str).digest('hex')
```

mononaut:
This PR resolves #4019 by adding support for P2PK pubkeys as a valid address type:

Screenshot 2023-07-22 at 5 54 34 PM

The address page detects if an address is a P2PK pubkey, and replaces requests to `/address/:address` and `/address/:address/txs` with requests to `/scripthash/:scripthash` and `/scripthash/:scripthash/txs` instead

### DIFF
--- a/backend/src/api/bitcoin/esplora-api.ts
+++ b/backend/src/api/bitcoin/esplora-api.ts
@@ -110,6 +110,14 @@ class ElectrsApi implements AbstractBitcoinApi {
     throw new Error('Method getAddressTransactions not implemented.');
   }
 
+  $getScriptHash(scripthash: string): Promise<IEsploraApi.ScriptHash> {
+    throw new Error('Method getAddress not implemented.');
+  }
+
+  $getScriptHashTransactions(scripthash: string, txId?: string): Promise<IEsploraApi.Transaction[]> {
+    throw new Error('Method getAddressTransactions not implemented.');
+  }
+
   $getAddressPrefix(prefix: string): string[] {
     throw new Error('Method not implemented.');
   }
--- a/backend/src/api/bitcoin/esplora-api.ts
+++ b/backend/src/api/bitcoin/esplora-api.ts
@@ -111,11 +111,11 @@ class ElectrsApi implements AbstractBitcoinApi {
   }
 
   $getScriptHash(scripthash: string): Promise<IEsploraApi.ScriptHash> {
-    throw new Error('Method getAddress not implemented.');
+    throw new Error('Method getScriptHash not implemented.');
   }
 
   $getScriptHashTransactions(scripthash: string, txId?: string): Promise<IEsploraApi.Transaction[]> {
-    throw new Error('Method getAddressTransactions not implemented.');
+    throw new Error('Method getScriptHashTransactions not implemented.');
   }
 
   $getAddressPrefix(prefix: string): string[] {
--- a/frontend/src/app/bitcoin.utils.ts
+++ b/frontend/src/app/bitcoin.utils.ts
@@ -281,3 +281,12 @@ export function isFeatureActive(network: string, height: number, feature: 'rbf' 
     return false;
   }
 }
+
+export async function calcScriptHash$(script: string): Promise<string> {
+  const buf = Uint8Array.from(script.match(/.{1,2}/g).map((byte) => parseInt(byte, 16)));
+  const hashBuffer = await crypto.subtle.digest('SHA-256', buf);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  return hashArray
+    .map((bytes) => bytes.toString(16).padStart(2, '0'))
+    .join('');
+}
--- a/frontend/src/app/bitcoin.utils.ts
+++ b/frontend/src/app/bitcoin.utils.ts
@@ -283,7 +283,10 @@ export function isFeatureActive(network: string, height: number, feature: 'rbf' 
 }
 
 export async function calcScriptHash$(script: string): Promise<string> {
-  const buf = Uint8Array.from(script.match(/.{1,2}/g).map((byte) => parseInt(byte, 16)));
+  if (!/^[0-9a-fA-F]*$/.test(script) || script.length % 2 !== 0) {
+    throw new Error('script is not a valid hex string');
+  }
+  const buf = Uint8Array.from(script.match(/.{2}/g).map((byte) => parseInt(byte, 16)));
   const hashBuffer = await crypto.subtle.digest('SHA-256', buf);
   const hashArray = Array.from(new Uint8Array(hashBuffer));
   return hashArray
--- a/frontend/src/app/components/address/address-preview.component.ts
+++ b/frontend/src/app/components/address/address-preview.component.ts
@@ -64,13 +64,15 @@ export class AddressPreviewComponent implements OnInit, OnDestroy {
           this.address = null;
           this.addressInfo = null;
           this.addressString = params.get('id') || '';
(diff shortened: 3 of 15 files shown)


backend/src/api/bitcoin/esplora-api.ts
```diff
@@ -110,6 +110,14 @@ class ElectrsApi implements AbstractBitcoinApi {
     throw new Error('Method getAddressTransactions not implemented.');
   }
 
+  $getScriptHash(scripthash: string): Promise<IEsploraApi.ScriptHash> {
+    throw new Error('Method getAddress not implemented.');
```
junderw, line 114: Error message fix

backend/src/api/bitcoin/esplora-api.ts
```diff
@@ -110,6 +110,14 @@ class ElectrsApi implements AbstractBitcoinApi {
...
+  }
+
+  $getScriptHashTransactions(scripthash: string, txId?: string): Promise<IEsploraApi.Transaction[]> {
+    throw new Error('Method getAddressTransactions not implemented.');
```
junderw, line 118: error message

frontend/src/app/bitcoin.utils.ts
```diff
@@ -281,3 +281,12 @@ export function isFeatureActive(network: string, height: number, feature: 'rbf'
     return false;
   }
 }
+
+export async function calcScriptHash$(script: string): Promise<string> {
+  const buf = Uint8Array.from(script.match(/.{1,2}/g).map((byte) => parseInt(byte, 16)));
```
junderw, line 286: Looping over match is fine, but it will produce some weird results if the input is weird.

```
> "aaf".match(/.{1,2}/g)
[ 'aa', 'f' ]

# even if we fix the regex
> "aaxxxxxxxxxxdd".match(/[0-9a-fA-F]{2}/g)
[ 'aa', 'dd' ]
```

frontend/src/app/components/address/address-preview.component.ts
```diff
@@ -64,13 +64,15 @@ export class AddressPreviewComponent implements OnInit, OnDestroy {
           this.address = null;
           this.addressInfo = null;
           this.addressString = params.get('id') || '';
-          if (/^[A-Z]{2,5}1[AC-HJ-NP-Z02-9]{8,100}$/.test(this.addressString)) {
+          if (/^[A-Z]{2,5}1[AC-HJ-NP-Z02-9]{8,100}|[A-F0-9]{130}$/.test(this.addressString)) {
```
junderw, line 67: The length is 130 or 66. There are some people who have spent to compressed p2pk.

To be specific it would be `(04[A-F0-9]{128}|(02|03)[A-F0-9]{64})`

We could just ignore compressed p2pk though.

frontend/src/app/components/address/address-preview.component.ts
```diff
@@ -64,13 +64,15 @@ export class AddressPreviewComponent implements OnInit, OnDestroy {
...
             this.addressString = this.addressString.toLowerCase();
           }
           this.seoService.setTitle($localize`:@@address.component.browser-title:Address: ${this.addressString}:INTERPOLATION:`);
 
-          return this.electrsApiService.getAddress$(this.addressString)
-            .pipe(
+          return (this.addressString.match(/[a-f0-9]{130}/)
```
junderw, line 72: See above

frontend/src/app/components/address/address.component.ts
```diff
@@ -72,7 +72,7 @@ export class AddressComponent implements OnInit, OnDestroy {
           this.addressInfo = null;
           document.body.scrollTo(0, 0);
           this.addressString = params.get('id') || '';
-          if (/^[A-Z]{2,5}1[AC-HJ-NP-Z02-9]{8,100}$/.test(this.addressString)) {
+          if (/^[A-Z]{2,5}1[AC-HJ-NP-Z02-9]{8,100}|[A-F0-9]{130}$/.test(this.addressString)) {
```
junderw, line 75: same as above

frontend/src/app/components/address/address.component.ts
```diff
@@ -83,8 +83,11 @@ export class AddressComponent implements OnInit, OnDestroy {
               .pipe(filter((state) => state === 2 && this.transactions && this.transactions.length > 0))
           )
           .pipe(
-            switchMap(() => this.electrsApiService.getAddress$(this.addressString)
-              .pipe(
+            switchMap(() => (
+              this.addressString.match(/[a-f0-9]{130}/)
```
junderw, line 87: same as above

frontend/src/app/components/address/address.component.ts
```diff
@@ -114,7 +117,9 @@ export class AddressComponent implements OnInit, OnDestroy {
           this.updateChainStats();
           this.isLoadingAddress = false;
           this.isLoadingTransactions = true;
-          return this.electrsApiService.getAddressTransactions$(address.address);
+          return address.is_pubkey
+              ? this.electrsApiService.getScriptHashTransactions$('41' + address.address + 'ac')
```
junderw, line 121: same as above

For compressed keys it would be `'21'` instead of 41

frontend/src/app/components/transactions-list/transactions-list.component.html
```diff
@@ -56,7 +56,9 @@
                       <span i18n="transactions-list.peg-in">Peg-in</span>
                     </ng-container>
                     <ng-container *ngSwitchCase="vin.prevout && vin.prevout.scriptpubkey_type === 'p2pk'">
-                      <span>P2PK</span>
+                      <span>P2PK <a class="address p2pk-address" [routerLink]="['/address/' | relativeUrl, vin.prevout.scriptpubkey.slice(2, 132)]" title="{{ vin.prevout.scriptpubkey.slice(2, 132) }}">
+                        <app-truncate [text]="vin.prevout.scriptpubkey.slice(2, 132)" [lastChars]="8"></app-truncate>
```
junderw, line 60: same as above

junderw:
Overall, awesome stuff. Just wondering about the support for compressed P2PK... should we bother. An overwhelming majority of P2PK are uncompressed, but compressed one's do exist.

junderw:
According to this, an overwhelming majority of P2PK are uncompressed.

https://transactionfee.info/charts/bitcoin-script-pubkey-compression/

1. When you see "Uncompressed (in output)" decrease sharply, that's the transition from P2PK to P2PKH (the pubkey moved to the input)
2. When you see "Compressed (in input)" increase is when wallets started using compressed keys in the P2PKH.
3. There are a few "Compressed (in output)" in 2014 onward, but most of that would be raw multisig (Counterparty and other metacoin protocols used dummy raw multisig keys for embedding data)

I think with this info it seems reasonable to only support uncompressed (65 byte) pubkeys.

softsimon (approved):
tested ACK @ [ae183210]

Issue #3810, opened 2023-05-26
Search bar autofocus should only apply in dashboards and only in desktop version


nymkappa:
### How to reproduce

- open https://mempool.space/block/00000000000000000005333afbed0873106f44169139786cddd9868136b6554f
- try to press the right/left arrows on the keyboard
- confirm you cannot go back/forward between blocks because the search bar is autofocused

Another related issue is that the search bar sometimes triggers the keyboard on android, which makes a bad UX when you open the website. So let's disable it on mobile all together.

### Solutions

Only autofocus the search bar if we are on a dashboard
OR
Only autofocus the search bar if we are not in /block

PR #4028, opened 2023-07-23
[search bar] only autofocus when in `/`, `/mining` and `/lightning`

nymkappa:
Proposal to fix https://github.com/mempool/mempool/issues/3810

### DIFF
--- a/frontend/src/app/components/search-form/search-form.component.html
+++ b/frontend/src/app/components/search-form/search-form.component.html
@@ -1,7 +1,7 @@
 <form [formGroup]="searchForm" (submit)="searchForm.valid && search()" novalidate>
   <div class="d-flex">
     <div class="search-box-container mr-2">
-      <input autofocus (focus)="focus$.next($any($event).target.value)" (click)="click$.next($any($event).target.value)" formControlName="searchText" type="text" class="form-control" i18n-placeholder="search-form.searchbar-placeholder" placeholder="Explore the full Bitcoin ecosystem">
+      <input #searchInput (focus)="focus$.next($any($event).target.value)" (click)="click$.next($any($event).target.value)" formControlName="searchText" type="text" class="form-control" i18n-placeholder="search-form.searchbar-placeholder" placeholder="Explore the full Bitcoin ecosystem">
       <app-search-results #searchResults [hidden]="dropdownHidden" [results]="typeAhead$ | async" (selectedResult)="selectedResult($event)"></app-search-results>
     </div>
     <div>
--- a/frontend/src/app/components/search-form/search-form.component.ts
+++ b/frontend/src/app/components/search-form/search-form.component.ts
@@ -65,13 +65,15 @@ export class SearchFormComponent implements OnInit {
     this.stateService.networkChanged$.subscribe((network) => this.network = network);
     
     this.router.events.subscribe((e: NavigationStart) => { // Reset search focus when changing page
-      if (e.type === EventType.NavigationStart) {
+      if (this.searchInput && e.type === EventType.NavigationStart) {
         this.searchInput.nativeElement.blur();
       }
     });
 
-    this.stateService.searchFocus$.subscribe(focus => {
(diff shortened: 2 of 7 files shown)


frontend/src/app/components/search-form/search-form.component.ts
```diff
@@ -56,7 +57,9 @@ export class SearchFormComponent implements OnInit {
     private apiService: ApiService,
     private relativeUrlPipe: RelativeUrlPipe,
     private elementRef: ElementRef,
-  ) { }
+  ) {
+    this.searchAutofocus = ['/', '/mining', '/lightning'].includes(this.router.url);
```
softsimon, line 61: This probably doesn't work with languages added `/fr/`

frontend/src/app/components/search-form/search-form.component.html
```diff
@@ -1,7 +1,7 @@
 <form [formGroup]="searchForm" (submit)="searchForm.valid && search()" novalidate>
   <div class="d-flex">
     <div class="search-box-container mr-2">
-      <input autofocus (focus)="focus$.next($any($event).target.value)" (click)="click$.next($any($event).target.value)" formControlName="searchText" type="text" class="form-control" i18n-placeholder="search-form.searchbar-placeholder" placeholder="Explore the full Bitcoin ecosystem">
+      <input [autofocus]="searchAutofocus" (focus)="focus$.next($any($event).target.value)" (click)="click$.next($any($event).target.value)" formControlName="searchText" type="text" class="form-control" i18n-placeholder="search-form.searchbar-placeholder" placeholder="Explore the full Bitcoin ecosystem">
```
softsimon, line 4: I believe you can replace the whole custom component with a conditional attribute like:

`[attr.autofocus]="boolean"`

softsimon (changes requested):
:)

frontend/src/app/components/search-form/search-form.component.html
nymkappa: Does not work. But I replaced with a `.focus()` call in the component instead of using a directive, less code.

frontend/src/app/components/search-form/search-form.component.ts
nymkappa: Replaced this approach by leveraging the state service instead

nymkappa:
Re-did the implementation:
- Focus is triggered by the dashboards component directly through the StateService
- We don't auto focus when a touch screen is detected (tested on iPhone 13 / safari only)

frontend/src/app/shared/pipes/bytes-pipe/utils.ts
```diff
@@ -309,3 +309,29 @@ export function takeWhile(input: any[], predicate: CollectionPredicate) {
...
+      hasTouchScreen = true; // deprecated, but good fallback
+    } else {
+      // @ts-ignore - Only as a last resort, fall back to user agent sniffing
+    const UA = navigator.userAgent;
+      hasTouchScreen =
+        /\b(BlackBerry|webOS|iPhone|IEMobile)\b/i.test(UA) ||
+        /\b(Android|Windows Phone|iPad|iPod)\b/i.test(UA);
+    }
+  }
+  console.log(hasTouchScreen);
```
softsimon, line 335:
```suggestion

File: frontend/src/app/shared/pipes/bytes-pipe/utils.ts [hunk 4]
```diff
@@ -332,6 +332,5 @@ export function hasTouchScreen(): boolean {
         /\b(Android|Windows Phone|iPad|iPod)\b/i.test(UA);
     }
   }
-  console.log(hasTouchScreen);
```
nymkappa: Thanks

softsimon (changes requested):
There is an issue now with clicking on the page and selecting text. It will always force the focus. It should only force the focus on first page load.

softsimon (approved):
tested ACK @ [6d5be78dd]

PR #4030, opened 2023-07-23
Use more reliable Github Action for Rust toolchain install

junderw:
@knorrium This is probably more well maintained.

We need it whitelisted though:

`dtolnay/rust-toolchain@*`

This should be added to the Actions whitelist.


knorrium:
utACK until the list is updated by @wiz

knorrium (approved):
Tested ACK through CI

Issue #4032, opened 2023-07-23
[liquid] block list breaks when using pagination


nymkappa:
- Go to https://liquid.place/blocks
- Use the pagination to browse page 3+
- Page breaks

Issue #1823, opened 2022-06-07
See if we can upgrade the blocks list component when `DATABASE.ENABLED: false`


nymkappa:
Now that mining pools are stored into RAM when the database is disabled (see https://github.com/mempool/mempool/pull/1822), we should be able to display the full blocks list data set even if indexing is not available.

So instead of showing 

Height | Timestamp | Mined | Transactions | Size
-- | -- | -- | -- | --

We can probably show this even if database is disabled

Height | Pool | Timestamp | Mined | Reward | Fees | TXs | Size
-- | -- | -- | -- | -- | -- | -- | --

PR #4033, opened 2023-07-23
[block list] improve block list when db = false

nymkappa:
Fixes https://github.com/mempool/mempool/issues/1823
Also fixes https://github.com/mempool/mempool/issues/4032

Please check if the block list works fine in the following condition:
- mainnet with db
- mainnet without db
- liquid

Without db, we can show the full table details now since we index all the data on the fly
Screenshot 2023-07-23 at 12 19 09

Also re-enabled the fee column when audit is not available.
image

On liquid, it should simply not break (no changes) 😄 
Screenshot 2023-07-23 at 12 03 54


softsimon:
@nymkappa please rebase

softsimon (approved):
tested ACK @ [e6236d8c]

tested on mainnet and Liquid and it works with/without DB

PR #4037, opened 2023-07-23
Change colour of recently broadcasted transactions in expected block overview graph

orangesurf:
Current colors for recently broadcasted and removed transactions are quite similar (#8a3480 & #f344df).

This pull request is intended to change recently broadcasted to a different colour (#c5a75e) to make the distinction more apparent.

The change in tx-view.ts switches from a relative colour (using desaturation & darken) to an explicit hex code.

I have not built so I can't confirm that this has the intended result.

proposed-expected-block-colour-scheme-change

### DIFF
--- a/production/install
+++ b/production/install
@@ -1793,7 +1793,7 @@ case $OS in
 esac
 
 # wait for mysql to start
-sleep 5
+sleep 10
 
 mysql << _EOF_
 create database mempool;


mononaut:
Thanks for the PR!

I'm not a huge fan of that light brown color for recently broadcast transactions. The current similarity in colors is somewhat intentional, since "recently broadcast" and "removed" are both subsets of "in the expected template, but missing from the actual block".

orangesurf:
Thanks for reviewing - The similar colour makes sense from that perspective. 

To me the current colour similarity makes it easy to mistake small missing transactions for recently broadcast transactions when glancing at the expected block overview graph, but this is only my opinion. 

For example http://mempool.space/block/0000000000000000000051456952317f445661dad9214d450aa71545065e6746

image

mononaut:
That's true. Maybe we should dim the "recently broadcast" color a bit more.

PR #4039, opened 2023-07-24
Fix missing sigops

mononaut:
Fixes a typo which accidentally disabled sigop calculations for new transactions on mainnet, and recalculates those missing sigops when loading from disk cache.


softsimon (approved):
tested ACK @ [44f2217a6]

PR #4041, opened 2023-07-24
Sanitize channel id search

softsimon:
Filters for IDs and Short IDs eg;

`765939x1133x0`
`842158836741373952`


mononaut (approved):
utACK

PR #4043, opened 2023-07-24
Mempool inSync status

mononaut:
_(builds on PR #4025)_

This PR makes the mempool inSync status more conservative (i.e. it stays out of sync until the whole mempool *and* templates are definitely ready.

It also forces the mempool blocks and fee recommendations to show as skeleton loaders whenever the mempool is not in sync:

https://github.com/mempool/mempool/assets/83316221/7c3194c8-74dc-4d66-94e1-965bc0590147


softsimon (approved):
tested ACK @ [36fe5627]

wiz (approved):
Tested ACK @ v3.0.0-dev [36fe5627]

PR #4045, opened 2023-07-24
Faster blocks

mononaut:
This PR uses a new `/block/:hash/txs` endpoint added in https://github.com/mempool/electrs/pull/24 to speed up block processing by loading block transactions in a single request instead of individually.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
+++ b/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
@@ -8,6 +8,7 @@ export interface AbstractBitcoinApi {
   $getBlockHeightTip(): Promise<number>;
   $getBlockHashTip(): Promise<string>;
   $getTxIdsForBlock(hash: string): Promise<string[]>;
+  $getTxsForBlock(hash: string): Promise<IEsploraApi.Transaction[]>;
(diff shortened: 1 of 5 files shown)


wiz (approved):
Tested ACK @ d35d04b2e82f134f90ae2875d368326a2b7a5585

softsimon (changes requested):
We must use the existing mempool transactions on new block

softsimon (changes requested):
The fee spans are now gone

Screenshot 2023-07-25 at 00 26 02

mononaut:
Rebased on master, and refactored the `$getTransactionsExtended` function as follows:
 - try to look up transactions in the mempool cache first.
 - if esplora is enabled and we still need to fetch > 500 transactions, fetch them in bulk using the new mempool/electrs endpoint.
 - fetch any remaining missing transactions individually.
 
The other logic should be unchanged, except that I've removed the `memPool.hasPriority()` short-circuit on non-esplora backends. With the more conservative inSync behavior from #4043 the mempool always has priority while resyncing the last `INITIAL_BLOCKS_AMOUNT` blocks, so the short-circuit just wrecks those block templates.

softsimon (approved):
tested ACK @ [25925751e]

PR #4046, opened 2023-07-25
Exclude all conflicting transactions from audit score

mononaut:
Following from the discussion on issue #4038, this PR adjusts the block audit to exclude *all* conflicting transactions between mined blocks and our expected templates.

The implementation reuses the "full rbf" database column to avoid a migration, and labels all conflicted missing/added transactions as "Conflicting":

Screenshot 2023-07-25 at 2 45 19 PM


softsimon (approved):
tested ACK @ [d7b874ac4]

PR #3937, opened 2023-07-10
Replace client-side stale blocks

mononaut:
_(builds on PR #3932)_

This PR refactors the frontend `stateService.blocks$` observable so that we can swap out the client-side cached blocks when a reorg occurs and always show the latest main chain data.

https://github.com/mempool/mempool/assets/83316221/cac77695-c3e6-4475-ae0f-13b7fe89e8bf

PR #4048, opened 2023-07-25
Fix key navigation subscription leak

mononaut:
After navigating to and from the next mempool block using the arrow keys, a stray rxjs subscription persists which will trigger a navigation to the latest mined block when the blocks observable next changes.

Previously the `blocks$` observable was a `BehaviorSubject` which always immediately returned a result, so the leak did not produce visible effects. However #3937 changed `blocks$` to a simple `Observable` which causes the weird navigation effects.

This PR fixes the issue by not subscribing to `blocks$` on every arrow key event and fetching the latest value of `blocksSubject$` directly instead.

Before:

https://github.com/mempool/mempool/assets/83316221/2a1d3d34-398f-4134-9a98-0915ac6ea2bc

After:

https://github.com/mempool/mempool/assets/83316221/7955defe-459e-40ef-8b24-3ac5a175779d


softsimon (approved):
tested ACK @ [e15c0c6c7a]

PR #4050, opened 2023-07-27
Handle failures while fetching block transactions

mononaut:
Fixes a bug where the newly refactored `$getTransactionsExtended` would filter out failed transaction responses instead of retrying and/or throwing errors, which could produce templates etc with missing transactions.

This PR retries failed tx requests (falling back to Core), and adds sanity checks to ensure the final list of transactions is fully formed.

If any transactions are missing, or the coinbase is malformed, the function throws an error.

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -105,11 +105,16 @@ class Blocks {
       }
     }
 
-    // Skip expensive lookups while mempool has priority
     if (onlyCoinbase) {
       try {
-        const coinbase = await transactionUtils.$getTransactionExtended(txIds[0], false, false, false, addMempoolData);
-        return [coinbase];
+        const coinbase = await transactionUtils.$getTransactionExtendedRetry(txIds[0], false, false, false, addMempoolData);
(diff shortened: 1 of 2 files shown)


backend/src/api/blocks.ts
```diff
@@ -134,26 +139,43 @@ class Blocks {
...
     }
 
     if (!quiet) {
       logger.debug(`${foundInMempool} of ${txIds.length} found in mempool. ${totalFound - foundInMempool} fetched through backend service.`);
     }
 
+    // Require the first transaction to be a coinbase
+    const coinbase = transactionMap[txIds[0]];
+    if (!coinbase || !coinbase.vin[0].is_coinbase) {
+      console.log(coinbase);
```
softsimon, line 163:
`````suggestion

mononaut: fixed

softsimon (approved):
tested ACK @ [589adb95]

Tested using Esplora and Bitcoin Core mode.
Didn't test what happens on failure.

PR #4053, opened 2023-07-28
[mining] use .slug to load pool logo

nymkappa:
Small refactoring. Right now, we can't rename a mining pool without breaking its logo.


softsimon (approved):
tested ACK @ [3f3f0db2f2]

PR #4054, opened 2023-07-28
Show new mined transactions on the address page

### LINKED ISSUE #4051: The address page doesn't add new block transactions
The address page does add new mempool transactions dynamically, and update the status of existing transactions when they get mined into a block, but it does NOT add new block transactions to the page if they weren't in the mempool already, for example the Coinbase transaction.

mononaut:
Fixes #4051


softsimon (approved):
tested ACK @ [9b65fbd98]

PR #4058, opened 2023-07-29
Current fiat prices API

### LINKED ISSUE #4052: Expose the current fiat prices using an API
1. Expose the current price of Bitcoin in our supported fiat currencies via an api endpoint.

Suggested endpoint and data output, similar to the current https://mempool.space/api/v1/historical-price

`/api/v1/price`.

```
 {
   "USD": 29366,
   "EUR": 26502,
   "GBP": 22683,
   "CAD": 38732,
   "CHF": 25313,
   "AUD": 43447,
   "JPY": 4115960,
   "time": 1690423200, (latest price update)
}
```

2. If it's not done already, use median instead of average for the exchange prices. https://github.com/mempool/mempool/issues/3986
3. Increase the update frequency from current hourly (?)

mononaut:
Resolves #4052.

This PR adds a `/api/v1/prices` API and makes the price update frequency configurable (default 1hr, but set to 5 mins on production mainnet).

I changed my mind about using the median for prices - we already filter out zero prices and absurdly high price outliers, and I think a median might not always be representative from only a few data sources.


backend/src/__fixtures__/mempool-config.template.json
```diff
@@ -92,6 +92,7 @@
     "PASSWORD": "__SOCKS5PROXY_PASSWORD__"
   },
   "PRICE_DATA_SERVER": {
+    "UPDATE_FREQUENCY": "__PRICE_DATA_SERVER_UPDATE_FREQUENCY",
```
softsimon, line 95:
````suggest
    "UPDATE_FREQUENCY": "__PRICE_DATA_SERVER_UPDATE_FREQUENCY__",

File: backend/src/api/prices/prices.routes.ts [hunk 2]
```diff
@@ -0,0 +1,19 @@
+import { Application, Request, Response } from 'express';
+import config from '../../config';
+import pricesUpdater from '../../tasks/price-updater';
+
+class PricesRoutes {
+  public initRoutes(app: Application): void {
+    app.get(config.MEMPOOL.API_URL_PREFIX + 'prices', this.$getCurrentPrices.bind(this));
+  }
+
+  private async $getCurrentPrices(req: Request, res: Response): Promise<void> {
```
softsimon, line 10:
```suggestion
  private $getCurrentPrices(req: Request, res: Response): void {

**Comment by softsimon** [comment]
I noticed the PRICE_DATA_SERVER stuff is not used at all. Should we instead erase everything there from the configs and add a new PRICE_UPDATE_FREQUENCY to the general MEMPOOL section instead?

**Comment by mononaut** [comment]
yeah, maybe

File: backend/src/tasks/price-updater.ts [hunk 3]
```diff
@@ -103,13 +103,13 @@ class PriceUpdater {
...
    */
-  private async $updatePrice(): Promise<void> {
+  private async $updatePrice(storeInDb: boolean): Promise<void> {
     if (this.lastRun === 0 && config.DATABASE.ENABLED === true) {
       this.lastRun = await PricesRepository.$getLatestPriceTime();
     }
 
-    if ((Math.round(new Date().getTime() / 1000) - this.lastRun) < 3600) {
-      // Refresh only once every hour
+    if ((Math.round(new Date().getTime() / 1000) - this.lastRun) < Math.min(config.MEMPOOL.PRICE_UPDATE_FREQUENCY, 3600)) {
```
nymkappa, line 111: If you make the refresh time configurable, don't cap the value. Maybe I only want to refresh prices once a day 👍🏻

backend/src/repositories/PricesRepository.ts
```diff
@@ -65,7 +65,7 @@ class PricesRepository {
     
     try {
       await DB.query(`
-        INSERT INTO prices(time,             USD, EUR, GBP, CAD, CHF, AUD, JPY)
+        INSERT IGNORE INTO prices(time,             USD, EUR, GBP, CAD, CHF, AUD, JPY)
```
nymkappa, line 68: Better catch errors you want to silence rather than silence everything, see for example: https://github.com/mempool/mempool/blob/master/backend/src/repositories/BlocksRepository.ts#L180

But if your main issue is duplicated entries, here it would make more sense to use `ON UPDATE` to update the latest price entry in the db.

backend/src/api/prices/prices.routes.ts
```diff
@@ -0,0 +1,19 @@
...
+
+class PricesRoutes {
+  public initRoutes(app: Application): void {
+    app.get(config.MEMPOOL.API_URL_PREFIX + 'prices', this.$getCurrentPrices.bind(this));
+  }
+
+  private $getCurrentPrices(req: Request, res: Response): void {
+    res.header('Pragma', 'public');
+    res.header('Cache-control', 'public');
+    res.setHeader('Expires', new Date(Date.now() + 1000 * 300).toUTCString());
```
nymkappa, line 13: Since we know how often we will update the price (`PRICE_UPDATE_FREQUENCY`), maybe you can use its value there instead of a hardcoded 30 seconds.

nymkappa (changes requested):
It seems to be working fine. But I've left a few comments on the code.

mononaut:
Tested NACK @ [813f3dc]

Price stops updating at the top of the hour

softsimon:
> Tested NACK @ [[813f3dc](https://github.com/mempool/mempool/commit/813f3dc09deb7d4ada81c3776d253615e2fd0c55)]
> 
> Price stops updating at the top of the hour

Hopefully fixed now.
Along with setting the cache expiry time based on the setting as Kappa suggested.

docker/backend/start.sh
```diff
@@ -35,7 +35,7 @@ __MEMPOOL_MAX_BLOCKS_BULK_QUERY__=${MEMPOOL_MAX_BLOCKS_BULK_QUERY:=0}
 __MEMPOOL_DISK_CACHE_BLOCK_INTERVAL__=${MEMPOOL_DISK_CACHE_BLOCK_INTERVAL:=6}
 __MEMPOOL_MAX_PUSH_TX_SIZE_WEIGHT__=${MEMPOOL_MAX_PUSH_TX_SIZE_WEIGHT:=4000000}
 __MEMPOOL_ALLOW_UNREACHABLE__=${MEMPOOL_ALLOW_UNREACHABLE:=true}
-
+__MEMPOOL_PRICE_UPDATES_PER_HOUR__=${MEMPOOL_PRICE_UPDATES_PER_HOUR:=3600}
```
nymkappa, line 38:
```suggestion
__MEMPOOL_PRICE_UPDATES_PER_HOUR__=${MEMPOOL_PRICE_UPDATES_PER_HOUR:=1}
```

mononaut:
Tested ACK @ [07f95acc]

PR #4059, opened 2023-07-29
fix websocket connection state observable

### LINKED ISSUE #4044: Retry fetching 2h graph data after reconnection
Instead of being stuck on forever spinners. Retry when we go back online and in sync agian.

Screenshot 2023-07-24 at 17 22 33

mononaut:
Fixes #4044

The websocket `connectionState$` observable wasn't always properly updated, and the `mempoolStats$` pipeline in the dashboard didn't handle API errors so if it failed during init it would exit and not retry when the connection was restored.


softsimon (approved):
tested ACK @ [354c119e]

PR #4070, opened 2023-07-31
Misc Redis fixes

mononaut:
Fixes Redis tx deletion log prints, and speeds up deletion by using `UNLINK` instead of `DEL` and raising the batch size.


softsimon (approved):
tested ACK @ [49db63d]

PR #4071, opened 2023-07-31
Fast indexing

mononaut:
This PR aims to speed up indexing of blocks, block summaries and CPFP clusters by:
 - Reduce mempool update poll frequency 10x while indexing is in progress.
 - Use the new `mempool/electrs` bulk block txs endpoint to fetch transactions when possible, instead of getting verbose blocks from Bitcoin Core.
 - Indexing CPFP clusters and block summaries together, which:
     - Avoids fetching block transactions twice.
     - Ensures we include effective fee rate data in block summaries.

and also fixes a bunch of bad indexing debug logs to show the actual rate of progress.


wiz (approved):
Tested ACK @ 0d25ef0b5b8a05a83fde361addaae6c2ad41110f

Re-indexed past 3000 blocks or so, was around 3x faster!

PR #4072, opened 2023-07-31
Fix scripthash lookup for Electrum*

softsimon:
"Fixes" scripthash lookup although it returns empty balance and no transactions using `romanz/electrs`, it might work with Fulctrum though.


mononaut (approved):
Tested ACK @ [bed7c1b2]

PR #4074, opened 2023-07-31
Fix rust gbt docker  typo

knorrium:
There was a bug in the unit test that caused it to return early and not catch all the keys in the MEMPOOL config, so the RUST_GBT key was wrong.


softsimon (approved):
utACK

PR #4075, opened 2023-07-31
Make sed commands platform agnostic

knorrium:
GNU sed and BSD sed are different. Making all commands platform agnostic so we can run the Docker scripts on Macs as well.

Builds on #4074


knorrium:
No longer needed.

PR #4068, opened 2023-07-31
Get blocks from electrs again

mononaut:
Resolves #4055

PR #4078, opened 2023-08-01
Fix: Use bits to calculate difficulty instead of floating points

junderw:
Apparently bringing the difficulty float from electrs after PR 5 caused problems because we use difficulty (which is inherently inaccurate) to calculate the difficulty changes.

This PR should remove the reliance on the difficulty float in important calculations.

See this comment: https://github.com/mempool/electrs/pull/5#issuecomment-1657856555

Things to verify in review:

Will this affect a running instance's caches etc.? If so, how can we migrate the existing database? (or if we just want to re-index, how long will that take?)

Note:

We could have hit this issue due to a change in the compiler used by Bitcoin Core etc. as the float JSON representation is not guaranteed. This is why the actual mining code in Core only uses integers.

### DIFF
--- a/backend/src/__tests__/api/difficulty-adjustment.test.ts
+++ b/backend/src/__tests__/api/difficulty-adjustment.test.ts
@@ -1,4 +1,8 @@
-import { calcDifficultyAdjustment, DifficultyAdjustment } from '../../api/difficulty-adjustment';
+import {
+  calcBitsDifference,
+  calcDifficultyAdjustment,
+  DifficultyAdjustment,
+} from '../../api/difficulty-adjustment';
 
 describe('Mempool Difficulty Adjustment', () => {
   test('should calculate Difficulty Adjustments properly', () => {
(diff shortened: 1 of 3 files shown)


mononaut:
Tested NACK @ [b11e996]

This breaks difficulty adjustment indexing (as expected, since the difficulty is now calculated differently to what we already indexed)

Screenshot 2023-08-01 at 1 24 37 PM

I think rebasing this PR on top of #4068 (or merging that first) should fix the issue.

wiz (commented):
@junderw is this going to work for mainnet, testnet, and signet? it looks like it makes some assumptions from consensus rules, which might need to be changed?

junderw:
> @junderw is this going to work for mainnet, testnet, and signet? it looks like it makes some assumptions from consensus rules, which might need to be changed?

As is shown here: https://github.com/bitcoin/bitcoin/blob/v25.0/src/pow.cpp#L74-L123

1. The min max of x4 and /4 is hard coded in this logic, regardless of what each of the params values (the thing that changes between networks). It does x4 and /4 to a params value, but then later in the function it nullifies that params value by dividing by it.
2. The max value being `0x1f0000ff` is because an exponent of 0x20 would not be able to generate a u256, and ff is the highest value that can be made. This is just the maximum value for this type of compact integer (in a u32 sense).
3. The min value being 1 is also hard coded, if nBits is 0 the PoW check returns early with a false.
4. The 3 bytes being signed (and therefore avoiding using the top bit) is also network agnostic.

Since this new function doesn't deal with the difficulty period integer and just says "only use on nBits from blocks in two adjacent difficulties" as long as no one misuses it, it should work on any network.

junderw:
> I think rebasing this PR on top of #4068 (or merging that first) should fix the issue, since then we detect adjustments using the bits target instead of the difficulty.

That's what I figured (and why I made this a draft heh)

wiz:
@junderw now that #4068 is merged did you want to rebase / ready for review this one?

junderw:
Ready for review

mononaut (approved):
Tested ACK @ [b365ad3]

PR #4081, opened 2023-08-01
fix null coinbase fees in block summary

mononaut:
Fixes a bug where the coinbase fee field in the block summaries could be left `null` instead of set to zero.

Block summaries created with the bug will need to be reindexed to fix.

Remember to also clear the redis block summaries cache before testing!
```
redis-cli DEL block-summaries
```


wiz (approved):
Tested ACK

PR #4084, opened 2023-08-01
Show tool list when no other networks available in footer

hunicus:
In some cases, such as on Raspberry Pi distros, no other networks will show in the networks list, making the footer look like this:

Screenshot from 2023-08-01 18-42-35

This PR adds another list in the footer to maintain aesthetic balance:

Screenshot from 2023-08-01 18-42-18


softsimon (approved):
tested ACK @ [9373fb3d]

PR #4085, opened 2023-08-02
use bulk mempool post api to batch mempool update requests

mononaut:
This PR uses the new `/mempool/txs` POST api added in https://github.com/mempool/electrs/pull/31 to fetch missing mempool transactions from mempool/electrs in batches, which can be significantly faster than requesting transactions individually during the mempool update loop.


softsimon (approved):
tested ACK @ [ca0c6b5e6e]

Tested both with `electrs` and Core as backend.

PR #4087, opened 2023-08-02
Superfast parallel block indexing

mononaut:
This PR uses nodejs worker threads to parallelize the `$generateBlockDatabase` and `$generateBlocksSummariesDatabase` functions.

We create a temporary pool of workers to saturate available CPU cores, and push block indexing tasks through them as quickly as possible.

This PR uses the new mempool/electrs bulk block txs method to fetch transactions for the block summaries/CPFP indexing. However, using Core's verbose block RPC instead was actually faster on my machine for some reason. I'm not sure that will hold true in production, but I'll open an alternative PR that always uses verbose block so we can compare performance.

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -410,6 +410,8 @@ class Blocks {
       return;
     }
 
+    const workerPool: Worker[] = [];
+
     try {
       // Get all indexed block hash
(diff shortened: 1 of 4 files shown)


wiz:
@softsimon any reason this wasn't merged?

PR #4091, opened 2023-08-03
Update liquid-master-page.component.html

dni:
ngIf for ```BISQ_ENABLED``` was missing


softsimon (approved):
utACK

But CLA agreement is missing.

PR #4092, opened 2023-08-03
BUGFIX: docker mempool/backend copy rust-gbt outside /package

dni:
```COPY --from=builder --chown=1000:1000 /build/package/rust-gbt ./rust-gbt/``` this is still needed inside docker because its included like below, which means inside the package context its outside of the package.

https://github.com/mempool/mempool/blob/cac2a984abac022cf20630ce44465cca7383f85f/backend/src/api/mempool-blocks.ts#L1

otherwise after you run the docker image you will get: 

screenshot-1691089880

### DIFF
--- a/docker/backend/Dockerfile
+++ b/docker/backend/Dockerfile
@@ -22,6 +22,7 @@ FROM node:16.16.0-buster-slim
 WORKDIR /backend
 
 RUN chown 1000:1000 ./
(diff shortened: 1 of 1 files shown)


junderw:
NACK. This needs a more permanent fix.

These 3 things need to be changed.

1. The dependency entry in package.json
2. The various script entries in package.json that move rust-gbt around
3. The position of the rust-gbt folder

I am also curious as to why this causes an issue in Docker but not when deploying in prod.

PR #4093, opened 2023-08-04
Fix: Rust-GBT packaging needs fixing

junderw:
Fixes #4092 

The packaging around rust-gbt was a little wonky, and relied on some of the weirder path resolution quirks of NodeJS.

This should fix the Dockerfile issue as well as make things a bit more correct.

### DIFF
--- a/backend/npm_package.sh
+++ b/backend/npm_package.sh
@@ -1,0 +1,14 @@
+#/bin/sh
+set -e
+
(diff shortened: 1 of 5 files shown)


softsimon (approved):
partial tested ACK @ [96628ce]

Didn't run the Docker stuff but the other changes works fine.

PR #4095, opened 2023-08-04
Refactor websocket address tracking

mononaut:
This PR refactors how we handle the `track-address` websocket subscription in the backend to reduce code duplication and scale better for many simultaneous subscriptions.

Instead of searching the full list of transactions separately for every subscribed client, we build a mini address index cache once up front, from which we can look up tracked addresses in the client update loop in ~constant time.

Building the cache adds a small constant overhead to the `$handleMempoolChange` and `handleNewBlock` functions, but the fast lookups maintain performance much better when there are many open address subscriptions.

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -192,7 +192,7 @@ class WebsocketHandler {
               if (/^04[a-fA-F0-9]{128}$/.test(parsedMessage['track-address'])) {
                 client['track-address'] = null;
                 client['track-scriptpubkey'] = '41' + matchedAddress + 'ac';
-              } else if (/^|(02|03)[a-fA-F0-9]{64}$/.test(parsedMessage['track-address'])) {
+              } else if (/^(02|03)[a-fA-F0-9]{64}$/.test(parsedMessage['track-address'])) {
                 client['track-address'] = null;
                 client['track-scriptpubkey'] = '21' + matchedAddress + 'ac';
               } else {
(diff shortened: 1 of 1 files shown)


backend/src/api/websocket-handler.ts
```diff
@@ -958,6 +888,52 @@ class WebsocketHandler {
...
+
+  private async getFullTransactions(transactions: MempoolTransactionExtended[]): Promise<MempoolTransactionExtended[]> {
+    for (let i = 0; i < transactions.length; i++) {
+      try {
+        transactions[i] = await transactionUtils.$getMempoolTransactionExtended(transactions[i].txid, true);
+      } catch (e) {
+        logger.debug('Error finding transaction in mempool: ' + (e instanceof Error ? e.message : e));
+      }
+    }
+    return transactions;
```
softsimon, line 934: If this errors, the non-extended transaction will return here instead of not at all as in the current implementation. May be an issue but not sure.

mononaut: hmm, I'm not sure what the best solution is.

If we filter out failures, those transactions will just never show up on the address page. But the non-extended transactions are missing prevouts, so the inputs show as "UNKNOWN".

softsimon: If non existing prevouts are sent to the frontend it will crash the frontent

mononaut: the frontend doesn't actually seem to throw any errors for me, but the transaction shows up like this:

Screenshot 2023-08-07 at 11 50 37 AM

softsimon (approved):
tested ACK @ [38e9021e8]

PR #4096, opened 2023-08-04
Speed up RBF detection

mononaut:
`Common.findRbfTransactions()` detects RBF events by iterating over added and deleted transactions in a nested loop and checking each combination of transactions for conflicting inputs.

This is extremely fast for a few transactions, but the `O(N²)` complexity blows up when there are many added and deleted txs (as there might be after restarting the server).

This PR adds an alternative RBF detection algorithm which should run in `~O(N)`.

It has some extra constant overhead, so we only use it when the number of transactions is large enough to outweigh that cost.


softsimon (approved):
tested ACK @ [ a4e721921 ]

PR #4098, opened 2023-08-05
Fix: Difficulty calculations for Liquid networks must be NaN

junderw:
From @nymkappa 

```
Aug  5 10:43:07 [97086] DEBUG: <liquidtestnet> Exception in runMainUpdateLoop() (count: 1). Retrying in 1 sec. Reason: Invalid bits. Stack trace: Error: Invalid bits
    at verifyBits (/Users/nymkappa/Documents/mempool/backend/dist/api/difficulty-adjustment.js:34:19)
    at calcBitsDifference (/Users/nymkappa/Documents/mempool/backend/dist/api/difficulty-adjustment.js:37:5)
    at Blocks.$updateBlocks (/Users/nymkappa/Documents/mempool/backend/dist/api/blocks.js:635:102)
    at async Server.runMainUpdateLoop (/Users/nymkappa/Documents/mempool/backend/dist/index.js:200:38)
```

The problem is that Liquid blocks don't have bits or difficulty attributes, but TypeScript doesn't catch that and incorrectly says that there will always be a number.

Previously, we were performing calculations on `undefined` which was returning NaN.

Making this behavior more explicit by checking for the network and explicitly setting it to NaN (like we used to).

### DIFF
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -674,7 +674,11 @@ class Blocks {
           this.updateTimerProgress(timer, 'got previous block hash for initial difficulty adjustment');
           const previousPeriodBlock: IEsploraApi.Block = await bitcoinApi.$getBlock(previousPeriodBlockHash);
           this.updateTimerProgress(timer, 'got previous block for initial difficulty adjustment');
-          this.previousDifficultyRetarget = calcBitsDifference(previousPeriodBlock.bits, block.bits);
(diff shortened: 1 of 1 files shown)


nymkappa (approved):
Tested ACK @ 9da9c2750d0db4e7d2c6c1c446e338a045a231a5

The error is gone.

softsimon (approved):
tested ACK @ [ea926660]

PR #4100, opened 2023-08-05
Feature: Build Rust during build script AND install

junderw:
Developer eXperience fell due to the removal of rust build step from the `npm run build` script.

Since Rust caches build artifacts anyways, 2 build calls in quick succession will see the 2nd one only take a few ms anyways... so I put it back.

It was annoying to have to delete /package and `git restore rust-gbt` every time I tested the npm run package command, so I add a few QoL things. (cp -R instead of mv, and adding to .gitignore)

### DIFF
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -45,3 +45,6 @@ testem.log
 #System Files
 .DS_Store
 Thumbs.db
+
+# package folder (npm run package output)
+/package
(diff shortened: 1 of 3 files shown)


softsimon (approved):
LGTM

wiz (commented):
Tested ACK @ e994aac

PR #4108, opened 2023-08-06
Handle historical price API calls when using any testnet

softsimon:
This PR fixes:

* A failing DB lookup and bad http backend response when the testnet frontend is asking for historical prices
* The frontend will not longer call the historical prices API and error out

### DIFF
--- a/backend/src/api/mining/mining-routes.ts
+++ b/backend/src/api/mining/mining-routes.ts
@@ -41,6 +41,10 @@ class MiningRoutes {
       res.header('Pragma', 'public');
       res.header('Cache-control', 'public');
       res.setHeader('Expires', new Date(Date.now() + 1000 * 300).toUTCString());
+      if (['testnet', 'signet', 'liquidtestnet'].includes(config.MEMPOOL.NETWORK)) {
+        res.status(400).send('Prices are not available on testnets.');
+        return;
+      }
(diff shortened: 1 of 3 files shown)


nymkappa:
I don't understand what this PR fixes. Could you elaborate a bit more please?

✅ If I go on https://mempool.space/testnet/, a single call to the historical price API is made (which returns an empty response, should be a 204 here instead of a 200, but it works fine)
✅ If I browse blocks or transaction, no further API call is made, as expected
✅ All fiat value are displaying zero, as expected. I also don't see any error in the console.

softsimon:
> I don't understand what this PR fixes. Could you elaborate a bit more please?

The NodeJS Backend log shows this when the frontend loads fiat prices

`Aug  9 01:09:05 [24372] ERR: <testnet> Cannot fetch historical prices from the db. Reason Cannot read properties of undefined (reading 'USD')`

It's a bad SQL query and its error in the http request isn't handled properly.
Instead I disabled the API for testnet and throwing 400 .. But this caused the frontend to make 20 repeated error requests, that's why I disabled the request completely. Why make a request at all?

nymkappa (approved):
Tested ACK @ cbebbd40f16e48c7efa220854441c695ad69ee0c

PR #4110, opened 2023-08-06
Backend unit test tweaks

knorrium:
Reduce the config test spam by setting a CI=true flag


softsimon (approved):
tested ACK @ [adc46b6a]

PR #4114, opened 2023-08-08
Scroll blockchain in faq so divider is visible

hunicus:
Sometimes mostly (or only) mempool blocks show on the mempool explore FAQ...

#### Desktop

Screenshot from 2023-08-08 13-56-43

#### Mobile

Screenshot from 2023-08-08 13-57-18

#### Desktop

Screenshot from 2023-08-08 13-58-26

#### Mobile

Screenshot from 2023-08-08 13-58-13


hunicus:
nymkappa told me this isn't working as intended on some devices so I'm investigating.

nymkappa (changes requested):
I think on most screen it would work fine, but it really breaks when there are too many pixels in there.

For example, using a screen resolution of 2056x1329 (highest resolution on my macOS setup), with a Firefox browser zoom set to 67%, the separator is gone.

Maybe @mononaut can take a look at this as it's more familiar with this crazy `mempool | blockchain` component.

image

softsimon:
Fixed in https://github.com/mempool/mempool/pull/4274

PR #4121, opened 2023-08-09
Refine footer design/layout

hunicus:
Most parts of the footer have yet to be translated but I did leave some extra room for longer strings & allow for text wrapping.

#### Current

Screenshot from 2023-08-09 16-51-37

#### Proposed

Screenshot from 2023-08-09 16-28-21

#### Current

Screenshot from 2023-08-09 16-52-06

#### Proposed

Screenshot from 2023-08-09 16-27-57


softsimon (approved):
I like this new designed.

Tested ACK @ [db90e77a]

PR #4136, opened 2023-08-10
use new services api to fetch chad/whale profile image

nymkappa:


wiz:
how about just removing the username and only having the md5 in the url to remove the query parameter

nymkappa:
> how about just removing the username and only having the md5 in the url to remove the query parameter

We don't keep the md5 string in the db right now, so we'd need to md5 all images in the db to find the matching one, too expensive.

But that's something I can improve later one aka have a table dedicated to images.

PR #4137, opened 2023-08-10
Multi address tracking

mononaut:
_(builds on #4209)_

This PR adds support for subscribing to multiple addresses and scripthashes over the websocket.

The behavior of the existing `track-address` subscription is unchanged.

The PR adds two new subscriptions:
 - 'track-addresses', which expects an array of addresses/pubkeys as payload.
 - 'track-scriptpubkeys`, which expects an array of hexadecimal scriptpubkeys strings as payload.

Each client may have *one* active subscription of each type (`track-address`, `track-addresses`, and `track-scriptpubkeys`). Subsequent requests replace open subscriptions of the same type.

Sending a `track-addresses` or `track-scriptpubkeys` request with an empty array payload (`[]`) ends the current subscription.

Relevant events are passed to subscribed clients under the `multi-address-transactions` and `multi-scriptpubkey-transactions` keys respectively, in the format:
```
{
  ...
  'track-address-transactions': {
    <address-1>: {
      mempool: MempoolTransactionExtended[],
      confirmed: MempoolTransactionExtended[],
      removed: MempoolTransactionExtended[],
    },
    <address-2>: {
      mempool: MempoolTransactionExtended[],
      confirmed: MempoolTransactionExtended[],
      removed: MempoolTransactionExtended[],
    },
    ...
  }
  ...
}
```

where for each subscribed address, `mempool` is a list of relevant new transactions entering the mempool, `confirmed` is a list of transactions included in a new block, and `removed` is a list of transactions evicted from the mempool.

Note that transactions may be `confirmed` without a corresponding prior `mempool` event.


softsimon:
I can see how this can be abused with no limits. Perhaps there needs to be a backend config with a default limit of 10 (or something) maximum subscriptions per client, with the possibility to raise the limits on servers dedicated to enterprise customers.

mononaut:
I'm not sure this is any more vulnerable to abuse than individual subscriptions, and might actually be less so since it reduces the overhead of maintaining different websockets, although I suppose it depends how effective our rate limiting is for websocket initiations.

We also improved performance and scalability quite a bit in #4095.

mononaut:
rebased on current master

mononaut:
Added a `config.MEMPOOL.MAX_TRACKED_ADDRESSES` config setting, with default 1.

Attempts to subscribe to more than that number of addresses or scriptpubkeys will result in an error message returned in the websocket response object.

CharliesAngel1 (commented):
approve

backend/src/api/websocket-handler.ts
```diff
@@ -195,21 +201,46 @@ class WebsocketHandler {
...
               }
-              if (/^04[a-fA-F0-9]{128}$/.test(parsedMessage['track-address'])) {
-                client['track-address'] = '41' + matchedAddress + 'ac';
-              } else if (/^(02|03)[a-fA-F0-9]{64}$/.test(parsedMessage['track-address'])) {
-                client['track-address'] = '21' + matchedAddress + 'ac';
-              } else {
-                client['track-address'] = matchedAddress;
+            }
+            if (Object.keys(addressMap).length > config.MEMPOOL.MAX_TRACKED_ADDRESSES) {
+              response['track-addresses-error'] = `too many addresses requested, this connection supports tracking a maximum of ${config.MEMPOOL.MAX_TRACKED_ADDRESSES} addresses`;
```
softsimon, line 221: This is not sent back as valid JSON like other data.

`Unexpected token 'o', ..."-error": too many ad"... is not valid JSON`

mononaut: Fixed

mononaut:
Fixed bad error response and rebased on master

softsimon (approved):
tested ACK @ [0cddfcfa6]

PR #4152, opened 2023-08-14
Select unfurl mode with route prefix

mononaut:
This PR offloads the user-agent matching used to select SEO vs preview rendering modes from the unfurler itself to the reverse proxy server.

Instead of checking against an explicit list of user-agents, the unfurler mode is now controlled by ~~an `X-Unfurl-Type` request header~~ a route prefix.

~~Set `X-Unfurl-Type: preview` to render for link preview bots. All other requests default to SEO-friendly full server side rendering for webcrawlers.~~

Route to `/unfurl/...` to render for link preview bots. Route to `/slurp/...` for SEO-friendly server side rendering for webcrawlers.


wiz (changes requested):
Can you please fix the canonical URL:
```
# curl -i -s -H 'User-Agent: Twitterbot' https://node202.tk7.mempool.space/|grep canonical
    <link rel="canonical" href="https://mempool.space/unfurl/" />
```

wiz (approved):
Tested ACK @ 1653f003bc4b2b68e75affafc3288a086983b4d9

Please fix canonical URL pointing to /unfurl/ separately, need to merge this now to fix googlebot crawling

PR #4154, opened 2023-08-15
Record purging rate in statistics

mononaut:
This PR starts recording the `mempoolminfee` in the statistics table (from the second minfee node, if available).

The data isn't used yet, but this is a first step towards adding a chart of the purging rate over time.


softsimon (approved):
tested ACK @ [68854d56c]

Screenshot 2023-10-14 at 22 35 35

PR #4156, opened 2023-08-15
Batch esplora outspends requests

mononaut:
This PR uses the bulk `/txs/outspends` endpoint implemented in https://github.com/mempool/electrs/pull/36 to fetch batches of outspends from the `transactions-list` and flow diagram components.

The new endpoint allows us to bypass the nodejs backend entirely for these requests, and retrieve the outspends directly from `mempool/electrs`.

The PR continues to support the `/api/v1/outspends` endpoint for backwards compatibility, switches frontend requests over to the new `/api/txs/outspends` path, and adds support for that route for non-esplora backends.


wiz (approved):
Tested ACK @ v3.0.0-dev [9b720b68e] using mempool-electrs 3.0.0-dev-3858f5a

softsimon (dismissed):
Deprecate the old API and remove the NodeJS proxy

mononaut:
removed the old API & proxy, and rebased on master

softsimon (approved):
utACK with the latest changes

PR #4164, opened 2023-08-16
Fix Docker start script and unit test

knorrium:
The unit test wasn't working properly with array values, so the ESPLORA FALLBACK config was missing and the test did not fail.


softsimon (approved):
utACK, tests successful

PR #4183, opened 2023-08-18
Don't overload core with mempool tx requests

### LINKED ISSUE #4178: WARNING: request rejected because http work queue depth exceeded, it can be increased with the -rpcworkqueue= setting
Getting core spamming this warning recently. It sometimes even forces core to abort. Reproduced with this very light config.

```
  "MEMPOOL": {
    "ENABLED": true,
    "NETWORK": "mainnet",
    "BACKEND": "none",
    "HTTP_PORT": 8999,
    "INDEXING_BLOCKS_AMOUNT": 0,
    "BLOCKS_SUMMARIES_INDEXING": false,
    "AUTOMATIC_BLOCK_REINDEXING": false,
    "AUDIT": false,
    "ADVANCED_GBT_AUDIT": false,
    "ADVANCED_GBT_MEMPOOL": false,
    "RUST_GBT": false,
    "CPFP_INDEXING": false
  },
  "LIGHTNING": {
    "ENABLED": false
  },
```

mononaut:
Fixes #4178 by fetching mempool transactions from Core sequentially instead of making 10k requests in parallel.


backend/src/api/transaction-utils.ts
```diff
@@ -74,8 +74,18 @@ class TransactionUtils {
...
-      return (results.filter(r => r.status === 'fulfilled') as PromiseFulfilledResult<MempoolTransactionExtended>[]).map(r => r.value);
+      const results: MempoolTransactionExtended[] = [];
+      for (const txid of txids) {
+        try {
+          const result = await this.$getMempoolTransactionExtended(txid, addPrevouts, lazyPrevouts, forceCore);
+          if (result) {
+            results.push(result);
+          }
+        } catch {
+          // skip failures
```
nymkappa, line 85: Would be good to log something in DEBUG, or do we already log in another function?

nymkappa: Also for the dev who will land on this comment in a few months, could be good to elaborate a bit why we're fine skipping failures

mononaut: we already log progress in the outer function (although those logs might be a bit sparse with a slow core backend)

junderw:
Perhaps using https://www.npmjs.com/package/p-limit might be better?

Basically it creates N slots for promises to run concurrently, then when one resolves, it fills the slot with the next promise in the queue.

(You pass a closure that returns the promise `limit(() => myPromise())` so that the promise doesn't start (send a request) until the queue puts it in a slot)

It's pretty useful.

junderw:
https://github.com/mempool/mempool/commit/ebf43bd074be315f524f32ccf88b742bc822f146

Joining together the two packages, It's only 142 lines of code. So I just added it with TypeScript definitions. Should be useful.

Feel free to merge it into your branch if you want.

32 concurrent seems like a reasonable amount for bitcoind... maybe a little lower might be good?

mononaut:
added, thanks!

32 concurrent requests still saturates the default Core maximum RPC work depth, so I've dropped that down to a more conservative 8, which should be fast enough while leaving plenty of spare room for other tasks and applications.

nymkappa (approved):
Tested ACK @ b9838fda8db62aae69a261ec8f15a5d5b10a7133

PR #4192, opened 2023-08-20
accept CLA

djahwork:
accept CLA and add fee % in transaction fee section

Capture d’écran du 2023-08-20 19-02-47

visual proof that ordinals are regarded

Capture d’écran du 2023-08-20 19-23-22

3ceb28ac9598dcb76640bb611900f6aacabc2aa2d5dc1dad73a58e9acddc25bb

### DIFF
--- a/contributors/djahwork.txt
+++ b/contributors/djahwork.txt
@@ -1,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of August 20, 2023.
+
+Signed: djahwork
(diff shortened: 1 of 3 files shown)


softsimon:
We discussed this PR briefly and weren't super stoked on this feature addition. We're trying to keep the transaction page as clean as possible and the % number doesn't immediately tell the user what it represents. Perhaps a better solution is a tooltip over the fee number.

DCMTOKEN:
I feel the same way. I didn’t quite understand when it ask for the percentage, assuming everyone would know what it’s replying to. 


Sent from Yahoo Mail for iPhone


On Thursday, November 9, 2023, 7:03 PM, softsimon ***@***.***> wrote:




We discussed this PR briefly and weren't super stoked on this feature addition. We're trying to keep the transaction page as clean as possible and the % number doesn't immediately tell the user what it represents. Perhaps a better solution is a tooltip over the fee number.

djahwork:
Ok, thanks for the feedback.
I thought it was useful to have the % of sats spent for a given tx.
I think it is even more relevant than the USD price of the transaction.

PR #4195, opened 2023-08-21
Update rust-toolchain

junderw:
## Background

#4194 changed from 1.70 to 1.63. It also added a `[workspace]` to `rust-gbt/Cargo.toml` which broke the existing workspace.

## Issues fixed

- Missed changing the rust-toolchain
- Linked the CI toolchain to the rust-toolchain file to prevent missing it in the future
- Removed the `[workspace]` that broke some assumptions of our build
- Added an argument to README, if you use the argument on a non-affected version of npm, it does nothing, but if you are on a version of npm affected by the bug (9.0.0 to 9.4.1) this extra argument is necessary to build properly. (Debian stable is on a version of npm with the bug)


junderw:
@TheBlueMatt pin

junderw:
https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

It tells rustup to ignore its global system-wide settings, and while inside this workspace it should only use a specific toolchain.

The `1.63` toolchain is set to the latest `1.63.x` version.

Without changing this, the overwhelming majority of developers that tries to work on mempool (and install rust via the instructions on the official rust website) will be forced into using 1.70.x when we want to force them into using 1.63.x to make sure (ie. the let-else you removed) new language features don't compile for them locally.

junderw:
> It does seem strange to me to force users to download yet another toolchain

In the rustup world, Gigamegs flow freely, and people with slow internet connections are banished to the depths of the underworld. :rofl: 

But tbh, this is just an issue of "should it break locally first or do we just catch it in CI?"

Developers who care about downloading too many toolchains probably won't use rustup, in which case this file is meaningless to them. (ie. yourself)

It's a pretty common practice, especially for crates that require specific versions of nightly.

junderw:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

I wonder if using `rust-version` might also help catch people trying to compile with `let-else` etc.

junderw:
I was checking things locally and found the `[workspace]` commit.

This is problematic, because we have a workspace file at the git root, and the npm run package command does cleanup under the assumption that the target directory (which we REALLY want to clean up post-build if it's there) is not there.

When running npm install, cargo automatically finds the Cargo.toml at he root of git and builds fine with the Cargo.lock in the root of git, placing the target directory in the root, and `.gitignore` ignores it properly.

Can you explain how not having `[workspace]` broke your build somehow?

junderw:
See: https://github.com/mempool/mempool/blob/master/Cargo.toml

junderw:
Also, here are the results of using `rust-version = "1.63"`

```
With 1.55:

The package requires the Cargo feature called `edition2021`, but that feature is not stabilized in this version of Cargo (1.55.0 (32da73ab1 2021-08-23)).

With 1.56:

error: package `gbt v0.1.0 (/home/jonathan/git/mempool/backend/rust-gbt)` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.56.1

With 1.71 + Re-introducing let-else (a feature not allowed in 1.63):

No error. Built successfully.
```

I find that the build errors are much nicer for people using versions between 1.56 and 1.62 (Before that they get the 2021 edition error)

But considering the previous LTS of Debian (old-stable) uses 1.48, I'm not sure who would benefit from including that.

keeping the rust-toolchain in sync with CI is best imo. Which afaik is a manual thing.

(can github action yaml files include strings from other files?......)

junderw:
Hmm. Looks like it messes with the GitHub action whitelist.

junderw:
Perhaps this is a problem with some tool like rust-analyzer and how it interacts with your IDE?

Something on your system is looking through the symlink and getting confused.

Maybe old versions of npm copied instead of symlinking?

Here's my behavior.

```
[jonathan@jonathan-zenbook backend] (junderw/rusttoolchain $=)
$ rm -rf ../target/ node_modules/ dist/ rust-gbt/target/ rust-gbt/gbt.linux-x64-gnu.node 
[jonathan@jonathan-zenbook backend] (junderw/rusttoolchain $=)
$ npm install

added 650 packages, and audited 652 packages in 42s

68 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
[jonathan@jonathan-zenbook backend] (junderw/rusttoolchain $=)
$ ls -l node_modules/rust-gb*
lrwxrwxrwx 1 jonathan jonathan 11  8月 21 10:45 node_modules/rust-gbt -> ../rust-gbt
```

junderw:
> If npm is `cd`'d to the `node_modules` dir when it tries to build it may fail.

Does it actually do that on your machine? If you run `npm install` from the backend folder, does it fail? Could you post the error?

junderw:
https://github.com/npm/cli/pull/6142

This is an npm bug fixed in 9.4.2 onward.

To support npm 9.2.0 and its buggy behavior would require some extra workarounds, just adding `[workspace]` is not sufficient.

junderw:
Workaround for npm@9.2.0: `npm install --no-install-links`

Confirmed working:
```
[jonathan@jonathan-zenbook backend] (junderw/rusttoolchain $=)
$ npm -v
9.2.0
[jonathan@jonathan-zenbook backend] (junderw/rusttoolchain $=)
$ rm -rf ../target/ node_modules/ dist/ rust-gbt/target/ rust-gbt/gbt.linux-x64-gnu.node 
[jonathan@jonathan-zenbook backend] (junderw/rusttoolchain $=)
$ npm install --no-install-links

added 650 packages, and audited 652 packages in 41s

68 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
[jonathan@jonathan-zenbook backend] (junderw/rusttoolchain $=)
$ ls -l node_modules/rust-gb*
lrwxrwxrwx 1 jonathan jonathan 11  8月 21 11:11 node_modules/rust-gbt -> ../rust-gbt
```

junderw:
> the issues with simply removing the workspace are?

There's no issue with removing the workspace **and** changing all the affected areas.

However, only adding `[workspace]` to `rust-gbt/Cargo.toml` is not the same thing.

It is simply less work and simpler to revert the broken change and inform people using the broken `>=9.0.0 <9.4.2` versions of npm on the workaround.

junderw:
> that's really a critical bug?

No it's not critical. I never claimed it to be critical. But it is a bug and requires fixing in one way or another.

junderw:
> I'm still missing something here.

One example: docker image build, for instance, performs a lot of cleanup scripts to make the final image as small as possible. All we need is the final `*.node` file. After that, we can even delete all the rust source files, Cargo.toml, etc. The current script does not delete any target directories because it assumes the target directory is at the root.

Would having an extra 6 GB target file in the image break anything? No... but still, in the sense that the project is designed around not including the target folder in the image, a change that includes the target folder in the docker image is not a complete change.

It should have been caught in review, but I was asleep while it was proposed and swiftly merged.

junderw:
> Apologies if my comment above was unclear

No worries.

The build system of mempool is rather hairy, and getting rust-gbt to play nice with it, as well as npm, was very difficult.

Also, one of the main user groups for mempool are docker users, so imo docker image size is pretty important to the project.

junderw:
Also, I think originally we had the opposite problem. Rust was complaining about not having a workspace in the git root so we made a workspace in the git root as a solution to that problem, then started trial-and-erroring the other smaller issues around that as an assumption (location of Cargo.lock and target folder).

junderw:
> AFAIU, its more of a change than a bugfix

If they viewed it as a change, they would probably wait til v10 to include it instead of a patch release. It is potentially breaking, and the change from false to true was released in v9.0.0 so they seem to be aware that it's breaking, and yet decided it was urgent enough to release in a patch.

Which is why I called it a bug. But meh, semantics.

junderw:
For now I've filed https://github.com/npm/cli/issues/6726

We'll see how they react. From a quick glance, it says I need a Debian command to report Debian bugs...?

I'll look into debian reporting later, maybe after I see how npm reacts.

PR #4198, opened 2023-08-21
More unfurler fixes

mononaut:
- avoid throwing frontend errors if we can't get a webgl context
- adds more timing and debugging logs
- fixes unfurler resource proxy routing


wiz (approved):
Tested ACK @ 0d2df72621da50cbabbf408667b4479567134771

PR #4200, opened 2023-08-21
Update canonical link with block hash

mononaut:
Adds a `updateCanonical(path)` method to the seoService, and uses it to point the `rel="canonical"` link at the block *hash* when a block is accessed by height.


softsimon (approved):
tested ACK @ [368ab1dc6]

PR #4188, opened 2023-08-19
User menu + integrated accelerator if available

nymkappa:

PR #4202, opened 2023-08-22
Dynamic width chain

mononaut:
_(to merge into https://github.com/mempool/mempool/pull/4188)_

~~Makes the main `mempool | blockchain` component dynamically resize to fit its container, instead of the whole window, and automatically corrects the scroll position accordingly.~~

New approach: applies a translation to the `mempool | blockchain` component while the menu is open to compensate for the offset. The transform can be composited, so it animates as smoothly as the menu slide.

Before:

https://github.com/mempool/mempool/assets/83316221/e3ea85cc-581f-4ccc-83bb-a8ac99bd1f9e

After:

~~https://github.com/mempool/mempool/assets/83316221/3321b8d6-87f1-417a-8f7a-20525ace10be~~

https://github.com/mempool/mempool/assets/83316221/eefb125a-d93c-4eed-8578-04209f432250


mononaut:
Updated with a new approach

nymkappa:
Nice, so even tho it's "bouncing" a bit it's much better. Problem is that it broke support for window resizing.

For example if you resize the window smaller this is the result:

image

mononaut:
Sorry, I lost a HostListener decorator somewhere. It's fixed now.

mononaut:
and rebased

PR #4203, opened 2023-08-22
Fix unfurl fallback img routes

mononaut:
Fixes the fallback image paths in `/unfurl/...` responses, and adds a separate local image path to the unfurler route config for when we send fallback image files directly after failed `/render/...` requests.

PR #4204, opened 2023-08-22
Use symlink to avoid duplicate unfurler fallback images

mononaut:
Reimplementation of #4203 with a symlink to the `frontend/src/resources` folder to remove the need for duplicate images.


wiz (approved):
utACK

PR #4205, opened 2023-08-22
Unfurl fallback no-cache

mononaut:
_(builds on #4204)_

Sets a `Cache-control: no-cache` header to unfurler fallback-on-failure responses.


wiz (approved):
utACK

PR #4209, opened 2023-08-24
Notify address page of removed transactions

### LINKED ISSUE #4208: Replaced transactions are not removed from the address page
If an unconfirmed transaction is dropped from the mempool for whatever reason, it remains listed on the address page until the user navigates away or refreshes the page.

When the transaction was removed because it was replaced-by-fee, this can lead to duplicate/conflicting transactions being shown together, and incorrect (or even impossible) address balances:

https://github.com/mempool/mempool/assets/83316221/e0106e17-cd47-4eac-b480-7da989e8270f

mononaut:
Fixes #4208

This PR checks for removed mempool transactions involving subscribed addresses, and sends them to interested clients under the `address-removed-transactions` key in the websocket response.

The front end then removes them from the address page and updates the address sent/received/balance totals accordingly:

https://github.com/mempool/mempool/assets/83316221/b624d41a-9942-4bd5-b2bf-c58357997256

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -486,6 +486,7 @@ class WebsocketHandler {
 
     // pre-compute address transactions
     const addressCache = this.makeAddressCache(newTransactions);
+    const removedAddressCache = this.makeAddressCache(deletedTransactions);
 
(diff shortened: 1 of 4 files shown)


softsimon (approved):
tested ACK @ [9b17c163]

PR #4213, opened 2023-08-26
Send cpfp/effective fee rate changes to subscribed websocket clients

### LINKED ISSUE #4211: Automatically refresh CPFP info in transaction component
CPFP data is not displayed in real time on the transaction page. It seems that a page reload is required in order to get the updated data.

Is it possible to display updated CPFP info in real time?

mononaut:
_(resolves #4211)_

This PR monitors changes to CPFP info (including effective fee rate, ancestor/descendant or acceleration changes), and sends the new data to any websocket clients with matching `track-tx` subscriptions.

The frontend then uses this to update the displayed CPFP information without needing to refresh the page.


nymkappa (approved):
@softsimon please review the code, I only tested directly.

On my end I tested it on mainnet and the CPFP data is properly updated everything I spend an unconfirmed output, data is updated and arrow is also properly moved at the correct position.

Works well.

softsimon (approved):
tested ACK @ [6d643604]

Tried on testnet. Works nicely.

PR #4217, opened 2023-08-28
Implement pid file & checks

### LINKED ISSUE #137: Implement pid file + checking for "mempool already running" to prevent starting 2nd instance and corrupting database
pidfile should be "mempool-${databasename}.pid" or something to allow running per-network instances on same server

mononaut:
Resolves #137

Saves a PID file at `/${PID_DIR}/mempool-${databasename}.pid` if the database is enabled, and fails on startup if the PID file already exists:
```
ERR: Already running on PID 1337 (or pid file '.../mempool/backend/dist/mempool-mempool.pid' is stale)
```

The PID file is saved in the current directory (`__dirname`) by default, or under `config.DATABASE.PID_DIR` if configured.


backend/src/index.ts
```diff
@@ -306,6 +313,13 @@ class Server {
       this.lastHeapLogTime = now;
     }
   }
+
+  onExit(exitEvent): void {
+    DB.releasePidLock();
```
knorrium, line 318: @mononaut don't you have to check for `config.DATABASE.ENABLED` here?

mononaut: yes, I probably should.

although `DB.releasePidLock()` will have no effect unless the pid in the lock file belongs to the current process, which it shouldn't if we started with `config.DATABASE.ENABLED=false`.

knorrium:
@mononaut did you see the failing unit test for the `PID_DIR` override?

Mempool Backend Config › should override the default values with the passed values

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 0
    + Received  + 1

      Object {
        "DATABASE": "__DATABASE_DATABASE__",
        "ENABLED": false,
        "HOST": "__DATABASE_HOST__",
        "PASSWORD": "__DATABASE_PASSWORD__",
    +   "PID_DIR": "",
        "PORT": 18,
        "SOCKET": "__DATABASE_SOCKET__",
        "TIMEOUT": 3000,
        "USERNAME": "__DATABASE_USERNAME__",
      }

mononaut:
Fixed the `config.DATABASE.ENABLED` check and the config unit tests, and rebased on current master.

softsimon (approved):
tested ACK @ [75c5041]

`Oct 14 15:04:39 [74592] ERR: Already running on PID 74391 (or pid file '/Users/simon/repos/mempool/backend/dist/mempool-mainnet.pid' is stale)`

wiz (commented):
Tested NACK - Requested changes:
1) Need to add check if PID is *actually* still running, currently daemon fails to start if the PID file exists, despite the PID not actually still running
2) If the PID does exist, check to see if it's actually a mempool nodejs instance, in case a random other process got that same PID

wiz:
Also, a SIGTERM did not clear the PID files, it seems?

PR #4220, opened 2023-08-30
Accelerator dashboard

mononaut:
This PR adds a basic dashboard of public accelerator data at `/acceleration`:

acceleration-dash

The chart of acceleration fees and the list of recent accelerations also have full-page versions at `/graphs/acceleration/fees` and `/acceleration-list`.


mononaut:
Rebased on master

wiz:
@mononaut when you get a chance can you please rebase onto current master

mononaut:
rebased on current master

Fireskull81:
> rebased on current master

nymkappa:
image

I'd use one of the side to show the currently accelerating stats instead.

We can still show both timespan using a simple switch like this one:
image

nymkappa:
There is a small css issue here
image

nymkappa:
It'd be interesting to show the fiat values as well, like in other dashboards
image

image

nymkappa:
ALso, in the services backend, we consider an acceleration to be successful or failed only after 3 confirmations. So the following dashboard data is not valid. It shows 1 transaction as "accelerated" even tho the block was mined by a non participating mining pool. We are going to mark the acceleration as failed after 3 confirmations. This is to reduce the potential issue related to re-orgs.

Screenshot 2023-12-01 at 4 34 44 PM
Screenshot 2023-12-01 at 4 34 36 PM

nymkappa (changes requested):
### `/acceleration-list` is quite broken

I don't think this page is relevant for now since you can get similar info from the block audit. If you want to polish it feel free to do so, otherwise just disable the page for now (and remove the link to it from the dashboard)

image

### Color issue

I think the "faded" effect looks weird here. IMO, we can just display them all in normal colors, similar to the RBF replacement widget where status does not matter.
Even a "canceled" acceleration request had a role to play in the fee market so it's fine to show it in normal color.
Screenshot 2023-12-04 at 2 33 34 PM

### Need to find more relevant data point to show here

This represents the maximum fee delta that can be collected by mining pools. However it will rarely be the case as block median fee rate will most of the time be much lower that the maximum "fee rate set by the user".
image


Here maybe we can show the sum of actual fee paid by users upon successful acceleration
image

### Mobile

The widgets on mobile looks odd vs other dashboard

Screenshot 2023-12-04 at 2 48 04 PM

This is visually more appealing because it's a bit more centered
Screenshot 2023-12-04 at 2 47 54 PM

Recent acceleration widget does not show txid link at all on mobile
image

mononaut:
Updated with a new design:

Screenshot 2023-12-08 at 1 07 14 PM

wiz (approved):
Tested ACK @ v3.0.0-dev [aff44d90][44f883bc]

mihaiepure1977 (commented):
Je vous demande de ma ramburser mon argant

nymkappa:
> Je vous demande de ma ramburser mon argant

Nous ne faisons qu'afficher des données sur le réseau Bitcoin. On ne peux pas vous aider.

PR #4221, opened 2023-08-30
Add meta descriptions

hunicus:
Notes:

- switched attributes for Twitter metadata from `property` to `name` since that is valid HTML
- some pages like the blocks list, recent RBF replacements, privacy policy, terms of service, and trademark guidelines didn't have title set so these changes add title and descriptions to those pages


wiz (approved):
Tested ACK @ v3.0.0-dev [9fdb164b6]

wiz:
@hunicus pointed out a few of the seo service calls are not in the ngOnInit() but he will fix those in a separate PR later

PR #4224, opened 2023-08-30
Bump failover timeout to 5s

mononaut:
Increases the timeout for esplora failover monitoring checks from 2000ms to 5000ms


wiz (approved):
utACK

PR #4236, opened 2023-09-07
Add mempool transaction count line to graph

mononaut:
This PR extends the main mempool graph to include a data series showing the number of unconfirmed transactions in the mempool over time:

Screenshot 2023-09-08 at 1 35 56 AM

The line is shown by default, but can be toggled off via the dropdown filter menu:
Screenshot 2023-09-08 at 1 33 49 AM

The widget version of the graph on the main dashboard is unchanged:

Screenshot 2023-09-08 at 1 34 02 AM

We have been collecting this data since at least v2.0.0, so it should already be available for most graph timespans. This PR includes the count in the all of the `/api/v1/statistics/:duration` endpoints.

### DIFF
--- a/backend/src/api/statistics/statistics-api.ts
+++ b/backend/src/api/statistics/statistics-api.ts
@@ -220,6 +220,7 @@ class StatisticsApi {
   private getQueryForDays(div: number, interval: string) {
     return `SELECT
       UNIX_TIMESTAMP(added) as added,
+      CAST(avg(unconfirmed_transactions) as DOUBLE) as unconfirmed_transactions,
       CAST(avg(vbytes_per_second) as DOUBLE) as vbytes_per_second,
       vsize_1,
       vsize_2,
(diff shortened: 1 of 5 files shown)


softsimon:
The data is not exposed from for everything above 6M because of the second `getQueryForDays` method. Deliberate or not?

mononaut:
not deliberate, and should be fixed now.

softsimon:
It's nice to see the count in the tooltip, but why does it look like a checkbox? 

Screenshot 2023-09-20 at 17 22 29

mononaut:
it's supposed to be a color swatch to match the white line, like the fee rate band labels 😅

we can try a text-only label, I'm just not sure it'll be immediately obvious what the white line is without some similar clue.

softsimon (approved):
tested ACK @ [c80201e3d]

PR #4253, opened 2023-09-14
Pause recording statistics during esplora failover

mononaut:
When the backend fails over to another esplora instance, differences in mempool composition can produce spurious transaction statistics.

This PR extends the failover mechanism to report the active esplora host, as well as which host fulfilled a particular request, and then "pauses" statistics until the effect of a failover has passed.

While statistics are "paused", we skip writing new statistics rows to the database, and stop sending new statistics updates over the websocket.

Due to the moving window averaging uses for some of these statistic, the "pause" lasts at least `TX_PER_SECOND_SAMPLE_PERIOD` seconds after recovering from a failover, to ensure all bad data has timed out of the sample window.


mononaut:
rebased

wiz:
@softsimon any reason this wasn't merged?

PR #4254, opened 2023-09-14
Add REQUEST_TIMEOUT and FALLBACK_TIMEOUT esplora config options

mononaut:
This PR adds two new settings to the `config.ESPLORA` configuration:

`REQUEST_TIMEOUT`: the timeout in ms for regular API requests (previously hard-coded to 10000ms).

When a regular API request hits this timeout, it counts towards the failure rate used to trigger a failover.

`FALLBACK_TIMEOUT`: the timeout in ms for health check requests (previously hard-coded to 2000, then increased to 5000 in PR #4224).

If an esplora host takes longer than this to respond to a periodic health check, it is marked as unreachable and (if active) a failover is immediately triggered.


softsimon (approved):
tested ACK @ [a663cca3]

PR #4255, opened 2023-09-14
Send correct tx conf status in websocket msgs

mononaut:
Updates transactions in the mempool cache after confirmation, so that the `status` field is correct for websocket messages about confirmed transactions for e.g. address subscriptions or txConfirmed events.


softsimon (approved):
utACK

PR #4257, opened 2023-09-15
Sticky header non bitcoin

nymkappa:
Fixes https://github.com/mempool/mempool/issues/4223
Sticky header also applied on liquid and bisq


softsimon (approved):
tested ACK @ [14a41b310]

PR #4258, opened 2023-09-16
Sigops everywhere

mononaut:
Updates `extendMempoolTransaction(...)` to use the sigop count provided by `mempool/electrs` where possible (added by https://github.com/mempool/electrs/pull/43).

Also fixes an error in how the nodejs sigop calculation counts certain P2WSH multisig and P2SH inputs.

Then updates the front end to always display sigops and adjusted vsize on every transaction (when available from the backend). In most cases, the adjusted vsize will be identical to the regular vsize, but we need to put something there to keep the columns balanced.

Screenshot 2023-09-19 at 12 20 15 AM

Note that the adjusted vsize is not (yet) accounted for in the effective fee rate for mined transactions.

### DIFF
--- a/backend/src/api/bitcoin/esplora-api.interface.ts
+++ b/backend/src/api/bitcoin/esplora-api.interface.ts
@@ -6,6 +6,7 @@ export namespace IEsploraApi {
     size: number;
     weight: number;
     fee: number;
+    sigops?: number;
     vin: Vin[];
     vout: Vout[];
     status: Status;
--- a/backend/src/api/transaction-utils.ts
+++ b/backend/src/api/transaction-utils.ts
@@ -116,7 +116,10 @@ class TransactionUtils {
   public extendMempoolTransaction(transaction: IEsploraApi.Transaction): MempoolTransactionExtended {
     const vsize = Math.ceil(transaction.weight / 4);
     const fractionalVsize = (transaction.weight / 4);
-    const sigops = !Common.isLiquid() ? this.countSigops(transaction) : 0;
+    let sigops = transaction.sigops;
+    if (sigops == null) {
+      sigops = !Common.isLiquid() ? this.countSigops(transaction) : 0;
+    }
     // https://github.com/bitcoin/bitcoin/blob/e9262ea32a6e1d364fb7974844fadc36f931f8c6/src/policy/policy.cpp#L295-L298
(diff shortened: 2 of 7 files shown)


backend/src/api/bitcoin/esplora-api.interface.ts
```diff
@@ -6,6 +6,7 @@ export namespace IEsploraApi {
     size: number;
     weight: number;
     fee: number;
+    sigops?: number;
```
junderw, line 9: Why is this optional? Once merged in Electrs, it will always return the sigops.

mononaut: It's needed for compatibility with other backends, since Core, romanz/electrs, older esplora versions etc still won't supply sigops.

Ideally we'd just fill in the missing value using the local calculation, but we aren't guaranteed to have prevouts available everywhere the interface is instantiated.

junderw: Oh yeah, heh.

backend/src/api/transaction-utils.ts
```diff
@@ -116,7 +116,10 @@ class TransactionUtils {
   public extendMempoolTransaction(transaction: IEsploraApi.Transaction): MempoolTransactionExtended {
     const vsize = Math.ceil(transaction.weight / 4);
     const fractionalVsize = (transaction.weight / 4);
-    const sigops = !Common.isLiquid() ? this.countSigops(transaction) : 0;
+    let sigops = transaction.sigops;
```
junderw, line 119: Note: Currently, electrs will return sigops for Liquid as well, might want to add the check for Liquid before this.

backend/src/api/transaction-utils.ts
```diff
@@ -116,7 +116,7 @@ class TransactionUtils {
   public extendMempoolTransaction(transaction: IEsploraApi.Transaction): MempoolTransactionExtended {
```
softsimon, line 116: The Sigop data needs to be added to `extendTransaction` function to be shown on Transaction pages in Bitcoin Core only mode. Is it missing or does it have some issues with it?

softsimon (changes requested):

mononaut:
Rebased, and added sigops to mined tx pages for non-esplora backends

softsimon (approved):
tested ACK @ [29299e622]

Both with mempool/electrs and Bitcoin Core

PR #4260, opened 2023-09-17
Use new electrs outspends API

softsimon:
* When using esplora as backend, this requires the new outspends API https://github.com/mempool/electrs/pull/36
* In Bitcoin Core mode, it just changes the API to behave like the new one.
* Removed to Core to Esplora "proxy" in the backend


mononaut:
duplicate of #4156?

PR #4265, opened 2023-09-18
[lightning] fix js error when there is no geolocation available

nymkappa:
Screenshot 2023-09-18 at 09 46 25

## Fixed

image


softsimon (approved):
utACK

PR #4268, opened 2023-09-20
Fix mobile blockchain alignment on block/tx page load

mononaut:
(hopefully) fixes #4266 and #4017.

| Before | After |
|-|-|
| localhost_4200_block_000000000000000000058452bbe379ad4364fe8fda68c45e299979b492858095(iPhone 12 Pro) (1) | localhost_4200_block_000000000000000000058452bbe379ad4364fe8fda68c45e299979b492858095(iPhone 12 Pro) |

PR #4269, opened 2023-09-20
Reduce blockchain jumping

mononaut:
_(builds on #4268)_

Various changes to the scrollable blockchain components to reduce unwanted jumping when navigating between pages.


softsimon (approved):
tested ACK @ [1a6a0c12e]

Removes the jumping completely for me.. Great..

PR #4270, opened 2023-09-20
Limit assets download to Liquid

softsimon:
* Stop downloading the heavy Liquid assets when not running in Liquid base module
* Removing the Mempool fork of the liquid asset repository that is not in use


knorrium:
@softsimon I updated my other PR to include these Liquid changes. Would you mind reviewing that instead?

PR #4280, opened 2023-09-23
Standalone block visualizations

mononaut:
This PR adds two new pages at `/view/block/:hash-or-height` and `/view/mempool-block/:index`
which display the mined and projected block visualizations respectively on their own at maximum size.

Mouse-over tooltip and click-to-navigate features work as usual.

Useful for grabbing high-resolution screenshots, exploring finer details of a block, or as a kiosk/screensaver view.

frame_chrome_mac_dark


wiz (approved):
Tested ACK @ v3.0.0-dev [80ecf7727]

PR #4281, opened 2023-09-25
[ui] don't add 120px to blockchain component on liquid

nymkappa:


mononaut (approved):
Tested ACK @ 6b8db33a

Fixes the horizontal scroll issue on Liquid, and doesn't break mainnet

PR #4285, opened 2023-09-26
[nodejs backend] added /api/v1/health

nymkappa:
I've added a basic health check API to get an overview of the sync/indexing status for the whole stack.

### Example

`.mempool.withCpfp` data is optional because it's very heavy on mysql and very slow to generate.

```
[14:36:25] nymkappa @ /Users/nymkappa $ curl -s 'localhost:8999/api/internal/health?withCpfp=1' | jq
{
  "core": {
    "height": 863916,
    "txindex": {
      "synced": true,
      "best_block_height": 863916
    },
    "coinstatsindex": {
      "synced": true,
      "best_block_height": 863916
    }
  },
  "mempool": {
    "height": 863916,
    "indexing": {
      "enabled": true,
      "blocks": {
        "count": 11464,
        "progress": 1,
        "withCpfp": {
          "count": 0,
          "progress": 0
        },
        "withCoinStats": {
          "count": 11463,
          "progress": 1
        }
      }
    }
  }
}
```


knorrium:
@nymkappa should the healthcheck handle connection errors to the services better?

i.e, I have wrong core config, and the response is just `connect ECONNREFUSED 127.0.0.1:8332` instead of a well formed JSON saying that core is unreachable

then I had a typo on the port and got `Port should be >= 0 and < 65536. Received type number (284444).`

and with bad credentials I got `Invalid params, response status code: 401`

it would be nice to have the error under `{ "core": { "error": ""}}` or something

On the `indexing` section, perhaps you could include the indexing percentage along with the `indexedBlockCount`?

`{"core":{"height":1302,"txindex":{"synced":true,"best_block_height":1302}},"mempool":{"height":1302,"indexing":{"indexedBlockCount":977,"indexedBlockCountWithCPFP":1073,"indexedBlockCountWithCoinStats":0}}}`

When I had the database disabled all the `indexing` values were `-1`, but maybe it can be made more obvious like `{ "indexing": { "enabled": false}}`?

nymkappa:
> @nymkappa should the healthcheck handle connection errors to the services better?

"Worst" case scenario now dumps the following:

```
{
  "core": {
    "height": -1,
    "error": "connect ECONNREFUSED 127.0.0.1:8332"
  },
  "mempool": {
    "height": 0,
    "indexing": {
      "enabled": false,
      "indexedBlockCount": -1,
      "indexedBlockCountWithCPFP": -1,
      "indexedBlockCountWithCoinStats": -1
    }
  },
  "esplora": {
    "height": -1,
    "error": "connect ECONNREFUSED 127.0.0.1:3000"
  }
}
```

nymkappa:
> On the indexing section, perhaps you could include the indexing percentage along with the indexedBlockCount?

Also added indexing progress

```
  "mempool": {
    "height": 810283,
    "indexing": {
      "enabled": true,
      "blocks": {
        "count": 1890,
        "progress": 1,
        "withCpfp": {
          "count": 174,
          "progress": 0.17
        },
        "withCoinStats": {
          "count": 1890,
          "progress": 1
        }
      }
    }
  },
 ```

**Review by softsimon — CHANGES_REQUESTED** [review]
I think the problem with this solution is that every time someone pings the health API, we do calls to Bitcoin Core and multiple SQL queries.

Is it going to be private/internal?

Otherwise it has to be updated on some interval instead, or used cached response.

**Comment by nymkappa** [comment]
> I think the problem with this solution is that every time someone pings the health API, we do calls to Bitcoin Core and multiple SQL queries.
> 
> Is it going to be private/internal?
> 
> Otherwise it has to be updated on some interval instead, or used cached response.

Originally this PR was made for internal use only (custom request). I've changed the route to use /internal instead.

**Comment by nymkappa** [comment]
@wiz @knorrium is this still needed

**Comment by fubz** [comment]
I'd love to see this for internal use on my self hosted instances.

**Comment by nymkappa** [comment]
Still works fine but it is very very slow. It's limited to `/internal`. If @fubz is still interested maybe we can merge this at last lol

```
$ curl -s http://localhost:8999/api/internal/health | jq
{
  "core": {
    "height": 903272,
    "txindex": {
      "synced": true,
      "best_block_height": 903272
    },
    "coinstatsindex": {
      "synced": true,
      "best_block_height": 903272
    }
  },
  "mempool": {
    "height": 903272,
    "indexing": {
      "enabled": true,
      "blocks": {
        "count": 903273,
        "progress": 1,
        "withCpfp": {
          "count": 890943,
          "progress": 0.99
        },
        "withCoinStats": {
          "count": 903272,
          "progress": 1
        }
      }
    }
  },
  "esplora": {
    "height": 903272
  }
}
```

fubz:
If you push it i'll use it!  I will be uplifting my mempool instances "soon"

nymkappa:
Closing due to lack of interest

PR #4294, opened 2023-09-30
Docker updates

knorrium:
- Fix issue with missing `ACCELERATOR` config
- Update base node images
- Update mariadb image
- Update nginx image
- Drop support for armv7 images
- Cleanup unused files

Screenshot 2023-09-29 at 10 00 28 PM


wiz (approved):
utACK

PR #4300, opened 2023-09-30
Update node types to v20

knorrium:
When updating node versions we need to update the types as well.

Also removing the Cypress type package that has been deprecated.


knorrium:
Closing this stale PR

PR #4302, opened 2023-10-03
Fix block pagination for liquid

softsimon:
While testing another PR I noticed the amount of blocks loaded in the block list is mismatching the hard coded value in the blocks list frontend component, which is 15 here https://github.com/mempool/mempool/blob/master/frontend/src/app/components/blocks-list/blocks-list.component.ts#L136

You. can reproduce the bug by just going to liquid.network/blocks and go to page 2.

The `/api/v1/blocks` and `getBlocks` now returns 15 blocks both on mainnet and liquid, bisq.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -478,7 +478,7 @@ class BitcoinRoutes {
       }
 
       let nextHash = startFromHash;
-      for (let i = 0; i < 10 && nextHash; i++) {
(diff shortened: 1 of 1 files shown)


Dezzj (commented):
Veiwed

nymkappa (approved):
utACK @ [10be2c8](https://github.com/mempool/mempool/pull/4302/commits/10be2c8176813bd82ca6f14effeac78ab0143389)

Dezzj (commented):
Veiwed

PR #4306, opened 2023-10-05
Detect more multisig scripts

vostrnad:
Currently, only scripts that use `OP_CHECKMULTISIG` are labeled as multisig. n-of-n scripts that use `OP_CHECKSIGVERIFY`/`OP_CHECKSIG`, as well as Taproot scripts that use `OP_CHECKSIGADD` are not detected. This PR adds both those cases.

TXIDs to test on:
- 2-of-2: 905ecdf95a84804b192f4dc221cfed4d77959b81ed66013a7e41a6e61e7ed530
- 3-of-3: 10f726bd74800f9d0073d93fd0c65ef218fcfa84be09830341be3360395afa94
- 998-of-999: 7393096d97bfee8660f4100ffd61874d62f9a65de9fb6acf740c4c386990ef73

### DIFF
--- a/frontend/src/app/bitcoin.utils.ts
+++ b/frontend/src/app/bitcoin.utils.ts
@@ -124,6 +124,22 @@ export function parseMultisigScript(script: string): void | { m: number, n: numb
   if (!script) {
     return;
   }
+
+  if (/^(?:OP_PUSHBYTES_\d+ \w+ OP_CHECKSIGVERIFY )+OP_PUSHBYTES_\d+ \w+ OP_CHECKSIG$/.test(script)) {
+    const n = script.match(/OP_CHECKSIG/g).length;
+    return { m: n, n };
+  }
+
+  const checkSigAddMatch = script.match(/^OP_PUSHBYTES_32 \w{64} OP_CHECKSIG (?:OP_PUSHBYTES_32 \w{64} OP_CHECKSIGADD )+(?:OP_PUSHNUM_(\d+)|OP_PUSHBYTES_[1-4] (\w+)) (?:OP_NUMEQUAL|OP_GREATERTHANOREQUAL)$/);
+  if (checkSigAddMatch) {
+    const n = script.match(/OP_CHECKSIG/g).length;
+    const m = checkSigAddMatch[1]
+      ? parseInt(checkSigAddMatch[1], 10)
+      // convert from little-endian to big-endian
+      : parseInt(checkSigAddMatch[2].match(/.{2}/g).reverse().join(''), 16);
+    return { m, n };
+  }
+
   const ops = script.split(' ');
   if (ops.length < 3 || ops.pop() !== 'OP_CHECKMULTISIG') {
     return;


junderw:
NACK

Widening the meaning of "multisig" to mean "any script with multiple signatures" will mean that we are always going to be encountering false negatives. Especially with taproot, due to aggregation of signatures.

"any script that contains the OP_CHECKMULTISIG" definition is a much more manageable definition.

vostrnad:
@junderw I understand that with Taproot and signature aggregation it's no longer possible to detect all use of multisig, but that doesn't seem like a good reason to me to ignore that which we can.

Note that this isn't widening the definition of "multisig" to any script with multiple signatures, it's specifically the two script templates mentioned in [BIP342](https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki#cite_note-5) as replacements for `OP_CHECKMULTISIG`:

> **Alternatives to <code>CHECKMULTISIG</code>** There are multiple ways of implementing a threshold _k_-of-_n_ policy using Taproot and Tapscript:
> * **Using a single <code>OP_CHECKSIGADD</code>-based script** A <code>CHECKMULTISIG</code> script <code>m <pubkey_1> ... <pubkey_n> n CHECKMULTISIG</code> with witness <code>0 <signature_1> ... <signature_m></code> can be rewritten as script <code><pubkey_1> CHECKSIG <pubkey_2> CHECKSIGADD ... <pubkey_n> CHECKSIGADD m NUMEQUAL</code> with witness <code><w_n> ... <w_1></code>. Every witness element <code>w_i</code> is either a signature corresponding to <code>pubkey_i</code> or an empty vector. A similar <code>CHECKMULTISIGVERIFY</code> script can be translated to BIP342 by replacing <code>NUMEQUAL</code> with <code>NUMEQUALVERIFY</code>. This approach has very similar characteristics to the existing <code>OP_CHECKMULTISIG</code>-based scripts.
> * **Using a _k_-of-_k_ script for every combination** A _k_-of-_n_ policy can be implemented by splitting the script into several leaves of the Merkle tree, each implementing a _k_-of-_k_ policy using <code><pubkey_1> CHECKSIGVERIFY ... <pubkey_(n-1)> CHECKSIGVERIFY <pubkey_n> CHECKSIG</code>. This may be preferable for privacy reasons over the previous approach, as it only exposes the participating public keys, but it is only more cost effective for small values of _k_ (1-of-_n_ for any _n_, 2-of-_n_ for _n &ge; 6_, 3-of-_n_ for _n &ge; 9_, ...). Furthermore, the signatures here commit to the branch used, which means signers need to be aware of which other signers will be participating, or produce signatures for each of the tree leaves.

junderw:
So multisig is an alternative multisig to multisig?

The fact that it needs to be called an "alternative" should hint at it being something different.

Not to mention, even when you get a match, you don't know what the other script branches are, so calling it a 998-of-999 multisig is misleading at best.

What if there were other branches that allowed different combinations of keys that change the effective m or n values?

If we merge, it should be named something else.

"Taproot Script Condition: Alternative Multisig m-of-n"  etc. with a tool tip that mentions that we can't see the whole script, so this is just one possible condition.

vostrnad:
> The fact that it needs to be called an "alternative" should hint at it being something different.

The reason for a new opcode in Tapscript was to allow batch verification of multisig scripts and fixing `OP_CHECKMULTISIG`'s inefficiency. Even though the opcode name changed, it's still very much multisig as before.

> you don't know what the other script branches are

In this case you can, since the 998-of-999 input has no merkle branch steps in its control block, which means there could not have been any other scripts. In general you are right, and such is life with Taproot.

I think it's better to label things as best we can than to not label them at all. After all, we already do something similar (and arguably potentially much more misleading) when labeling specific script templates as e.g. "Lightning Force Close" even though not all revokable timelocks necessarily belong to Lightning.

junderw:
> it's still very much multisig as before.

It's not multisig. There could be a branch that signs with 1 key unconditionally. The key spend key could also be spendable.

Notice that the existing multisig parsing logic will return nothing if any extra ops are added before or after the CHECKMULTISIG template. any `OP_IF <single sig> OP_ELSE <multisig> OP_ENDIF` funny business is rejected. Why is that? I mean, what if the input used the else path? Why not label it multisig?

Answer: because it's not multisig.

> I think it's better to label things as best we can

Like I said, I'm ok with labelling it:

`Tapscript Singular Condition: Multi-party CHECKSIGADD script with 998 signatures required and 999 keys present`

is a perfectly fine label.

Shortening that into `multisig 998-of-999` is misleading and inaccurate.

vostrnad:
Why was this closed? If we agree that a CHECKSIGADD script could use a label but not what the label should be, I don't think it makes sense for me to create a second PR with a different label when I can just change it in this one.

PR #4309, opened 2023-10-08
Configurable unfurler config

knorrium:
Allow the unfurler to read from different config paths in preparation for dockerization

### DIFF
--- a/unfurler/src/config.ts
+++ b/unfurler/src/config.ts
@@ -1,7 +1,7 @@
 const fs = require('fs');
 const path = require('path');
 
-const configPath = process.env.UNFURLER_CONFIG || '../config.json';
+const configPath = process.env.UNFURLER_CONFIG || './config.json';
 const absolutePath = path.resolve(configPath);
 let config;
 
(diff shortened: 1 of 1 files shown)


unfurler/src/config.ts
```diff
@@ -1,4 +1,16 @@
-const configFile = require('../config.json');
+const fs = require('fs');
+const path = require('path');
+
+const configPath = process.env.UNFURLER_CONFIG || '../config.json';
```
mononaut, line 4: can we make the default path `'./config.json'`?

(`require('../config.json')` is relative to this file, but `path.resolve('../config.json')` is relative to the working directory, i.e. one level higher)

mononaut (changes requested):
Tested ACK, pending change to the default config path

unfurler/src/config.ts
knorrium: @mononaut good catch. I think I changed this when it was running off `dist`

mononaut (approved):
Tested ACK @ e9e507af

PR #4323, opened 2023-10-20
Hide mempool graph count line by default

mononaut:
Hides the new "count" line on the Mempool by vBytes graph by default


Dezzj (commented):
Veiwed

wiz (approved):
Tested ACK @ v3.0.0-dev [86458927d]

PR #4326, opened 2023-10-23
Fix memory cache cleanup logic

TKone7:
Keep (filter) cache entries with an expiry in the future.

Currently all active cache entries are immediately removed, while expired ones are kept in cache forever.


softsimon (approved):
utACK

PR #4328, opened 2023-10-25
Enforce purging rate minimum on recommended fees

mononaut:
Due to the way `mempoolminfee` works, when fee rates are highly compressed (e.g. 300MB+ of unconfirmed transactions paying within 1 s/vb of each other), it's possible for the purging rate to briefly exceed the actual contents of the mempool. (see https://bitcoin.stackexchange.com/a/120024)

In that scenario, the median rates of the next few projected blocks (and therefore our fee recommendations) could be *lower* than the purging rate.

To avoid recommended fee rates that would be rejected by default nodes, we should always recommend *at least* the current purging rate.

Therefore this PR enforces a minimum of `ceil(mempoolminfee)` on all recommended fee rates.

### DIFF
--- a/backend/src/api/fee-api.ts
+++ b/backend/src/api/fee-api.ts
@@ -3,21 +3,30 @@ import { Common } from './common';
 import mempool from './mempool';
 import projectedBlocks from './mempool-blocks';
 
+interface RecommendedFees {
+  fastestFee: number,
+  halfHourFee: number,
+  hourFee: number,
+  economyFee: number,
+  minimumFee: number,
+}
+
 class FeeApi {
   constructor() { }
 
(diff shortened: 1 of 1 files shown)


murchandamus:
Concept ACK

By the way, how do you establish the dynamic minimum mempoolfeerate? Do you have some nodes running with the default configuration just to be able to read that off, or is it a calculated value from the nodes with the big mempools and custom timeout?

mononaut:
Yes, we read the `mempoolminfee` off another node running with default `maxmempool` and `mempoolexpiry` configuration.

softsimon (approved):
tested ACK @ [83fde60]

PR #4332, opened 2023-10-28
Fix stray comma in docker backend config

mononaut:
Fixes a stray comma in the docker backend config template, which causes the generated config file not to be valid json.


knorrium (approved):
utACK

PR #4345, opened 2023-11-06
Shake the echarts tree

### LINKED ISSUE #785: Investigate if we can tree shake Echarts more
We added a 991 KB (315 KB compressed) extra bundle for the ECharts.

Investigate ways to tree shake this into a smaller payload for our usage.

Might be some hints here: https://github.com/apache/echarts/issues/15138

mononaut:
Resolves #785.

This PR changes how we use the echarts library to allow angular/webpack to tree-shake and only include the parts we actually use.

This seems to reduce the size of our largest webpack chunk (the graphs module) by about 50%.

~~Leaving in draft for now while I double-check that it doesn't break any of our charts.~~ seems ok so far?

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -85,7 +85,6 @@
     "clipboard": "^2.0.11",
     "domino": "^2.1.6",
     "echarts": "~5.4.3",
-    "echarts-gl": "^2.0.9",
     "lightweight-charts": "~3.8.0",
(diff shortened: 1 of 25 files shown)


softsimon (approved):
tested ACK @ [a33dc74c8]

softsimon (approved):
tested ACK @ v3.0.0-dev [2df99ef7e]

PR #4354, opened 2023-11-10
Better code-splitting

mononaut:
This PR splits the frontend build into finer-grained lazily-loaded modules, in order to reduce the initial download size for each page and minimize the time to *first contentful paint* on slower connections.

The main changes are:
 - splitting the largest page components into their own individual modules.
 - separating the bitcoin, liquid and bisq "master pages" and routes.
 - refactoring the routing modules to reduce code duplication.


softsimon (approved):
tested ACK @ [af7b9c0dc8]

Tested Bisq and Liquid using frontend proxy and that works too.

PR #4361, opened 2023-11-11
ApiService caching layer

### LINKED ISSUE #4360: Outspends API is called twice from the Transaction page
Opening any random Transaction and 2 outspends requests are triggered. This should be optimized.

```
INFO - handle GET /tx/198975a89cc20152051a5c9b65f894a6ebf32c217ef5324a9c1ebf4d6798eb12/outspends
INFO - handle GET /tx/198975a89cc20152051a5c9b65f894a6ebf32c217ef5324a9c1ebf4d6798eb12/outspends
```

mononaut:
Resolves #4360

This PR adds a `cachedRequest(function, expiresAfter, ...params)` function to the main ApiService, which can be used to temporarily cache any of the supported API functions via a map of replayed BehaviorSubjects.

Subsequent requests to the same function with the same params within the expiry window will return the cached result instead of making a fresh request.

This allows components to easily share API requests where it would otherwise be awkward to do so.

The PR then uses this new feature to prevent making duplicate outspend requests from the `transactions-list` and `tx-bowtie-graph` components on the transaction page.

### DIFF
--- a/frontend/src/app/components/transactions-list/transactions-list.component.ts
+++ b/frontend/src/app/components/transactions-list/transactions-list.component.ts
@@ -53,7 +53,6 @@ export class TransactionsListComponent implements OnInit, OnChanges {
     private assetsService: AssetsService,
     private ref: ChangeDetectorRef,
     private priceService: PriceService,
-    private cd: ChangeDetectorRef,
   ) { }
 
   ngOnInit(): void {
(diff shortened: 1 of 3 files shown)


softsimon (approved):
tested ACK @ [6740ab61f3]

PR #4368, opened 2023-11-13
Make search and select work for truncated text

### LINKED ISSUE #4367: Can't Ctrl+F search for addresses on transaction page - space/line-break in address
When Ctrl+F searching for e.g. `3QYepbTj2Dpi8Ef2usYhHBb3SNju5pdkxJ` (address associated with the first input) on https://mempool.space/tx/c9b5719131bfeac6378749243731c5e70f1ce56deabb7006a2b6539710866420 it does not find it as there's a space or line-break in the address. Removing a few characters at the end of the search helps. However, someone not aware of this might come to the conclusion that a transaction does not spent or pay to an address. This can be fixed by removing or changing the following truncation: 

https://github.com/mempool/mempool/blob/28113d11411efd1a11a324bd94976bb86f877b3c/frontend/src/app/components/transactions-list/transactions-list.component.html#L71C1-L73C29

softsimon:
fixes #4367

With this solution of a hidden transparent text, search and find is possible on transaction and address page for example.

Tried on Chrome, Firefox and Safari.

Screenshot 2023-11-13 at 17 43 39

### DIFF
--- a/frontend/src/app/shared/components/truncate/truncate.component.html
+++ b/frontend/src/app/shared/components/truncate/truncate.component.html
@@ -1,4 +1,5 @@
 <span class="truncate" [style.max-width]="maxWidth ? maxWidth + 'px' : null" [style.justify-content]="textAlign" [class.inline]="inline">
+  <div class="hidden">{{ text }}</div>
     <ng-container *ngIf="link">
       <a [routerLink]="link" class="truncate-link">
         <ng-container *ngIf="rtl; then rtlTruncated; else ltrTruncated;"></ng-container>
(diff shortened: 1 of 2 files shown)


nymkappa (changes requested):
Seems to be working fine overall. But tested NACK because it creates an overflow on mobile layout when checking an address.

image

softsimon:
> Seems to be working fine overall. But tested NACK because it creates an overflow on mobile layout when checking a transaction.

Should be fixed now.

nymkappa:
Tested ACK @ 86fe6a802b2bb069cd32bdb62f079fad961e90de

Next improvement would be to be able to search through the whole list of input/outputs which are hidden by default.

image

nymkappa (approved):
Tested ACK @ 86fe6a802b2bb069cd32bdb62f079fad961e90de

Next improvement would be to be able to search through the whole list of input/outputs which are hidden by default.

image

Issue #4245, opened 2023-09-11
The transaction submission chart in vB/s should be smoothed to be more readable


murchandamus:
### Description

The “Transaction vBytes per second (vB/s)” chart on the `…/graphs/mempool` page is very noisy to the point that it is neigh unreadable for most of the time intervals. I’ll refer to this chart as “Submission rate” in the following.

### Problem to be solved

Although the Submission Rate chart is juxtaposed with the mempool composition chart, the Submission Rate chart would be more readable if it had a less detailed resolution than the mempool composition chart. E.g. on the 1M chart, there appear to be datapoints every 30 minutes, but the submission rate can drastically jump between two half hour intervals. E.g. at 4:30PM on Sep. 7th it’s 971 vB/s, then at 5:00PM it’s 8,128 vB/s:

![image](https://github.com/mempool/mempool/assets/4060799/ed4f35e5-3b8f-4e6a-b3af-8ef8dbd4d848)
![image](https://github.com/mempool/mempool/assets/4060799/9d4407ba-f6f8-4678-84ff-e5c7ebb211b3)

### Proposed solution

 Perhaps the Submission Rate chart would be more useful, if it were smoothed to surface trends, e.g. if each 30 minute interval on the 1M chart showed the average of the prior three hours. Alternatively, it would be nice to be able to pick windows over which data is smoothed from a dropdown menu.

#### Additional info

Overview of all time windows:

2h:
![image](https://github.com/mempool/mempool/assets/4060799/1a81e61f-91a5-4c28-b01f-dadd0e6b3851)

24h:
![image](https://github.com/mempool/mempool/assets/4060799/dfb31c8b-fd30-4689-90bf-5f7ba310de83)

1W:
![image](https://github.com/mempool/mempool/assets/4060799/36f6646d-fbdd-437d-9c7f-f7b03f30ce0f)

1M:
![image](https://github.com/mempool/mempool/assets/4060799/0f39332d-d925-4576-8321-383676900464)

3M:
![image](https://github.com/mempool/mempool/assets/4060799/018b6f83-c779-4eeb-86a6-7f5ea8205c30)

6M:
![image](https://github.com/mempool/mempool/assets/4060799/5588b49e-1f18-4b3d-be43-246bba205373)

1Y:
![image](https://github.com/mempool/mempool/assets/4060799/ad2bbeb1-33c9-4be7-928f-9a729854f0ac)

2Y:
[...]

PR #4371, opened 2023-11-14
[graph] add toggle to show/hide outliers in transaction vBytes per second graph

nymkappa:
Related to, but does not fix https://github.com/mempool/mempool/issues/4245

I've added a toggle to show/hide outliers values. When enabled, we cap all values to 4 times the median value of the serie. See illustrations:

### `All`

Screenshot 2023-11-14 at 10 53 23 AM
Screenshot 2023-11-14 at 10 53 28 AM

### `1w`

Screenshot 2023-11-14 at 10 53 42 AM
Screenshot 2023-11-14 at 10 53 47 AM

### Mobile layout

image

### DIFF
--- a/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
+++ b/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
@@ -256,13 +256,11 @@ export class IncomingTransactionsGraphComponent implements OnInit, OnChanges, On
       ],
       yAxis: {
         max: (value) => {
-          if (!this.outlierCappingEnabled) {
+          if (!this.outlierCappingEnabled || value.max < this.medianVbytesPerSecond * OUTLIERS_MEDIAN_MULTIPLIER) {
             return undefined;
+          } else {
+            return Math.round(this.medianVbytesPerSecond * OUTLIERS_MEDIAN_MULTIPLIER);
           }
-          if (value.max < this.medianVbytesPerSecond * OUTLIERS_MEDIAN_MULTIPLIER) {
-            return undefined;
-          }
-          return Math.round(this.medianVbytesPerSecond * OUTLIERS_MEDIAN_MULTIPLIER);
         },
         type: 'value',
         axisLabel: {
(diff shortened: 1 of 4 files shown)


mononaut:
I'd prefer to just limit the yAxis range rather than actually edit the data points, e.g:

Screenshot 2023-11-15 at 3 22 55 AM

otherwise the capped outlier points tend not to extend all the way to the top of the graph viewport, and the tooltip shows the wrong values.

nymkappa:
> I'd prefer to just limit the yAxis range

Not sure why I did not think of this approach 🤣 will update.

nymkappa:
> I'd prefer to just limit the yAxis range

Updated the PR! Thanks for the better approach.

frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
```diff
@@ -232,6 +255,7 @@ export class IncomingTransactionsGraphComponent implements OnInit, OnChanges, On
         }
       ],
       yAxis: {
+        max: this.outlierCappingEnabled ? Math.round(this.medianVbytesPerSecond * OUTLIERS_MEDIAN_MULTIPLIER) : undefined,
```
mononaut, line 258:
```suggestion
        max: this.outlierCappingEnabled ? (value) => Math.min(value.max, this.medianVbytesPerSecond * OUTLIERS_MEDIAN_MULTIPLIER) : undefined,
```

mononaut: not tested, but just an idea to avoid this effect when there aren't any big outliers:


Screenshot 2023-11-15 at 7 54 59 AM
Screenshot 2023-11-15 at 7 54 51 AM

softsimon:
* Maybe this should even be default?
* And also store the checkbox setting in a localStorage like other graph settings.

frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.ts
nymkappa: Implemented this idea a bit differently, nice feedback thanks

nymkappa:
> * Maybe this should even be default?
> 
>     * And also store the checkbox setting in a localStorage like other graph settings.

Did not set this by default to avoid confusion for current users. Saved the state in localstorage.

Also fixed @mononaut mentioned issue.

Fireskull81:
03864ef025fde8fb587d989186ce6a4a186895ee44a926bfc370e2c366597a3f8f@3.33.236.230:9735

mononaut (approved):
Tested ACK @ [4cd8d70d]

PR #4373, opened 2023-11-14
Fix silently unhandled database exceptions

mononaut:
Adds missing try/catch handling to some `PricesRepository.$getLatestPriceId()` calls, and fixes and issue introduced by the pid cleanup exit handler where unhandled exceptions would cause the process to end silently without printing any error logs or traces.


softsimon (approved):
tested ACK @ [5c0a59d2f]

Unable to reproduce more crashes with a unstable database connection.

PR #4381, opened 2023-11-15
Minimum acceleration preview bar height

### LINKED ISSUE #4380: Accelerator preview min height for next block extra fee required
If the extra fee required to get into the next block is not much on the scale, it's height is too small to fix the text. Consider adding a min height so the text is always readable.

Screenshot 2023-11-15 at 4 57 27 PM

mononaut:
Resolves #4380

Make the next block fee bar always at least 30px high, and hides it when the value is negative (i.e. transactions already paying more than the projected next block fee rate)

| Before | After |
|-|-|
| Screenshot 2023-11-15 at 8 47 12 AM | Screenshot 2023-11-15 at 8 46 47 AM |

### DIFF
--- a/frontend/src/app/components/accelerate-preview/accelerate-fee-graph.component.scss
+++ b/frontend/src/app/components/accelerate-preview/accelerate-fee-graph.component.scss
@@ -18,6 +18,7 @@
       bottom: 0;
       left: 0;
       right: 0;
+      min-height: 30px;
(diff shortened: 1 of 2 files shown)


nymkappa (approved):
Tested ACK @ 19477c4ee308e28750bb8fe31f79582b21e275d3

PR #4386, opened 2023-11-17
Fix electrum p2pk/scripthash format

mononaut:
This PR fixes an issue with P2PK "address" lookups using electrum/fulcrum backends.

P2PKs do not have an address format, so we query their balances and transactions by scripthash.

Very helpfully, the esplora and electrum APIs use different endianness for their scripthashes, so API queries that work for esplora backends return no results from electrum/fulcrum backends.

This PR converts the scripthashes in these queries to little-endian before proxying them to the electrum/fulcrum backend, which resolves the issue.

Before:
Screenshot 2023-11-17 at 3 58 13 AM

After:

Screenshot 2023-11-17 at 3 56 01 AM

### DIFF
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -573,7 +573,9 @@ class BitcoinRoutes {
     }
 
     try {
-      const addressData = await bitcoinApi.$getScriptHash(req.params.scripthash);
+      // electrum expects scripthashes in little-endian
+      const electrumScripthash = req.params.scripthash.match(/../g)?.reverse().join('') ?? '';
+      const addressData = await bitcoinApi.$getScriptHash(electrumScripthash);
       res.json(addressData);
     } catch (e) {
       if (e instanceof Error && e.message && (e.message.indexOf('too long') > 0 || e.message.indexOf('confirmed status') > 0)) {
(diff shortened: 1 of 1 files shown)


junderw:
> Very helpfully, the esplora and electrum APIs use different endianness for their scripthashes

endianness is *always* helpful.

mononaut:
little-endian hex strings should be illegal

junderw:
What are you talking about? Having two ways to interpret data is *even better* when talking about hashes.

You get 2 hashes out of one piece of data for the price of one. It's magical.

softsimon (approved):
tested ACK @ [7717c15]

using `romanz/electrs` v0.10.1

Screenshot 2023-11-17 at 17 30 16

cculianu:
Electrum dev here -- I'm not really familiar with this codebase but a user reported something other than this 0 balance issue -- when drilling down onto a P2PK address, Fulcrum returned an error that the scriptHash is invalid (wrong length).  I audited the code to Mempool but I am not familiar with this codebase at all (nor do I really know TypeScript), but it appears there may be inconsistent usage of `scriptHash` in this file, in particular on this line it appears that an `encodeScriptHash` call needs to be made: https://github.com/mempool/mempool/blob/7717c15666bae20af8a04f45bb84c7a00beeb709/backend/src/api/bitcoin/electrum-api.ts#L170C12-L170C12

This may explain the error the user got -- see [Fulcrum issue #212](https://github.com/cculianu/Fulcrum/issues/212)

mononaut:
No worries, that code is a bit misleading.

Despite the parameter names, `$getScriptHashBalance` and `$getScriptHashHistory` actually take scriptPubKeys (not hashes), and then convert them to electrum-format scripthashes using `encodeScriptHash`.

Whereas, for compatibility reasons, `$getScriptHash` and `$getScriptHashTransactions` take actual scripthashes in esplora-format, which just need to be converted to little-endian.

Sorry about the misdirected bug report, and thanks for taking the time to investigate!

PR #4384, opened 2023-11-16
Fix P2PK balance display in address page

ncois:
This PR fixes a bug that prevents the correct display of the transferred balance in transactions with P2PK.
Before: 
genesis-p2pk
Examples: [uncompressed](https://mempool.space/address/04678afdb0fe5548271967f1a67130b7105cd6a828e03909a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112de5c384df7ba0b8d578a4c702b6bf11d5f) and [compressed](https://mempool.space/address/02f1a8c87607f415c8f22c00593002775941dea48869ce23096af27b0cfdcc0b69) P2PK

Now: 
genesis-p2pk-fixed

Sorry for the duplicate, I wanted to squash commits and ended up closing the previous PR by mistake.

PR #4389, opened 2023-11-17
Fix P2PK balance display in address page (duplicate)

ncois:
Follows #4384 with a new branch name


softsimon (approved):
tested ACK @ [76a07315]

PR #4390, opened 2023-11-17
Fix tooltip overflow by increasing width

### LINKED ISSUE #4359: Mempool graph tooltip overflow
This overflow may occur when skimming through the mempool graph.

Screenshot 2023-11-11 at 12 04 54

ncois:
Fixes #4359 by increasing the width of the tooltips in mempool and incoming tx graph: 
tooltip-mempool
tooltip-incoming-txs


nymkappa (changes requested):
The correct fix woudl be to have a tooltip that resizes properly based on its content. Making it wider just make it looks weird on the 2h charts.

image

Or as an alternative quick fix maybe we can use the 24h formatting for the 2h as well. @softsimon

ncois:
> The correct fix would be to have a tooltip that resizes properly based on its content.

I agree, making the tooltip wider is the most  straightforward but it is not optimal

> Or as an alternative quick fix maybe we can use the 24h formatting for the 2h as well.

I think it is better to keep the 2h formatting, the date would be a little useless on the 2h graph. 

Maybe a compromise would be to set the width of the tooltip based on the graph type (wider for graphs other than 2h)?

ncois:
Now with 2h timeframe: 
2h
And other timeframes: 
24h+

nymkappa (approved):
It's not perfect, definitely needs some more love (auto resizing based on the content instead of hardcoded width).

But this is **definitely better** than what we have in prod right now.

Tested ACK @ c5ce3167f39fc40834a0743f10de56398826577e

PR #4391, opened 2023-11-17
fix: incorrect `HTTP_PORT` docker compose field in docker README.md

0xBEEFCAF3:
The override is BACKEND_HTTP_PORT [defined here](https://github.com/mempool/mempool/blob/59c513f2a52a96d125944b42230e1fb7b13b130e/docker/backend/start.sh#L7)

Just found this while setting up docker compose. 
Idk if it makes sense to update `start.sh` instead. Lmk

### DIFF
--- a/contributors/0xBEEFCAF3.txt
+++ b/contributors/0xBEEFCAF3.txt
@@ -1,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of November 17, 2023.
+
+Signed: 0xBEEFCAF3
--- a/docker/README.md
+++ b/docker/README.md
@@ -124,7 +124,7 @@ Corresponding `docker-compose.yml` overrides:
     environment:
       MEMPOOL_NETWORK: ""
       MEMPOOL_BACKEND: ""
-      MEMPOOL_HTTP_PORT: ""
+      BACKEND_HTTP_PORT: ""
       MEMPOOL_SPAWN_CLUSTER_PROCS: ""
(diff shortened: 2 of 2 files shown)


nymkappa:
I think it's actually a typo in start.sh. It's the only instance of `BACKEND_HTTP_PORT` I can find in the source code.

@knorrium can you check?

knorrium:
I think we'll want to normalize all the variables (so the startup script looks up `MEMPOOL_HTTP_PORT`), but we'll have to update the config unit test and notify the community (especially Umbrel) when we release a new image. This was the original intent when we introduced more overrides, but some integrations were already using `BACKEND_HTTP_PORT` in the wild.

The current Umbrel setup is using `MEMPOOL_HTTP_PORT` in the compose file, but since script actually looks up `BACKEND_HTTP_PORT`, it's always using the default.

I found [this repo](https://github.com/BoltzExchange/boltz-client-python/blob/35939d17733e38e9ce88ead5e1a09f6d499d544b/docker/docker-compose.yml#L171) is using `latest` and `BACKEND_HTTP_PORT`, so we should give them a courtesy ping or perhaps make the script backwards compatible for a few versions before dropping `BACKEND_HTTP_PORT` completely.

knorrium (approved):
utACK, ok to merge to align the doc with reality for now.

nymkappa:
> @knorrium Great. I'll rebase later today. The cla job keep failing even after I added my file. Any ideas why?

That's because the CLA needs to be merged into master before you can open a PR without any warning. This is fine, no worry. Your future PRs won't have this issue.

0xBEEFCAF3:
@nymkappa Rebased!

PR #4396, opened 2023-11-21
Fix mempool graph tooltip width & vb precision

### LINKED ISSUE #4387: Fix decimals for mb at each sat/bytes steps.
image

I'd fix it to 0.00 so we can quickly see small numbers by their visual size. 

Ex:
```
   13 is bigger than 
 3.72
```

mononaut:
Resolves #4387

This PR always displays vsize counts on the mempool graph tooltip with 2 decimal places to improve alignment, and also slightly increases the width of the tooltip to avoid text wrapping with longer date strings.

| Before | After |
|-|-|
| Screenshot 2023-11-21 at 4 35 21 PM | Screenshot 2023-11-21 at 4 35 44 PM |


softsimon (approved):
tested ACK @ [ab079e9372]

PR #4398, opened 2023-11-21
Fix negative left margin on master page

mononaut:
| Before | After |
|-|-|
| localhost_4200_(iPhone SE) (2) | localhost_4200_(iPhone SE) (1) |


softsimon (approved):
tested ACK @ [d471616a2]

PR #4400, opened 2023-11-21
Add a form on address page to display only credit or debit transactions

natsee:
This PR adds a feature to the address page that allows the filtering of transactions based on whether they are credit or debit to the address. I tried to keep the original transactions structure untouched to avoid unwanted behaviour. 
Screenshot 2023-11-21 at 16 55 55

I think I took into account most edge cases, like what if there is no debit transaction, or what if there is no credit transaction loaded in the 50 fetched txs during the scroll, etc. But I might have missed some things, so any feedback is welcome.

### DIFF
--- a/frontend/src/app/components/address/address.component.html
+++ b/frontend/src/app/components/address/address.component.html
@@ -57,11 +57,30 @@
       <h2 class="text-left">
         <ng-template [ngIf]="!transactions?.length">&nbsp;</ng-template>
         <ng-template i18n="X of X Address Transaction" [ngIf]="transactions?.length === 1">{{ (transactions?.length | number) || '?' }} of {{ txCount | number }} transaction</ng-template>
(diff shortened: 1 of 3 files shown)


softsimon:
I haven't reviewed this PR because I suspect it will be too complex of a feature to maintain than that it's worth currently.

The address page is complicated due to various reasons, like we push update the status of transactions live, we add new transactions live and play a sound when it happens. Adding a filter to this might cause issues, and also the filtering of 50 transactions suddenly when switching can have other consequences.

You might already have thought of all these use cases.

ncois:
> I haven't reviewed this PR because I suspect it will be too complex of a feature to maintain than that it's worth currently.

Yes, I understand it's a quite complex feature to deal with. I thought this would be useful for example to look for mining pools transactions that spend all the coinbases accumulated over many transactions.
But I admit that it's a niche use case. So no problem!

softsimon:
Not planned

PR #4401, opened 2023-11-22
Fix truncated hidden text layout flow bug

mononaut:
The hidden text content added to the truncate component in #4368 has a weird effect on document flow that breaks the bottom nav positioning on the tx page when the side menu is open (!?).

This seems to be caused by using `position: absolute` with no directly positioned ancestor.

This PR adds `position: relative` to the outer truncate component, which seems to resolve the issue.


softsimon (approved):
tested ACK @ [94a7b710]

PR #4403, opened 2023-11-22
Fix health score when mempool was empty

mononaut:
We currently assign empty blocks a health score of 0%, even if the mempool may have been genuinely empty when the miner constructed their template.

e.g:
https://mempool.space/testnet/block/000000000000001d6b8026c28825d56acc8a2341b59fd5edfe2acc04bea38df5
Screenshot 2023-11-22 at 1 15 04 PM

This PR explicitly handles that edge case by assigning a health score of 100% if the mempool is empty or only contains recently broadcast transactions.

### DIFF
--- a/backend/src/api/audit.ts
+++ b/backend/src/api/audit.ts
@@ -144,7 +144,12 @@ class Audit {
 
     const numCensored = Object.keys(isCensored).length;
     const numMatches = matches.length - 1; // adjust for coinbase tx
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested NACK 

When mempool is empty `auditBlock` just returns 0 score directly

softsimon (approved):
tested ACK @ [75cc844676]

PR #4404, opened 2023-11-22
Fixes truncated links

softsimon:
Fixes:

* Links wasn't clickable on the left side
* Mobile width wasn't applied

Still some issues like when searching for a TXID or Address, only half of the address looks selected.


mononaut:
Tested ACK @ [dbcd9000]

Confirmed this fixes the unclickable links, and improves the hidden text width.

The width of the highlighted section is still pretty bad in some cases, but maybe best to fix that in a separate PR?

Screenshot 2023-11-22 at 5 02 09 PM

PR #4407, opened 2023-11-22
Standalone multi-block view page

mononaut:
at `/view/blocks`
Screenshot 2023-11-22 at 4 59 58 PM

supports query parameters to adjust formatting/animation:

- `numBlocks`: number of blocks to display
- `blockWidth`: size of each block visualization in px
- `autofit`: same as the single block page I made before, adjusts the grid dimensions to fit the blockWidth exactly
- `padding`: around each block, in px
- `wrap`: wrap to fit the page width (by default they'll stay in a single row)
- `animationDuration`: in ms
- `stagger`: delay between the transition animation for each subsequent block in ms (can set this negative to start from the other end)
- `test`: start at height 800,000 and simulate new blocks every ~10 seconds to test the animation more easily


wiz (changes requested):
Also please remove font

mononaut:
fixed

wiz (approved):
Tested ACK @ v3.0.0-dev [1a732f18f]

PR #4411, opened 2023-11-23
[menu] force background color when menu is open

nymkappa:
https://github.com/mempool/mempool/assets/9780671/afe085ed-07ad-4faa-bdb3-b7fad80d77f0


softsimon (approved):
tested ACK @ [e5d27887

PR #4415, opened 2023-11-23
[ui] only load empty sidebar in foss mainnet

nymkappa:


softsimon (approved):
tested ACK @ [7f4fd83ad]

PR #4419, opened 2023-11-23
[about page] fix wrapping perk

nymkappa:
Conflicts with https://github.com/mempool/mempool.space/pull/422, but easy enough to resolve when needed


softsimon (approved):
utACK

PR #4421, opened 2023-11-24
[about page] move CTA sponsors box into a sub component

nymkappa:
### Testing

Simply make sure the about page sponsors CTA boxes are unchanged vs prod.


frontend/src/app/components/about/about.module.ts
```diff
@@ -25,10 +26,14 @@ export class AboutRoutingModule { }
   imports: [
     CommonModule,
     AboutRoutingModule,
-    SharedModule,
+    SharedModule
```
softsimon, line 29: Leave the comma, so that it won't cause a extra row git change when another module import is added

frontend/src/app/components/about/about-sponsors.component.scss
```diff
@@ -0,0 +1,45 @@
...
+#become-sponsor-container .ng-fa-icon {
+  color: #2ecc71;
+  margin-right: 5px;
+}
+
+#become-sponsor-container .sponsor-feature {
+  text-align: left;
+  width: 250px;
+  margin: 12px auto;
+  white-space: nowrap;
```
nymkappa, line 36: https://github.com/mempool/mempool/pull/4419/files#diff-9daf43654b4e0bb2c925b6396c6e3a7e1b117bd854fe9c46290f025bfe0a762eR285

PR #4433, opened 2023-11-25
Make the acceleration preview dismissable

mononaut:
Currently after clicking on the "Accelerate" button, there's no way to get rid of the acceleration preview without navigating away from the transaction page.

This PR adds an 'X' button to dismiss the acceleration preview, and slightly reduces the width of the error/success alert banner to make room for it:

Screenshot 2023-11-25 at 11 48 00 AM

### DIFF
--- a/frontend/src/app/components/accelerate-preview/accelerate-preview.component.html
+++ b/frontend/src/app/components/accelerate-preview/accelerate-preview.component.html
@@ -239,7 +239,3 @@
     </div>
   </ng-container>
 </div>
-
-<button type="button" class="close dismiss-preview" (click)="onDismiss()">
-  <span aria-hidden="true">&times;</span>
-</button>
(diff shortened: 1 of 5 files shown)


softsimon:
For CPFP, Bowtie, Input/output details on the transaction page we all use the Button as a toggle to reveal a new panel.

So wouldn't it be more consistent (and look nicer) if the Accelerator button also is a toggle.

mononaut:
yeah, I suppose I was thinking the functionality is more like a "close" button than a show/hide toggle.

but maybe the same style as the other buttons would look better.

mononaut:
Rebased on master and changed to the toggle button style:

Screenshot 2023-12-01 at 3 01 59 PM

softsimon (approved):
tested ACK @ [1fe983a29]

PR #4438, opened 2023-11-26
Refactor difficulty reindexing to process blocks in height order

mononaut:
This PR slightly simplifies the `$indexDifficultyAdjustments()` function, and fixes an obscure bug which could fill the difficulty adjustments table with garbage, requiring a manual difficulty reindex.

The bug occurs when there is a gap in the blocks table for some reason when `$indexDifficultyAdjustments()` runs. In that case, an adjustment of `difficulty%` would be inserted into the DB for *every* block height below the gap.

This PR enforces processing blocks sequentially in ascending order during difficulty indexing, which allows us to simplify the reindexing logic (and fix the bug).

### DIFF
--- a/backend/src/api/mining/mining.ts
+++ b/backend/src/api/mining/mining.ts
@@ -15,6 +15,13 @@ import bitcoinApi from '../bitcoin/bitcoin-api-factory';
 import { IEsploraApi } from '../bitcoin/esplora-api.interface';
 import database from '../../database';
 
+interface DifficultyBlock {
+  timestamp: number,
+  height: number,
+  bits: number,
+  difficulty: number,
+}
+
 class Mining {
   private blocksPriceIndexingRunning = false;
   public lastHashrateIndexingDate: number | null = null;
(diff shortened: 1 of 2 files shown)


softsimon (approved):
tested ACK @ [66d88abd]

backend/src/api/mining/mining.ts
```diff
@@ -633,6 +645,17 @@ class Mining {
       default: return 86400 * scale;
     }
   }
+
+  // Finds the oldest block in a consecutive chain back from the tip
+  // assumes `blocks` is sorted in ascending height order
+  private getOldestConsecutiveBlock(blocks: DifficultyBlock[]): DifficultyBlock {
```
nymkappa, line 651: I feel this this duplicate is not necessary. Isn't `BlocksRepository::$getOldestConsecutiveBlock()` serving the same purpose?

mononaut: yes, but that makes another expensive DB query to fetch all of the blocks again

mononaut: maybe we can pull this out into a utility function we can also call from `BlocksRepository::$getOldestConsecutiveBlock()`, to avoid the code duplication?

nymkappa: > maybe we can pull this out into a utility function we can also call from `BlocksRepository::$getOldestConsecutiveBlock()`, to avoid the code duplication?

That seems a good way yes 👍🏻

PR #4439, opened 2023-11-26
Fix spurious ln channel typeahead matches

### LINKED ISSUE #4427: Lightning Channel Search Result Bug
### Description

Any search will return the same lightning channel list

#### Version

Occurring on currently deployed site (5d5e9e8)

### Steps to reproduce

Search any string starting with a letter. If a channel list appears it will be the one shown

### Expected behaviour

Channel list reflects user input

### Actual behaviour

Channel list is always the same

### Screenshots

image

Zion Search
image

Cit Search
image

mononaut:
Fixes #4427

This PR changes the `$searchChannelsById(search)` function to return no channels if the search query is empty after sanitization. Previously we would return the first 10 channels matching that empty query (i.e. just the first 10 channels in the DB).

| Before | After |
|-|-|
| Screenshot 2023-11-26 at 9 24 46 AM | Screenshot 2023-11-26 at 9 22 27 AM |
| Screenshot 2023-11-26 at 9 24 59 AM | Screenshot 2023-11-26 at 9 22 20 AM |

### DIFF
--- a/backend/src/api/explorer/channels.api.ts
+++ b/backend/src/api/explorer/channels.api.ts
@@ -80,7 +80,13 @@ class ChannelsApi {
 
   public async $searchChannelsById(search: string): Promise<any[]> {
     try {
-      const searchStripped = search.replace(/[^0-9x]/g, '') + '%';
+      // restrict search to valid id/short_id prefix formats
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK @ [0b449347c]

PR #4448, opened 2023-11-29
[accelerator] show wait list message in preview when logged in with no access

nymkappa:


softsimon (changes requested):
Seems like I get the "You are currently on the wait list"

Although I am not logged in yet.

nymkappa:
> Seems like I get the "You are currently on the wait list"
> 
> Although I am not logged in yet.

I don't believe you

image

softsimon (approved):
utACK, will test on staging

PR #4449, opened 2023-11-29
Fix some SEO strings

nymkappa:
Spotted few things while doing translations


softsimon (approved):
tested ACK @ [36398ca5]

Issue #4453, opened 2023-11-30
Add % ownership in lightning dashboard


nymkappa:
<img width="519" alt="image" src="https://github.com/mempool/mempool/assets/9780671/f02cd076-fb33-43de-9431-b73e681572ba">

PR #4460, opened 2023-11-30
Add % ownership in capacity table (LN dashboard)

ncois:
#4453 

Screenshot 2023-11-30 at 11 31 47

- Maybe the % value could be a little more small / greyish (like the BTC symbol)
- the % does not appear in the dedicated Liquidity Ranking table for now since that was not asked, but it can be added


nymkappa (changes requested):
Technically works, but need some polish.

### It does not look good visually, here is my suggestion using this simple css on the `%` part, eg:

```css
.capacity-ratio {
  font-size: 12px;
  color: darkgrey;
}
```

image

### Please apply the `%` in the node ranking per channel count as well.

image

### Finally, please also add the `%` in /lightning/nodes/rankings/liquidity` and `/lightning/nodes/rankings/connectivity`.
```

<img width="564" alt="image" src="https://github.com/mempool/mempool/assets/9780671/08666d6c-cfa4-4f5a-a507-c0af72a347f4">

**Comment by natsee** [comment]
Dashboard

Screenshot 2023-12-13 at 15 14 04

`/lightning/nodes/rankings/liquidity`

Screenshot 2023-12-13 at 15 14 23

`/lightning/nodes/rankings/connectivity`

Screenshot 2023-12-13 at 15 15 31

File: frontend/src/app/lightning/nodes-ranking/top-nodes-per-capacity/top-nodes-per-capacity.component.ts [hunk 1]
```diff
@@ -59,6 +62,11 @@ export class TopNodesPerCapacity implements OnInit {
         })
       );
     }
+
+    this.statistics$?.subscribe((data) => {
```
softsimon, line 66: We should try stick to using RxJS observables as much as possible, instead of subscribing and setting variables. Here is an area of refactor.

Remove this code and instead merge the statistics data with the `topNodesPerChannels$` like this:


```
this.topNodesPerChannels$ = combineLatest([
        this.apiService.getTopNodesByChannels$(),
        this.statistics$
      ])
      .pipe(
          map(([ranking, statistics]) => {
          for (const i in ranking) {
            ranking[i].geolocation = <GeolocationData>{
              country: ranking[i].country?.en,
              city: ranking[i].city?.en,
              subdivision: ranking[i].subdivision?.en,
              iso: ranking[i].iso_code,
            };
          }
          ranking['totalChannels'] = statistics.latest.channel_count;
          ranking['totalCapacity'] = statistics.latest.total_capacity;
          return ranking;
        })
      );
```

Just add totalChannels and totalCapacity as optional variables to the ITopNodesPerChannels interface.

The template can now just use `node.totalChannels` and `node.totalCapacity` and drop the NgIf*.

frontend/src/app/lightning/nodes-ranking/top-nodes-per-capacity/top-nodes-per-capacity.component.ts
```diff
@@ -14,11 +14,14 @@ import { LightningApiService } from '../../lightning-api.service';
 })
 export class TopNodesPerCapacity implements OnInit {
   @Input() nodes$: Observable<INodesRanking>;
+  @Input() statistics$: Observable<any>;
```
softsimon, line 17: Consider creating an interface into the `node-api.interface.ts` similar to `ITopNodesPerChannels` above, instead of using `any`. Basically do a json to ts interface on the server response data.

softsimon (changes requested):
It looks good now, just have some code suggestions.

natsee:
I deleted the subscriptions and instead kept observables following your suggestions @softsimon. I just have one comment: 
```         
 for (const i in ranking) {
  ranking[i].geolocation = <GeolocationData>{
    country: ranking[i].country?.en,
    city: ranking[i].city?.en,
    subdivision: ranking[i].subdivision?.en,
    iso: ranking[i].iso_code,
  };
}
ranking['totalCapacity'] = statistics.latest.total_capacity;
ranking['totalChannels'] = statistics.latest.channel_count;
```

This works fine, but isn't it confusing to set a property on `ranking` using this syntax `ranking["totalCapacity"] = ...` while `ranking` is an array?

softsimon:
> I deleted the subscriptions and instead kept observables following your suggestions @softsimon. I just have one comment:
> 
> ```
>  for (const i in ranking) {
>   ranking[i].geolocation = <GeolocationData>{
>     country: ranking[i].country?.en,
>     city: ranking[i].city?.en,
>     subdivision: ranking[i].subdivision?.en,
>     iso: ranking[i].iso_code,
>   };
> }
> ranking['totalCapacity'] = statistics.latest.total_capacity;
> ranking['totalChannels'] = statistics.latest.channel_count;
> ```
> 
> This works fine, but isn't it confusing to set a property on `ranking` using this syntax `ranking["totalCapacity"] = ...` while `ranking` is an array?

You're right.

A solution is to change the return object (`return ranking;`) to something like:

```
return {
   ranking: ranking,
   statatistics: {
     totalCapacity: statistics.latest.total_capacity,
     totalChannels: statistics.latest.channel_count,
  }
}
```

And the template has to be updated to read these sub objects. let node of data.nodes

nymkappa (approved):
Tested ACK @ 5c43c17261ebfdfa6f765460fba434a18649b514
@softsimon please check if the code change you've requested is good for you

softsimon (approved):
tested ACK @ [8cb49ba8]

Issue #4459, opened 2023-11-30
Search for Block by Date / Time


orangesurf:
### Description

Allow jumping to block by date

### Problem to be solved

If you want to find the chain state at a particular date you currently need to use an external tool or scroll through a lot of blocks. 

### Proposed solution

Add common date / time formats as valid search inputs, and use a binary search to identify the block closest to the time / first block at that date (or use a lookup table if only adding support for dates) 

(note: Block time is miner set and can diverge, so block times may not be in block order)

#### Additional info

Example of binary search 
https://gist.github.com/mutatrum/015d0022bc048d6c9ee07f36ab650c31#file-blocktime-sh

nymkappa:
We have all block timestamps indexed in the database. A simple MySQL query would do the job for that.

Note that this does not exactly give the nearest but would be good enough for the purpose of this task.

For example something like this: `SELECT * FROM blocks WHERE timestamp < _formatted user input_ ORDER BY timestamp DESC LIMIT 1`

softsimon:
We already has an API endpoint for this, just undocumented and underutilized.

[/mining/blocks/timestamp/:timestamp](https://mempool.space/api/v1/mining/blocks/timestamp/1354116278)

PR #4462, opened 2023-11-30
Search a block by date and timestamp

ncois:
Implements #4459 by allowing to search a block by date / time or timestamp.

Screenshot 2023-11-30 at 19 14 01

`YYYY-MM-DD` and `YYYY-MM-DD HH-MM` are the two compatible formats. Hyphen `-` or slash `/` can both be used:

Screenshot 2023-11-30 at 19 21 58

This will go to the first block **before** the date-time provided:

Screenshot 2023-11-30 at 19 25 52

It is also possible to search by timestamp:

Screenshot 2023-11-30 at 19 15 40

***

I wasn't sure about what to do if no block corresponds to the entry (example: a date in 2008 or a timestamp in the future, so an error is only raised in the console for now.


nymkappa:
@mononaut tagging you since you recently made improvements to the search feature would be good to get your review 👍🏻

nymkappa (approved):
Works fine on my end. I'll leave the code review for @mononaut and @softsimon.

Tested ACK @ 2584a1f2b00e0808fb068179dd8b52944cda26c8

PR #4465, opened 2023-12-01
Fix Liquid minfee defaults

mononaut:
Sets the default `mempoolminfee` and `minrelaytxfee` to 0.1 s/vb for Liquid, which resolves an issue with the Liquid recommended fees.


softsimon (approved):
utACK

PR #4451, opened 2023-11-29
Difficulty adjustment chart: format the hashrate and difficulty

ncois:
Fixes #4442 by displaying formatted values on all screen sizes

I also removed the rounding for the difficulty value, which caused to always display the difficulty with two zeroes fraction digits on mobile:

| Before (desktop) | Before (mobile) | After (mobile + desktop) |
| ------------- | ------------- | ------------- |
| Screenshot 2023-11-29 at 11 31 38 | Screenshot 2023-11-29 at 11 32 36  | Screenshot 2023-11-29 at 11 34 12|

PR #4466, opened 2023-12-01
Pool chart hashrate format in power of ten

ncois:
For coherence with #4451, this formats in power of ten the hashrate in pools charts (example https://mempool.space/mining/pool/binancepool)

| Before  | After |
|  Screenshot 2023-12-01 at 12 58 21 | Screenshot 2023-12-01 at 12 58 37 |

### DIFF
--- a/frontend/src/app/components/pool/pool.component.ts
+++ b/frontend/src/app/components/pool/pool.component.ts
@@ -163,10 +163,8 @@ export class PoolComponent implements OnInit {
           let hashratePowerOfTen: any = selectPowerOfTen(1);
           let hashrate = ticks[0].data[1];
 
-          if (this.isMobile()) {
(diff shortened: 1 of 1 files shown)


nymkappa (changes requested):
This works fine, but I realized that we should keep some precision longer. For example, when we switch from 999 PH/s to 1EH/s, we look lot of granularity. My suggestion would be in this case to display up to 9999 PH/s before switch to 10EH/s, so we keep a minimum level of details.

Here is an example implementation. We should also apply this idea to other hashrate charts.

Screenshot 2023-12-04 at 11 17 10 AM
Screenshot 2023-12-04 at 11 17 17 AM

Results in a finer details, without having too much either:
Screenshot 2023-12-04 at 11 17 47 AM

PR #4468, opened 2023-12-02
Fix liquid recommended fee rounding

mononaut:
completes the fix in #4465


softsimon (approved):
tested ACK @ [cd8e3e260]

Using Liquid and Mainnet

Issue #513, opened 2021-05-12
Miner detection on address page


AbelLykens:
### Description

If I look at https://mempool.space/address/bc1qx9t2l3pyny2spqpqlye8svce70nppwtaxwdrp4 for example, it would be nice if the mining pool is listed as user of this address.

#### Version
 v2.2.0-dev [934dd673]  (mempool.space)

### Steps to reproduce

View page
See no tag for miner

### Expected behaviour

View page
See tag for miner

### Actual behaviour

View page
See no tag for miner

### Screenshots

![image](https://user-images.githubusercontent.com/47113175/117952259-83f45100-b304-11eb-9dae-5236892261fa.png)
![image](https://user-images.githubusercontent.com/47113175/117952282-8c4c8c00-b304-11eb-8b40-ffc6636056f3.png)

#### Device or machine

FF on Linux

#### Additional info

<3

nymkappa:
Can be implemented once https://github.com/mempool/mempool/pull/3070 is merged because we are now indexing `coinbase_address` for each blocks. So we can easily generate all coinbase addresses for all mining pools (combined with the existing address known from `pools.json`) and display those in the address component.

nymkappa:
@ncois if you're interested in this feel free to try something out

PR #4473, opened 2023-12-03
Display mining pool on address page

natsee:
#513 

This PR: 
- adds an endpoint `mining/pools-addresses` that gives the list of alls miners with their indexed addresses
- if an address is related to a pool, displays its name on the address page

https://mempool.space/block/000000000000000000009ef7ac2c30976bc7e05c2cdab10e5a5de7efd96492a7

https://github.com/mempool/mempool/assets/46578910/33b1f00d-2d2f-4b69-be76-9286f86e2de8
***
There are some addresses that are not indexed in [pools-v2.json](https://github.com/mempool/mining-pools/blob/master/pools-v2.json). For example: 
- Block [200006](https://mempool.space/block/00000000000000c545b3420a66abe8bff779497cdcf10be0aa28cfb52fb2ee0c) is mined by BTCGuild with the coinbase address `14kKEh1sbFrJoByed6ZKGntNHLY47jXNJ7`, but the `BTCGuild` tag does not appear on the [address page](https://mempool.space/address/14kKEh1sbFrJoByed6ZKGntNHLY47jXNJ7).
- Block [700027](https://mempool.space/block/00000000000000000000ca53c6f68c660b89c8c31d5f6e843f7147f141a0feb7) is mined by Foundry with the coinbase address `19dENFt4wVwos6xtgwStA6n8bbA57WCS58` but this one is not indexed either
- etc.

To get the missing addresses I wrote a [script](https://github.com/ncois/mining-pools/blob/index-pools-addresses/scripts/pools-addresses-indexer.py) following @nymkappa's comment on #513: 
> [...] we are now indexing `coinbase_address` for each blocks. So we can easily generate all coinbase addresses for all mining pools (combined with the existing address known from `pools.json`) and display those in the address component.

I haven't had a chance to index all the addresses since the script needs the full `blocks` table in the db which takes time to index. Once it is done I will submit a PR to mempool/mining-pools with all the addresses found.


nymkappa (changes requested):
Thanks, seems to be great start. Quickly testing locally seem to do the job. Few comments.

We could avoid calling this new API endpoint for each address page load since we will return the same thing most of the time, and we don't really need real time update either. Maybe saving the response somewhere (a service? we could use it to tag other addresses later on maybe) to avoid calling this API all the time would be good.

Also you can increase the cache time to 1 hour (currently 60 seconds).

ncois:
> Thanks, seems to be great start. Quickly testing locally seem to do the job. Few comments.
> 
> We could avoid calling this new API endpoint for each address page load since we will return the same thing most of the time, and we don't really need real time update either. Maybe saving the response somewhere (a service? we could use it to tag other addresses later on maybe) to avoid calling this API all the time would be good.
> 
> Also you can increase the cache time to 1 hour (currently 60 seconds).

Thanks for the feedback! 
- I increased the cache time to 1 hour
- I stored the response with all pools addresses in `cache.service` 

I also added a way to display multiple mining pool on the address page when it is related to more than one: 

Screenshot 2023-12-04 at 15 21 27

ncois:
I've turned this PR into a draft because I am trying a different way of implementing the address indexing that will scale better when the number of addresses related to pools increases. Plus it should allow not to miss any address in the indexing.
- Create a table `pools_addresses` containing all addresses associated with one or more pools, and even the amount mined with each pool
- The data would need to be given to the backend the same way pools-v2.json is
- On the address page, a simple request to the backend that would look like `SELECT pools, values WHERE address LIKE our_address` would get all the data we need to display the mining pool(s) associated with the address (instead of querying the list of all pools addresses and then checking if the address is in the list which scales poorly)

Let me know if you have any comments or suggestions about this implementation.

ncois:
I think this PR is now ready for review. 

- I imported the file that indexes all addresses related to mining pools (see https://github.com/mempool/mining-pools/pull/44) to the database in a new table called `pools_addresses` 
- This table is used to make the endpoint `mining/pools-addresses/:address` that gives all pools related to the `address` param
- The miner related to the address is displayed on the address page using the new endpoint

softsimon:
Too much to maintain for little benefit. Probably not going to happen any time soon.

PR #4476, opened 2023-12-04
[accel preview] only show waitlist message if logged in

nymkappa:


wiz (approved):
utACK - will test on staging with master deployment

PR #4481, opened 2023-12-06
Improve acceleration tracking

mononaut:
Fixes some issues with the way we were tracking accelerated transactions.

Before, sometimes the backend would receive the updated list of accelerations *before* the block in which an acceleration was confirmed, and so fail to include the acceleration in the block audit.

This PR takes a similar approach to #3655, by fetching the list of accelerations first, then checking for new blocks, before finally applying the update.

It also switches to use axios directly instead of using the `axios-query` wrapper, which could have caused unnecessary delays and exceptions in the mempool update loop since it's intended for more async tasks.


wiz (approved):
Tested ACK @ b7c4cd43fce1c03e3ebca04b1bd3e377a4b1bd87

PR #4482, opened 2023-12-06
Fix negative accelerated fee rate, simplify fee rate table

mononaut:
This PR fixes an issue where we could show a negative accelerated fee rate for the first few blocks after an accelerated transaction is confirmed.

It also simplifies the accelerated/effective fee rate part of the fee table to reduce duplication and make the template logic slightly easier to understand.


wiz (approved):
Tested ACK @ v3.0.0-dev [a46e779d][5f41af47]

PR #4483, opened 2023-12-06
fix: RTL issues related to sidenav and global footer

TechMiX:
Tried to fix some of the issues introduced by sidenav and the site options in the footer, in right-to-left mode.

Before:

before-small

After:

after-small


nymkappa (changes requested):
Thanks for checking this. Problem is you're working on a very old version of mempool. Things are quite different now. Can you update your fork and update your PR please.

For info, the master branch of your work is from March 13th. https://github.com/TechMiX/mempool/tree/master

TechMiX:
@nymkappa I've already merged the latest master.

TechMiX:
rebased it again.

frontend/src/app/shared/components/global-footer/global-footer.component.scss
```diff
@@ -63,7 +63,6 @@ footer .row.main .links .category:not(:first-child) {
 }
 
 footer .site-options {
-  float: right;
```
TechMiX, line 66: @nymkappa @softsimon I don't know where this needs to be raised, but please use the `float-right` or `float-left` CSS classes on your elements, instead of manually setting it like this. Because if we try to abide by this simple rule, then everything would :sparkles: automagically :sparkles:  be correct in the RTL mode. (thanks to overrides that `.rtl-layout` is doing [here](https://github.com/mempool/mempool/blob/4169e1053ffbdf1ec9fcbe00016db257766fcd5d/frontend/src/styles.scss#L862-L876))

nymkappa (changes requested):
Okay. So what confused me is that empty menu with the `Sign in` button. This is not a thing anymore. It's actually dead code so I need to delete it.

The footer fix is good so we could just merge that already, but the menu is not working. That's expected since you were not aware of the implementation details. See the following screenshot. The menu is in "open" state but nothing shows except a weird black thing to the upper right area.

Screenshot 2023-12-07 at 9 31 12 PM

In order to get something similar to prod, can you try the following (for testing only, please do not commit this of course):

In `menu.component.ts`, at line 37, copy/paste this line, in order to hard code the menu content (you also need to import `of`):

```
this.userMenuGroups$ = of([{"title":"Accelerator","i18n":"","items":[{"title":"History","i18n":"","faIcon":"history","link":"/services/accelerator/history"},{"title":"Overview","i18n":"","faIcon":"wallet","link":"/services/accelerator/overview"}]},{"title":"Account","i18n":"","items":[{"title":"Swag","i18n":"","faIcon":"tshirt","link":"/services/account/swag"},{"title":"Subscriptions","i18n":"","faIcon":"user-clock","link":"/services/account/subscriptions"},{"title":"Settings","i18n":"","faIcon":"wrench","link":"/services/account/settings"},{"title":"Logout","i18n":"","faIcon":"sign-out-alt","link":"logout"}]}]);
```

Also, you need some fake auth. To do so open your browser debugger, and open the localstorage. Please add the following key/value pair:

Key: `auth`
Value: `{"token":"1234","user":{"userId":1,"username":"test1234","imageMd5":""}}`

Those two things should allow you to get something similar to what we have in production so you can work in proper condition.

Thanks a lot!

nymkappa:
As a side note, if you think it's too hacky to make the menu work + main page layout work properly in RTL, please let us know. We are considering re-implementing it using a much simpler but different approach, similar to github menus style, because we have had aleady quite a few issues with the current implementation.

nymkappa (approved):
Thanks a lot and sorry for the confusion. We will have to polish a bit more but it fixes current prod issues.

Tested ACK @ 9406103ed1d895a15895d0e1d5eada60731b7160 (using both build types @wiz).

Screenshot 2023-12-08 at 4 45 27 AM
Screenshot 2023-12-08 at 4 45 39 AM
Screenshot 2023-12-08 at 4 45 48 AM
Screenshot 2023-12-08 at 4 45 56 AM

nymkappa (changes requested):
Thanks a lot. I found one issue left. If you run the project without the hacks I've mentioned in my previous comments, the layout breaks. Also make sure your `mempool-frontend-config.json` has the following settings:

```
  "OFFICIAL_MEMPOOL_SPACE": false,
  "ACCELERATOR": false
```

image

I found it can be fixed using those changes, feel free to apply them.

image

TechMiX:
@nymkappa Fixed the `empty-sidenav` as well.

nymkappa (approved):
Great! I've noticed other issues with RTL but that's out of the scope of this PR. This PR fixes current prod issue so we can merge it.

Tested ACK @ bbcc683a3c2c9db98b67878d4b09105709c66792

softsimon (approved):
tested ACK @ [9406103ed]

Tested with and without the hybrid build.

Thanks a ton for helping out again with the RTL layout @TechMiX

Issue #4486, opened 2023-12-07
Fix message in mining pool page when hashrate is not yet indexed


nymkappa:
`Indexing blocks` is not the correct message to display here.

<img width="1404" alt="image" src="https://github.com/mempool/mempool/assets/9780671/490f904c-575f-471f-8e28-80eb663c68d6">

PR #4487, opened 2023-12-08
[mining] show proper message when we have less than 2 weeks worth of mining pool hashrate data

nymkappa:
Fixes https://github.com/mempool/mempool/issues/4486

Unless we have at least two data point for the mining pool hashrate history chart, show a more genertic `Not enough data yet` info message. This message will also be shown during re-indexing (this is also correct).

Screenshot 2023-12-08 at 5 00 51 PM


softsimon (approved):
tested ACK @ [01131360]

PR #4488, opened 2023-12-09
Acceleration tab

softsimon:
Make sure frontend config has

`  "ACCELERATOR": true,`

image


wiz:
will this fit on mobile devices with small width?

wiz:
@softsimon can you fix it so the rocket is in between mempool and mining dashboards and also the mobile layout concern, maybe we have to reduce the width of the buttons to fit on small devices?

mononaut:
I've opened PR #4586 onto this branch to implement the requested changes, after which I think it might be ready for review?

PR #4497, opened 2023-12-12
Acceleration preview loader

softsimon:
On slow connections there was no indication that the acceleration preview was being loaded.

Screenshot 2023-12-12 at 20 48 35


nymkappa:
It works fine. How about showing some message as well, something like `We are calculating a price for you blabla` @hunicus ?

nymkappa (approved):
Tested ACK @ d3f7190b7f68b6a269eae48bdbc4debad4e986be

hunicus:
> How about showing some message as well, something like `We are calculating a price for you blabla`

To be honest I don't think it's needed.

PR #4500, opened 2023-12-13
Projected block filters

mononaut:
This PR adds basic filtering functionality to the projected block visualizations:

https://github.com/mempool/mempool/assets/83316221/86e9c2fd-9117-4ab4-b667-f84811316bb2

---

### UI

Mousing over/tapping the projected block visualization reveals the "filter" icon button in the top left, which toggles a filter menu overlay.

The menu consists of 24 toggleable filter tags arranged in logical-ish groups:

Screenshot 2023-12-13 at 11 08 49 AM

Any active filters are listed at the top of the visualization next to the filter icon, and continue to show after the menu is closed (so that it's obvious which filters are applied in screenshots etc).

Transactions matching the active filters are displayed in their regular colors, while everything else has opacity reduced to 20% to create a dimming/highlighting effect.

Screenshot 2023-12-13 at 11 06 45 AM

Filters are *conjunctive* - i.e. transactions are highlighted *only* if they match *every* active filter. Perhaps in future we can add more complex logic, or support multiple sets of filters in different colors.

Some pairs of filters are mutually exclusive, and so act like radio buttons (e.g. RBF enabled/disabled and transaction versions).

---

### Implementation

In the backend, mempool transactions are classified according to the 24 different filter criteria, and the result stored compactly as bitwise flags encoded in a single new numeric "flags" field.

Then in the frontend, the transactions in the block visualization are efficiently tested against the selected set of filters by applying the filter to the flags field as a bitmask.

---

### Limitations

So far, the filters are only implemented for projected blocks. Supporting mined blocks will require a lengthy reindexing of block templates, so I want to iterate on and finalize the filter spec first.

I've tried to keep the transaction classification algorithm as efficient as possible, but it does involve iterating over all of the inputs, outputs and witnesses of each transaction. If that turns out to be unacceptably slow we might need to optimize further and maybe disable some of the more expensive filters (like the sighash flags).

The coinjoin, consolidation and batch payment flags rely on fairly arbitrary heuristics. We'll want to iterate on these over time.

For the sake of efficiency, signature detection for the sighash flags also relies on heuristics instead of fully parsing scripts, which may produce occasional false positives.

Many of the filters operate at the input/output level, but they're actually applied to whole transactions. This means some combinations of filters can be misleading/unintuitive - e.g. filtering for `sighash_single` and `sighash_anyonecanpay` will match transactions where one input is signed using `sighash_single` and a different input is signed with `sighash_anyonecanpay`, rather than only transactions with `sighash_single | acp` signatures.

On very small screens, there's limited space within the block visualization area, so we can only show a subset of filter options.

We should probably add an FAQ entry and some tooltips to explain all of this.

---

### Future improvements/extensions
 - Make the active filters persistent when you leave and return to the projected blocks page.
 - Extend filter functionality to mined blocks
 - Add the full list of matched filters to the "features" badge area on the transaction page.
 - Add filter options for other transaction and script types (lightning channel operations, timelocks, multisig etc etc).

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -171,8 +171,8 @@ export class Common {
     // heuristic to detect probable DER signatures
     return (w.length >= 18
       && w.startsWith('30') // minimum DER signature length is 8 bytes + sighash flag (see https://mempool.space/testnet/tx/c6c232a36395fa338da458b86ff1327395a9afc28c5d2daa4273e410089fd433)
-      && ['01, 02, 03, 81, 82, 83'].includes(w.slice(-2)) // signature must end with a valid sighash flag
-      && (w.length === parseInt(w.slice(2, 4), 16) + 6) // second byte encodes the combined length of the R and S components
+      && ['01', '02', '03', '81', '82', '83'].includes(w.slice(-2)) // signature must end with a valid sighash flag
+      && (w.length === (2 * parseInt(w.slice(2, 4), 16)) + 6) // second byte encodes the combined length of the R and S components
     );
   }
 
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -140,6 +140,65 @@ export class Common {
     return matches;
   }
 
+  static setSchnorrSighashFlags(flags: bigint, witness: string[]): bigint {
+    // no witness items
+    if (!witness?.length) {
+      return flags;
+    }
+    const hasAnnex = witness.length > 1 && witness[witness.length - 1].startsWith('50');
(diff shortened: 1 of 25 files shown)


softsimon (approved):
tested ACK @ v3.0.0-dev [14e440da6]

wiz:
ok please DM me when you guys merge this

mononaut:
> The filters are displayed on "Actual block" but not working.

should be fixed now

softsimon (approved):
tested ACK @ v3.0.0-dev [1fb3de9dc]

Issue #4475, opened 2023-12-04
Duplicate API call when loading the lightning dashboard


nymkappa:
1. Open a new empty tab
2. Open the debugger network tab
3. Load https://mempool.space/lightning
4. Observe duplicate api call

<img width="327" alt="image" src="https://github.com/mempool/mempool/assets/9780671/33846753-d9dc-4aad-be97-4a03e8a66ecf">

takuabonn:
hi @nymkappa 
I want to work in this issue. will u assign it to me?

takuabonn:
hi @nymkappa 
this cause is to reuse  LightningStatisticsChartComponent and NodesNetworksChartComponent on lightning-dashboard.component.html.

LightningStatisticsChartComponent and NodesNetworksChartComponent dispatch same API.

because use in same page, Duplicate API call.

how to this resolve  necessary below.
change to use one component in lightning-dashboard.component.html.

softsimon:
> change to use one component in lightning-dashboard.component.html.

We can also use something like `this.apiService.cachedRequest(this.lightningApiService.getLatestStatistics$())` but we might have to add the cachedRequest methods into the `lightningApiService` first.

takuabonn:
thanks @softsimon !
That's an implementation like api.service.ts?

takuabonn:
hi @softsimon, @nymkappa 
cannot pass verification/cla-signed though have sigin in commit.

softsimon:
> hi @softsimon, @nymkappa cannot pass verification/cla-signed though have sigin in commit.

Ignore the bot, it doesn't detect it when it's in the same PR. It should be fine.

takuabonn:
hi @softsimon 
Am I correct that the correction is now complete?

PR #4501, opened 2023-12-13
update to use cachedRequest

takuabonn:
## bugfix
[Duplicate API call when loading the lightning dashboard](https://github.com/mempool/mempool/issues/4475)


softsimon (approved):
tested ACK @ [1fb3de9d]

PR #4502, opened 2023-12-13
Search result: do not offer "Go to" if block height > chain tip

### LINKED ISSUE #4474: Don't offer "go to block" search result if you search for a block height > current block height
Searching for obviously non existing block should not present any result

image

The result leads to this page which is expected but unnecessary.

Screenshot 2023-12-04 at 1 20 01 PM

Instead, why not simply showing an appropriate string in the result area?

image

natsee:
Fixes #4474 

Screenshot 2023-12-13 at 19 24 25

### DIFF
--- a/frontend/src/app/components/search-form/search-form.component.ts
+++ b/frontend/src/app/components/search-form/search-form.component.ts
@@ -174,7 +174,7 @@ export class SearchFormComponent implements OnInit {
           const addressPrefixSearchResults = result[0];
           const lightningResults = result[1];
 
-          const matchesBlockHeight = this.regexBlockheight.test(searchText);
+          const matchesBlockHeight = this.regexBlockheight.test(searchText) && parseInt(searchText) <= this.stateService.latestBlockHeight;
           const matchesDateTime = this.regexDate.test(searchText) && new Date(searchText).toString() !== 'Invalid Date';
           const matchesUnixTimestamp = this.regexUnixTimestamp.test(searchText);
           const matchesTxId = this.regexTransaction.test(searchText) && !this.regexBlockhash.test(searchText);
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/search-form/search-results/search-results.component.html
```diff
@@ -1,10 +1,13 @@
 <div class="dropdown-menu show" *ngIf="results" [hidden]="!results.hashQuickMatch && !results.addresses.length && !results.nodes.length && !results.channels.length">
-  <ng-template [ngIf]="results.blockHeight">
+  <ng-template [ngIf]="results.blockHeight && !results.blockHeightOutOfRange">
     <div class="card-title" i18n="search.bitcoin-block-height">Bitcoin Block Height</div>
     <button (click)="clickItem(0)" [class.active]="0 === activeIdx" type="button" role="option" class="dropdown-item">
       <ng-container *ngTemplateOutlet="goTo; context: { $implicit: results.searchText }"></ng-container>
     </button>
   </ng-template>
+  <ng-template #noBlockFound [ngIf]="results.blockHeight && results.blockHeightOutOfRange">
```
softsimon, line 8: We can remove this part.

We never show "Lightning Channel not found" or similar when posting wrong info ... We just display what it matches.

softsimon (changes requested):
See comment

natsee:
Removed the "No block found" message:

Screenshot 2023-12-14 at 09 42 20

softsimon (changes requested):
I realized we don't need the `blockHeightOutOfRange` stuff anymore.. just adding this to the existing matchesBlockHeight check.


`const matchesBlockHeight = this.regexBlockheight.test(searchText) && parseInt(searchText) <= this.stateService.latestBlockHeight;`

softsimon (approved):
tested ACK @ [b0c02b16]

PR #4505, opened 2023-12-15
Fix overflow on mining pool ranking table

### LINKED ISSUE #4479: UI overflow mining pool ranking
image

natsee:
Fixes #4479 by hiding the Average Health column on small screen width on 24h time window.

| Before  | After |
| Screenshot 2023-12-15 at 12 26 13 | Screenshot 2023-12-15 at 12 26 47 |

Another overflow occured when there were long pool name in the table: it is fixed by truncating the pool name on small screen.
 

| Before  | After |
| Screenshot 2023-12-15 at 11 37 04 | Screenshot 2023-12-15 at 12 27 52 |


nymkappa (approved):
It does not look very balanced overall but it's better than an overflow. Maybe we should align some column on the right instead of having everything the center. But that's another issue.

Tested ACK @ edb4c8d9ad041256dcd839c958dd53f1693d777b
@softsimon

PR #4506, opened 2023-12-15
Mempool Goggles icon, tooltip, FAQ

mononaut:
Adds some polish to the mempool block filters added in #4500 by changing the icon, adding a "beta" tag and an info tooltip which links to a new FAQ entry.

I couldn't find a good "goggles" or "glasses" icon in the usual library, so this uses a custom SVG image.

See also #4510 and #4515.

Screenshot 2023-12-15 at 1 38 36 PM

Screenshot 2023-12-15 at 1 27 32 PM


frontend/src/app/shared/components/svg-icons/goggles.component.ts
```diff
@@ -0,0 +1,34 @@
+import { Component, Input } from '@angular/core';
```
softsimon, line 1: We can't use the current `svg-images-component.html` ?

Jean68172: Yes

mononaut: yes absolutely. I didn't realize what that component did when I checked it earlier - I thought it was just for the bitcoin/liquid network logos.

mononaut:
Rebased on current master and changed the svg icon to use the existing svg-images component

softsimon (approved):
tested ACK @ [78857102f]

PR #4510, opened 2023-12-15
Smoother goggles color change transition

mononaut:
Fixes an issue that caused laggy color transitions when changing goggle filters.

Before:

https://github.com/mempool/mempool/assets/83316221/914b090a-3ad4-4dce-9b94-8931e4d9396c

After:

https://github.com/mempool/mempool/assets/83316221/eae38651-f16d-417e-a441-2980e065219d


softsimon (approved):
tested ACK @ [7bdb2f41]

PR #4512, opened 2023-12-15
Acceleration sounds

softsimon:
Not sure about the detection of a confirmed transaction, whether it actually was accelerated or not.


frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -439,7 +439,11 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
           block_time: block.timestamp,
         };
         this.stateService.markBlock$.next({ blockHeight: block.height });
-        this.audioService.playSound('magic');
+        if (this.accelerationInfo && ['accelerating', 'mined', 'completed'].includes(this.accelerationInfo.status)) {
```
mononaut, line 442:
```suggestion
        if (this.tx.acceleration || (this.accelerationInfo && ['accelerating', 'mined', 'completed'].includes(this.accelerationInfo.status))) {
```

mononaut (commented):
The "acceleration requested" sounds plays well, but I couldn't get it to trigger the "acceleration successful" sound - I think we just need to check if the transaction is currently being accelerated, in case the acceleration history hasn't updated yet.

mononaut (approved):
Tested ACK @ 1d74a99d2ffaecef14f15124451e258dc883f906

PR #4515, opened 2023-12-17
Faster goggles

mononaut:
Optimizes transaction classification by:
 - Checking if transactions have already been classified before re-running the static analysis.
 - Recording replacements as they occur, instead of performing an expensive RBF cache lookup at classification time.


softsimon (approved):
tested ACK @ [e8c48f6f2]

PR #4523, opened 2023-12-20
fix bug in backend Docker `start.sh` script

jamesblacklock:
In the mainline, when `ESPLORA_UNIX_SOCKET_PATH` is not specified, it gets the literal string value `"null"`. Then the Esplora host gets set to `config.ESPLORA.UNIX_SOCKET_PATH || config.ESPLORA.REST_API_URL`, i.e. `"null"`, which is truthy.

My fix is to default `ESPLORA_UNIX_SOCKET_PATH` to an empty string instead.

I also added the copied files from the `docker/init.sh` script to `.gitignore` which seems appropriate.


jamesblacklock:
@knorrium This has been open for a month. Is there no interest in this PR? This is an actual bug! The Docker `start.sh` script will not work if you are using Esplora for the backend!

jamesblacklock:
@knorrium

knorrium (approved):
Tested ACK @ 770b6df

PR #4525, opened 2023-12-21
Fix missing mempool fee chart when webgl disabled

mononaut:
When webgl is disabled, the next block visualization is hidden and the `/mempool-block/...` pages only show the fee distribution chart.

However the transaction data required for that chart was requested by the block visualization component, so the chart would show up empty in that case.

This PR fixes the websocket subscription to ensure the correct data is available whether or not webgl is enabled.

| Before | After |
|-|-|
| Screenshot 2023-12-21 at 1 05 48 PM | Screenshot 2023-12-21 at 1 05 38 PM |


softsimon (approved):
tested ACK @ [0370cc89]

Thanks for noticing this and fixing

PR #4526, opened 2023-12-21
Mined Goggles

mononaut:
This PR extends Mempool Goggles functionality to mined blocks and block audits.

Changes:
 - The backend now includes Goggles classification flags in the newly saved block summaries (we were already saving the flags in audit templates).
 - The Goggles UI is enabled on mined block and audit visualizations when classification data is available.
 - The users' choice of Goggles filters now persists while navigating around between different block visualizations.
 
This PR **_does not_** yet reindex older blocks and templates to add Goggles classification data. I will implement Goggles reindexing separately.

Screenshot 2023-12-21 at 4 24 36 PM


softsimon (approved):
tested ACK @ [82920f6]

PR #4534, opened 2023-12-28
Fix duplicate address in search results

natsee:
This PR avoids to display both Bitcoin Address "Go to ..." and Bitcoin Addresses when searching an address: 

| Before  | After |
| Screenshot 2023-12-28 at 15 04 46 | Screenshot 2023-12-28 at 15 19 40 |


softsimon (approved):
tested ACK @ [5291828e]

PR #4541, opened 2023-12-31
Optimize websocket updates on new block

mononaut:
This PR optimizes the websocket updates performed in `handleNewBlock`.

In particular:
 - Uses a dictionary lookup instead of `.indexOf(...)` to check if subscribed transactions have confirmed.
 - Eliminates some large intermediate variables to reduce garbage collection.
 - If the mempool block delta is very large, prefer sending a full copy of the new mempool block instead (saves including the long list of `removed` txids).

The PR also adds a new periodic log message which lists the numbers of active websocket subscriptions (for `track-tx`, `track-rbf` and `track-mempool-block`), to help identify where further optimization might be most useful.

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -282,8 +282,7 @@ class WebsocketHandler {
           }
 
           if (Object.keys(response).length) {
-            const serializedResponse = this.serializeResponse(response);
-            client.send(serializedResponse);
+            client.send(this.serializeResponse(response));
           }
         } catch (e) {
           logger.debug(`Error parsing websocket message from ${client['remoteAddress']}: ` + (e instanceof Error ? e.message : e));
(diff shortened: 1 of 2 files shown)


wiz (approved):
Tested ACK @ v3.0.0-dev [23ececca9]
Tested ACK @ v3.0.0-dev [23ececca][1a294f8d]

@mononaut one thing I noticed is that if I view a mempool block and then click on a TX, the stats show `track-tx: 1, track-mempool-block: 1` but if I just go to the main dashboard and click on a random TX the stats show `track-tx: 1, track-mempool-block: 0` as expected - is there a chance (or perhaps reason why) we're not unsubscribing from the mempool-block after clicking thru to a TX page?

mononaut:
hmm, that's definitely possible. I will investigate.

mononaut:
yep #4542

PR #4542, opened 2024-01-01
unsubscribe from mempool block tracking

mononaut:
Adds a missing `stopTrackMempoolBlock()` call when navigating away from the projected mempool block page.


softsimon (approved):
tested ACK @ [c1785993a]

Fixes the missing unsubscribe when webgl is disabled.

Issue #4529, opened 2023-12-27
Can not search for a block by hash on signet


natsee:
On signet, using the search bar to search for a block by hash fails. See below with block [175513](https://mempool.space/signet/block/000000fff24a5235cf5fced785f7c208bd4e519b8ef32c5cef8fae18c3e24fd7): 

<img width="501" alt="Screenshot 2023-12-27 at 15 49 44" src="https://github.com/mempool/mempool/assets/46578910/d8f92479-e48c-4a86-85c2-f542fe9fc4e2">

When proceeding to the search:

<img width="1083" alt="Screenshot 2023-12-27 at 15 52 22" src="https://github.com/mempool/mempool/assets/46578910/326ec300-73c2-4b82-aa3d-2f463195199c">

This happens because the block hash regex used in the search form component requires 8 leading zeroes to match, whereas block hashes in signet usually have 5 or 6 (difficulty around 0.003). 

A solution could be to set `regexBlockhash = /^[0]{5}[a-fA-F0-9]{59}$/` on signet.

softsimon:
There's a work in progress to fix this here https://github.com/mempool/mempool/pull/2401

PR #4543, opened 2024-01-01
Search bar feature: network based regex and address matches

natsee:
- Builds on [[Feature]: Improve Search bar with better regex and network based address matches](https://github.com/mempool/mempool/pull/2401)
- Fixes [Can not search for a block by hash on signet](https://github.com/mempool/mempool/issues/4529)
- Includes [Fix duplicate address in search results](https://github.com/mempool/mempool/pull/4534)

This PR introduces custom regexes so that the search bar is now aware of the current network: 

Example on mainnet: 

| Before  | After |
| Screenshot 2024-01-01 at 14 06 41 | Screenshot 2024-01-01 at 14 07 45  |
| Screenshot 2024-01-01 at 14 07 09 (Still on mainnet) | Screenshot 2024-01-01 at 14 08 22 (Switched network to testnet)|

It is also possible to search for cross-environment addresses:
- search a liquid address on mempool: 

https://github.com/mempool/mempool/assets/46578910/56755af3-c48c-489f-8659-142d7ae331b7

- search a liquid testnet address on bisq: 

https://github.com/mempool/mempool/assets/46578910/b70dbcb9-7586-4e0d-812d-99f9c97e0de5


softsimon (changes requested):
We need to check the frontend flags (`TESTNET_ENABLED` etc) if the networks are available before we let users route to them. If they are not available either not show the search result, or show them but grey it out and don't let users navigate to it.

A default Mempool install don't have anything but Mainnet.

softsimon (approved):
tested ACK @ [0722c221c]

PR #4537, opened 2023-12-29
Highlight accelerations in block previews & non-audit mode

mononaut:
This PR extends acceleration highlighting to the regular (non-audit) block visualization, and the corresponding opengraph preview images.

Screenshot 2023-12-29 at 2 19 41 PM
Screenshot 2023-12-29 at 2 19 25 PM

PR #4545, opened 2024-01-02
Acceleration flag check on block/tx view

softsimon:
https://github.com/mempool/mempool/pull/4537 broke the block visualization view if Acceleration flag is not enabled.

Also fixes an error on the transaction page when a transaction is confirmed.


mononaut (approved):
Tested ACK @ a498f920

Issue #4547, opened 2024-01-03
Liquid: page /assets/featured unreachable


natsee:
### Description

I can not access the page https://liquid.network/assets/featured by clicking on the "Featured" button:

<img width="618" alt="Screenshot 2024-01-03 at 14 11 29" src="https://github.com/mempool/mempool/assets/46578910/6367064e-d82c-4aba-8c86-d5d3b43c55bd">

It automatically redirects to https://liquid.network/assets/all.

It is probably a routing problem, I can look into it.

PR #4548, opened 2024-01-03
[Liquid] Fix missing assets/featured route

natsee:
Fixes https://github.com/mempool/mempool/issues/4547

The featured assets page is now reachable: 

Screenshot 2024-01-03 at 14 23 22


softsimon (approved):
tested ACK @ [f73350f]

PR #4554, opened 2024-01-06
Handle unmineable transactions in GBT implementations

mononaut:
Fixes an edge case that could cause GBT implementations to get stuck when the mempool contains unmineable transactions (e.g. testnet transactions larger than 3996000 WU).


softsimon (approved):
tested ACK @ [30d58d9]

jaonoctus (commented):
ACK

PR #4555, opened 2024-01-06
Limited GBT

mononaut:
### Summary

This draft PR is a reimplementation of the core idea from #3071, to improve GBT performance when the mempool is very large.

When `config.MEMPOOL.LIMIT_GBT` and `config.MEMPOOL.USE_SECOND_NODE_FOR_MINFEE` are both set to `true`, the backend behaves as follows:
  - On each mempool update, the node additionally fetches the a list of all txids in the minfeenode's  mempool.
  - A "candidate set" is calculated, defined as the intersection of the primary and minfeenode mempools.
  - Set reconciliation is performed to identify transactions entering and leaving the candidate set on each update.
  - The full GBT algorithm is run on the candidate set only (i.e. at most 300MB of transactions)
  - All remaining transactions are added (unsorted) to the "stack of blocks" without any CPF or effective fee rate data.
  - When users request CPFP data for transactions that are in the main mempool but not in the minfeenode (i.e. purged transactions), that information is calculated on-demand.
    - Applies to both `/api/v1/cpfp/:txid` REST API requests and websocket track-tx subscriptions.
    - Some anti-DoS measures are enforced:
        - A maximum of 50 in-mempool relatives are considered.
        - CPFP results are cached for the entire cluster.
        - CPFP data is calculated a maximum of once per minute per cluster.
    - Only a very small fraction of the purged mempool is ever queried, and the results are mostly irrelevant except to provide CPFP data and effective fee rates on the transaction page, so this should save a lot of wasted effort.

---

One other major change with this PR is that the old `getSimonTemplate` GBT implementation is completely removed, along with the `ADVANCED_GBT_AUDIT` AND `ADVANCED_GBT_MEMPOOL` config settings.

Now that Rust GBT is stable, there's really no reason to keep the old implementation, and it no longer seems worth the maintenance burden when adding new GBT-related features.

The non-Rust advanced GBT options is preserved to support users who can't run Rust for whatever reason, and that option also supports `LIMIT_GBT`.

---

### TODO

There are a few things left to complete before this is ready to leave draft:
 - [x] Handle accelerated transactions below the purge rate.
 - [x] Test the various combinations of backend and config options.
 - [x] Verify that the projected templates are still sufficiently good with `LIMIT_GBT` enabled.
   - No noticable difference from full GBT templates so far.
 - [x] Check that calculating CPFP data on-demand is not too expensive
   - With 8000 active `track-tx` websocket subscriptions, on-demand CPFP calculations still consume no measurable CPU time.


softsimon:
@mononaut please rebase.

we need this, but disabled for now

mononaut:
rebased on current master

softsimon (approved):
partially tested ACK [3732406c]

Without being enabled it doesn't break current behavior. Only deprecating the GBT settings making it default.

PR #4562, opened 2024-01-09
Journald

nikicat:
This PR adds support for logging via journald on Linux.

### DIFF
--- a/backend/package-lock.json
+++ b/backend/package-lock.json
@@ -20,6 +20,7 @@
         "mysql2": "~3.6.0",
         "redis": "^4.6.6",
         "rust-gbt": "file:./rust-gbt",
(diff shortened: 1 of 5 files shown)


wiz (changes requested):
NACK for adding new npm dependencies

nikicat:
> NACK for adding new npm dependencies

Is it bad? Why? Is it some kind of project policy?

wiz:
Yeah we're trying to remove npm deps, not add more

nikicat:
So how this or a similar feature should be implemented? By vendoring 3rd party lib or other way? Are there some written policies about it?

wiz:
Yeah sorry, if you can implement it without using third party npm deps then please make a new PR

PR #4566, opened 2024-01-10
Clean rust-gbt directory before build

mononaut:
Deletes `*.node`, `index.d.ts`, `index.js` files and the `target` directory in the `rust-gbt` folder before building the rust module, to ensure the latest version is always built.


softsimon (approved):
tested ACK @ [8336a00ae]

Ciornenichii:
**
Ciornenichii **

Yappadappa (commented):
What bonus

PR #4578, opened 2024-01-13
Add seconds to the block view timestamp

knorrium:
Small feature request from Twitter:

Screenshot 2024-01-13 at 1 52 40 PM
Screenshot 2024-01-13 at 1 52 29 PM


mononaut (approved):
Tested ACK @ [65236c7b]

This makes it much easier to compare close block timestamps - I've resorted to checking the raw API responses in the past to check sub-minute differences.

PR #4579, opened 2024-01-14
Add seconds to block timestamp in the mining pool view and blocks list

knorrium:
Same as #4578 but on the mining pool view and blocks list

Screenshot 2024-01-13 at 6 10 02 PM
Screenshot 2024-01-13 at 6 09 41 PM
Screenshot 2024-01-13 at 6 07 44 PM
Screenshot 2024-01-13 at 6 03 09 PM


softsimon (approved):
tested ACK @ [a8442a34]

PR #4583, opened 2024-01-15
Fix goggles inscription detection bug

mononaut:
Mempool Goggles currently fails to correctly identify certain strangely constructed inscriptions, e.g. the vertical band of small txs around the selected transaction in the screenshot are all inscriptions, but not identified as such.

Screenshot 2024-01-15 at 2 18 26 AM

All of the misidentified transactions are signature-less inscriptions without any extra witness data, like:
https://mempool.space/tx/c6f4eb5c3c485ba4a76ae1bdd2f049552be307ff86036885b8a299bd4aaf0cb5

It turns out that the classification code assumed valid taproot script path spends always have at least three witness items (the control block, the script itself, and a signature).

But of course script spends without signature checks don't need to include that third witness item.

This PR fixes the bug by more explicitly validating the conditions for a script path spend, including logic to handle the presence/absence of a taproot annex.

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -263,8 +263,13 @@ export class Common {
         case 'v0_p2wsh': flags |= TransactionFlags.p2wsh; break;
         case 'v1_p2tr': {
           flags |= TransactionFlags.p2tr;
-          if (vin.witness.length > 2) {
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK @ [00b8f001b0]

PR #4559, opened 2024-01-08
[bitcoin core] add internal routes to bitcoin core rpc

nymkappa:
Context https://github.com/mempool/mempool.space/issues/512

PR #4584, opened 2024-01-15
[bitcoin core] Add internal routes to bitcoin core

nymkappa:
Redo of https://github.com/mempool/mempool/pull/4559 after removing it from the mega branch

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-core.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin-core.routes.ts
@@ -1,0 +1,221 @@
+import { Application, NextFunction, Request, Response } from 'express';
+import logger from '../../logger';
+import bitcoinClient from './bitcoin-client';
+
+/**
+ * Define a set of routes used by the accelerator server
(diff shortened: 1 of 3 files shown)


backend/src/api/bitcoin/bitcoin-core.routes.ts
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * Define a set of routes used by the accelerator server
+ * Those routes are not designed to be public
+ */
+class BitcoinBackendRoutes {
+  private static tag = 'BitcoinBackendRoutes';
+
+  public initRoutes(app: Application) {
+    app
+      .get('/api/internal/bitcoinCore/' + 'getMempoolEntry', this.disableCache, this.$getMempoolEntry)
```
softsimon, line 14: Perhaps we shouldn't leave the current kebab-casing standard and move to a camelCase standard? `bitcoin-core`

softsimon (approved):
utACK

PR #4594, opened 2024-01-20
Liquid: add indexing process of Federation utxos

natsoni:
This draft implements the indexing of all of the Liquid Federation UTXOs backing the circulating L-BTC on Liquid. 

All of the UTXOs that supplied the coins backing the peg-ins transactions are already indexed in the table `elements_pegs`. It is worth noting that these UTXOs are spent by the Federation shortly after their creation, but generally to their own wallets. See https://help.blockstream.com/hc/en-us/articles/900002386446-How-does-the-Liquid-Federation-s-multisig-work under the section "Timelock Refresh" to understand how the Federation manages these UTXOs and why they are regularly spent or "refreshed". 

From these "genesis" UTXOs, the goal is to get all of the currently held UTXOs by the Federation to determine the total supply of BTC backing the L-BTC: this is not a trivial task since, as said above, the UTXOs held by the Federation change regularly. 

To do so, we need to scan the history of these UTXOs referenced in `elements_pegs` and _follow their spending_ until we find unspent outputs. This allows to compute the exact number of BTC held by the Federation from the inception of the Liquid network up to now. 

We store all of the UTXOs that the Federation ever held in a new table called `federation_txos`: 

| txid     | txindex | bitcoinaddress | amount | blocknumber | blocktime      | unspent | lastblockupdate |
| -------- | ------- | -------------- | ------ | ----------- | -------------- | ------- | --------------- |
| 0090f...  | 1      | 3EiAcrz...   | 50331 | 707244 | 1635499995  | 0 |  708543 | 

This row shows the output `0090f63ccfaed6dfd9e077c6e2fcc80209159feeabf1cc72074bff28c098d7c7:1` that was held by the Federation on the address `3EiAcrzq1cELXScc98KeCswGWZaPGceT1d` since block 707244. It was spent on block 708543. 

The indexing process involves scanning Bitcoin blocks all the way to the tip of the blockchain. It populates the table `federation_txos` which contains all of the UTXOs that the Federation ever held, and it takes around a day to complete. Once completed, indexing follows the growth of the blockchain with very little computation.

⚠️: This PR bumps the schema version to 67 to create the tables `federation_addresses` and `federation_txos` which uses the former as a foreign key. The new schema also empties the `elements_pegs` table to re-index it during the initial sync.


backend/src/api/database-migration.ts
```diff
@@ -558,6 +558,21 @@ class DatabaseMigration {
       await this.$executeQuery('ALTER TABLE `statistics` ADD min_fee FLOAT UNSIGNED DEFAULT NULL');
       await this.updateToSchemaVersion(66);
     }
+
+    if (databaseSchemaVersion < 67 && config.MEMPOOL.NETWORK === "liquid") {
+      // Drop and re-create the elements_pegs table
+      await this.$executeQuery('DROP table IF EXISTS elements_pegs;');
```
softsimon, line 564: Instead of drop and recreate... `TRUNCATE TABLE elements_pegs` should be enough

natsee: Yes truncate would be better, but in this case I also added a primary key on `(txid, txindex)` to `elements_pegs` so drop / create also adds the primary key to the table

softsimon: In that case it's better to do a TRUNCATE and then
`ALTER TABLE elements_pegs ADD PRIMARY KEY (txid, txindex);`
To make the migration flow consistent.

PR #4597, opened 2024-01-21
Cache build assets to prevent rate limiting

knorrium:
This PR adds a new step to the CI workflow that runs sync-assets, zips them, upload them as an artifact for the rest of the build and caches it for future runs.

The e2e tests have been flattened into this workflow so only one action runs now.

The sync-assets.js script was also updated with a small fix to the path resolution.


softsimon (approved):
tested ACK

PR #4605, opened 2024-01-23
Goggle Indexing

mononaut:
_(builds on PR #4526)_

This PR adds an indexing stage to retroactively add Goggles classification flags to block summaries and projected block templates, which enables the Goggles view on historic mined blocks.

Classification requires full copies of the relevant transactions, so this is only enabled with `mempool/electrs` backends, and takes a very very long time to complete (although it should speed up a bit once it reaches blocks from before the audit feature was enabled, and continue accelerating into the earlier parts of the blockchain).

It should be possible to parallelize a lot of this for maybe a 3-5x speedup by maxing out electrs threads and doing the classification in separate nodejs workers, however this would probably render the mempool/electrs instance unusable for normal workloads.

Another alternative is not to preemptively classify historic blocks at all, and simply fill in the data lazily as and when blocks are requested.

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -7,7 +7,7 @@ import cpfpRepository from '../repositories/CpfpRepository';
 import { RowDataPacket } from 'mysql2';
 
 class DatabaseMigration {
-  private static currentVersion = 66;
+  private static currentVersion = 67;
   private queryTimeout = 3600_000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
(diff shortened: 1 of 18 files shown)


backend/src/api/database-migration.ts
```diff
@@ -558,6 +558,14 @@ class DatabaseMigration {
       await this.$executeQuery('ALTER TABLE `statistics` ADD min_fee FLOAT UNSIGNED DEFAULT NULL');
       await this.updateToSchemaVersion(66);
     }
+
+    if (databaseSchemaVersion < 67) {
```
softsimon, line 562:
```suggestion
    if (databaseSchemaVersion < 67  && isBitcoin === true) {
```

mononaut:
Fixed issues related to missing block summaries and non-bitcoin networks

softsimon (approved):
tested ACK @ [82920f62]

PR #4608, opened 2024-01-23
Fix version update & error handling in Goggles indexing

mononaut:
Fixes some bugs in the Goggles indexing code from #4605:
 - The new template/summary "version" column was not updated properly, so indexing would repeat indefinitely
 - Added some missing error handling


softsimon (approved):
tested ACK @ [8ca2b2b5]

Issue #4606, opened 2024-01-23
Channel search results showing up during TX search


softsimon:
This is due to the channel search regex that clears out characters turning

`a5f4fa1e76c503ee6b07660e530ac24b6e03daae894c5cc8c1e1e3d4d8c2eff6`

.. into a search for `LIKE 5%`

<img width="406" alt="Screenshot 2024-01-23 at 17 37 47" src="https://github.com/mempool/mempool/assets/8561090/f9af9184-9464-4f10-990d-f18db3570836">

afahrer:
Hello, I'd like to work on this issue if possible.

afahrer:
Opened PR #4609

softsimon:
Fixed in https://github.com/mempool/mempool/pull/4609

PR #4609, opened 2024-01-24
Channel search results showing up during TX search

afahrer:
Fix for #4606


afahrer:
Transaction ID examples with new regex.

image

image

Examples with old regex.

image

image

These would have queried for lightning channels with the wildcard character.

'5%'
image
'59%'
image

softsimon (approved):
tested ACK

PR #4612, opened 2024-01-24
Update rust-toolchain to 1.65

mononaut:
Updates the rust-toolchain version to 1.65, to resolve the following build/install error:
```
npm ERR! error: package `napi-derive-backend v1.0.58` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.63.0
```

Also updates the rust-gbt npm/cargo versions, and pins `@napi-rs/cli` to version "2.16.1".


softsimon (approved):
tested ACK @ [4e007acd8]

PR #4615, opened 2024-01-25
Fix halving fireworks

mononaut:
- Fixes and improves the "special block" fireworks effect
- Runs the fireworks for 8 blocks, instead of for 60 minutes after the special block is mined (or after the page is loaded while the special block is at the tip)

Before:

https://github.com/mempool/mempool/assets/83316221/14c635d1-6a0c-46c6-877e-6b217c96b480

After:

https://github.com/mempool/mempool/assets/83316221/ebc64cb3-241c-41a6-9052-cf431612fec4

### DIFF
--- a/frontend/src/app/components/start/start.component.html
+++ b/frontend/src/app/components/start/start.component.html
@@ -2,8 +2,9 @@
 
 <ng-container *ngIf="specialEvent">
   <div class="pyro">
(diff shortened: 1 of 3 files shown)


softsimon (approved):
tested ACK @ [dbe7ebe53]

PR #4619, opened 2024-01-27
[mining] update halving widget to be more human readable

nymkappa:
~I've included a copy/pasted small javascript that helps formatting the string http://countdownjs.org/demo.html~ _fixed by mononaut_

Need feedback on the UX. Hovering with the mouse will show the block count left. Also @mononaut can you confirm if `blocksUntilHalving - 1 === 0` condition is correct for `one block left before halving` ?

### Right now

image

### At height 839999

(uses average block time)

image

### At height 840000

image

### At height 900000

image

### At height 1000000

image


softsimon:
ACK for the idea and design.. I want this.

NACK for the implementation using a new 32kB file to generate the time left. It has to be possible with current code and at the same time use our already i18n translated strings for months, weeks, minutes, seconds etc.

mononaut:
> can you confirm if blocksUntilHalving - 1 === 0 condition is correct for one block left before halving ?

I'd be tempted to do `blocksUntilHalving === 1`, but yes that's correct.

I agree with @softsimon, concept ACK, but I'd prefer to just use the `<app-time>` component instead of adding the whole countdown.js file.

nymkappa:
> > can you confirm if blocksUntilHalving - 1 === 0 condition is correct for one block left before halving ?
> 
> I'd be tempted to do `blocksUntilHalving === 1`, but yes that's correct.
> 
> I agree with @softsimon, concept ACK, but I'd prefer to just use the `<app-time>` component instead of adding the whole countdown.js file.

Does the `app-time` supports calculating multiple elements? I did not dig into it much but it seemed to be that it was only capable to displaying one unit at a time.

mononaut:
no it doesn't, but if we really need the double units it shouldn't be too difficult to add support.

mononaut:
I just opened a quick PR to this branch with a possible way of implementing this via `<app-time>`: https://github.com/mempool/mempool/pull/4625

softsimon (approved):
tested ACK @ [2d189ca]

PR #4621, opened 2024-01-27
Fix acceleration list css, move to /acceleration/list

mononaut:
Fixes the responsive layout for the standalone Acceleration List page, and changes the URL to `/acceleration/list`.

Before:
Screenshot 2024-01-27 at 4 47 43 PM

After:
Screenshot 2024-01-27 at 4 48 24 PM


softsimon (approved):
utACK

PR #4623, opened 2024-01-27
Goggles indexing config & rate limit

mononaut:
This PR adds a `config.MEMPOOL.GOGGLES_INDEXING` config setting to enable/disable indexing Goggles data for historic blocks and templates (defaults to false, but enabled in the bitcoin mainnet production config).

The PR also adds a 250ms timeout after saving a template or block summary during Goggles indexing, to reduce the strain on MySQL.


softsimon (approved):
tested ACK @ [09e38ac1]

PR #4625, opened 2024-01-27
dual-unit app-time halving countdown

mononaut:
_(targets PR #4619)_

An alternative approach to #4619, extending the `<app-time>` component for dual-unit time strings instead of using countdown.js.

This reuses the existing translation strings, but the way the two units are combined probably won't be grammatically correct in some other languages.

Screenshot 2024-01-27 at 7 15 57 PM


nymkappa:
Thanks. Let me test it now.

nymkappa (approved):
It works fine, give slightly different results but that's good enough for me, can always be improved later if we want more accuracy. Thank you!

PR #4626, opened 2024-01-27
Fix pools dominance chart

mononaut:
This PR fixes the pools dominance chart, which is currently broken when any pools mined blocks in some periods but not others.

The PR also explicitly sorts pools in the chart by % hashrate.

Before:
Screenshot 2024-01-27 at 10 13 31 PM

After:
Screenshot 2024-01-27 at 10 25 32 PM

### DIFF
--- a/frontend/src/app/components/hashrates-chart-pools/hashrate-chart-pools.component.ts
+++ b/frontend/src/app/components/hashrates-chart-pools/hashrate-chart-pools.component.ts
@@ -11,6 +11,13 @@ import { MiningService } from '../../services/mining.service';
 import { download } from '../../shared/graphs.utils';
 import { ActivatedRoute } from '@angular/router';
 
+interface Hashrate {
+  timestamp: number;
+  avgHashRate: number;
+  share: number;
+  poolName: string;
+}
+
 @Component({
   selector: 'app-hashrate-chart-pools',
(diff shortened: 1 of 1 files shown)


nymkappa (changes requested):
I don't have this bug locally, so I guess that's what you mean by "explicitly sorts pools", the result can have some randomness?

I think the new ordering is not good either. What we should aim for is keeping the color (the ordering) consistent between all timespan. I think that's how I was doing it before, ordering by their latest weekly hashrate share.

With your PR, Foundry is red on the 6m chart and at the bottom, but on the all timespan, it becomes light blue and is moved in the middle of the chart, making it very hard to track things.

Screenshot 2024-01-28 at 09 42 14

Screenshot 2024-01-28 at 09 42 22

mononaut:
This PR sorts by average % hashrate over the whole chart, but yeah I'll switch to sorting by % hashrate in the most recent data point so the ordering/colors are stable across different timespans (although we'll need some sort of stable tie-breaker for pools with equal/zero hashrate in the latest point).

nymkappa:
Just go with a good old alphanumeric tie break hehe

mononaut:
Updated with stable sorting

nymkappa (approved):
Tested ACK @ 2637636f27306e5a1d55d0600cb54a72bfbcec18

PR #4628, opened 2024-01-28
Display pool tags on dashboard

softsimon:
Trying out this idea of always displaying the pool tags, except when we focus on a block, dim the tag behind instead of removing it so it doesn't look too weird.

The reason is so I don't constantly have to switch to the miner dashboard and back, just to see the pool name.

Default dashboard
Screenshot 2024-01-29 at 01 51 07

Block open

Screenshot 2024-01-29 at 01 52 07

### DIFF
--- a/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
+++ b/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
@@ -56,7 +56,7 @@
               <app-time kind="since" [time]="block.timestamp" [fastRender]="true" [precision]="1" minUnit="minute"></app-time></div>
           </ng-container>
         </div>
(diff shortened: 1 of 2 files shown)


mononaut:
Concept ACK

I was thinking recently about letting users choose between regular or mining-dashboard-style blocks everywhere, but this is a much better solution.

I'm not so sure about dimming/removing the tag on selected blocks - it would be nice to see who mined a transaction at a glance from the tx page, for example. Perhaps we could shift the tag down below the arrow, or move the rest of the page content down a bit to make room?

nymkappa:
Might as well just completely hide it since the miner tag is shown in the summary right below anyway.

mononaut:
not on the transaction page though

wiz (approved):
utACK

PR #4629, opened 2024-01-29
ops: Add new prod servers for fallback/replication

wiz:
@mononaut can the health checks handle this many servers? I remember we previously had issues where it would fallback before completing all the health checks, so I imagine adding more servers would cause that bug to reproduce more


wiz:
also, should we enable replication for the other non-mainnet networks?

mononaut:
> @mononaut can the health checks handle this many servers? I remember we previously had issues where it would fallback before completing all the health checks, so I imagine adding more servers would cause that bug to reproduce more

I think it should be fine. If it tries to failover before the first health check is complete, it'll just continue using the default mempool/electrs backend.

We make all of the health check requests concurrently, so that could be a limitation. Otherwise the only effect should be that the health check loop takes a bit longer and runs slightly less often.

mononaut:
> also, should we enable replication for the other non-mainnet networks?

Sure, why not. Although I haven't stress tested audit replication for e.g. testnet block storms, so I don't know how it'll handle that.

PR #4631, opened 2024-01-29
Fix missing optional chain on bestDescendant access

mononaut:
Should fix an occasional error in the frontend:
```
ERROR TypeError: Cannot read properties of null (reading 'bestDescendant')
```

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -540,7 +540,7 @@
               </ng-container>
             </ng-template>
           </div>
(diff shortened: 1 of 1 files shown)


Alirezadd65:
mandado: ST24012600249231584040092

use o PIX para Pagar 10000 R$

Abra o app com sua chave PIX cadastrada, escolha Pagar com Pix e escaneie o QR Code ou copie e cole o texto

softsimon (approved):
utACK

Couldn't find a real world example to reproduce but it should now work if `bestDescendant` is missing from the cpfpInfo

PR #4632, opened 2024-01-30
Handle Redis errors and disconnects

mononaut:
This PR refactors the Redis cache module to better tolerate client errors, automatically reconnect when necessary, and catch up with missed cache operations once Redis is available again.

This should prevent the main loop from getting stuck waiting for cache operations to complete while Redis is unavailable.


softsimon (approved):
tested ACK @ [086ae697]

PR #4639, opened 2024-01-31
Fix error handling in acceleration data polling

mononaut:
Fixes an issue with API polling on the accelerator dashboard that prevents future updates after any request fails.


softsimon (approved):
tested ACK @ [0bf0d79ee]

Issue #4630, opened 2024-01-29
Liquid: Frontend undefined crash


softsimon:
Going directly to any random older TX, a javascript error occurs. Although different in Firefox and Chrome. In Firefox when trying to go to the Dashboard right after, it is broken.

On current prod build v3.0.0-dev [[6553de53b](https://github.com/mempool/mempool/commit/6553de53b)]

1. Open https://liquid.network/tx/df1efeb8dbdb85856a69b79cc45cd18219f87d31fdd945724480997d928a6339
2. Wait 2 seconds.. and check the JS console for errors
3. Navigate to the Dashboard

Firefox
<img width="2141" alt="Screenshot 2024-01-29 at 14 42 33" src="https://github.com/mempool/mempool/assets/8561090/0f2bf2fc-eb47-4b3b-8648-23fb877d1cfc">

Chrome

<img width="565" alt="Screenshot 2024-01-29 at 14 46 01" src="https://github.com/mempool/mempool/assets/8561090/067ea517-a608-46c3-bdda-a1d068d1f11b">

PR #4640, opened 2024-02-01
Fix Liquid frontend crash

natsoni:
Fixes https://github.com/mempool/mempool/issues/4630 by adding a check for the mempool blocks length in the mempoolBlocks subscription of the transaction component.

The bug only happened if there is no transaction in the mempool; that explains why it only manifested on Liquid 😅.


softsimon (approved):
tested ACK @ [229e122d]

PR #4642, opened 2024-02-02
Multiple acceleration arrows

mononaut:
This draft PR estimates the position of accelerated transactions in different pool mempools, and displays each as a separate arrow in the mempool blocks.

If any pool accepted all active accelerations, then we know exactly where each transaction should be, so the corresponding arrows are placed using the GBT-calculated positions.

The PR also calculates an approximate un-accelerated mempool position, and displays the regular white arrow in that position.

https://github.com/mempool/mempool/assets/83316221/4f1c3725-07bd-4f01-b3c9-dcb890bdc69d

drawbacks:
 - the accuracy of the un-accelerated arrow is pretty bad, especially when the transaction is one of many with a similar fee rate.
 - most of the time accelerated transactions will end up in more or less the same place in all accepting pools, and so the accelerating arrows will strongly overlap.
     - maybe it would be better to simply only ever show one arrow for the original position and one for the accelerated position, and accept that this won't be a perfect representation in more complex conditions.

### DIFF
--- a/backend/src/api/mempool-blocks.ts
+++ b/backend/src/api/mempool-blocks.ts
@@ -7,6 +7,7 @@ import { Worker } from 'worker_threads';
 import path from 'path';
 import mempool from './mempool';
 import { Acceleration } from './services/acceleration';
+import PoolsRepository from '../repositories/PoolsRepository';
 
 const MAX_UINT32 = Math.pow(2, 32) - 1;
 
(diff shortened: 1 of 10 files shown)


wiz:
@softsimon please don't let these PRs go ignored, either request changes or merge and iterate later in future PRs, but don't let the PR go stale and have merge conflicts 😅

mononaut:
If I can get a Concept ACK, I'll tidy this up and get it ready to merge.

I think the decisions to make are:
 - do we want this feature at all?
 - if so, separate arrows for each pool partner (like this PR) or just a single generic accelerated arrow?
 - if separate arrows, should they be labelled by pool (bearing in mind that they'll normally heavily overlap).

softsimon:
Let's scrap this idea because an arrow can never represent when a certain pool is going to mine a block.

Instead add some kind of power boost animation from the actual arrow all the way signifying that it's being "boosted".

image

PR #4646, opened 2024-02-04
[sync-assets] More logging, add MEMPOOL_CDN and DRY_RUN options

knorrium:
- Improved logging (better indentation of log entries)
- Add new DRY_RUN option for local runs (still hits the GH API for the assets list)
- Add new MEMPOOL_CDN option to download assets from the CDN instead of GH


softsimon (approved):
tested ACK @ [195eeaa7]

PR #4652, opened 2024-02-06
Fix infinite address scroll

mononaut:
The address page currently stops trying to load more transactions after fetching `chain_stats.tx_count`.

This PR fixes that and only stops trying to load more after we actually reach the end of the API pagination.

### DIFF
--- a/frontend/src/app/components/address/address.component.ts
+++ b/frontend/src/app/components/address/address.component.ts
@@ -31,8 +31,7 @@ export class AddressComponent implements OnInit, OnDestroy {
   addressLoadingStatus$: Observable<number>;
   addressInfo: null | AddressInformation = null;
 
-  totalConfirmedTxCount = 0;
-  loadedConfirmedTxCount = 0;
+  fullyLoaded = false;
   txCount = 0;
   received = 0;
   sent = 0;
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/address/address.component.ts
```diff
@@ -252,16 +248,20 @@ export class AddressComponent implements OnInit, OnDestroy {
...
     this.electrsApiService.getAddressTransactions$(this.address.address, this.lastTransactionTxId)
       .subscribe((transactions: Transaction[]) => {
-        this.lastTransactionTxId = transactions[transactions.length - 1].txid;
-        this.loadedConfirmedTxCount += transactions.length;
-        this.transactions = this.transactions.concat(transactions);
+        if (transactions && transactions.length) {
+          this.lastTransactionTxId = transactions[transactions.length - 1].txid;
+          this.transactions = this.transactions.concat(transactions);
+        }
+        if (transactions?.length == null || transactions.length < 50) {
```
softsimon, line 262: This address transaction API in romanz/electrs mode via nodejs, we only fetch like 10 transactions at a time which might cause an issue here?

mononaut: fixed

PR #4653, opened 2024-02-06
Staggered esplora fallback health checks

mononaut:
This PR makes esplora fallback health checks sequentially separated by a 50ms delay to avoid spamming too many simultaneous requests when there are many fallback servers.


wiz (approved):
utACK

PR #4654, opened 2024-02-07
fix syntax error in mempool-config.json

jamesblacklock:
fixed syntax error in docker/mempool-config.json that was introduced a few weeks ago


knorrium (approved):
Added a CI step to validate the JSON, which was used to validate this change.

Tested ACK @ [1547c2c](https://github.com/mempool/mempool/pull/4654/commits/1547c2c4775330b639a8ad072e87df6e6db615b0) and [95233f0](https://github.com/mempool/mempool/pull/4654/commits/95233f0ef378e6394d7d2f3b17533a84984e9798)

PR #4655, opened 2024-02-07
modify docker/init.sh for macOS compatibility

jamesblacklock:
These minor modifications to `docker/init.sh` make the script work under macOS while preserving compatibility with Linux.


wiz (approved):
utACK

PR #4656, opened 2024-02-07
configurable prefix for API backend URL

jamesblacklock:
This change enables straightforward configuration of a "prefix" for mempool URLs.

For a working example, please see https://jamesblacklock.net/bitcoin-mempool

Note that resources are loaded from, e.g., https://jamesblacklock.net/bitcoin-mempool/resources/favicons/favicon.ico

This is achieved by using relative paths instead of absolute paths and modifying the "base href."

Note also that API calls are sent to https://jamesblacklock.net/bitcoin-mempool/api/v1/[METHOD]

This works because of the newly added `API_PREFIX` config.

By default, of course, nothing changes. But this makes it much easier to use a URL prefix across the whole application. I thought this could likely be useful to others. Open to suggestions on how to improve this.


jamesblacklock:
It's not passing CI, but it works for me locally just fine.

If there is interest in merging this change, I will look into resolving the CI issues. If not, I will not bother. Lmk.

wiz (commented):
I don't think it's nearly this easy to implement this feature without breaking lots of things :)

jamesblacklock:
> I don't think it's nearly this easy to implement this feature without breaking lots of things :)

Why do you say that? It doesn't seem so difficult...

wiz:
> It's not passing CI, but it works for my deployment just fine.

This about sums it up :)

jamesblacklock:
Okay, AFAICT there is no interest in this PR from the mempool team. Therefore I will close it.

PR #4657, opened 2024-02-07
Add a step to validate the JSON config generated for Docker

knorrium:
CI should fail if there are errors in the resulting JSON for Docker installs


softsimon (approved):
utACK

PR #4658, opened 2024-02-07
Blockchain component: display dash when feerate is undefined

natsoni:
This PR adds a placeholder in the blockchain component in the case there is no fee data for the block (e.g. on Liquid).

Before: 

Screenshot 2024-02-07 at 11 34 52

After: 

Screenshot 2024-02-07 at 11 33 49


softsimon (approved):
tested ACK @ [df7f1cc86b]

PR #4670, opened 2024-02-08
Reduce the network size of mempool block websocket updates

mononaut:
This PR reduces the network size of the data sent for the next block mempool visualizations by packing the transaction objects into compact arrays.

We could optimize this further by cutting out some of the non-essential fields (fee, value) and replacing the txid with a truncated version or a short internal uid, but the additional complexity that requires doesn't seem worth it at this point.


softsimon (approved):
tested ACK @ [46f2509c]

PR #4672, opened 2024-02-09
[auth] properly refresh user when `auth` object changes

nymkappa:
This PR aims to fix the following bug. You can easily reproduce it locally as well.

### Repro the bug

- Go to mempool.space
- Make sure you are logged out
- Open https://mempool.space/login
- Login with Twitter or your credentials

Result: user name and subscription level badge is not showing, a page refresh is required

image

### DIFF
--- a/frontend/src/app/services/services-api.service.ts
+++ b/frontend/src/app/services/services-api.service.ts
@@ -4,7 +4,7 @@ import { HttpClient, HttpParams } from '@angular/common/http';
 import { StateService } from './state.service';
 import { StorageService } from './storage.service';
 import { MenuGroup } from '../interfaces/services.interface';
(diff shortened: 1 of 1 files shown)


frontend/src/app/services/services-api.service.ts
```diff
@@ -59,6 +64,13 @@ export class ServicesApiServices {
     }
 
     this.getUserInfo$().subscribe();
+    this.router.events.subscribe((event) => {
```
softsimon, line 67: How about this style instead :)

```
this.router.events
      .pipe(
        filter((event) => event instanceof NavigationStart && this.currentAuth !== localStorage.getItem('auth')),
        switchMap(() => this.getUserInfo$()),
      )
      .subscribe();
```

softsimon (approved):
tested ACK @ [6933d3c39]

PR #4676, opened 2024-02-10
Log mempool stats on new blocks

softsimon:
With this PR, we run the mempool statistics log twice when a new block arrives to reflect in the graph what exactly happened to the mempool queue in the graph. The once a minute log is still in effect but it will skip if the last block was found within the last 30 seconds to reduce datapoints.

This will lead to up to 288 (+20%) more data points per day in the `statistics` table.

The result is this cleaner looking mempool graph, at least in the 2H view.

Before
Screenshot 2024-02-10 at 20 51 19

After
Screenshot 2024-02-10 at 20 42 56

Zoomed in / Filtered fee bands 
Screenshot 2024-02-10 at 21 01 14

### DIFF
--- a/backend/src/api/statistics/statistics-api.ts
+++ b/backend/src/api/statistics/statistics-api.ts
@@ -285,7 +285,7 @@ class StatisticsApi {
 
   public async $list2H(): Promise<OptimizedStatistic[]> {
     try {
-      const query = `SELECT *, UNIX_TIMESTAMP(added) as added FROM statistics ORDER BY statistics.added DESC LIMIT 120`;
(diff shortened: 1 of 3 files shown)


mononaut (changes requested):
Concept ACK, but as implemented this causes the API to return slightly wrong timespans for the 2hr and 24hr statistics.

I've opened a PR to this branch with an attempted fix: #4681

mononaut:
https://github.com/mempool/mempool/pull/4681

mononaut (approved):
Tested ACK @ 6dedecad

PR #4678, opened 2024-02-11
Temporarily disable tagging images as latest

knorrium:
Temporary workaround to skip tagging images as `latest`, in preparation for a patch release.

I'll submit a follow up PR to rebuild the image and tag as `latest` with a release workflow soon (tm).


wiz (approved):
utACK

PR #4685, opened 2024-02-12
Add Goggles filter for nVersion=3

mononaut:
Adds a Goggles filter toggle for v3 transactions.

Screenshot 2024-02-12 at 8 08 25 PM

Now that v3 transaction policy has been merged into Bitcoin Core, we may start to see version 3 transactions appear on testnet and signet.

The feature isn't yet included in any tagged releases, and I understand nVersion=3 transactions will remain non-standard for now, so it may be a while yet before any v3 transactions hit mainnet, but it doesn't hurt to be prepared.


softsimon (approved):
utACK

PR #4693, opened 2024-02-18
Add Public-Pool to pools.json

benjamin-wilson:
Added public-pool to pools.json. 

Is there anything else to be done so the pool can be queried like https://mempool.space/mining/pool/FoundryUSA ?


softsimon:
The file you updated is only used in the testing framework.

Use this repo for updates to the mining pool identification

https://github.com/mempool/mining-pools

PR #4698, opened 2024-02-19
Continue other indexing tasks while Goggles classification runs

mononaut:
Currently other indexing tasks (hashrates etc) are prevented from making progress until Goggles classification completes.

This PR runs the Goggles indexing in the background instead so that other tasks can continue to run every hour.


softsimon (approved):
tested ACK @ [69dc21d2]

PR #4700, opened 2024-02-19
Add pagination to Lightning nodes list and fix table overflow

### LINKED ISSUE #4661: Add pagination to Lightning nodes list (country and isp)

natsoni:
Fixes #4661

Screenshot 2024-02-19 at 17 09 13

We still fetch all of the nodes data on load to fill the map and the header table. The pagination allows not to display a buggy table with 3000+ entries.


softsimon (approved):
tested ACK @ [17697e66]

PR #4708, opened 2024-02-24
Add Acceleration Goggles

mononaut:
This PR adds a new Goggles filter for accelerated transactions.

Screenshot 2024-02-24 at 8 28 04 PM

This doesn't change the backend classification, but rather incorporates our existing acceleration data into the frontend filters.

If the `ACCELERATOR` setting is disabled in the frontend, then the filter will not appear.


wiz (approved):
utACK

PR #4712, opened 2024-02-26
Add non-standard Goggles filter

mononaut:
This PR adds a new Goggles filter for non-standard transactions.

This PR does **not** bump the block template version, so the new filter will not apply to blocks mined before deployment without manually triggering a re-index (e.g. by resetting the version column to zero in `blocks_summaries`).

Screenshot 2024-02-26 at 1 51 32 AM

Once this is deployed, most non-standard transactions will be detected and filterable in mined blocks.

The PR detects the following kinds of non-standardness
 - version > 2
 - `tx-size` (more than 100kvb)
 - `tx-size-small` (non-witness size < 65 bytes)
 - `bad-txns-too-many-sigops`
 - `scriptsig-size` (scriptsig > 1650 bytes)
 - `scriptsig-not-pushonly` non-push opcodes in scriptsig
 - `bad-txns-nonstandard-inputs` due to excessive P2SH sigops or non-standard type
 - `scriptpubkey`
     - unknown output type
     - bare multisig with non-standard threshold
     - op_return over default 83 byte datacarriersize limit
 - `dust`
 - `multi-op-return`

But does not yet implement
 - `bad-txns-nonstandard-inputs` due to failed P2SH script evaluation
 - `bad-witness-nonstandard` (witness script resource usage limits)
 - `non-mandatory-script-verify-flag` (non-standard DER encodings or non-null dummy arguments)

The filter is currently hidden/disabled on projected blocks (since we're running default mempool policy).

---

### TODO

I suspect people will want to know *why* a given transaction is non-standard. Perhaps we could record the reason in a new transaction flag field, or re-run the standardness tests client-side to identify and display the specific reasons on the transaction page.

If Libre Relay gains more traction among miners, it might be worth switching our nodes over too to get the widest possible view of pending transactions. In that case, we should enable the non-standard filter on projected blocks too.

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -7,6 +7,24 @@ import { isIP } from 'net';
 import transactionUtils from './transaction-utils';
 import { isPoint } from '../utils/secp256k1';
 import logger from '../logger';
+import { getVarIntLength, opcodes, parseMultisigScript } from '../utils/bitcoin-script';
+
+// Bitcoin Core default policy settings
+const TX_MAX_STANDARD_VERSION = 2;
+const MAX_STANDARD_TX_WEIGHT = 400_000;
+const MAX_BLOCK_SIGOPS_COST = 80_000;
+const MAX_STANDARD_TX_SIGOPS_COST = (MAX_BLOCK_SIGOPS_COST / 5);
+const MIN_STANDARD_TX_NONWITNESS_SIZE = 65;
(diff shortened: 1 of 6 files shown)


wiz:
Please fix failing tests

mononaut:
fixed

softsimon (approved):
tested ACK @ [eadbc139]

PR #4715, opened 2024-02-26
[accelerator] improve acceleration list/dashboard

nymkappa:
To merge alongside https://github.com/mempool/mempool.space/pull/656

- Added pagination support in `/acceleration/list`
- Added a `Requested` column to know when the acceleration request was sent
- Differentiate "mined" and "completed"
- Use `container-lg` instead of `container-xl` so there is not too much empty space, until we add more data in there
- Cleaned up a bit the responsiveness (`/acceleration/list` and in the dashboard widgets `/acceleration`). TXID link is always visible now
- Compute 1m stats into the backend

image

Screenshot 2024-02-26 at 16 41 46

Screenshot 2024-02-26 at 16 41 28


nymkappa:
Putting back into draft because I actually need to fix the stats calculation before this can be merged.

mononaut (approved):
Tested ACK @ 867e9594

PR #4718, opened 2024-02-28
Accelerator audit

mononaut:
This PR adds a service to independently calculate, save and serve data related to the Mempool Accelerator, including the "Bid Boost" out-of-band fees for accelerated transactions.

This allows local instances to audit the accelerator and determine the out-of-band fees paid per-transaction, per-block and per-pool.

The new data is available via three new API endpoints:
 - `/api/v1/accelerations/pool/:slug`
 - `/api/v1/accelerations/block/:height`
 - `/api/v1/accelerations/recent/:interval` (accepts the standard statistics API interval strings like '24h', '1w', '1m' etc)

which return an array of objects for any matching accelerations, like
```
[
  ...
  {
    txid: string,
    height: number,
    pool: {
      id: number,
      slug: string,
      name: string,
    },
    effective_vsize: number,
    effective_fee: number,
    boost_rate: number,
    boost_cost: number,
  },
  ...
]
```

The new data is not currently exposed anywhere in the frontend. I plan to add entries on the _block_, _mining pool_ and maybe _transaction_ pages in a separate PR.

_The new functionality currently requires `config.MEMPOOL_SERVICES.ACCELERATIONS` to be enabled, as the Bid Boost calculations require access to the list of accelerated transactions and their maximum bids._

_We can implement a fully trustless version where accelerations are detected and an approximate Bid Boost inferred without this dependency in a future PR._

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -7,7 +7,7 @@ import cpfpRepository from '../repositories/CpfpRepository';
 import { RowDataPacket } from 'mysql2';
 
 class DatabaseMigration {
-  private static currentVersion = 68;
+  private static currentVersion = 69;
   private queryTimeout = 3600_000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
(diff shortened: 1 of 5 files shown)


backend/src/api/database-migration.ts
```diff
@@ -1123,6 +1128,23 @@ class DatabaseMigration {
     ) ENGINE=InnoDB DEFAULT CHARSET=utf8;`;
   }
 
+  private getCreateAccelerationsTableQuery(): string {
+    return `CREATE TABLE IF NOT EXISTS accelerations (
+      txid varchar(65) NOT NULL,
+      added date NOT NULL,
```
softsimon, line 1134: Should the `date` value not be more specific than down to the day?

mononaut: yes, it should be a datetime. I'll fix that.

softsimon (approved):
tested ACK @ [f923ee4ed]

PR #4721, opened 2024-02-29
Only display acceleration tab on mainnet

### LINKED ISSUE #4719: [menu] accelerator should be mainnet only
Anything accelerator related should be visible/accessible on mainnet only (?)

image

image

softsimon:
fixes #4719


softsimon:
self tested ACK

PR #4723, opened 2024-02-29
Liquid dashboard: Filter for expired UTXOs - Match layout with main dashboard

natsoni:
This PR:
- adds to the Liquid dashboard an overview of expired UTXOs of the Federation
- classifies expired UTXOs as dust depending on their value and the current fee rate
- removes expired UTXOs from the holdings of the Federation
- adds data about usage of the emergency keys by Blockstream since Liquid started
- makes the Liquid dashboard layout fit the mempool dashboard

⚠️ need a re-indexing the Federation UTXOs table

Screenshot 2024-02-29 at 15 06 41

*********************

Screenshot 2024-02-29 at 15 08 32


wiz:
Instead of a checkbox to view the expired UTXOs, how about another tab ?

wiz:
Also "expired" is probably a bit confusing to use here, perhaps "timelock expired" or something hmm

natsoni:
> Instead of a checkbox to view the expired UTXOs, how about another tab ?

Sure :) 



> Also "expired" is probably a bit confusing to use here, perhaps "timelock expired" or something hmm

Yep, just need to make sure it fits on all layouts

natsoni:
Replaced the checkbox with a tab under the button `Timelock-Expired UTXOs`:
Screenshot 2024-03-04 at 20 39 25

Also added a tooltip on the main dashboard: 
Screenshot 2024-03-04 at 20 41 40

softsimon (approved):
tested ACK @ [e8c5d478f]

PR #4729, opened 2024-03-01
[accelerator] fix accel bid boost history chart

nymkappa:
Built on top of/includes https://github.com/mempool/mempool/pull/4728

Screenshot 2024-03-01 at 12 57 27

I've added the `3m` button timeframe as well in that chart

Screenshot 2024-03-01 at 12 57 41

### If there is nothing to show

image


wiz (approved):
utACK

PR #4733, opened 2024-03-03
Hide effective rate on tooltip where irrelevant

mononaut:
PR #4670 rounded the `rate` field in mempool block websocket updates to 2 decimal places, which broke some naive logic in the tooltip component for detecting whether the "effective fee rate" ought to be displayed.

This PR adds a bit of tolerance to that condition, and then also checks the CPFP goggles flags to correct any false negatives.

| Before | After |
|-|-|
| Screenshot 2024-03-03 at 4 49 23 PM | Screenshot 2024-03-03 at 4 49 09 PM |


wiz (approved):
utACK

PR #4736, opened 2024-03-05
Show accelerator audit total on block page

mononaut:
This PR makes use of some of the new locally audited acceleration info APIs added in #4718 to display any additional out-of-band fees on the block page:

Screenshot 2024-03-05 at 12 52 54 AM

The out-of-band fees also contribute to the % fee difference.

### DIFF
--- a/backend/src/repositories/AccelerationRepository.ts
+++ b/backend/src/repositories/AccelerationRepository.ts
@@ -56,6 +56,9 @@ class AccelerationRepository {
   }
 
   public async $getAccelerationInfo(poolSlug: string | null = null, height: number | null = null, interval: string | null = null): Promise<PublicAcceleration[]> {
+    if (!interval || !['24h', '3d', '1w', '1m'].includes(interval)) {
+      interval = '1m';
+    }
     interval = Common.getSqlInterval(interval);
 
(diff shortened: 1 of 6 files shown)


wiz:
Concept ACK - is this ready for review?

mononaut:
yes, ready for review.

I'm not 100% sure this is the best way to display this data, but we can iterate.

softsimon (changes requested):
The call for accelerations on the block page needs to be limited to prod with services backend.

wiz:
Right, it should be using the backend-configured hostname which defaults to mempool.space
Also, it needs to respect if the accelerator feature is disabled

wiz:
And it seems a test is failing

mononaut:
The purpose of this is to showcase the new open-source calculation/audit of acceleration fees, so it doesn't rely on the services API, and should work on any instance where `config.MEMPOOL_SERVICES` is properly configured to fetch public acceleration data from mempool.space.

If that's disabled on the backend, then those new APIs return `[]` and nothing is displayed on the frontend.

But yeah maybe we should add a new frontend config setting to enable/disable local non-services acceleration features (like this, highlighting accelerations in the block visualizations, etc), separate from the acceleration dashboard and official accelerator stuff.

mononaut:
fixed

wiz (approved):
utACK

PR #4737, opened 2024-03-05
Bump rust-gbt deps

junderw:
This will update deps for rust-gbt in hopes that it makes things better.


wiz (approved):
Tested ACK @ 0f1434e32

softsimon (commented):
tested ACK @ [0f1434e3]

PR #4740, opened 2024-03-05
Polish /nodes and /network pages

mononaut:
Adds ranking numbers and flag emojis, fits the pages to mobile screens, makes iframes full-width, and fixes some other UI issues.
Screenshot 2024-03-05 at 11 17 12 PM
Screenshot 2024-03-05 at 11 18 24 PM


wiz (approved):
utACK

PR #4743, opened 2024-03-06
Fix: Refresh Liquid icons every 15 seconds to match electrs.

junderw:
This will fix the problem where icons aren't getting refreshed.


softsimon (approved):
tested ACK @ [1df5c5df6]

PR #4744, opened 2024-03-06
More status page polish

mononaut:
- Moves status pages to /nodes and /monitoring
- Adds a native mempool | blockchain component to the top of each page
- Removes pages titles
- Adds a "last updated" column (larger screens only)

/monitoring
Screenshot 2024-03-06 at 7 35 03 PM

/nodes
Screenshot 2024-03-06 at 7 35 14 PM


wiz (approved):
utACK

PR #4745, opened 2024-03-06
Fix missing mempool blocks

mononaut:
This PR works around an Angular bug which can cause the mempool blocks component to fail to render in certain conditions.

That happens most often after switching to the block page when there is only one block pending and relatively little mempool activity (which is most of the time on e.g. signet and liquid).

The bug seems to be caused by buggy change detection when using `async` pipes inside the `ngForOf` directive, so the workaround is to subscribe to the observable in an `ngIf` instead, one level higher.

The change to the `mempoolBlocks$` observable isn't as substantial as it looks from the diff - it just removes the `merge(
      of(true),
      fromEvent(window, 'resize')
    )` part, which doesn't really do anything right now, removes the superfluous `markForCheck()` call from the end, and then fixes the indentation.
    
Before:

https://github.com/mempool/mempool/assets/83316221/0d5d2898-411a-405d-a0f4-91a342c1f16a

After:

https://github.com/mempool/mempool/assets/83316221/ae1d5f81-f740-48ec-97e5-b69059049b69


softsimon (approved):
tested ACK @ [3178d30f2a]

PR #4747, opened 2024-03-07
Use fx rates to add more currencies support and improve fiat prices 

natsoni:
Builds on #4723 to avoid db versioning conflicts

This PR adds support to 26 new fiat currencies (full list of at the end) by using a free fx rates API ([freecurrencyapi.com](https://freecurrencyapi.com)).
It also fills the missing values for historical prices of current currencies in the db using fx rates (e.g. if `BTCJPY` data is not available, set `BTCJPY = BTCUSD * USDJPY`)

**Pros:** 
- more currencies supported = good 👍 
- allows to give more relevant fiat values for older data: 
As far as I know, currently if a historical data for `BTCJPY` is not set (say on December 2010), the fiat value displayed for this period is `historical BTCUSD * current USDJPY` which is inaccurate because `USDJPY` fluctuates between 2010 to today. To fix this issue, this PR displays `historical BTCUSD * historical USDJPY` instead. 

**Cons:** 
- `historical-price` request more than quadrupled (!) in size, from 250 KB -> 1.1 MB. To limit this problem, maybe we don't need to add the full list of 26 currencies?

- the need for an external api key adds complexity

**Limits:** 
- the currency conversion rates have yearly granularity to prevent exceeding the external API's limits
- for some currencies (eg ABC), `BTCABC` does not exactly equals `BTCUSD * USDABC`

--------------------------------
26 added currencies: 
- BGN - Bulgarian Lev
- BRL - Brazilian Real
- CNY - Chinese Yuan
- CZK - Czech Koruna
- DKK - Danish Krone
- HKD - Hong Kong Dollar
- HRK - Croatian Kuna
- HUF - Hungarian Forint
- IDR - Indonesian Rupiah
- ILS - Israeli New Shekel
- INR - Indian Rupee
- ISK - Icelandic Króna
- KRW - South Korean Won
- MXN - Mexican Peso
- MYR - Malaysian Ringgit
- NOK - Norwegian Krone
- NZD - New Zealand Dollar
- PHP - Philippine Peso
- PLN - Polish Złoty
- RON - Romanian Leu
- RUB - Russian Ruble
- SEK - Swedish Krona
- SGD - Singapore Dollar
- THB - Thai Baht
- TRY - Turkish Lira
- ZAR - South African Rand


wiz:
Excellent work, but I agree that we probably don't need all those currencies - perhaps you can find a list of the most popular fiat currencies and keep the top 10 or 15 or so?

softsimon:
> but I agree that we probably don't need all those currencies

I think we need them all.

But the frontend only need to fetch and get updates on the "current" selected currency.

wiz:
good point, no need to fetch them all at once

natsoni:
Improved the price fetching mechanism: 
- we query only the selected currency, at the timestamp we need (depending on block or transaction date)
- we never query the full historical data except on the address page when the address has more than 10 transactions (to avoid sending too many requests at once), and even then we only query the history of the selected currency)

softsimon (changes requested):
See DM

natsoni:
Changed the fx rates granularity to monthly. A mempool instance with an API key from [freecurrencyapi.com](https://freecurrencyapi.com) (5k requests/month on free plan) will use: 
-  About 170 requests to load the monthly exchange rates from 2010 to today on first run
- 30 requests per month afterwards to fetch the exchange rates once per day

Which is way under the free plan limits for single self hosted instances.

mempool.space instances using the same API key will need more than 6k requests to load historical fx data and will run out of requests. This will merely result in older data being unavailable for a month until the API limits are reset. Then, the older historical data will be fetched and all the instances will be able to run normally using each 30 requests per month.

softsimon (approved):
tested ACK @ [0fcfc5e78d]

Issue #4422, opened 2023-11-24
Show mining related data on transaction page


orangesurf:
### Description

Show information from the block audit view on individual transaction pages.

### Problem to be solved

There are interesting flags in the block audit which are not currently shown on the transaction page. These flags could add useful context, for example whether a pool included a 

### Proposed solution

Add mining related tags to the transaction page.

Example with recent high fee transaction
- Show that transaction was seen in mempool prior to being mined
- Show that transaction was in the expected block

<img width="1216" alt="image" src="https://github.com/mempool/mempool/assets/91332210/78277b40-8f0c-4bc2-bc8e-9b4626cb9f95">

Example with recent censored transaction
- Show that transaction was seen in mempool prior to being mined
- Show that transaction was delayed (not in expected block)
<img width="1216" alt="image" src="https://github.com/mempool/mempool/assets/91332210/c2df55ff-e464-4a70-bc84-c0753c4837a0">

Example of an added transaction (Luxor mined 3.94MWU tx)
- Show that transaction was not seen in the mempool prior to mining
- Show that the transaction was added (not expected in the mempool.space next block template)
- Show the mining pool that mined the transaction (could be added to all transaction pages) 

<img width="1216" alt="image" src="https://github.com/mempool/mempool/assets/91332210/107988ce-cbfb-4804-8120-34f75c00d123">

Example of an accelerated transaction (k3tan 1 sat/vB demo)
- As above but also show that transaction was accelerated

<img width="1185" alt="image" src="https://github.com/mempool/mempool/assets/91332210/f1a36543-1541-4b17-b03a-dff5a8053109">

The tags could link to relevant pages
- "Delayed" tag could link to the block audit for the block in which it was expected (by mempool.space)
- "Added" tag could link to the block audit for the block in which it was included 
- "Accelerated" tag could link to details of the acceleration (inc. amount paid)

AjUnstoppable:
hi, signed up for accelerator long ago but its still showing me as being on the waitlist. how much longer will it take? as one of my transaction is stuck from past 3+ days now due to low gas n i really need to access funds from my wallet. Kindly please help urgently

orangesurf:
builds on  https://github.com/mempool/mempool/issues/513

wiz:
@mononaut this would be cool to have

PR #4748, opened 2024-03-07
Display basic mining info on transaction page

mononaut:
This PR is a partial implementation of the feature described in #4422.

It adds a new "mining" row to the details panel on the transaction page, which includes a miner tag and (if available) some other audit-related status badges.

This feature is only enabled on Bitcoin mainnet, and the audit-related functionality is further restricted to instances with the `AUDIT` and `MINING_DASHBOARD` settings enabled, and for blocks above the `MAINNET_BLOCK_AUDIT_START_HEIGHT`.

The audit-related badges all have mouse-over tooltips with more information.

Screenshot 2024-03-07 at 11 36 41 PM
Screenshot 2024-03-07 at 11 34 40 PM
Screenshot 2024-03-07 at 11 35 30 PM
Screenshot 2024-03-07 at 11 34 26 PM

The audit status badges aren't an exact match for the functionality in #4422. Specifically:
 - No "Delayed" badge, since we don't currently have a good way to identify that case.
 - The "(Not) Seen in mempool" and "Added" badges have slightly different semantics:
    - "Added" matches the current usage in the block audit (either added out-of-band, or intentionally prioritized)
    - "Seen in mempool" is displayed if the transaction is neither "added" nor a "conflict", and the "Expected in Block" badge is not already displayed. This more or less matches the "Marginal fee rate" badge in the main block audit.
        - Since we don't currently record whether transactions were seen in the mempool or not, this is necessarily a bad heuristic.
 - Additional "Conflict" badge, with the same meaning as the main block audit.

When we overhaul the main block audit feature to disambiguate some of these cases, we can update these tags on the transaction page at the same time.

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -78,7 +78,8 @@
                       {{ pool.name }}
                     </a>
                     <ng-container *ngIf="auditStatus">
-                      <span *ngIf="auditStatus.expected; else seen" class="badge badge-success mr-1" i18n-ngbTooltip="Expected in block tooltip" ngbTooltip="This transaction was projected to be included in the block" placement="bottom" i18n="tx-features.tag.expected|Expected in Block">Expected in Block</span>
+                      <span *ngIf="auditStatus.coinbase; else expected" class="badge badge-primary mr-1" i18n="tx-features.tag.coinbase|Coinbase">Coinbase</span>
+                      <ng-template #expected><span *ngIf="auditStatus.expected; else seen" class="badge badge-success mr-1" i18n-ngbTooltip="Expected in block tooltip" ngbTooltip="This transaction was projected to be included in the block" placement="bottom" i18n="tx-features.tag.expected|Expected in Block">Expected in Block</span></ng-template>
                       <ng-template #seen><span *ngIf="auditStatus.seen; else notSeen" class="badge badge-success mr-1" i18n-ngbTooltip="Seen in mempool tooltip" ngbTooltip="This transaction was seen in the mempool prior to mining" placement="bottom" i18n="tx-features.tag.seen|Seen in Mempool">Seen in Mempool</span></ng-template>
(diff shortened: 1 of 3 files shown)


orangesurf:
utACK 

The logic makes sense and the tooltips are clear.

orangesurf (approved):
utACK
The logic makes sense and the tooltips are clear.

nymkappa:
image

nymkappa:
Also I'm confused about badges between the `Recently broadcasted` and `Marginal fee rate` status, but could be my lack of understanding about the audit status.

- A `Recently broadcasted` tx shows `Expected in Block` badge
- A `Marginal fee rate` tx shows `Seen in Mempool` badge

What's the difference ?

wiz (changes requested):
Screenshot 2024-03-08 at 21 38 04

mononaut:
> Also I'm confused about badges between the `Recently broadcasted` and `Marginal fee rate` status, but could be my lack of understanding about the audit status.
> 
> * A `Recently broadcasted` tx shows `Expected in Block` badge
> * A `Marginal fee rate` tx shows `Seen in Mempool` badge
> 
> What's the difference ? Even though a tx has a marginal fee rate, we could assume it was expected in the block right. It just happen that our algo did not include it in our block template while the miner did (for whatever reason).

"Recently broadcasted" is a status applied to transactions in the *projected template*, so it doesn't directly correspond to anything here (the "Expected in Block" badge you're seeing is from a later block where the transactions *was* mined)

"Marginal fee rate" implies "Seen in Mempool", because our current audit system applies that status when a transaction is missing from the projected template (so not "Expected in Block"), but *was* somewhere in the second projected mempool block. We don't currently track which transactions we did or didn't see in the mempool beyond the second block, so this is the best we can do for now.

Similarly, some of the "Added" transactions may actually have been in the mempool, we just can't tell which with the data currently available.

mononaut:
Fixed

Screenshot 2024-03-08 at 3 23 37 PM

nymkappa:
> "Recently broadcasted" is a status applied to transactions......we just can't tell which with the data currently available.

Thanks. I just noticed we actually have a FAQ for that. Clicking on those badge should point to it 🙏 so I can get out of the noob zone.

wiz (approved):
utACK (based on previous test)

PR #4749, opened 2024-03-08
Add fake p2wsh goggles

mononaut:
"Stamps" recently launched a [new data encoding scheme](https://github.com/mikeinspace/stamps/blob/main/OLGA.md) using fake P2WSH outputs.

For example, [this transaction](https://mempool.space/tx/9910643c85cdfd06459231c3361e36327f74bcde40573d2fdd841b35d320bb6a) encodes a 6.4kb SVG image across 199 × 32 byte script hashes.

This PR adds a new Goggles filter under the "Data" category to detect these transactions.

Only a few hundred of these transactions have been mined so far, but it might be worth having the Goggle filter ready in case this new method of data embedding suddenly becomes popular the way previous methods have.

Screenshot 2024-03-08 at 12 53 38 AM

The nature of P2WSH means it isn't actually possible to reliably detect fake outputs, so this implementation simply detects this one specific encoding scheme.

Alternatively, it might be possible to detect enough fake outputs to catch most data-embedding transactions using something like a threshold level of Shannon entropy, on the assumption that real hashes are uniformly distributed and embedded data is not, but I expect it would be tricky to find the right balance of false positives and false negatives to be useful.

This PR also runs a data migration to trigger reclassification of mined block templates since height 832000, in order to properly classify the transactions that are already confirmed. This updates the templates in-place, so there's no need to take servers out of production, although the changes won't be reflected in recent blocks until they age out of the cache.


softsimon (approved):
tested ACK @ [6d3c429b]

PR #4752, opened 2024-03-08
Fix stuck Goggles on websocket reconnect

mononaut:
This PR fixes an issue where we could fail to resume the `track-mempool-block` subscription after a websocket reconnection.

The PR also adjusts the transition on the block visualization loading spinner, so that it stays visible for a bit longer until the transactions finish animating in.


wiz (approved):
utACK

PR #4753, opened 2024-03-08
Hide mining pools on clock blockchain

mononaut:
Before:
Screenshot 2024-03-08 at 7 17 40 PM

After:
Screenshot 2024-03-08 at 7 17 28 PM


wiz (approved):
utACK

PR #4754, opened 2024-03-09
[accelerator] set accel graph bar min height

nymkappa:
Screenshot 2024-03-09 at 09 23 19

Make sure we always show something even if the value is really small in comparison to other bars

Screenshot 2024-03-09 at 09 23 32


wiz (approved):
utACK

PR #4756, opened 2024-03-09
i18n: fix incorrect translation for `Mining Dashboard` and `Accelerator Dashboard` since #4220

GoudanWoo:
> I'm sure that this is not a simple pull request for translations. So I do not submit it on Transifex.

----

## What

incorrect translation for `Mining Dashboard` when use Non-English.

### Non-English like Chinese (incorrect)

image

image

### Non-English like French (incorrect)

image

image

### English (correct)

image

image

----

## Why

In #4220, @mononaut copied `:@@a681a4e2011bb28157689dbaa387de0dd0aa0c11:Accelerator Dashboard` from `:@@a681a4e2011bb28157689dbaa387de0dd0aa0c11:Mining Dashboard` but not modify the id. 

https://github.com/mempool/mempool/blob/6f97a2ef741bc140c8aeb4ec9440e879324f3d49/frontend/src/app/components/accelerator-dashboard/accelerator-dashboard.component.ts#L16

https://github.com/mempool/mempool/blob/6f97a2ef741bc140c8aeb4ec9440e879324f3d49/frontend/src/app/components/mining-dashboard/mining-dashboard.component.ts#L20

And then In 6671746f, a new i18n is be extracted with wrong translation id for `Mining Dashboard` and `Accelerator Dashboard`.

image

image

----

## How

So we must reextract i18n and modify the correct id for `Accelerator Dashboard` in `accelerator-dashboard.component.ts`.

### DIFF
--- a/frontend/src/app/components/acceleration/accelerator-dashboard/accelerator-dashboard.component.ts
+++ b/frontend/src/app/components/acceleration/accelerator-dashboard/accelerator-dashboard.component.ts
@@ -40,7 +40,7 @@ export class AcceleratorDashboardComponent implements OnInit {
     private serviceApiServices: ServicesApiServices,
     private stateService: StateService,
   ) {
-    this.seoService.setTitle($localize`:@@a681a4e2011bb28157689dbaa387de0dd0aa0c11:Accelerator Dashboard`);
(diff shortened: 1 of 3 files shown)


softsimon:
Thanks for reporting these, but I will fix this in an alternative way soon.

PR #4757, opened 2024-03-09
Add audit=[bool] param to /block/[hash] url

russeree:
**Request**

Currently the audit view of a block doesn't have a URL parameter to temporality disable/enable audit view when using a ```/block/<hash>```. This is because that parameter is stored in a session storage variable. This is opposed to when a user clicks on the details view of a block and mempool will add the URL parameters ```"showDetails=false&view=actual#block"``` which will activate a details view of the block when the url is passed to another device. This is useful for consistent visualizations or presentation. 

**Solution**
Add a check for a URL parameter ```audit=[true/false]``` being passed to a ```/block/<hash>``` url that will temporarily display the audit view based on that parameter. 

**Design Choices**

1.  URL Parameter is passed and the page will show the requested view audit [false/true]
2. This does not overwrite the users existing preference for their audit view, so even if they are sent a link to show a block without the audit view. As soon as they refresh the page without the URL param it displays their session saved version
3. Clicking the audit view on or/off while inside of a param'd url will modify the session storage object and be reflected on the the next non param visit of another block.

### DIFF
--- a/frontend/src/app/components/block/block.component.ts
+++ b/frontend/src/app/components/block/block.component.ts
@@ -127,7 +127,6 @@ export class BlockComponent implements OnInit, OnDestroy {
     if (this.auditSupported) {
       this.isAuditEnabledFromParam().subscribe(auditParam => {
         if (this.auditParamEnabled) {
-          console.log(`auditParamEnabled: ${auditParam}`);
           this.auditModeEnabled = auditParam;
         } else {
           this.auditPrefSubscription = this.stateService.hideAudit.subscribe(hide => {
--- a/frontend/src/app/components/block/block.component.ts
+++ b/frontend/src/app/components/block/block.component.ts
@@ -61,6 +61,7 @@ export class BlockComponent implements OnInit, OnDestroy {
   transactionsError: any = null;
   overviewError: any = null;
   webGlEnabled = true;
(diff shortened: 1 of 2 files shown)


frontend/src/app/components/block/block.component.ts
```diff
@@ -124,9 +125,16 @@ export class BlockComponent implements OnInit, OnDestroy {
     this.setAuditAvailable(this.auditSupported);
 
     if (this.auditSupported) {
-      this.auditPrefSubscription = this.stateService.hideAudit.subscribe((hide) => {
-        this.auditModeEnabled = !hide;
-        this.showAudit = this.auditAvailable && this.auditModeEnabled;
+      this.isAuditEnabledFromParam().subscribe(auditParam => {
+        if (this.auditParamEnabled) {
+          console.log(`auditParamEnabled: ${auditParam}`);
```
mononaut, line 130: stray console.log
```suggestion
```

mononaut (approved):
Tested ACK @ [fffffffdb3826]

This works well, and doesn't interfere with other query params on this page or the user's existing audit preference.

russeree:
You guys killed off the stray console.log too fast, TY for the merge.

PR #4761, opened 2024-03-10
Fix block acceleration fees calculation & api bugs

mononaut:
This PR should fix a bug in the accelerator audit calculation from https://github.com/mempool/mempool/pull/4718 which was interpreting transaction fees in the wrong units (BTC instead of sats).

Since the results so far are incorrect, **this PR includes a database migration to truncate the local accelerations table**.

The PR also fixes an issue with the `/api/v1/accelerations/pool/:slug` and `/api/v1/accelerations/block/:height` APIs which were mistakenly returning results for `/api/v1/accelerations/recent/24h`.

### DIFF
--- a/backend/src/api/acceleration.ts
+++ b/backend/src/api/acceleration.ts
@@ -33,7 +33,7 @@ interface GraphTx {
   vsize: number;
   weight: number;
   fees: {
-    base: number;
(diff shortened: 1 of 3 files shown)


wiz (approved):
utACK

PR #4766, opened 2024-03-11
Audit past acceleration fees

mononaut:
This PR uses the paginated mempool services backend API to fetch past accelerations, and independently calculate the expected Bid Boost fees paid for each.

The task runs as part of the regular indexing, and so will also catch up on accelerations mined while the backend was offline or missed for other reasons.

Accelerations are processed block-by-block, oldest first, with progress tracked in a "last_acceleration_block" row in the `mempool.state` db table. If you need to reindex the acceleration data for any reason, truncate the `mempool.accelerations` table (or drop the relevant rows) and reset the "last_acceleration_block" to 0 (or some other starting height) in the `mempool.state` table.

The implementation supports both Bid Boost V1 and V2 calculations, with a hardcoded switch-over at block height 831580.

The PR also includes a small bugfix for the existing V2 calculation related to vsize rounding, which could sometimes cause results to be off by a few sats.

Note that due to variability in the median fee rate used for the V1 calculation, the fees calculated for these older accelerations may differ slightly from the amounts actually charged (as reported by the services backend public APIs).

### DIFF
--- a/backend/src/api/acceleration.ts
+++ b/backend/src/api/acceleration.ts
@@ -238,7 +238,7 @@ class AccelerationCosts {
   private convertToGraphTx(tx: MempoolTransactionExtended): GraphTx {
     return {
       txid: tx.txid,
-      vsize: tx.vsize,
+      vsize: Math.ceil(tx.weight / 4),
       weight: tx.weight,
       fees: {
         base: 0, // dummy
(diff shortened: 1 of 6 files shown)


rawadbadr:
هل اسم جيثوب يخص محفظتي ام ماذا

wiz (approved):
utACK

PR #4767, opened 2024-03-12
Add unfurler routes for /enterprise and /sponsor

hunicus:
Sponsors image needs to be tweaked but let's see how this looks.


unfurler/src/routes.ts
```diff
@@ -119,6 +110,10 @@ const routes = {
     title: "RBF",
     fallbackImg: '/resources/previews/rbf.jpg',
   },
+  sponsors: {
```
wiz, line 113: are you sure?

wiz (approved):
utACK

PR #4773, opened 2024-03-15
Simplify recent transactions observable

mononaut:
This PR simplifies the `stateService.transactions$` observable used for the Recent Transactions widget, and also fixes the ordering of transactions there (so that new transactions enter from the top, not the bottom).

Before:

https://github.com/mempool/mempool/assets/83316221/0a1d3c25-2042-4d6a-a6c7-d91fa30f08b8

After:

https://github.com/mempool/mempool/assets/83316221/c0d40fc9-91ac-4dca-ac95-ba58480a7367


wiz (approved):
utACK

PR #4774, opened 2024-03-15
Show accelerator audit totals on pools page

mononaut:
This PR adds a new `/accelerations/total` API endpoint for the locally audited acceleration info and uses it to display total out-of-band fees on the mining pool page.

The PR also deduplicates some tables on the pool page to make the component easier to work with.


wiz (approved):
utACK

PR #4775, opened 2024-03-15
Increase precision of out-of-band fees on the block page

mononaut:
Increases the maximum precision of the out-of-band fees on the block page from 4 to 8 decimal places.

Before:
Screenshot 2024-03-15 at 7 47 55 AM

After:
Screenshot 2024-03-15 at 7 47 36 AM


wiz (approved):
utACK

PR #4783, opened 2024-03-17
Fix pool oob fees table mobile layout

mononaut:
Before:
Screenshot 2024-03-17 at 8 00 39 AM

After:
Screenshot 2024-03-17 at 8 00 29 AM


wiz (approved):
utACK

softsimon (approved):
tested ACK @ [a8fa7dcb2a]

PR #4785, opened 2024-03-17
Display more accurate price data on tx bowtie tooltip

natsoni:
This PR adds some data on the tx bowtie tooltip: 
- the number of blocks elapsed between the input and prevout or the output and the input spending it (if exists)
- if the user chooses to view fiat instead of btc value, the value displayed in the tooltip uses the BTC price at the time of the corresponding transaction (price data queried on hover)

*********************

- Prevout: 
Screenshot 2024-03-17 at 16 55 48

- Input: 
Screenshot 2024-03-17 at 16 55 41

- Output: 
Screenshot 2024-03-17 at 16 55 33

- Next input spending the output: 
Screenshot 2024-03-17 at 16 55 25

### DIFF
--- a/frontend/src/app/components/amount/amount.component.html
+++ b/frontend/src/app/components/amount/amount.component.html
@@ -8,8 +8,13 @@
     }}
   </span>
   <ng-template #noblockconversion>
-    <span class="fiat">{{ addPlus && satoshis >= 0 ? '+' : '' }}{{ (conversions[currency] > -1 ? conversions[currency] : 0) * satoshis / 100000000 | fiatCurrency : digitsInfo : currency }}
-    </span>
+    <ng-container *ngIf="!forceBlockConversion; else zeroValue">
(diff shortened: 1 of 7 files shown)


mononaut (approved):
Tested ACK @ [2dc6f6ff]

Small nit re. the wording of the labels, but otherwise this is a great feature and it works well.

It can generate a *lot* of API requests on more complex transactions, so we might need to consider some kind of batch API or more aggressive caching if we see users getting rate limited.

mononaut (approved):
Tested ACK @ [2dc6f6ff]

Small nit re. the wording of the labels, but otherwise this is a great feature and it works well.

It can generate a *lot* of API requests on more complex transactions, so we might need to consider some kind of batch API or more aggressive caching if we see users getting rate limited.

softsimon (approved):
tested ACK @ [a04d685]

PR #4790, opened 2024-03-18
Account menu overlays UI instead of pushing it

### LINKED ISSUE #4662: Left menu should overlay without pushing the UI
* Sticky header
* Sticky menu
* Scroll-able menu
* Overlay the site and not pushing, reducing complexity and hackyness

An example of such behavior can be find using the mobile layout on this site https://pool.braiins.com/en/stats/btc

natsoni:
Fixes #4662 

Screenshot 2024-03-18 at 16 56 13

**********************

Screenshot 2024-03-18 at 16 56 43

### DIFF
--- a/frontend/src/app/components/master-page/master-page.component.html
+++ b/frontend/src/app/components/master-page/master-page.component.html
@@ -81,10 +81,10 @@
     <app-search-form [hamburgerOpen]="user != null" class="search-form-container" location="top" (searchTriggered)="collapse()"></app-search-form>
   </div>
 </nav>
+<app-menu *ngIf="servicesEnabled" [navOpen]="menuOpen" (loggedOut)="onLoggedOut()" (menuToggled)="menuToggled($event)"></app-menu>
 </header>
 
 <div class="d-flex" style="overflow: clip">
-  <app-menu *ngIf="servicesEnabled" [navOpen]="menuOpen" (loggedOut)="onLoggedOut()" (menuToggled)="menuToggled($event)"></app-menu>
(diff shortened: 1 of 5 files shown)


wiz (commented):
hmm are you sure this won't break anything in the customer control panel?

nymkappa:
Does not seem to break anything.

I wonder if there is a way to disable this stupid scrolling behavior when you hit the top or bottom of a page.

image

Also, on mobile safari when you scroll down you have this artifact. It happens because the viewport height changes due to the url input are beeing hidden while scrolling. This also happens on github.com.

natsoni:
> * I wonder if there is a way to disable this stupid scrolling behavior when you hit the top or bottom of a page.

These lines added in `frontend/src/styles.scss` solves this, but I don't know if we want that behaviour (there is no more "bounce effect" when scrolling up). 
```
body {
  overscroll-behavior-y: none;
}
```

mononaut:
Is there any reason not to make the menu absolutely positioned *inside* the sticky main `<header>`?

That would fix the bouncy overscroll behavior

softsimon (approved):
tested ACK @ [393181f753]

wiz:
> hmm are you sure this won't break anything in the customer control panel?

@softsimon @nymkappa as per my previous message, on desktop while inside the control panel, this might hide stuff depending on the screen resolution - please test/polish more in various screen resolutions to prevent this:

Before:
Screenshot 2024-03-22 at 12 02 27

After:
Screenshot 2024-03-22 at 11 59 00

PR #4798, opened 2024-03-19
Flex justified navbar

softsimon:
Before

Screenshot 2024-03-19 at 15 13 00

After

Screenshot 2024-03-19 at 15 13 11


softsimon:
tested with Kappa on desktop, Samsung S24 and a small iPhone

PR #4809, opened 2024-03-22
[menu] fix css

nymkappa:
Horrible css but that seems to fix current issue. Would be good to completely rewrite this stuff at some point.


softsimon (approved):
tested ACK @ [c47822ca]

PR #4810, opened 2024-03-22
Show goggles badges on block overview tooltip

mononaut:
This PR adjusts the layout of the block visualization tooltip, and adds badges for most Mempool Goggles categories:

Screenshot 2024-03-22 at 7 20 26 AM

The badges are color-coded to match the active Goggles filters, so you can see at a glance why a particular transaction has been highlighted.

The existing purple "acceleration" tag is incorporated into this new badge area, but keeps the pulsing purple background.

Some less informative filter categories are excluded to avoid making the tooltip excessively large:
 - address types (most transactions match multiple types)
 - `sighash_all` and `sighash_default` (almost every transaction matches either or both)

### DIFF
--- a/frontend/src/app/components/block-overview-graph/block-overview-graph.component.html
+++ b/frontend/src/app/components/block-overview-graph/block-overview-graph.component.html
@@ -12,6 +12,8 @@
       [clickable]="!!selectedTx"
       [auditEnabled]="auditHighlighting"
       [blockConversion]="blockConversion"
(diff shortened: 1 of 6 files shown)


softsimon (approved):
tested ACK @ [15062b80]

Took the liberty to remove the goggles icon for now as it looks like a feature tag and not a section header. Looks more cleaner also.

Screenshot 2024-03-23 at 14 07 01

PR #4812, opened 2024-03-23
Add Goggles filter badges to the transaction page

mononaut:
This PR adds Mempool Goggles classification tags to the transaction page, in a new row of the main details panel:

Screenshot 2024-03-23 at 6 41 31 AM
Screenshot 2024-03-23 at 6 41 41 AM

Only the following filter badges are shown:
 - Coinjoin
 - Consolidation
 - Batch payment
 - OP_RETURN
 - Fake pubkey
 - Fake scripthash
 - Inscription
 - P2PK
 - Bare multisig
 - Non-standard
 
Badges for other address types, RBF settings, CPFP etc are not shown, to avoid overlapping with other fields.

Sighash flags are also not shown, as they might be confusing. Maybe we can display these on individual inputs/signatures instead in a later PR.

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -71,19 +71,15 @@
                   </td>
                 </tr>
                 <tr *ngIf="network === ''">
-                  <td class="td-width" i18n="transaction.mining">Mining</td>
+                  <td class="td-width" i18n="transaction.audit">Audit</td>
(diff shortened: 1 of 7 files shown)


mononaut:
Updated to split the "Miner" and "Audit" badges into separate rows, and also fix the color of the "Added" audit badge.

softsimon (approved):
tested ACK @ [40069ebe]

PR #4815, opened 2024-03-24
testmempoolaccept

mononaut:
This PR adds a `POST /api/txs/test` endpoint for non-esplora backends to match the new mempool/electrs API https://github.com/mempool/electrs/pull/81, and also adds a new page as a frontend interface for the new endpoints.

Users can paste in a list of comma-separated raw transactions, and the page renders the results in a nice table:

Screenshot 2024-03-24 at 9 00 05 AM

Or displays a short error message if the transactions fail basic pre-check validation:

Screenshot 2024-03-24 at 9 00 20 AM


softsimon (approved):
tested ACK @ [e45a0372]

works both with just bitcoin core, and electrs

PR #4817, opened 2024-03-25
Fix unnecessary request of blocks on dashboard load

### LINKED ISSUE #4469: Unnecessary loading of blocks
When loading the dashboard. 3 pages of early blocks are initially loaded which is not needed.

Screenshot 2023-12-02 at 15 37 12

natsoni:
Fixes #4469

Screenshot 2024-03-25 at 12 11 58

I checked that this PR did not re-introduce some bugs fixed by #4269 (since I reversed one of its changes: `chainTip = -1` instead of `chainTip = 100` in `start.component.ts`), but I could have missed something.

### DIFF
--- a/frontend/cypress/fixtures/mainnet_mempoolInfo.json
+++ b/frontend/cypress/fixtures/mainnet_mempoolInfo.json
@@ -1,554 +1,771 @@
 {
     "mempoolInfo": {
-        "loaded": true,
-        "size": 15168,
-        "bytes": 6070217,
-        "usage": 30922560,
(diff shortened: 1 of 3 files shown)


softsimon (changes requested):
When the RBF test is running, opening the TX page, the blockchain skeleton loader never goes away.

Either the tests are a bit sensitive / change detection is handled differently in the test environment, or there is a problem with this PR.

softsimon (approved):
tested ACK @ [6bebc76]

PR #4818, opened 2024-03-25
Add halving countdown widget to main dashboard

mononaut:
This PR adds a toggle to switch the difficulty adjustment widget on the main dashboard to show a halving countdown instead:

Screenshot 2024-03-25 at 8 35 58 AM
Screenshot 2024-03-25 at 8 36 03 AM

The toggle defaults to "difficulty mode", until the next halving is closer than the next difficulty adjustment.

We don't yet save the user's preference anywhere, so the widget resets to the default view after navigating away from the page.


softsimon (approved):
tested ACK @ [0ece8bb2a]

PR #4824, opened 2024-03-26
Set coherent units for hashrate and difficulty units on hashrate graph

### LINKED ISSUE #2491: The difficulty and hashrate graph are not on equivalent units.
https://mempool.space/graphs/mining/hashrate-difficulty

The scales used for the different units should match so that the hashrate is overlapping the difficulty adjustments.

Example https://bitcoin.sipa.be/speed-lin-10k.png
image

image

I'm not sure of the exact hashrate to difficuly conversion but if reported hashrate is 100% stable the 0% difficulty adjustment line should end up on top of each others.

natsoni:
Fixes #2491

| Before  | After |
| Screenshot 2024-03-26 at 16 58 36 | Screenshot 2024-03-26 at 16 58 59 |
| Screenshot 2024-03-26 at 16 59 52 | Screenshot 2024-03-26 at 17 00 16 |


softsimon (approved):
tested ACK @ [fa2dd3c7cb]

PR #4829, opened 2024-03-28
Add an option to always show the block fees on infinite blockchain

natsoni:
This PR adds an option for the user to always display block fees on the infinite blockchain (and not only on the mining dashboard): 
Screenshot 2024-03-28 at 12 19 55

I was not sure about where to display the option so it is currently invisible above the blockchain divider and visible on hover: 
| Default mode  | Fees mode |
| Screenshot 2024-03-28 at 12 20 55  | Screenshot 2024-03-28 at 12 20 55  |

Default mode still displays the block fees when on the mining dashboard, and the user's preference is stored in local storage.

### DIFF
--- a/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
+++ b/frontend/src/app/components/blockchain-blocks/blockchain-blocks.component.html
@@ -40,12 +40,14 @@
                 <app-fee-rate unitClass=""></app-fee-rate>
               </div>
             </ng-template>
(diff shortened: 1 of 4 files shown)


natsoni:
Removed the button above the divider. Now the infinite blockchain will keep showing the block fees if the user was on the mining dashboard before clicking on a block / tx:

https://github.com/mempool/mempool/assets/46578910/8e96eae8-abe3-465f-b11c-f23925e698fa

Block fees will stay visible until the user comes back to the main dashboard.

softsimon (approved):
tested ACK @ [66309813]

PR #4831, opened 2024-03-28
Add multi-address wallet page

mononaut:
_~~(uses https://github.com/mempool/electrs/pull/82)~~_
_(builds on #5375)_

This draft PR adds a page at `/wallet/:wallet` with combined balance and transaction info for named wallets.

The embeddable wallet summary widget page that was previously on that path is moved to `/widget/wallet?...`

localhost_4200_wallet

As well as the combined balance chart and summary stats, the page includes a treemap showing the distribution of balances across the different addresses (colored by the number of transactions on each address).

~~The PR relies on some new mempool/electrs multi-address APIs to be added in https://github.com/mempool/electrs/pull/82 to allow fetching transactions and balance summary data without making an absurd number of requests and/or opening the backend to DoS risks.~~

**Other limitations**: We don't currently have a scalable way to get the total number of transactions for a set of addresses (since there's no way to deduplicate transactions involving multiple addresses).

### DIFF
--- a/backend/src/api/bitcoin/esplora-api.ts
+++ b/backend/src/api/bitcoin/esplora-api.ts
@@ -305,7 +305,7 @@ class ElectrsApi implements AbstractBitcoinApi {
   }
 
   $getAddress(address: string): Promise<IEsploraApi.Address> {
-    throw new Error('Method getAddress not implemented.');
+    return this.failoverRouter.$get<IEsploraApi.Address>('/address/' + address);
   }
 
   $getAddressTransactions(address: string, txId?: string): Promise<IEsploraApi.Transaction[]> {
--- a/backend/src/api/services/wallets.ts
+++ b/backend/src/api/services/wallets.ts
@@ -8,7 +8,14 @@ import { TransactionExtended } from '../../mempool.interfaces';
 interface WalletAddress {
   address: string;
   active: boolean;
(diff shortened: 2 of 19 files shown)


mononaut:
refactored to support named wallets, and rebased on #5375.

the transaction list has been removed, as we don't have a good way to fetch transactions for a large number of addresses without https://github.com/mempool/electrs/pull/82. I am investigating alternatives, and will update the PR once I've found a good approach.

mononaut:
Taking out of draft.

The more limited functionality via the named wallet APIs from #5375 should work well enough for a mostly hidden page (it's only linked to from one of the custom dashboard widgets, and requires the corresponding wallet to be configured in the backends)

mononaut:
I've opened a [separate draft PR](https://github.com/mempool/mempool/pull/5562) to restore the transactions list feature, so that this part isn't blocked by the dependency on https://github.com/mempool/electrs/pull/82

junderw:
Memo from Matrix:

- Add cache key for nginx on the batch endpoints: `proxy_cache_key "$request_uri|$request_body";`

wiz:
please rebase

mononaut:
rebased

wiz (approved):
Tested ACK @ v3.1.0-dev [8514fb9bd][344dfc24]

PR #4834, opened 2024-03-29
hide empty audit row in tx page details panel

mononaut:
Before:
Screenshot 2024-03-29 at 6 52 50 AM

After:
Screenshot 2024-03-29 at 6 52 40 AM


softsimon (approved):
tested ACK @ [cbcb61d3fa]

PR #4836, opened 2024-03-31
Display first seen time on block visualisation tooltips

mononaut:
This PR adds the first seen time to tooltips on both the projected and mined block visualizations.

Screenshot 2024-03-31 at 3 46 29 AM

In the case of mined blocks, the time is displayed relative to the block timestamp where possible (this means that first seen times are now included in the block template/summary blobs).

Screenshot 2024-03-31 at 3 47 53 AM


softsimon (changes requested):
Use "Confirmed: After X time" like the TX page.

softsimon (approved):
tested ACK @ [bf22f4515f]

PR #4837, opened 2024-03-31
Goggles age filter

mononaut:
This PR adds a new Goggles mode for displaying the ages of transactions in projected and mined blocks using a gradient of opacity.

Screenshot 2024-04-04 at 7 03 31 AM

This helps to see at a glance which transactions are newly arrived, and which have been pending in the mempool for some time.

The filter also works on mined blocks (if we have first seen data available), with the opacity representing age of the transaction at the time of mining:

Screenshot 2024-04-04 at 7 04 05 AM

To avoid clashes with the way we currently use opacity for the other Goggles filters, when 'age' mode is enabled, regular Goggles filters use a fixed dark blue for the 'non-matching' color:

Screenshot 2024-04-04 at 7 04 44 AM

Regular "color-by-fee-rate" mode is unaffected:

Screenshot 2024-04-04 at 7 06 54 AM

### DIFF
--- a/frontend/src/app/app.constants.ts
+++ b/frontend/src/app/app.constants.ts
@@ -39,8 +39,6 @@ export const mempoolFeeColors = [
   'ae005b',
 ];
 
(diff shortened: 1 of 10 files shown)


mononaut:
Updated with a new approach

mononaut:
Screenshot 2024-04-04 at 7 43 12 AM

softsimon (approved):
tested ACK @ [fc5312549]

beage666 (commented):
How do I log in

PR #4838, opened 2024-03-31
Don't exclude accelerated txs from fee graph & fee statistics

mononaut:
Previously we excluded accelerated transactions from the fee distribution graph and the block-level fee statistics (median & fee percentiles, and by extension the fee recommendations).

With Bid Boost V2, this is no longer necessary, and instead we should incorporate accelerated fees into these statistics.


softsimon (approved):
utACK

Issue #4839, opened 2024-03-31
Track multiple transactions in the backgground


softsimon:
- [ ] The backend needs a new WebSocket API for multiple transactions tracked at the same time
- [ ] The frontend need a global multi transaction tracker functionality that will live in LocalStorage and across pages
- [ ] A way to "pin" transaction to the global transactions tracker with UI similar to the mock preview below

Later, we probably want to extend this to address tracking as well later, and also adding multiple "shadow arrows" to show all tracked transactions below the mempool blocks.

![Screenshot 2024-03-31 at 15 12 54](https://github.com/mempool/mempool/assets/8561090/41342e5f-759f-43f8-b1ab-dedf495d5985)

mononaut:
I added the websocket subscription for this in #4840

PR #4840, opened 2024-03-31
Add track-txs websocket subscription

mononaut:
This PR addresses the first task in issue #4839, by adding a new `track-txs` websocket subscription for tracking multiple transactions.

The subscription is started by passing an array of txids in the `track-txs` key of the websocket message, and responds with an dictionary of `TxTrackingInfo` objects in the `tracking-txs` key of the response:

```typescript
interface TxTrackingInfo {
  replacedBy?: string,
  position?: { block: number, vsize: number, accelerated?: boolean },
  cpfp?: {
    ancestors?: Ancestor[],
    bestDescendant?: Ancestor | null,
    descendants?: Ancestor[] | null,
    effectiveFeePerVsize?: number | null,
    sigops: number,
    adjustedVsize: number,
  },
  utxoSpent?: { [vout: number]: { vin: number, txid: string } },
  accelerated?: boolean,
  confirmed?: boolean
}

{
  "tracking-txs": {
    <txid>: TxTrackingInfo
  }
}
```

There's currently a bug in master which causes the cpfp field to always be returned, but after I fix that in a separate PR the response will only return information that changed since the last update.

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -565,7 +565,7 @@ class WebsocketHandler {
       }
       if (client['track-txs']) {
         for (const txid of client['track-txs']) {
-          trackedTxs.add(client['track-tx']);
+          trackedTxs.add(txid);
         }
       }
     });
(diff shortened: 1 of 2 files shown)


softsimon (approved):
tested ACK @ [7deee5d5f]

PR #4841, opened 2024-03-31
Clear cpfp dirty status

mononaut:
Fixes a bug (possibly introduced in the Limited GBT PR (#4555)) where the `cpfpDirty` flag would not get properly cleared, leading to unnecessary websocket messages and on-demand CPFP re-calculations.


softsimon (approved):
utACK

PR #4842, opened 2024-03-31
Fix next block subsidy calculation

mononaut:
Calculate the actual next halving subsidy instead of using a hard-coded 3.125 BTC


knorrium:
Tested ACK @ [9270298](https://github.com/mempool/mempool/pull/4842/commits/9270298374364a307c195e6c8b4c7a3f33dbe7da)

Screenshot 2024-03-31 at 5 59 33 PM

knorrium (approved):

softsimon (approved):
utACK

PR #4848, opened 2024-04-01
Add backend type flag, disable address graphs for non-esplora

### LINKED ISSUE #4843: Disable balance history view when not on electrs
image

mononaut:
Resolves #4843.

This PR adds the backend mode (esplora, electrum or core) to our init data, so that we can tell which features to support in the client.

Then it disables the address balance graph on non-esplora backends.


softsimon (approved):
tested ACK @ [1a3b8580]

PR #4849, opened 2024-04-01
Add backend mode to backend-info

mononaut:
Adds the backend mode (esplora, electrum or none) to the backend-info API response


softsimon (approved):
tested ACK @ [df72829fd]

PR #4852, opened 2024-04-01
Better audits

mononaut:
_(builds on PR #4836)_

This PR makes some improvements to the block audits:
 - Better "marginal fee rate" classification on the mined block side, by looking back more than one mempool block for matches, and allowing a larger margin of error.
   - This should reduce the number of marginal fee rate transactions incorrectly marked "added".
 - New "Prioritized" category for transactions seen in the mempool but unexpectedly included in the block.
   - This helps to differentiate between *out of band transactions* and *in-band transactions with out of band fees*. (previously both cases would have been marked "added").
   - "Prioritized" transactions are highlighted in desaturated blue.
 - Use the new category to improve the accuracy of the "Seen in mempool" tag on the transaction page.
 
 
Screenshot 2024-04-01 at 8 08 39 AM
Screenshot 2024-04-01 at 8 08 16 AM


softsimon (approved):
tested ACK @ [bf22f4515f]

PR #4853, opened 2024-04-01
Fix Liquid crash

natsoni:
This PR fixes a crash happening on the Liquid dashboard since #4829 got merged. It happens when clicking on the recent peg or the federation wallet table. 

Screenshot 2024-04-01 at 18 25 04


softsimon (approved):
utACK

PR #4854, opened 2024-04-01
Fix frontend errors on block loading

natsoni:
This PR fixes a bug where some requests randomly fails when quickly navigating through the blocks:

https://github.com/mempool/mempool/assets/46578910/e495fd84-2f9d-4b8c-8fac-78843b5e4cbd


softsimon (approved):
tested ACK @ [974eaeb02]

PR #4855, opened 2024-04-01
Run sync-assets-dev when building

knorrium:
The Angular build process wipes out the `dist` folder so production builds always have to redownload assets. 

We will now run the `sync-assets-dev` target to keep a local copy of the assets in `src/resources`.


wiz (approved):
Tested ACK

Issue #4860, opened 2024-04-02
Do not attempt to load acceleration data on liquid blocks


knorrium:
When loading a liquid block, a 404 error is logged as we try to load acceleration data

GET https://liquid.network/liquid/api/v1/accelerations/block/1 404 (Not Found)

softsimon:
https://github.com/mempool/mempool/issues/4861

PR #4865, opened 2024-04-02
Public acceleration config

### LINKED ISSUE #4861: Block accelerations should only load on mainnet
Opening a block on Testnet or Liquid will trigger a accelerations API call.

mononaut:
Fixes #4861, #4860

This PR adds a new frontend config setting `PUBLIC_ACCELERATIONS` to enable/disable locally audited acceleration data (from the new-ish `/api/v1/accelerations/...` endpoints) separately from the acceleration dashboard and services apis.

The PR also restricts frontend calls to those APIs to mainnet.

`PUBLIC_ACCELERATIONS` is false by default, but enabled in the production mainnet frontend config.


wiz (approved):
utACK

PR #4868, opened 2024-04-03
Fix footer position

### LINKED ISSUE #4867: Footer not always at the very bottom
If you flip your Studio Display vertical, you will find that the global footer is not at the very bottom as it should be.

Screenshot 2024-04-02 at 21 08 41

natsoni:
Fixes #4867

The footer is incorrectly positioned on frontends with services enabled (official instance and accelerator set to true). This PR fixes this by always displaying the `empty-sidenav` element, instead of only when the services are disabled.


wiz (approved):
utACK

PR #4871, opened 2024-04-03
Fix Liquid testnet broken blocks

### LINKED ISSUE #4797: Blocks break on Liquid Testnet
Click on the latest block and this happens. liquid.network/testnet

Screenshot 2024-03-19 at 14 11 39

natsoni:
Fixes #4797

The bug was due to the `networkChanged$` observable incorrectly emitting whenever the URL changed on Liquid testnet. This caused the blocks and transactions data to be reset, resulting in the blockchain disappearing and data on the main dashboard not loading anymore. 

https://github.com/mempool/mempool/assets/46578910/94ceeeec-4cda-45d6-a2b6-34624da08aff


softsimon (approved):
utACK

PR #4874, opened 2024-04-03
Fix the Docker workflow for the new Rust steps

knorrium:
We are using relative paths and changing directories when building the backend now, so that broke the current Docker tagging workflow.

I tried to keep the changes at a minimum to get a working build again so we can test on Umbrel again.


softsimon (approved):
utACK

Issue #4870, opened 2024-04-03
Acceleration dashboard backend APIs


softsimon:
Acceleration dashboard should work by either proxy the requests to the services backend or ideally just use the local synchronized accelerated transactions database to give the same API responses.

Marcus8319:
How do i with drawl funds

PR #4879, opened 2024-04-04
[accelerator] proxy acceleration api to prod

nymkappa:
Fixes https://github.com/mempool/mempool/issues/4870


softsimon (changes requested):
`X-Total-Count:` is not passed on for the history api so the pagination isn't working.

softsimon (approved):
tested ACK @ [b421be3315]

PR #4880, opened 2024-04-04
Enable fiat price conversion again

knorrium:
This flag needs to be enabled by default. The other conversion feature is gated by the API key flag.


softsimon (approved):
utACK

PR #4887, opened 2024-04-04
Local acceleration data

mononaut:
Adds a `config.MEMPOOL_SERVICES.PREFER_LOCAL` setting (default `true`), which uses locally indexed acceleration data for the `/api/v1/services/accelerator/accelerations` and `/api/v1/services/accelerator/accelerations/history` endpoints instead of proxying to prod (as in #4879)

These local implementations don't support the full range of query params and data, but only just enough to make the current front end usages work.

the `/history/aggregated` and `/stats` endpoints are still proxied.


mononaut:
Rebased on master, and removed the `PREFER_LOCAL` config. Now these endpoints will always server local acceleration data (but prod will still proxy requests directly to the services backend via nginx)

softsimon (changes requested):
acceleration.boost is NaN because feePaid etc is missing from the local API so the Bid Boost column shows empty result instead of `0 sat`, when boostCost is 0

mononaut:
fixed & rebased on master

softsimon (approved):
tested ACK @ [11eaa0ca50]

PR #4889, opened 2024-04-04
Add support for inscription goggles for non-esplora backends

mononaut:
On non-esplora backends, we generally don't have access to prevout data.

This means we can't identify input addresses or address types, which we then also use to help identify inscriptions.

Adding prevout data is not feasible, so address type Goggles filters can't be supported, but this PR enables inscription detection by checking the witness data directly when prevouts are not available.


softsimon (approved):
tested ACK @ [04f6490ee]

PR #4891, opened 2024-04-05
Transaction details template refactor

mononaut:
This PR refactors the transaction page details tables, using the new Angular 17 control flow syntax to simplify our template logic, reduce duplication, and make it easier to work with these UI elements in future.

The PR then uses this opportunity to finally fix the hack we've been using to ensure alternating row colors on mobile (by merging the two columns into a single table instead of stacking the tables and adding phantom elements)

Apart from fixing that alternating row color issue, and slightly improving the skeleton loader logic, the PR should preserve all existing UI behavior on the transaction page.

| Before | After |
|-|-|
| Screenshot 2024-04-05 at 6 34 31 AM | Screenshot 2024-04-05 at 6 34 44 AM |

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -40,64 +40,39 @@
   </div>
   <div class="clearfix"></div>
 
-  <ng-template [ngIf]="!isLoadingTx && !error">
-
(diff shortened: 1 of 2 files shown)


mononaut:
Fixed skeleton loader issue

softsimon (approved):
tested ACK @ [2df47640]

PR #4895, opened 2024-04-05
Add note about accelerator waitlist to faq

hunicus:
## Current

> This website only provides data about the Bitcoin network, so there is nothing it can do to help you get your transaction confirmed quicker.

## Proposed

> Another option to get your transaction confirmed more quickly is Mempool Accelerator™. This service is still in development, but you can <a _ngcontent-ng-c571986414="" href="https://mempool.space/accelerator">sign up for the waitlist</a> to be notified when it's ready.

Screenshot from 2024-04-05 16-51-40


softsimon:
This text makes sense for the mempool space website. But it will be wrong on all devices this version is installed on, for possibly a year.

softsimon (approved):
utACK

PR #4905, opened 2024-04-06
Mini miner cpfp

mononaut:
Until now, we've been using a fast heuristic method to find CPFP clusters and effective fee rates in mined blocks.

The heuristic relies on a quirk of Core's block building algorithm, which causes CPFP clusters to be predictably arranged in contiguous sequences.

Recently there's been an increase in transactions prioritized out-of-band, and when these transactions have same-block ancestors they tend to break the heuristic.

e.g, here a pair of parent-child transactions are prioritized at the very top of a block. Since the parent pays *more*. this isn't actually a CPFP, but is identified as one by our heuristic, causing us to display the wrong effective fee rate:

https://github.com/mempool/mempool/assets/83316221/e3f42b10-6e43-4fd5-a1ba-297e0a3e2d4c

Screenshot 2024-04-06 at 8 11 49 AM

In theory this could also happen with non-prioritized transactions, it's just much less likely that non-CPFP children would ever be included immediately after a parent.

Note that the obvious fix of adapting the heuristic to check if the child actually pays more isn't a complete solution, since clusters with more than two elements aren't necessarily included in order of descending fee rate.

---

This PR abstracts out the mini-miner algorithm from the acceleration and cpfp modules, and uses it to quickly perform GBT on a block's transactions to identify the exact actual CPFP relationships and effective fee rates.

This should be much more robust than the heuristic method, and also works for "weird" blocks with prioritized transactions or unusual template construction where the heuristic method would fail.

The PR also fixes some other minor CPFP-related bugs caused by vsize rounding and sticky Goggles flags.


softsimon:
Getting a slight change in `effectiveFeePerVsize` calculation from the backend with this PR

From
`    "effectiveFeePerVsize": 6.966216216216216,`

To
`    "effectiveFeePerVsize": 6.989830508474577,`

Is that fine?

mononaut:
rebased on master, and added (currently unused) code to automatically reindex CPFP clusters/effective fee rates in old blocks using the new method.

softsimon (approved):
tested ACK @ [39859382]

CPFP classifications and effective fee looks accurate now

PR #4911, opened 2024-04-07
fix(address): clicking on the Balance History chart within the /testnet/address/:id page would incorrectly navigate to /tx/:tx

daweilv:
# What :computer: 
* This PR fixes a bug where clicking on the Balance History chart within the /testnet/address/:id page would incorrectly navigate to /tx/:tx. 
* After the fix, it correctly navigates to /testnet/tx/:tx.

# Evidence :camera:
* Before (online)
image
image

*After (my local branch)
image
image


softsimon (approved):
tested ACK

PR #4912, opened 2024-04-08
Fix accelerated fee rates

mononaut:
Fix conflict between accelerated and effective fee rates on the transaction page.


wiz (approved):
utACK

PR #4914, opened 2024-04-08
Remove another faq reference to tx-acc

hunicus:
## Current

Screenshot from 2024-04-08 14-17-16

## Proposed

Screenshot from 2024-04-08 14-19-54


softsimon (approved):
utACK

PR #4921, opened 2024-04-08
Fix new acceleration visualization color change

mononaut:
Fixes a bug where newly accelerated transactions would not be highlighted if they were already in the block visualization.


softsimon (approved):
utACK

PR #4924, opened 2024-04-08
Fix missing accelerated fee rate again

mononaut:
Fixes blank accelerated fee rate on mined transaction:

| Before | After |
|-|-|
| Screenshot 2024-04-08 at 11 38 22 AM | Screenshot 2024-04-08 at 11 37 54 AM |


wiz (approved):
utACK

PR #4928, opened 2024-04-08
Make bid boost widget link clickable on mobile

hunicus:
Link isn't currently clickable on mobile.

bid-boost-graph-fix


wiz (approved):
utACK

PR #4929, opened 2024-04-08
Fix accelerated rate labels

### LINKED ISSUE #4927: Accelerated fee rate / effective fee rate
An accelerated transaction mined by a non partner pool is labeled as having an accelerated fee rate, rather than an effective fee rate. 

Left = mempool.space, Right = 2.5.0

image

mononaut:
Fix incorrect "Accelerated fee rate" labels on transaction which were accelerated, but mined by another pool (fixes #4927).

Also fixes a bug where "accelerated" status was saved to the cached block summaries, causing incorrect highlighting in the visualization.


wiz (approved):
utACK

PR #4930, opened 2024-04-09
Filter accelerations for matching pool

mononaut:
Second part of #4929, filtering accelerations by pool at the block level to fix any remaining cases of incorrect highlighting in the mined block visualization.

**Test procedure:**
 - Navigate to block 00000000000000000002f3a330d44ad274ac55409a47ebebf31c3599ec04d8a7
 - Confirm there are no purple transactions in the mined template.


wiz (approved):
utACK

Issue #4908, opened 2024-04-07
"2 of 1 transactions"


softsimon:
How to reproduce:

1. Open any new address that has no funds
2. Send a transaction while watching

![IMG_20240407_162103_795.jpg](https://github.com/mempool/mempool/assets/8561090/3d8a959e-3c96-4d32-873e-f237768d427e)

natsoni:
Can't reproduce on testnet:

https://github.com/mempool/mempool/assets/46578910/f3b8bebd-6166-4b53-bfc4-c3ce066a4e2f

PR #4933, opened 2024-04-09
Fix subscription management in address component

natsoni:
This PR hopefully fixes the intermittent bug described in #4908 by unsubscribing to mempool transactions on destroy to avoid duplicated subscriptions.


softsimon (approved):
utACK

PR #4935, opened 2024-04-09
Fix accelerated fee rate in mined block tooltips

mononaut:
Fixes an issue where the wrong rate would be displayed in the block visualization tooltip for some accelerated transactions.


wiz (approved):
utACK

PR #4937, opened 2024-04-09
disable accelerate button after submission

mononaut:
gray out the acceleration panel and hide the sticky button after submitting an acceleration to avoid duplicate requests


wiz (approved):
utACK

PR #4939, opened 2024-04-09
Make room for natsoni as project member

hunicus:
Screenshot from 2024-04-09 20-47-47

### DIFF
--- a/frontend/src/app/components/about/about.component.scss
+++ b/frontend/src/app/components/about/about.component.scss
@@ -177,6 +177,10 @@
     }
   }
 
+  #project-members a.project-member-avatar img {
+    margin: 40px 20px 10px;
+  }
+
   .copyright {
     text-align: left;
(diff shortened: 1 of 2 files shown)


frontend/src/app/components/about/about.component.scss
```diff
@@ -177,6 +177,10 @@
     }
   }
 
+  .project-member-avatar img {
+    margin: 40px 20px 10px !important;
```
wiz, line 181: use of `!important` is discouraged, can't you fix this in a cleaner way

hunicus: Removed with [6daffe5](https://github.com/mempool/mempool/pull/4939/commits/6daffe5b13ffd59eabc65c8d92ffa38f78b76225).

softsimon (approved):
utACK

PR #4941, opened 2024-04-10
Polish global footer

natsoni:
This PR polishes the global footer to make place for the new theme selector and fixes an issue where the theme selector did not update properly on screen resize.


softsimon (approved):
tested ACK

PR #4942, opened 2024-04-10
Fix header color theme

natsoni:
This PR makes the navbar color adapt to the current theme so that it always matches the background color of the dashboard widgets. 

On high contrast / wiz theme: 

| Before | After |
| Screenshot 2024-04-10 at 12 29 37 | Screenshot 2024-04-10 at 12 30 58 |


softsimon (approved):
tested ACK @ [cabe629f1]

PR #4948, opened 2024-04-11
it hasn't been 54 years since this transaction was submitted

mononaut:
Restore the !== 0 check for transaction first seen times

54years


softsimon (approved):
tested ACK @ [fa9aaf0423]

PR #4949, opened 2024-04-11
Add sequence number to mempool block updates

mononaut:
In some browsers, when mempool is running in a background tab the process can get put to sleep and silently drop websocket messages.

Since our mempool block visualization subscription assumes reliable transport and only sends deltas most of the time, this can cause the visualization to get stuck showing stale transactions.

This PR adds a sequence number to the websocket `projected-block-transactions` messages, so that the frontend can detect when we've missed websocket updates and trigger a resync.


softsimon (approved):
tested ACK @ [3e6d3865]

Billym45:
I give softsimon all permission on my account to merge fix repair all my downloads he is second in command

PR #4950, opened 2024-04-11
Update prices table  'time' field to datetime

natsoni:


softsimon (approved):
tested ACK @ [79bfe9c866]

PR #4958, opened 2024-04-13
Add new VA1/FRA lightning nodes

wiz:


wiz (commented):
Tested ACK @ v3.0.0-dev [80c75f9ae]

PR #4965, opened 2024-04-13
[accelerator] remove simple mode from tx page

nymkappa:


wiz (approved):
I really like the sticky button, but ok utACK

mononaut:
utACK

PR #4968, opened 2024-04-14
Fix first seen loading time in tx tracker

mononaut:
Fix broken skeleton loader logic in the first seen time on the tx tracker page:
| Before | After |
|-|-|
| Screenshot 2024-04-14 at 3 51 09 AM | Screenshot 2024-04-14 at 3 50 52 AM |


wiz (approved):
utACK

PR #4969, opened 2024-04-14
[accelerator] re-add missing accelerator button on mobile, set locati…

nymkappa:
Fixes https://github.com/mempool/mempool.space/issues/908


wiz (approved):
utACK

PR #4979, opened 2024-04-16
Fix cropped transaction tooltip in block overview on mobile

natsoni:
This PR allows block overview tooltip not to be cropped on mobile.

| Before  | After |
|  |


softsimon (approved):
tested ACK @ [89ef5fe33d]

PR #4981, opened 2024-04-17
Fix miner data loading in transaction details

natsoni:
This PR fixes an issue where miner data would take some time before appearing in the transaction details when the audit feature is not available (in older blocks for example): 

https://github.com/mempool/mempool/assets/46578910/08b966c5-9922-4bbd-b7bf-31ca3194a024

This was fixed by using two separate subscriptions for the mining and audit data.


softsimon (approved):
tested ACK @ [bbaa5dafb]

PR #4985, opened 2024-04-18
Fix audit blockIndex increment bug

mononaut:
Fixes a bug where `projectedBlocks[blockIndex]` could be null due to incrementing `blockIndex` prematurely.


nymkappa (approved):
utACK, just reading the code change clearly shows the possible access to `undefined` here. Thanks.

softsimon (approved):
tested ACK @ [233f19d23]

PR #4995, opened 2024-04-19
Fix table pagination colors on custom themes

natsoni:
| Before  | After |
| IMG_1448 | IMG_1449 |


softsimon (approved):
utACK

PR #4997, opened 2024-04-19
Fix dropdown menu navigation with arrow keys

natsoni:
This PR fixes a small issue where the dropdown menu would not highlight the selected item when using arrow keys on contrast themes.


softsimon (approved):
utACK

PR #5004, opened 2024-04-21
Add block display mode toggle

natsoni:
This PR adds a toggle above the blockchain divider to choose between block fees or block vsize on blocks.
- the displaying of block fees does not depend on which dashboard we are anymore (by default we display vsize even on the mining and acceleration dashboards)
- block vsize is displayed by default, and user preference is stored in local storage

https://github.com/mempool/mempool/assets/46578910/4857becc-dfdf-48cc-b17d-56e9719e033f


softsimon (approved):

softsimon (approved):
tested ACK @ [1c9e2b546]

PR #5010, opened 2024-04-22
Transparently redirect direct mobile tx visits to pizza tracker

mononaut:
_(requires https://github.com/mempool/mempool.space/pull/1205)_

This PR moves the new "Pizza Tracker" page under the same `/tx/:txid` route path as the main transaction page, and adds a `canMatch` route guard to selectively render the pizza tracker component if all of the following conditions are met:
 - Initial page load (i.e. not an internal navigation)
 - On mobile (`window.innerWidth <= 767.98`)
 - *NOT* overridden by query parameter `?mode=details`
 
If the mode=details query parameter is set in the URL, then the full transaction page will always be displayed, so this can be used to explicitly link to the main page instead of the pizza tracker.

When navigating to a /tx page internally on mobile, `?mode=details` is automatically added to prevent forcing the user back to the tracker view on page reload/refresh.

~~Leaving in draft for now because the tracker page needs some more polish before we funnel all of our `/tx/` traffic there.~~

### DIFF
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -215,10 +215,6 @@ let routes: Routes = [
     loadChildren: () => import('./bitcoin-graphs.module').then(m => m.BitcoinGraphsModule),
     data: { preload: true },
   },
-  {
-    path: '**',
-    redirectTo: ''
-  },
 ];
 
 if (browserWindowEnv && browserWindowEnv.BASE_MODULE === 'liquid') {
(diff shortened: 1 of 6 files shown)


mononaut:
rebased on master, and taking out of draft (pending PR #5137)

Zoiediez:
> rebased on master, and taking out of draft (pending PR #5137)

mononaut:
rebased on current master & resolved merge conflicts

mononaut:
this doesn't work with the hybrid build - I'll reopen once I figure out a fix

mononaut:
hybrid build issues have been resolved, but the PR now needs to be tested & merged in tandem with https://github.com/mempool/mempool.space/pull/1205

wiz (approved):
utACK

PR #5013, opened 2024-04-24
Remove Mining dashboard from Liquid footer

natsoni:
The mining dashboard appears in the footer on Liquid base module. This PR adds a check to display it only on mempool base module with mining dashboard enabled.


softsimon (approved):
tested ACK @ [cd689afa]

PR #5014, opened 2024-04-24
Minor changes on Liquid frontend

natsoni:
This PR brings minor changes to the Liquid audit. 

Until now, the frontend claims a `unpeg event` as soon as the Federation's holdings are 1 sat lower than the L-BTC supply, which is a little dramatic imo:
Screenshot 2024-04-24 at 14 27 23
To fix this we add a 5% tolerance to the trigger, meaning that the the Federation's holdings needs to be less than `0.95 * L-BTC supply` to announce an `unpeg event`. 

This PR also fixes some color issues in the Federation UTXOs table, and an issue where the gauge pointer disappears on hover.


softsimon (approved):
utACK

PR #5019, opened 2024-04-28
Add new color theme

natsoni:
This PR adds a new color theme, alongside some CSS improvements for more color customisation using the themes. 
The new theme is not perfect but ready for review, and we can iterate from here for any improvements.


softsimon:
Merged in https://github.com/mempool/mempool/pull/5044

PR #5023, opened 2024-04-29
Change block display default mode to fees

natsoni:
Vox populi, vox dei?


softsimon (approved):
tested ACK @ [57b466d8]

PR #5025, opened 2024-04-29
Replace bisq with onbtc for about alliances

hunicus:
Screenshot from 2024-04-29 22-38-53


frontend/src/app/components/about/about.component.html
```diff
@@ -343,8 +343,8 @@ <h3 i18n="about.alliances">Community Alliances</h3>
       <a href="https://opencrypto.org/" title="Coppa - Crypto Open Patent Alliance">
         <img class="copa" src="/resources/profile/copa.png" />
       </a>
-      <a href="https://bisq.network/" title="Bisq Network">
-        <img class="bisq" src="/resources/profile/bisq.svg" />
+      <a href="" title="Oficina Nacional del Bitcoin">
```
softsimon, line 346: If it's not going to be clickable (it's still clickable but go to the start page), better just remove the a link and add the title attribute to the img tag directly.

softsimon (changes requested):

frontend/src/app/components/about/about.component.html
wiz: we can add the URL, just wait a few days

hunicus:
Screenshot from 2024-05-12 17-20-05

softsimon (approved):
tested ACK @ [bf541f089]

PR #5026, opened 2024-04-29
Polish address balance graph, add period toggle

mononaut:
This PR adds a bit of polish to the address balance graph, including:
 - a toggle between 'all' history and 'recent' (last 30 days)
   - (the toggle is hidden and defaults to 'all' if there is no activity on the address within the last 30 days)
 - moves the heading outside of the panel to match the rest of the site
 - minor adjustments to padding etc
 - fixes a couple of bugs

Before:
Screenshot 2024-04-29 at 8 59 59 PM

After:
Screenshot 2024-04-29 at 9 02 28 PM


softsimon (approved):
tested ACK @ [a582a3b0]

PR #5032, opened 2024-05-03
Fix mempool animation desync

mononaut:
This PR lightly refactors the frontend mempool block visualization update pipeline & animation scheduling to fix a number of lingering issues from PR #4556, #4949.

In particular:
 - the mempool visualization should no longer get out of sync or fall behind on long-running pages or background tabs.
 - the animations should be smoother and more consistent
 - the animation should now tolerate faster websocket updates (including lower `POLL_RATE_MS` settings).
 - No more flashes of empty templates when switching between mempool blocks.


softsimon (approved):
seems to work fine,

tested ACK @ [44116424]

PR #5033, opened 2024-05-03
Add block fees vs subsidy bar graph

### LINKED ISSUE #4998: Mining reward graph
Add a new Bitcoin mining reward graph that visualize the ratio between subsidy and fees.

image

natsoni:
Closes #4998

This PR adds the `Block Fees vs Subsidy` graph to the mining graphs. 
The data is not averaged between blocks, so we limit to display 20,000 blocks at a time. Also the navigation is done by block, not by time interval. 
The accelerations data is displayed if the frontend config has the `ACCELERATOR: true` flag

Screenshot 2024-05-03 at 12 36 31

This PR also adds an accelerations category to the graphs page.

Screenshot 2024-05-03 at 12 43 33


softsimon (changes requested):
Since I don't have a full local index, I get this span as a default with 19758 as highest block height, but graph show the latest block heights.

It loads 18 000 data points which causes the graph to be a bit laggy. Is there or should there be some limit to how much it can fetch? 

Screenshot 2024-05-04 at 14 33 16

If Lightning and Accelerations are disabled, this is how the menu looks like

Screenshot 2024-05-04 at 14 33 20

natsoni:
> Since I don't have a full local index, I get this span as a default with 19758 as highest block height, but graph show the latest block heights.

> If Lightning and Accelerations are disabled, this is how the menu looks like

These bugs should be fixed now

> It loads 18 000 data points which causes the graph to be a bit laggy. Is there or should there be some limit to how much it can fetch?

There is a limit of 20,000 data points on each page of the graph, I thought that was a good compromise between lag and usability of the graph. Do you think we should limit to 15,000 or even 10,000?

softsimon:
> There is a limit of 20,000 data points on each page of the graph, I thought that was a good compromise between lag and usability of the graph. Do you think we should limit to 15,000 or even 10,000?

When we got to a long timespan for Mempool data there is like 3000 data points, so 20K could be a bit much. 10K should be the data limit, if it doesn't lower the overall experience.

natsoni:
> 10K should be the data limit, if it doesn't lower the overall experience.

Updated to 10k blocks, the graph looks smooth enough to me with this limit.

natsoni:
Given feedback, I reverted the graph to use classic timespans. It looks ok even close to halvings:

Screenshot 2024-05-11 at 16 29 24

Will now work to add the feature to get smaller granularity as we zoom in the graph (and will try to make it easy to apply to other block-based graphs)

natsoni:
Better granularity is displayed as the user zooms in the graph: 

https://github.com/mempool/mempool/assets/46578910/fffb0d05-4a89-4e20-8bb9-dea373d81f29

frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
```diff
@@ -0,0 +1,551 @@
...
+    });
+  }
+
+  @HostListener('document:mouseup', ['$event'])
+  onMouseUp(event: MouseEvent) {
+    if (this.updateZoom) {
+      this.onZoom();
+      this.updateZoom = false;
+    }
+  }
```
mononaut, line 399: note that triggering `onZoom()` from a `mouseup` event won't work on touchscreen devices:

https://github.com/mempool/mempool/assets/83316221/ab0e8b26-a671-487e-af5c-03346ca12b9e

switching to `pointerup` might be good enough, otherwise you could add a debounced call directly from the datazoom event handler to catch changes more generically.

frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
```diff
@@ -0,0 +1,551 @@
...
+    const endTimestamp = timestamps[option.dataZoom[0].endValue];
+
+    // Avoid to fetch new data if not needed
+    if (option.dataZoom[0].end - option.dataZoom[0].start >= this.zoomSpan + 0.1
+       || this.getTimeRangeFromTimespan(Math.floor(startTimestamp / 1000), Math.floor(endTimestamp / 1000), false) as number === this.getTimeRange(this.zoomTimeSpan)
+      ) {
+      this.zoomSpan = option.dataZoom[0].end - option.dataZoom[0].start;
+      this.zoomTimeSpan = this.getTimeRangeFromTimespan(Math.floor(startTimestamp / 1000), Math.floor(endTimestamp / 1000), true) as string;
+      return;
+    }
```
mononaut, line 428: I suspect this logic could get us into trouble once we have a few different granularities of data all spliced together.

for example, after playing with the chart for a bit I ended up in this slightly confusing situation where each half of screen is on a different timescale:
Screenshot 2024-05-14 at 9 22 39 PM

mononaut (commented):
Very slick!

I've left a couple of code comments with minor issues I noticed, but overall this is a really nice addition to the mining charts.

frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
natsoni: This is because we don't reload data when the user zooms out: this allows to keep showing precise data on zoom out, but it's quite confusing as you showed.

frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
natsoni: Looks like `pointerup` works on mobile (tested on iOS safari).
I didn't use a debounced call because I didn't find an easy way to prevent the scenario where the `onZoom()` function gets triggered when the user pauses for longer than the debounce timer while zooming. Using `pointerup` ensures that the user is done zooming before continuing.

natsoni:
Thanks for the review @mononaut! 
The mechanism should now work on touchscreens, and it will fetch new data on every zoom to fix the weird situation with different granularities on view.

frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
mononaut: making a fresh API call on *every* change of zoom seems a little excessive, but it is probably the safest approach.

otherwise maybe it's possible to do something like:
 - cache the data in separate "layers" for each zoom level (1 min, 5 min, 30 min etc).
 - fetch any data that's missing from the target layer/zoom level within this datetime range.
 - replace the echart data series with the result of flattening the cached layers <= the target zoom level.
 
then you're only fetching data you don't have already, and never have to worry about random spans of 1 min data messing up the time axis on your 2hr view etc.

but tbh the extra complexity might not be worth the benefit 🤷

natsoni: I thought about doing something similar, and came up to the same conclusion as you 😅 but that's definitely the most optimal solution

softsimon (approved):
tested ACK @ [86c3ef68e]

I'll let @mononaut give the final ACK

mononaut (approved):
Tested ACK @ [86c3ef68e422b32a91b4e5439e538fb210f272a9]

PR #5034, opened 2024-05-03
Prevent address page to load more txs unnecessarily

natsoni:
This PR fixes a small issue where the address page would try to unnecessary load more transactions when addresses have less transactions than the number of transactions in each page (50 on prod):

https://github.com/mempool/mempool/assets/46578910/de10374c-805e-4722-9980-4163d60e316e


softsimon (approved):
tested ACK @ [66f431d3]

PR #5035, opened 2024-05-04
Replace acceleration API polling with websocket

mononaut:
This PR adds a new `'track-accelerations'` websocket subscription, and updates the acceleration dashboard to receive updates pushed through this subscription instead of polling the services backend API directly.

This makes the dashboard much more responsive to changes in the set of pending accelerations, and also reduces the load on the services backend.

### DIFF
--- a/backend/src/api/mempool.ts
+++ b/backend/src/api/mempool.ts
@@ -404,6 +404,10 @@ class Mempool {
 
       const newAccelerationMap: { [txid: string]: Acceleration } = {};
       for (const acceleration of newAccelerations) {
+        // skip transactions we don't know about
+        if (!this.mempoolCache[acceleration.txid]) {
+          continue;
+        }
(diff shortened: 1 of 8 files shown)


nymkappa (changes requested):
I've resolved conflicts. I've noticed the block template does not render the purple accelerated transaction unless I reload the page. But I think it might been an old bug (?)

Screenshot 2024-05-04 at 09 47 06

And vice versa

image

mononaut:
Fixed broken merge conflict resolution

nymkappa (approved):
Tested ACK @ [a29b293](https://github.com/mempool/mempool/pull/5035/commits/a29b29300e772e9e1f974156b9300c06e39db7c8)

Would be nice to refresh the `Acceleration stats` widgets as well when:
- A new tx is being accelerated
- A currently accelerated tx is not anymore (for whatever reason)

Issue #5036, opened 2024-05-04
[lightning] socket too long


nymkappa:
```
May  4 10:02:05 [53540] DEBUG: <lightning> database query "INSERT INTO nodes(
          public_key,
          first_seen,
          updated_at,
          alias,
          alias_search,
          color,
          sockets" failed!
May  4 10:02:05 [53540] ERR: <lightning> $saveNode() error: Data too long for column 'sockets' at row 1
```

wizard872:
i have same issues my self hosted mempool server

![image](https://github.com/mempool/mempool/assets/36080488/7d852b25-954f-4a2d-912d-3979735c4a4e)

softsimon:
> i have same issues my self hosted mempool server
> 

It should be fixed now on latest master

PR #4991, opened 2024-04-19
[lightning] unique sockets for ln nodes

nymkappa:

PR #5037, opened 2024-05-04
[lightning] apply #4991 at node creation as well

nymkappa:
Fixes https://github.com/mempool/mempool/issues/5036


softsimon (approved):
tested ACK @ [2e49ef38c9]

PR #5042, opened 2024-05-07
Add statistics to replication service

### LINKED ISSUE #2831: Server data synchronization service
A service that synchronizes and fills in missing data gaps. Starting with block audit data and mempool minute graph.

**Description**
When servers restart, there's always risk of the database missing a minute of the mempool graph data, and audit data if a block was found at the same time, and it takes a minute to catch up to the mempool synced state.

**Proposed solution**
Every server should be able to detect it's own gaps and reach out to other servers to fetch them, and fill in the local database.
A new config with list of remote servers to synchronize from in priority order.

Possibly new special endpoints is needed that can fetch specific datapoints that we want, that needs to be locked down behind some API key to not be abused.

natsoni:
Resolves #2831

This PR adds support for statistics data in the replication service. It works similarly to the audit replication by fetching statistics from trusted servers.

**How to use the service:** 
```
"REPLICATION": {
  "ENABLED": true,
  "STATISTICS": true,
  "STATISTICS_START_TIME": <timestamp> // to fill the statistics table from <timestamp> to now
  "SERVERS": [ ... ]
}
```

To fully fill the `statistics` table with data from 2017 up to now on new instances, we can set
```
"STATISTICS_START_TIME": 1481932800
```

It is also possible to fill data up to a relative time in the past, by setting `STATISTICS_START_TIME` to one of these timespans: 
```
'24h', '1w', '1m', '3m', '6m', '2y', 'all'
```

To only fill missing rows in the `statistics` table after server restart (assuming servers are not down more than 24h), we can set 
```
"STATISTICS_START_TIME": "24h"
``` 

This will fill missing rows in the statistics table from 24h ago to now.


mononaut (commented):
Tested, this worked great to fill a bunch of gaps in my local statistics.

I left a small comment about the approach for `$getMissingStatisticsInterval()`, which probably also applies to `$syncStatistics` to a lesser degree, but otherwise this looks good to me.

backend/src/replication/StatisticsReplication.ts
```diff
@@ -0,0 +1,217 @@
...
+          missingTimes.push(time);
+        }
+      }
+
+      return missingTimes;
+    } catch (e: any) {
+      logger.err(`Cannot fetch missing statistics times from db. Reason: ` + (e instanceof Error ? e.message : e));
+      throw e;
+    }
+  }
```
mononaut, line 172: It feels like there should be a cleaner/more efficient way to find the missing points than iterating over the entire list of data each time with `getClosestElement(...)`.

My instinct would be to round everything off to the nearest `step`, throw it in a Set or a dictionary, and then do a single pass over the expected *rounded* values to see what's missing.

For the longer intervals, you can probably get the points out of the DB pre-rounded to the correct `step` intervals by using the same `GROUP BY` logic as the statistics APIs you'll be hitting later to fetch the missing data:

https://github.com/mempool/mempool/blob/12d212b89f4a57948bf34ae01943d30ce09ecafa/backend/src/api/statistics/statistics-api.ts#L269-L271

natsoni:
Thanks for the kind review! Indeed my implementation was terrible, we can do better than a quadratic complexity for this kind of task 😅
I re-implemented the service according to your feedback, so the computation to find the missing points should now have linear complexity.

backend/src/replication/StatisticsReplication.ts
```diff
@@ -0,0 +1,231 @@
...
+            try {
+              await statistics.$create(statistics.mapOptimizedStatisticToStatistic([stat])[0], true);
+              synced++;
+            } catch (e: any) {
+              logger.err(`Failed to insert statistics row at ${time} (${interval}) from ${syncResult.server}. Reason: ` + (e instanceof Error ? e.message : e));
+            }
+          } else {
+            missed++;
+          }
+        }
```
mononaut, line 101: this approach works fine, but it does mean we lose a bit of precision in the 2h/24h charts by discarding some of the extra points before and after new blocks (see https://github.com/mempool/mempool/pull/4676).

| Screenshot 2024-05-21 at 11 26 13 PM | vs | Screenshot 2024-05-21 at 11 28 10 PM |
|-|-|-|

an alternative would be to iterate over the fetchedTimes instead of the missingTimes, and accept anything within half a step of a missingTime entry. that should still avoid duplicate data, while filling any gaps with the best level of detail.

natsoni: > iterate over the fetchedTimes instead of the missingTimes, and accept anything within half a step of a missingTime entry

Right, doing so fixes the issue: 

Trusted server: 
Screenshot 2024-05-22 at 18 18 36

Replication (the drop at the end is my local mempool not being the same as the trusted server): 
Screenshot 2024-05-22 at 18 18 58

mononaut (approved):
Tested ACK @ [60a07fb0]

PR #5044, opened 2024-05-08
onbtc theme

mononaut:
_(extends and supercedes #5019)_

updates onbtc theme and branding


softsimon (approved):
tested ACK @ [c15393981d]

PR #5045, opened 2024-05-08
Fix custom dashboard address graph

mononaut:
Fixes a regression in the custom dashboard address balance widget that causes it to load forever & throw errors when trying to navigate away.


softsimon (approved):
tested ACK @ [9495d7d2e]

PR #5049, opened 2024-05-08
Fix transparent navbar

natsoni:


softsimon (approved):
tested ACK @ [fc7614608]

PR #5053, opened 2024-05-09
Fix unfurler resources build

mononaut:
- `/resources/sv/` -> `/resources/onbtc/`
- clone unfurler resources directly from `../frontend/dist/mempool/browser/resources`
- no more symlinks


wiz (approved):
utACK

PR #5054, opened 2024-05-09
Fix statistics graphs

mononaut:
Since #4676, we log additional statistics points before and after new blocks. This breaks some assumptions in the statistics charts:
 a) that there would always be 60 points per hour, and
 b) no duplicate timestamps. 

This PR fixes assumption (a) in the incoming transactions dashboard widget by truncating data by time instead of by number of points.

It also fixes (b) by de-duplicating items in the tooltip legend for the incoming transactions and main mempool graphs.

And also fixes an unrelated old `undefined` tooltip title bug I noticed while I was in there.


softsimon (approved):
tested ACK @ [cd9e33616]

PR #5055, opened 2024-05-09
Update custom sv html

mononaut:
Updates the custom onbtc index.html, moves it to `index.mempool.onbtc.html`, and updates `generate-config.js` to use that as the base for frontend builds when customization is enabled.


wiz (approved):
utACK

PR #5061, opened 2024-05-11
update empty block explainer, add contributor info

BitcoinMechanic:


Retropex (commented):
ACK

mononaut (commented):
Thanks for the PR!

Do you know which mining hardware specifically suffers from this issue of persisting on old work, and roughly what proportion of total hashrate that represents?

I did some digging into the statistics on empty blocks, and while there's certainly evidence of some miners continuing to work on empty templates for 60 seconds or more, I have yet to convince myself that this phenomenon explains the *majority* of empty blocks.

For example, this histogram of block arrival time intervals shows that most empty blocks arrive within ~4 seconds, with the frequency falling off exponentially beyond ~1.5 seconds.

Screenshot 2024-05-12 at 11 15 06 PM


And from monitoring stratum v1 updates, it's fairly normal to see delays of 1-2 seconds in receiving a full template job after an empty one (and surprisingly not *un*common to observe delays as long as 20 or 30 seconds from certain pools).

wizkid057:
> Do you know which mining hardware specifically suffers from this issue of persisting on old work, and roughly what proportion of total hashrate that represents?

All modern multi-chip miners do this to some extent. Some worse than others. Hardware made by the same folks that run the pool with the highest overall number of empty blocks tend to be the worst, however, with some variants of that hardware worse than others.

I'm unsure what that would be total hashrate wise off hand, but its clearly a significant portion.

> I did some digging into the statistics on empty blocks, and while there's certainly evidence of some miners continuing to work on empty templates for 60 seconds or more, I have yet to convince myself that this phenomenon explains the _majority_ of empty blocks.
> 
> For example, this histogram of block arrival time intervals shows that most empty blocks arrive within ~4 seconds, with the frequency falling off exponentially beyond ~1.5 seconds.

Of course it becomes less likely as time passes.  Let me run through what happens with a bit of a simplified/contrived example which may help. You'll see that it fits well with your graph.

- Let's say we have a miner with 4 ASIC chips on it.
- These are crunching away at work that takes them roughly 4 seconds to complete before they can switch to the next work (in real hardware this will vary and drift).
- T+0s: A new job comes in for a new block with empty block work.
- The miner software immediately generates work for these chips to start on the empty work.
- T+0.1s: ASIC 0 finishes its current job (old block) and switches to the new empty block work.
- T+1.1s: ASIC 1 finishes its current job (old block) and switches to the new empty block work.
- T+1.5s: The pool sends new work with a full template
- T+2.1s: ASIC 2 finishes finishes its current job (old block) and switches to the new FULL block work.
- T+3.1s: ASIC 3 finishes finishes its current job (old block) and switches to the new FULL block work.
- T+4.1s: ASIC 0 finishes its current job (new empty block work) and switches to the new FULL block work.
- T+5.1s: ASIC 1 finishes its current job (new empty block work) and switches to the new FULL block work.

It took about 5 seconds for this machine to get itself 100% on to the new full block work, even though in theory it should have been < 2 seconds.

Now scale that up to ~100 chips on a machine, toss in an underpowered SoC running the show, and stock firmware that doesn't care to optimize this situation _at all_... and you end up with some chips potentially on the empty block work for 30s or more, depending on how timings work out (which is effectively random in real hardware, and the chips are never sync'd in line like the contrived example above).  I've tested some machines hanging on to jobs for well over 60 seconds.  

Keep in mind that almost all modern ASIC chips have no method to actually interrupt currently running work.  They can only specify the next work in a double-buffered fashion.  The amount of time between job changes at the chip level varies based on firmware, clock speeds, chip, type, etc etc.  Most pools also use a work update interval around 30 seconds, so the firmware tends to setup the staggering of work headed into the ASICs with this in mind.

So what you see in your graph fits perfectly with this situation, since the chips running the empty block job will fall off pretty quickly, but not entirely.  The hash rate working on the empty work might be super low after 3 seconds, but not zero.

As a bit of preemptive clarification: Sending stratum work with clean=true doesn't actually change how work is handled at the chip level on most hardware. With clean=true, all it does is just not send valid nonces to the pool when work from a pre-clean job is found. In the case many suggest of empty block work with clean=true followed by full block work with clean=true, this would mean the firmware would now just discard valid empty blocks instead of submitting those to the pool. It doesn't stop the machine from working on them nor get it on the full work any faster.

Hopefully this helps.

mononaut:
Thanks for the great explanation, that clarifies things a lot!

I think I'm satisfied that a good chunk of empty blocks are explained by this delay in switching over the new work buffer, and if the timing varies not just between machines but even between different chips in the same machine, then the exponential decay we see in the graph definitely makes sense.

I still have some reservations about making strong claims that this is the *main* cause of empty blocks vs simple delays in receiving a full template, especially during the first few seconds.

> Miners sometimes get lucky and find the empty block in the interim period before having the full template _**however this is rarely the reason for an empty block**_ as it's such a small amount of time.
>
>_**The usual cause is that**_ some mining hardware will continue to work on old work for up to 60 seconds despite having newer work.

But if we just soften that a little, I think it would be fine. Perhaps something like:

> When a new block is found, mining pools send miners a block template with no transactions so they can begin doing useful work as soon as possible. The pool will send a block template full of transactions soon after, but it takes slightly longer to reach miners, and some mining hardware will not switch to the new work immediately.
>
>Miners sometimes get lucky and find a block using the empty template during this interim period.

It's probably also worth mentioning your point about how `clean=true` simply discards valid work, since it's a frequent point of misunderstanding that empty blocks are somehow wasteful or prevent miners from getting to work on full templates.

BitcoinMechanic:
@mononaut updated with your suggestions - thanks

mononaut (approved):
Tested ACK @ [6b2b3459f7c57637529c6ba3a7dcee53140441ad]

PR #5062, opened 2024-05-12
Configurable threshold for esplora tip check

mononaut:
This PR adds a new optional config setting `ESPLORA.MAX_BEHIND_TIP`, which sets how far the `mempool/electrs` backend is allowed to fall behind the best tip before it's considered out of sync. (the previous hardcoded maximum was 2 blocks).

The PR also redefines the current best tip height in the electrs health checks to be the maximum height currently reported by healthy electrs peers, rather than the maximum height *ever seen*. This improves fallback behavior in the event of a reorg to a lower height.


backend/mempool-config.sample.json
```diff
@@ -59,7 +59,8 @@
     "RETRY_UNIX_SOCKET_AFTER": 30000,
     "REQUEST_TIMEOUT": 10000,
     "FALLBACK_TIMEOUT": 5000,
-    "FALLBACK": []
+    "FALLBACK": [],
+    "MAX_BEHIND_TIP": 2,
```
softsimon, line 63:
```suggestion
    "MAX_BEHIND_TIP": 2
```

softsimon (approved):
tested ACK @ [fda40cad]

backend/mempool-config.sample.json
Zoiediez: This was actually stolen from me is it 30,000 Maddox

Zoiediez: Matrix

Zoiediez: I have two other portfolios I need to look at as well

Zoiediez: Okay ty

PR #5067, opened 2024-05-12
Replace twitter logo with x logo

hunicus:
Now that there is a site hosted by Mempool with X, references to the bird site should be consistent.

Screenshot from 2024-05-12 16-24-39


softsimon (approved):
tested ACK @ [4729a87b3]

PR #5068, opened 2024-05-12
Fix error handling for non-critical services API requests

mononaut:
Fixes forever-loading mined block visualizations on mainnet when the services backend is down, caused by failed services API requests interrupting the rxjs pipeline.

Before:

https://github.com/mempool/mempool/assets/83316221/6d5ec05f-c20f-4cfe-ad0b-6088e4419bce

After:

https://github.com/mempool/mempool/assets/83316221/50639833-3788-4c80-ba05-139109af78f0


softsimon (approved):
tested ACK @ [c93aa4d82]

PR #5071, opened 2024-05-13
Fix nginx caching for customize.js resource

mononaut:
Updates the foss build nginx caching for `/resources/customize.js` to match the existing rules for `/resources/config.js`


wiz (approved):
utACK

PR #5072, opened 2024-05-14
Fix index.html generation for custom frontend builds

mononaut:


softsimon (approved):
tested ACK @ [2ffaeb50d]

PR #5074, opened 2024-05-15
Preserve site-specific meta title/descriptions/canonical urls

mononaut:
This PR updates the SeoService and fixes canonical link overrides to preserve whatever title, description and canonical domain is set in the base `index.html` meta tags, instead of overriding them with the default content.

These changes should also be reflected in slurped content served to SEO crawlers, which should fix the page descriptions etc for custom instances in search engine results.

Existing titles/descriptions/canonical links for the main site and Liquid should be unchanged.


softsimon (approved):
tested ACK @ [69e1d6150]

PR #5080, opened 2024-05-17
[faucet] polish code and UX - Move Twitter login component into FOSS

nymkappa:
Screenshot 2024-05-17 at 20 23 35
Screenshot 2024-05-17 at 20 23 47
Screenshot 2024-05-17 at 20 24 03
Screenshot 2024-05-17 at 20 25 03
Screenshot 2024-05-17 at 20 29 30
image
Screenshot 2024-05-17 at 20 45 18

### DIFF
--- a/frontend/src/app/components/faucet/faucet.component.html
+++ b/frontend/src/app/components/faucet/faucet.component.html
@@ -41,7 +41,7 @@
     }
 
     @if (!loading) {
-      <form [formGroup]="faucetForm" class="formGroup" (submit)="requestCoins()" [style]="(error || !this.user) ? 'opacity: 0.3' : ''">
+      <form [formGroup]="faucetForm" class="formGroup" (submit)="requestCoins()" [style]="(error || !this.user) ? 'opacity: 0.3; pointer-events: none' : ''">
         <div class="row">
           <div class="col">
             <div class="form-group mb-0">
--- a/frontend/src/app/components/faucet/faucet.component.html
+++ b/frontend/src/app/components/faucet/faucet.component.html
@@ -4,43 +4,45 @@
     <h1 i18n="testnet4.faucet">Testnet4 Faucet</h1>
   </div>
 
-  @if (error) {
-    <div class="alert alert-danger">
-      @switch (error) {
-        @case ('faucet_too_soon') {
-          Too many requests! Try again later.
-        }
-        @case ('faucet_maximum_reached') {
-          You have exceeded your testnet4 allowance. Try again later.
-        }
-        @case ('faucet_not_available') {
-          The faucet is not available right now. Try again later.
-        }
-        @default {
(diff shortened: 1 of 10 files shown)


frontend/src/app/components/faucet/faucet.component.html
```diff
@@ -4,43 +4,52 @@
...
+      </div>
+    }
+    
+    @else if (error) {
+      <!-- User can request -->
+      <app-mempool-error class="w-100" [error]="error"></app-mempool-error>
+    }
+
+    @if (!loading) {
+      <form [formGroup]="faucetForm" class="formGroup" (submit)="requestCoins()" [style]="(error || !this.user) ? 'opacity: 0.3' : ''">
```
mononaut, line 44: When the form is grayed out, we should probably also disable user interaction with the inputs, and/or hide the validation errors

Screenshot 2024-05-17 at 7 27 41 PM

frontend/src/app/components/faucet/faucet.component.ts
```diff
@@ -1,66 +1,89 @@
...
   ) {
+    this.faucetForm = this.formBuilder.group({
+      'address': ['', [Validators.required, Validators.pattern(getRegex('address', 'testnet4'))]],
+      'satoshis': [0, [Validators.required, Validators.min(0), Validators.max(0)]]
+    });
+  }
+
+  ngOnDestroy() {
+    if (this.faucetStatusSubscription) {
+      this.faucetStatusSubscription.unsubscribe();
```
mononaut, line 51: there's no need to unsubscribe from HttpClient observables (it happens automatically)

frontend/src/app/components/faucet/faucet.component.ts
```diff
@@ -78,59 +101,37 @@ export class FaucetComponent implements OnInit, OnDestroy {
...
   }
 
-  ngOnDestroy(): void {
-    this.stateService.markBlock$.next({});
-    this.websocketService.stopTrackingTransaction();
-    if (this.mempoolPositionSubscription) {
-      this.mempoolPositionSubscription.unsubscribe();
-    }
-    if (this.confirmationSubscription) {
-      this.confirmationSubscription.unsubscribe();
```
mononaut, line 124: ...but we do still need to clean up all of these things

mononaut (commented):
UI changes look great! and thanks for fixing the error alerts - I forgot we had a component for that already :)

wiz:
looks great but I need to remind you that the FOSS build should not show a faucet at all because it won't work without the services backend

mononaut (approved):
Tested ACK @ [f58c2e6fe5a3b0f1d888ae0b441098e99b8b0f80]

there's a minor issue with the form validation not triggering when the withdrawal amount is set using the quick buttons, but otherwise LGTM

PR #5081, opened 2024-05-17
Update lightning-api.service.ts - fix typo

hans-crypto:
You have this line twice in the file:

```
LINE 20: this.apiBaseUrl = ''; // use relative URL by default

LINE 24: this.apiBaseUrl = ''; // use relative URL by default
```

line 24 should be removed

### DIFF
--- a/contributors/hans-crypto.txt
+++ b/contributors/hans-crypto.txt
@@ -1,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of May 17, 2025.
+
+Signed: hans-crypto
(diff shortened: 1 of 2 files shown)


mononaut:
those two lines are different - one assigns to `this.apiBaseUrl` and the other to `this.apiBasePath`.
https://github.com/mempool/mempool/blob/db34ca6a5f37f8c420ce961f0ccd96bba9205dfd/frontend/src/app/lightning/lightning-api.service.ts#L16-L28

technically the second assignment is superfluous since the field is already initialized to that value at declaration, but imo it improves clarity.

hans-crypto:
Sorry, `apiBaseUrl` and `apiBasePath` looked too similar. My fault! 😃 👍

PR #5084, opened 2024-05-18
[tests] testnet -> testnet4

nymkappa:
_I have no idea what I'm doing but I'll try anyway because it's Saturday_

Tests pass again.


softsimon (approved):
All tests for the new testnet4 pass.

PR #5085, opened 2024-05-18
Refactor block transactions

softsimon:
What this PR / Refactor does:

1. The Transactions list on the Block page is now it's own separate component reducing the complexity of Block component slightly.
2. The new Angular defer view feature is utilized. So only when the Transactions list on the block page is visible in the viewport, it will render and load the transactions saving rendering time and server traffic, that commonly isn't required as mobile phones need to scroll down to view transactions. As soon as the transactions list is in view, the old behaviour of loading the transactions and lazy-loading the previous block kicks in, so people with long screens shouldn't notice any difference.

Opening a new Block page (on a default screen where the transactions list isn't visible), 6 API requests are now being triggered, instead of 10 currently!

### DIFF
--- a/frontend/src/app/components/block/block-transactions.component.ts
+++ b/frontend/src/app/components/block/block-transactions.component.ts
@@ -19,6 +19,7 @@ export class BlockTransactionsComponent implements OnInit {
   @Input() previousBlockHash: string;
   @Input() block$: Observable<any>;
   @Input() paginationMaxSize: number;
+  @Output() blockReward = new EventEmitter<number>();
 
   itemsPerPage = this.stateService.env.ITEMS_PER_PAGE;
   page = 1;
@@ -50,6 +51,13 @@ export class BlockTransactionsComponent implements OnInit {
             return of([]);
         }))
       ),
+      tap((transactions: Transaction[]) => {
+        // The block API doesn't contain the block rewards on Liquid
+        if (this.stateService.isLiquid() && transactions && transactions[0] && transactions[0].vin[0].is_coinbase) {
+          const blockReward = transactions[0].vout.reduce((acc: number, curr: Vout) => acc + curr.value, 0) / 100000000;
+          this.blockReward.emit(blockReward);
+        }
+      })
     );
 
     this.txsLoadingStatus$ = this.route.paramMap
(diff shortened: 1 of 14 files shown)


softsimon:
Ok, of course tests fail now when them expect the block transactions list to be loaded even if it's outside of the cypress test view.

frontend/src/app/components/block/block-transactions.component.ts
```diff
@@ -0,0 +1,90 @@
...
+          this.blockReward.emit(blockReward);
+        }
+        this.unsubscribeNextBlockSubscriptions();
+        if (this.previousBlockHash) {
+          setTimeout(() => {
+            this.nextBlockTxListSubscription = this.electrsApiService.getBlockTransactions$(this.previousBlockHash).subscribe();
+          }, 100);
+        }
+      })
+    );
```
mononaut, line 67: this whole section feels a little out-of-place in the new component.

I wonder if we could:
 - just grab the block reward directly from the `BlockExtended` object in the main block component, if we need it there?
 - move the next block transactions preloading stuff either back to the main block component or offload it to a service or something?

frontend/src/app/components/block/block-transactions.component.ts
```diff
@@ -0,0 +1,90 @@
...
+  ) { }
+
+  ngOnInit(): void {
+    this.transactions$ = combineLatest([this.block$, this.route.queryParams]).pipe(
+      tap(([_, queryParams]) => {
+        this.page = +queryParams['page'] || 1;
+      }),
+      switchMap(([block, _]) => this.electrsApiService.getBlockTransactions$(block.id, (this.page - 1) * this.itemsPerPage)
+        .pipe(
+          startWith(null),
```
mononaut, line 47: this `startWith(null)` causes the `tap` block below to trigger twice on load (first with null, then again with the actual transactions when the API call returns), which means we end up making duplicate requests to `getBlockTransactions$(...)`.

mononaut (changes requested):
Concept ACK: the deferred load is a good idea, and this implementation works well.

I just have a few minor comments below to resolve.

mononaut (approved):
Tested ACK @ [ea0a4d1d67a41f31eb93e284c0a1fe860a5dcd75]

PR #5086, opened 2024-05-19
 Update global-footer.component.html - fix broken link

hans-crypto:
the link to 3rdpartylicenses.txt is not part of the angular routing


softsimon:
Thanks, fixed in a separate commit. https://github.com/mempool/mempool/commit/8bac2db12c3e838bddd50781d661c942a88a3aeb

hans-crypto:
What was the reason that my PR could not be used directly? Was it because of the failed confirmation from this cla-bot?

PR #5091, opened 2024-05-21
Requested changes to testnet4 faucet UI

mononaut:
This PR makes some minor changes to the faucet UI:
 - s/"to limit abuse"/"to use the faucet"/.
 - only disable the submit button instead of graying out the whole form.
 - set dummy min/max values until/unless the user's limits are available.
 - minor refactor of the form initialization (just update the validators, instead of the whole formGroup, when possible).
 - refresh the faucet status after a successful payout, to keep the user's limits up to date.
 - support displaying both the *last* success banner and a *current* error (e.g. after the user's last permitted withdrawal).
 - properly validate amounts set via the quick buttons.
 

https://github.com/mempool/mempool/assets/83316221/a696cb09-a2a7-4a44-9da3-2e004e111863

Screenshot 2024-05-21 at 12 37 15 AM

https://github.com/mempool/mempool/assets/83316221/714b32a1-b2e0-4801-997b-91bc2ea351b0

Screenshot 2024-05-21 at 12 54 54 AM

### DIFF
--- a/frontend/src/app/components/faucet/faucet.component.ts
+++ b/frontend/src/app/components/faucet/faucet.component.ts
@@ -156,6 +156,7 @@ export class FaucetComponent implements OnInit, OnDestroy {
     } else {
       this.faucetForm.get('address').setValidators([Validators.required, Validators.pattern(getRegex('address', 'testnet4')), this.getNotFaucetAddressValidator(faucetAddress)]);
       this.faucetForm.get('satoshis').setValidators([Validators.required, Validators.min(min), Validators.max(max)]);
+      this.faucetForm.get('satoshis').setValue(Math.max(min, this.faucetForm.get('satoshis').value));
       this.faucetForm.get('satoshis').updateValueAndValidity();
       this.faucetForm.get('satoshis').markAsDirty();
     }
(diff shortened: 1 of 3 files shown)


frontend/src/app/components/faucet/faucet.component.ts
```diff
@@ -135,9 +129,40 @@ export class FaucetComponent implements OnInit, OnDestroy {
...
+    this.faucetForm = this.formBuilder.group({
+      'address': ['', [Validators.required, Validators.pattern(getRegex('address', 'testnet4')), this.getNotFaucetAddressValidator(faucetAddress)]],
+      'satoshis': [min, [Validators.required, Validators.min(min), Validators.max(max)]]
+    });
+
+    this.loading = false;
+    this.cd.markForCheck();
+  }
+
+  updateForm(min, max, faucetAddress: string): void {
```
nymkappa, line 153: Also set the default values here. EG: We set min to 1000 sats by default, but if status send a min of 5000 sats, you can still attempt a request and it ends up with a form validation error in the backend

```
May 21 10:10:44 [25666] DEBUG: <services> Validation error: The string input 1000 failed the validation function
May 21 10:10:44 [25666] DEBUG: <services> Invalid format for sats
```

nymkappa (changes requested):

frontend/src/app/components/faucet/faucet.component.ts
mononaut: fixed

nymkappa (approved):
Tested ACK @ 6bb2f06118ebc9647070b38a334c51094b7645c7

PR #5094, opened 2024-05-21
 Remove reference to bisq in unfurler 

hans-crypto:
Since bisq was removed, this code is also not needed anymore.


softsimon (approved):
utACK

PR #5098, opened 2024-05-23
Add check for pool addresses and regexes

### LINKED ISSUE #5079: Pools.json import errors
```
May 17 19:10:29 [26486] DEBUG: <testnet> database query "
        INSERT INTO pools
        SET name = ?, link = ?, addresses = ?, regexes = ?, slug = ?, unique_id = ?" failed!
May 17 19:10:29 [26486] ERR: <testnet> Cannot insert new mining pool into db. Reason: Column 'addresses' cannot be null
```

I suppose it should be okay to have no addresses, but the frontend probably need a special handling case.

natsoni:
Fixes #5079 

This PR adds a check before adding mining pools to the database: 

- if both fields `addresses` and `tags` are not defined, log an error and skip
- if both fields `addresses` and `tags` are an empty array, log an error and skip
- if one of the fields is an empty array, log a warn

### DIFF
--- a/backend/src/api/pools-parser.ts
+++ b/backend/src/api/pools-parser.ts
@@ -52,6 +52,28 @@ class PoolsParser {
         continue;
       }
 
+      // One of the two fields 'addresses' or 'regexes' must be a non-empty array
+      if (!pool.addresses && !pool.regexes) {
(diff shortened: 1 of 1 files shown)


nymkappa (approved):
utACK

softsimon (approved):
tested ACK @ [35215c774]

PR #5100, opened 2024-05-25
Add BitBanana to community integrations

michaelWuensch:
This PR adds BitBanana to community integrations.

BitBanana allows to use mempool
- as block explorer
- as source for fee estimation
- as source for exchange rates

It supports both, self-hosted or official mempool.space as source.


softsimon:
I'm very sorry but although many Bitcoin apps and services integrate Mempool, we will ourself pick which projects to be featured on the about page.

PR #5110, opened 2024-05-28
Slightly lift up blockchain toggle button

natsoni:
This PR slightly lifts up the blockchain upper toggle button so that it is not cropped on screenshots: 

Screenshot 2024-05-28 at 16 17 53


softsimon (approved):
tested ACK @ [dcd94d86]

PR #5111, opened 2024-05-28
Add explanatory tooltips to incoming transactions widget

natsoni:
This PR adds explanatory tooltips in the incoming transactions widget: 

Screenshot 2024-05-27 at 15 51 03
<br>
Screenshot 2024-05-27 at 15 55 56


softsimon:
The first tooltip is fine.

But not sure about the second one claiming it may exceed the default limit, because it's something that only happens in Prod when using Dual node setup. Perhaps we should just remove the `/ 300MB` thing in prod instead since we have no limit, and the tooltip will only explain that this is the memory usage of the mempool of Bitcoin Core.

PR #5112, opened 2024-05-28
Polish acceleration pie chart section

mononaut:
- Fixes mobile layout
- Better purple color contrast
- Fixes issue with data not updating after an acceleration request is confirmed.
Screenshot 2024-05-28 at 9 36 23 PM


wiz (approved):
utACK

Issue #4993, opened 2024-04-19
Lightning Node search results should be on top


softsimon:
I'm not sure if it makes sense or work in practice.. But when searching for something matching a Lightning Node, it should default the Lightning result to the top, as it will likely not be an address.

Obviously if people search for "10" it should suggest block height 10 .. and not a node ID or channel id first ....

<img width="499" alt="Screenshot 2024-04-19 at 17 53 55" src="https://github.com/mempool/mempool/assets/8561090/baf3b19b-e6df-4f92-9f5e-75b41b0be80f">

PR #5117, opened 2024-05-30
Add mining pools to search results

natsoni:
This PR adds mining pools support to the search bar. Active pools (which mined a block in the last year) are shown first. 

Screenshot 2024-05-30 at 15 07 15

******
~~I also updated the order of the results displayed to fix https://github.com/mempool/mempool/issues/4993. Now, if the search is exact match with a public key, the address result is displayed at the end, and the other networks results are hidden~~

| Before  | After |
| Screenshot 2024-05-30 at 15 12 52  | Screenshot 2024-05-30 at 15 11 59 |

*******
There are other issues I will fix in separate PRs to simplify the review for this one.

- The pool page does not update properly when switching pool: 

https://github.com/mempool/mempool/assets/46578910/67eb8144-de6e-4d3d-b197-5664e8293bc2

- The lightning search API frequently times out when the search text is only one or two characters long. This causes all results to take several seconds to appear even if the address prefix and the mining pools queries returned almost instantly. A simple solution would be to set a minimum number of characters required before initiating a lightning search. Or we could separate all queries on the frontend (address prefix, lightning search, and now mining pools).

https://github.com/mempool/mempool/assets/46578910/01e60926-8231-443e-ad7a-b136f9564d6c

### DIFF
--- a/frontend/src/app/components/search-form/search-form.component.ts
+++ b/frontend/src/app/components/search-form/search-form.component.ts
@@ -195,12 +195,17 @@ export class SearchFormComponent implements OnInit {
           const matchesTxId = this.regexTransaction.test(searchText) && !this.regexBlockhash.test(searchText);
           const matchesBlockHash = this.regexBlockhash.test(searchText);
           const matchesAddress = !matchesTxId && this.regexAddress.test(searchText);
+          const publicKey = matchesAddress && searchText.startsWith('0');
           const otherNetworks = findOtherNetworks(searchText, this.network as any || 'mainnet', this.env);
           const liquidAsset = this.assets ? (this.assets[searchText] || []) : [];
           const pools = this.pools.filter(pool => pool["name"].toLowerCase().startsWith(searchText.toLowerCase())).slice(0, 10);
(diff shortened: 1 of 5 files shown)


softsimon (changes requested):
The pool search works great.

But the change of the order should probably have been in a separate PR, because there is one issue with it. The search result isn't selecting the top result, so pasting a Lightning nod pubkey and press enter still takes you to an address page. Is it possible to move the search highlight index to the top without any issues in this case? Either that or move it out from this PR for now.

Screenshot 2024-05-31 at 12 54 55

natsoni:
> But the change of the order should probably have been in a separate PR, because there is one issue with it. The search result isn't selecting the top result, so pasting a Lightning nod pubkey and press enter still takes you to an address page. Is it possible to move the search highlight index to the top without any issues in this case? Either that or move it out from this PR for now.

Reverted the changes made to the results order, will work on that in another PR.

softsimon (approved):
tested ACK @ [f3bd50d4a]

PR #5118, opened 2024-05-30
Fix pool page update

natsoni:
Builds on #5117

The pool page now updates properly when switching pool using the search bar:

https://github.com/mempool/mempool/assets/46578910/c3ae2b84-9324-4c2b-8f0c-bed602898fac


softsimon (approved):
tested ACK @ [692edea1c]

PR #5122, opened 2024-05-31
Fix pool ranking table

natsoni:
This PR fixes an issue where the rank table would not appear in the pool ranking component when data is fetched from cache:

https://github.com/mempool/mempool/assets/46578910/a938aef8-b5ec-4157-a1bf-5c79df58d404


softsimon (approved):
tested ACK

PR #5123, opened 2024-05-31
Improve search results ordering

natsoni:
Builds on #5117

Fixes https://github.com/mempool/mempool/issues/4993

This PR updates the order of the results displayed. If the search is exact match with a public key: 
- the address result is displayed at the end
- the other networks results are hidden
- pressing enter will direct to the node page

| Before  | After |
| Screenshot 2024-05-30 at 15 12 52  | Screenshot 2024-05-31 at 15 15 03 |


softsimon (approved):
tested ACK @ [1c73dc59f9]

PR #5124, opened 2024-05-31
Searchbar: wait for 3 characters before requesting lightning data

natsoni:
Builds on #5117

The lightning search API frequently times out when the search text is only one or two characters long. This causes all results to take several seconds to appear even if the address prefix and the mining pools queries returned almost instantly. 
To mitigate this issue, this PR makes the frontend only query lightning nodes and channels if the search text is 3 or more characters.


softsimon (approved):
tested ACK @ [67afda7dcf]

PR #5131, opened 2024-06-03
Correctly expand environment variables in the config files in the docker images.

danrmiller:
The docker images that get uploaded to dockerhub under mempool/frontend don't accept environment variables for FRONTEND_HTTP_PORT, BACKEND_MAINNET_HTTP_HOST, and BACKEND_MAINNET_HTTP_PORT and instead just contain the placeholder string.


danrmiller:
I'm closing this and opening an issue instead https://github.com/mempool/mempool/issues/5133 as I am not sure I understand the docker build workflow and variable mapping.

PR #5137, opened 2024-06-04
Pizza tracker RBF support

mononaut:
Adds UX support for replaced transactions to the pizza tracker:

| localhost_4200_tracker_86e408c4bb441e4a1d827b6a1b8cd943a387c78660a2c8fb5631005f2db9d1fc(iPhone SE) | unlike the main transaction page, the "replaced by ..." alert here links directly to the most relevant transaction (either the latest pending replacement, or whichever version actually got confirmed), rather than the "next" transaction in the sequence of replacements.|
|-|-|


wiz (approved):
utACK

PR #5120, opened 2024-05-30
Multi-pool ETA

mononaut:
This PR improves the ETA calculation for transactions being accelerated by multiple different pools.

Screenshot 2024-05-30 at 9 28 07 PM

It leverages the "approximate accelerated position" calculations from closed PR #4642 to figure out which projected mempool block the transaction lies in for each subset of global hashrate, and then calculates the expected time until confirmation as follows:

#### ETA Calculation
let
  - $H(n)$ be the proportion of hashrate for which the transaction is in mempool block ≤ $n$
  - $S(n)$ be the probability of the transaction being mined in block $n$
    - by definition, $S(max) = 1$ , where $max$ is the maximum depth of the transaction in any mempool, and therefore $S(n>max) = 0$, so we only need to calculate $max$ different values.
  - $Q$ be the expected number of blocks until the transaction is confirmed.
  - $E$ be the expected time until the transaction is confirmed.

then
  - $S(0) = H(0)$
  - $S(i) = H(i) \times (1 - \sum\limits_{j=0}^{i-1} S(j))$
    - the probability of mining a block including the transaction at this depth, multiplied by the probability that it hasn't already been mined at an earlier depth.
  - $Q = \sum\limits_{i=0}^{max} S(i) \times (i+1)$
    - number of blocks, weighted by the probability that the block includes the transaction
  - $E = Q \times T$
    - expected number of blocks, multiplied by the avg time per block
    
    
As with our existing simple unaccelerated ETAs, this does not attempt to account for the rate of new transactions entering the mempool(s).

PR #5141, opened 2024-06-05
[liquid] don't fetch pools

nymkappa:
Fix liquid bug introduced in https://github.com/mempool/mempool/pull/5120 where it's trying to fetch pools on liquid


wiz (approved):
utACK

PR #5143, opened 2024-06-05
Fix tx acceleration labels

mononaut:
hotfix for missing acceleration labels on the transaction page


softsimon:
Fixed in https://github.com/mempool/mempool.space/pull/1018

PR #5144, opened 2024-06-06
fix: set http protocol

CharlieC3:
Sets the HTTP protocol in Nginx for /api paths. Without this, the very old HTTP 1.0 protocol is used and may cause HTTP 426 errors in service meshes.
Tested and verified working locally.

### DIFF
--- a/nginx-mempool.conf
+++ b/nginx-mempool.conf
@@ -60,9 +60,11 @@
 	}
 	location /api/v1 {
 		proxy_pass http://127.0.0.1:8999/api/v1;
+		proxy_http_version 1.1;
 	}
 	location /api/ {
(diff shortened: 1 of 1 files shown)


CharlieC3:
I submitted the contributor txt file to this PR, but the bot doesn't seem to like it for some reason 🤷

softsimon:
> I submitted the contributor txt file to this PR, but the bot doesn't seem to like it for some reason 🤷

The bot is wrong. Just ignore :)

wiz:
Thanks for the suggestion. Ideally we should migrate to unix sockets on localhost if possible, but if a network socket is needed than http 1.1 is probably the way to go - just need to double check it doesn't cause any issues if the connection is kept alive

PR #5147, opened 2024-06-06
Acceleration preview hashrate pie chart

mononaut:
_(builds on #5146)_
_(requires https://github.com/mempool/mempool.space/pull/1019)_

This PR adds the hashrate % pie chart and a forecast of estimated acceleration confirmation times to the acceleration preview component:

Screenshot 2024-06-06 at 7 27 50 PM

### DIFF
--- a/frontend/src/app/components/accelerate-preview/accelerate-preview.component.html
+++ b/frontend/src/app/components/accelerate-preview/accelerate-preview.component.html
@@ -65,10 +65,21 @@
         </div>
       </div>
       <br>
+      <h5 *ngIf="estimate?.pools?.length" i18n="accelerator.how-much-faster">How much faster?</h5>
+      <div class="row">
+        <div class="col">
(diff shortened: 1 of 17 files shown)


wiz:
does the pie chart only appear if the user is logged in?

mononaut:
the whole "how much faster" section appears if and only if the list of pools included in the estimate returned by the services backend is not empty.

but maybe it would be better to strictly limit it to only logged in users instead.

wiz:
well, you didn't answer the question tho - what does a non-logged-in user see when they press the button running this PR?

mononaut:
> the whole "how much faster" section appears _**if and only if**_ the list of pools included in the estimate returned by the services backend is not empty.

they either see a pie chart of the total hashrate corresponding to the sum of all the pools returned by the services backend for that acceleration estimate, or if that list is empty, the whole section is hidden.

so it depends how the pools are currently configured in the services backend.

right now, I expect they would see a pie chart with a sliver of purple representing a few % of total hashrate.

wiz:
ok thanks, let's think about this a bit before merging

softsimon (approved):
utACK

wiz (approved):
Concept ACK

PR #5149, opened 2024-06-08
Accurate timestamps on hover

mononaut:
We use the `app-time` component in lots of places to render friendlier, human-formatted relative time strings.

However, it would often be useful to also see the actual timestamp associated with that string.

This PR adds a `showTooltip` property to the component which makes it render the actual timestamp in a tooltip on hover. This remains off by default, but I've enabled it everywhere I thought it made sense.

e.g.

| Screenshot 2024-06-08 at 11 27 26 PM | Screenshot 2024-06-08 at 11 27 55 PM|
|-|-|


Zoiediez (commented):
How long does it take for assets to transfer to another account thanks ?

softsimon (approved):
tested ACK @ [3149199c]

PR #5151, opened 2024-06-10
Address page redesign phase 1

mononaut:
_**(STATUS: ready for review)**_

This PR is the first in what I hope will be a series of changes and improvements to make the address page more useful.

This initial stage focuses on refactoring some of the underlying logic of the address page component, and redesigning the top details section to focus on more immediately relevant data, by:

 - making "Balance" the headline item
 - hiding the QR code behind a mouse-over icon/button
 - removing the "total received" and "total sent" items
 - adding items for the number of UTXOs and the address type (including special address labels where relevant).
 - splitting balance and utxo data into separate confirmed & pending fields, to emphasize that unconfirmed funds/utxos are not safe or guaranteed.
 
This refactor/redesign hasn't yet been polished to fully support Liquid, P2PK addresses, or non-esplora backends, so I'm leaving it in draft for now.

If the general idea gets a concept ACK, I'll finish up those parts and take the PR out of draft.

Screenshot 2024-06-10 at 11 14 46 PM

| Before | After |
|-|-|
| localhost_4200_address_bc1q5x9p784q6jkj4w3cw56h8vejz326yrf6u4tuksgg8kjnfkpg7l5qe8kpz6(iPhone SE) (1) | localhost_4200_address_bc1q5x9p784q6jkj4w3cw56h8vejz326yrf6u4tuksgg8kjnfkpg7l5qe8kpz6(iPhone SE) |

https://github.com/mempool/mempool/assets/83316221/fcfb132c-9ab7-4c54-8dfa-06e56d8133dc

### DIFF
--- a/frontend/cypress/e2e/liquid/liquid.spec.ts
+++ b/frontend/cypress/e2e/liquid/liquid.spec.ts
@@ -72,20 +72,6 @@ describe('Liquid', () => {
       });
     });
 
-    it('renders unconfidential addresses correctly on mobile', () => {
-      cy.viewport('iphone-6');
-      cy.visit(`${basePath}/address/ex1qqmmjdwrlg59c8q4l75sj6wedjx57tj5grt8pat`);
-      cy.waitForSkeletonGone();
-      //TODO: Add proper IDs for these selectors
-      const firstRowSelector = '.container-xl > :nth-child(3) > div > :nth-child(1) > .table > tbody';
-      const thirdRowSelector = '.container-xl > :nth-child(3) > div > :nth-child(3)';
-      cy.get(firstRowSelector).invoke('css', 'width').then(firstRowWidth => {
-        cy.get(thirdRowSelector).invoke('css', 'width').then(thirdRowWidth => {
-          expect(parseInt(firstRowWidth)).to.be.lessThan(parseInt(thirdRowWidth));
-        });
-      });
-    });
-
     describe('peg in/peg out', () => {
       it('loads peg in addresses', () => {
         cy.visit(`${basePath}/tx/fe764f7bedfc2a37b29d9c8aef67d64a57d253a6b11c5a55555cfd5826483a58`);
--- a/frontend/src/app/components/address-labels/address-labels.component.html
+++ b/frontend/src/app/components/address-labels/address-labels.component.html
@@ -4,7 +4,7 @@
       <a [routerLink]="['/lightning/channel' | relativeUrl, channel.id]">
         <span 
           *ngIf="label"
-          class="badge badge-pill badge-warning"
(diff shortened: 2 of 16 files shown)


wiz:
Concept ACK

Maybe change pending to unconfirmed or even "pending unconfirmed" to make it more clear what that means?

softsimon:
Concept ACK

If we're going to use _pending_ we should define when and where. Unconfirmed is what we use for transactions in general so it should be Unconfirmed balance here as well.

Perhaps Total Received is more easy to understand than Volume, other explorers use that too.

mononaut:
Updated to switch "pending" to "unconfirmed balance/utxos".

I'm not 100% sure about this yet, because "unconfirmed balance" could be interpreted as "this is what the balance will be after all of the unconfirmed transactions settle", while "pending" slightly less ambiguously suggests "this is the unresolved *change* in balance".

but we can paint that bikeshed later.

Screenshot 2024-06-11 at 8 50 54 PM

I'm also leaving the "Volume" field as-is for the moment, because a) I think it more clearly captures what people want to know, which is "how much activity has there been on this address" (including self-sends etc), b) "total received" can be misinterpreted as a measure of net income or something like that, and c) I only have space free for one field atm, and "total received" works better as a pair together with "total sent".

I'm planning to add an extra details toggle in a later PR for some of the less important metrics, so the separate "total received/sent" fields might find a home there instead.

mononaut:
Added some abstractions for dealing with addresses/scripts, so that we can parse out basic address types immediately without relying on transaction data, and then update with more precise labels as that data comes in.

Screenshot 2024-06-12 at 6 16 02 PM
Screenshot 2024-06-12 at 6 16 21 PM
Screenshot 2024-06-12 at 6 17 43 PM

mononaut:
Taking this out of draft, since it seems to work ok with Liquid, P2PK addresses, and non-esplora backends now.

One of the Liquid e2e tests is failing because of the layout change, but I'm not 100% sure what the intent of the test was, so I don't want to delete it myself.

nymkappa:
Might be hard on mobile to hit the qr code button

mononaut:
good point. I've just pushed a change to make both icons slightly larger on mobile, and the QR button easier to tap.

| Before | After |
|-|-|
| localhost_4200_address_bc1qwvlnk4hwyn9usgfz2ppsl25hupnsxawtpwn8j4juce6a8c502w2qn39sj7(iPhone SE) | localhost_4200_address_bc1qwvlnk4hwyn9usgfz2ppsl25hupnsxawtpwn8j4juce6a8c502w2qn39sj7(iPhone SE) (2) |

softsimon (approved):
tested ACK @ [1219526e]

removed the Liquid test as it measures the width of the QR code block for some reason which is not there anymore

PR #5159, opened 2024-06-16
Handle services backend failures in block component

mononaut:
In order to properly account for & display accelerations in the block audits, the block component requests acceleration history from the services API.

If the services backend is unreachable for some reason, this request stalls the whole block audit observable pipeline until it eventually resolves or times out:

Screenshot 2024-06-16 at 2 38 32 PM

Since this is "bonus data" rather than anything essential for the block audits, this PR splits the services API acceleration history request into its own separate observable/subscription, so that we can render the block audits ASAP independently of services availability.

The tradeoff is that now the frontend will run the audit rendering calculations and block visualization setup twice for blocks containing accelerations - once when the main audit requests resolve, and again when the acceleration history arrives - but this shouldn't be noticeable.


softsimon (approved):
tested ACK @ [c5b12e3]

PR #5160, opened 2024-06-16
Fix incorrect non-standard label on reserved segwit output types

mononaut:
Fixes a minor bug in the standardness detection logic which incorrectly classifies reserved segwit output types as non-standard, e.g. https://mempool.space/tx/20b984492b5264162a4c92c9a34bc7fa08b67d669de7b4c5982ad3cb28aaecf6

This PR immediately fixes client-side classification (e.g. for the badges on the transaction page), but past block template and audit data will not be updated until the next time it is manually reindexed (not worth doing just for this).


softsimon (approved):
utACK

PR #5162, opened 2024-06-18
fix hex2ascii.pipe

hans-crypto:
Early exit if `hex` is undefined. 
In rare cases `undefined` gets into the pipe, and unessary errors appear in the dev tools.

Screenshot 2024-06-18 at 23 12 41


softsimon:
Do you have a transaction or similar to reproduce the error in order to test the PR?

hans-crypto:
Of course, I should have made better notes. Error happened on the transaction details page. It was during local development, after the backend has started but the not fully synced. I placed a breakpoint at that position and will report back, when I can reproduce it myself.

softsimon:
With test case. I assume this bug to be fixed.

hans-crypto:
Breakpoint still set. I will report again when I see this constellation again. 🫡

PR #5169, opened 2024-06-19
Implement $getTxsForBlock for Core backends

### LINKED ISSUE #5166: $indexPastAccelerations() fails on non-esplora backends
$indexPastAccelerations() fails due to $getTxsForBlock not being implemented in the bitcoinApi on non-esplora backends.

`Jun 18 15:58:14 [658189] ERR: Failed to process accelerations for block 790245. Reason: Method getTxsForBlock not supported by the Bitcoin RPC API.
`

mononaut:
Resolves #5166

This PR implements $getTxsForBlock for non-esplora backends, using the verbose block RPC and existing transaction conversion utility functions.

This is _**slow**_, and puts a bit of strain on the local core node, so we should avoid relying on it whenever possible.


softsimon (approved):
tested ACK @ [e95e64a4]

PR #5177, opened 2024-06-20
Deprecate pool_unique_id, fixing accelerations sync

softsimon:
The acceleration history sync was broken, as the pool_unique_id had been migrated to pool number array.

### DIFF
--- a/backend/src/repositories/AccelerationRepository.ts
+++ b/backend/src/repositories/AccelerationRepository.ts
@@ -308,10 +308,10 @@ class AccelerationRepository {
         }
         const accelerationSummaries = accelerations.map(acc => ({
           ...acc,
-          pools: acc.pools.map(pool => pool.pool_unique_id),
+          pools: acc.pools,
         }))
         for (const acc of accelerations) {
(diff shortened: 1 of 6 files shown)


backend/src/repositories/AccelerationRepository.ts
```diff
@@ -308,10 +308,10 @@ class AccelerationRepository {
         }
         const accelerationSummaries = accelerations.map(acc => ({
           ...acc,
-          pools: acc.pools.map(pool => pool.pool_unique_id),
+          pools: acc.pools,
         }))
         for (const acc of accelerations) {
-          if (blockTxs[acc.txid] && acc.pools.some(pool => pool.pool_unique_id === block.extras.pool.id)) {
+          if (blockTxs[acc.txid] && acc.pools.some(pool => pool === block.extras.pool.id)) {
```
nymkappa, line 314: I suppose you could change that to an `.includes` instead since it's a simple array of number now.

mononaut (approved):
Tested ACK @ [85006a5b]

PR #5178, opened 2024-06-20
Recover from esplora failover after a reorg to lower height

mononaut:
Hopefully fixes an issue where esplora failover hosts could get stuck in an "outOfSync" state after a reorg to a lower height.


wiz (approved):
utACK

PR #5183, opened 2024-06-22
Missing difficulty adjustment causes invalid json response

softsimon:
When a node is still syncing, and we don't have the Difficulty Adjustment data yet.

The websocket is returning invalid data _undefined_.

Returning an empty string instead that should be falsy for the frontend.

### DIFF
--- a/backend/src/api/websocket-handler.ts
+++ b/backend/src/api/websocket-handler.ts
@@ -1300,7 +1300,7 @@ class WebsocketHandler {
   // and zips it together into a valid JSON object
   private serializeResponse(response): string {
     return '{'
-        + Object.keys(response).map(key => `"${key}": ${response[key]}`).join(', ')
+        + Object.keys(response).filter(key => response[key] != null).map(key => `"${key}": ${response[key]}`).join(', ')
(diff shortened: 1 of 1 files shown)


mononaut:
`undefined` in this context is supposed to mean "don't include this key in the websocket response at all.

so I think a cleaner solution would be to explicitly filter out nullish values from the response during serialization like:
```typescript
// takes a dictionary of JSON serialized values
// and zips it together into a valid JSON object
private serializeResponse(response): string {
  return '{'
      + Object.keys(response).filter(key => response[key] != null).map(key => `"${key}": ${response[key]}`).join(', ')
      + '}';
}
```

softsimon:
> `undefined` in this context is supposed to mean "don't include this key in the websocket response at all".
> 
> so I think a cleaner solution would be to explicitly filter out nullish values from the response during serialization like:

Fixed and confirmed it works.

mononaut (approved):
Tested ACK @ [c581be0e]

I couldn't reproduce the missing difficulty adjustment, but this fix doesn't seem to break anything else.

PR #5184, opened 2024-06-22
Always show clearing rate line on incoming tx chart

### LINKED ISSUE #5148: Set 1.1mb bandwidth equivalent line as max resize level (it can zoom in too much.
image

When I look at that graph it looks like there's big spikes but in reality it's really down below the baseline of new blocks bandwidth.

It it would not zoom in more than 1mb baseline it would be obvious that it's very low. 

This is also a symptom of only looking at last 2 hours... Hope we get to be able to zoom out to a 12-24h

mononaut:
Fixes #5148


softsimon (approved):
tested ACK @ [7c806b4b2]

PR #5185, opened 2024-06-22
configurable block size & count in rust gbt

mononaut:
use `config.MEMPOOL.BLOCK_WEIGHT_UNITS` and `config.MEMPOOL.MEMPOOL_BLOCKS_AMOUNT` config settings in rust gbt instead of hardcoded constants.


softsimon (approved):
tested ACK @ [fa9a8bdba]

PR #5186, opened 2024-06-22
Fix hardcoded median weight units in calcEffectiveFeeStatistics

mononaut:


softsimon (approved):
tested ACK @ [6b84dc2b]

PR #5187, opened 2024-06-22
Prevent websocket reconnect on custom root instances

softsimon:


softsimon:
tested ACK @ [1c1c93abfc]

PR #5188, opened 2024-06-22
Adapt address page skeleton

natsoni:
This PR adds some skeleton loader to the address page table to match its new formatting: 
Screenshot 2024-06-22 at 15 33 19


softsimon (approved):
tested ACK @ [7f2987f2]

PR #5190, opened 2024-06-22
Address page ux updates

softsimon:
* Volume is now Total received (total confirmed balance)
* Capitalised unconfirmed labels
* Address type is now a label box
* Copy to clipboard is now aligned right of the QR code button

Before
Screenshot 2024-06-22 at 17 30 23

After
Screenshot 2024-06-22 at 17 30 27


wiz (approved):
utACK

PR #5193, opened 2024-06-23
Refactor address table to improve cell wrapping

mononaut:
Fixes a text wrapping issue in the new address page for addresses with very large balances/volumes, and refactors the address table itself to avoid inconsistent row heights.

Before:
Screenshot 2024-06-23 at 3 18 19 AM

After:
Screenshot 2024-06-23 at 3 18 27 AM


softsimon (approved):
utACK

PR #5194, opened 2024-06-23
Fix mempool transactions being stripped

### LINKED ISSUE #5150: vin.scriptsig_asm might be undefined with Core only
We got the report that the backend threw 500  undefined (reading 'split') ..  I was not able to reproduce that, but the frontend was. So it should be related. Basically **scriptsig_asm** can be undefined.

```
TypeError: Cannot read properties of undefined (reading 'split')
    at calcSegwitFeeGains (bitcoin.utils.ts:26:50)
    at TxFeaturesComponent.ngOnChanges (tx-features.component.ts:40:42)

```

softsimon:
fixes #5150

This happens when:

* Using Bitcoin Core
* Redis is enabled

### DIFF
--- a/backend/src/api/redis-cache.ts
+++ b/backend/src/api/redis-cache.ts
@@ -155,7 +155,7 @@ class RedisCache {
     const toAdd = this.cacheQueue.slice(0, this.txFlushLimit);
     try {
       const msetData = toAdd.map(tx => {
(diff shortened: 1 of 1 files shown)


mononaut (approved):
Tested ACK @ [3a7d3915]

PR #5195, opened 2024-06-23
Address balance graph: hide usd on non-mainnet networks

natsoni:
Screenshot 2024-06-23 at 19 00 18


softsimon (changes requested):
Use `stateService.isAnyTestnet()` instead and make it work with the new rootNetwork thing

natsoni:
> Use `stateService.isAnyTestnet()` instead and make it work with the new rootNetwork thing

Done using only `stateService.isAnyTestnet()`. It works without additional logic (tested with `ROOT_NETWORK="signet"`)

softsimon (approved):
tested ACK @ [8a43ed1a]

PR #5197, opened 2024-06-24
Sha256 P2PK secure context workaround

softsimon:
Because the native library `crypto.subtle` doesn't work over Non-SSL connections, affecting local self hosted instances, instead we use a local sha256 function.

Causing P2PK address lookups to fail.

Reference https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle


mononaut (approved):
Tested ACK @ [5d1877a275a4b0ca29854540878209e0078bdbfb]

PR #5200, opened 2024-06-24
no trailing spaces

mononaut:
Updates the .eslintrc config for both front and backend to disallow trailing whitespace.


softsimon (approved):
utACK

PR #5203, opened 2024-06-24
Romanz support for address page

softsimon:
Hiding the rows with data we don't have when running with `romanz/electrs`

Desktop
Screenshot 2024-06-24 at 16 25 32

Mobile
Screenshot 2024-06-24 at 16 25 37


wiz (approved):
utACK

PR #5204, opened 2024-06-24
Fix tx position frontend error

softsimon:
Got random frontend errors related to methods called when `this.mempoolBlocks` is an empty array.

The `this.etaService.mempoolPositionFromFees` for example as it requires at least one mempoolBlock.


wiz (approved):
utACK

PR #5205, opened 2024-06-24
Add coinbase_addresses to extended blocks & table

mononaut:
This PR adds a new field to the extended blocks object/DB table containing an array of *all* coinbase output addresses for each block.

Many pools now use multiple coinbase output addresses, and do not send the majority of their rewards to the first output, so the single `coinbase_address` field is becoming less useful.

Indexing the full list of coinbase output addresses will also allow us to very quickly update pool assignments when pools-v2.json changes in future, without re-fetching block/coinbase tx data. (see PR #5206)

The PR also adds a step to the indexer to backfill missing coinbase addresses. which should take an hour or two to complete on the first run.


softsimon (approved):
tested ACK @ [03036bf59]

PR #5207, opened 2024-06-24
Pool reindexing

mononaut:
This PR reimplements the mining pool update/reindexing process to quickly refresh pool assignments when `pools-v2.json` changes with as little downtime as possible.

The PR relies on the new `coinbase_addresses` column in the blocks table added by PR #5205, so that needs to be deployed and allowed to finish indexing before attempting a `--update-pools` using this PR.

### Implementation

The PR works by fetching a full list of blocks from the DB for each pool affected by a change to `pools-v2.json` (each pool where the list of addresses or regexes changed, plus the 'unknown' pool).

For each block, we run the pool matching function against the indexed coinbase scriptsig and output addresses, and then update the resulting pool_id only if it has changed.

After this all finishes, we request a hashrate reindex from the mining service like before.

---

**question**: now that we aren't deleting any blocks from the DB, do we still need to request a difficulty adjustment reindex?


nymkappa:
Concept ACK. I'm not sure why I've never took the time to implement this. I think we never really felt the need for it since we can just take servers out of production. AFAIK most of the original indexing code has not been optimized/changed since my original naive implementation. Good to see it being optimized 👍🏻, and I'm sure there is lot of room for improvement in other places as well.

mononaut:
afaict, #5205 is deployed and finished indexing already, so this should be ready to go now.

PR #5209, opened 2024-06-25
Fix coinbase address type error

mononaut:
Fixes an ambiguous type error introduced in #5205


nymkappa (approved):
Tested ACK @ 3e057f2db1c853c24933338c5aee78ddbed18261

PR #5210, opened 2024-06-26
Force amounts to BTC unit in pool page

natsoni:
Before: 
Screenshot 2024-06-26 at 10 54 31

After:
Screenshot 2024-06-26 at 10 53 51


softsimon (approved):
tested ACK @ [ad2d7af084]

PR #5213, opened 2024-06-26
Add page in URL to accelerations table

natsoni:
This PR brings a few changes to the accelerations table:

- adds the page number as URL parameter (e.g. https://mempool.space/acceleration/list/42)
- adds support for keyboard navigation within the table
- fixes a loading indicator issue
- adds an SEO title to the page

This also update modifies the blocks table and the recent pegs table (Liquid) to use `/:page` instead of a query parameter for tracking the page, and fixes an issue with the loading indicator.


softsimon (approved):
tested ACK @ [122b4b05c4]

PR #5216, opened 2024-06-26
Align "Accelerated to / by" fields on mobile

natsoni:
| Before  | After |
| Screenshot 2024-06-26 at 18 05 23 | Screenshot 2024-06-26 at 18 05 05 |

### DIFF
--- a/frontend/src/app/components/acceleration/active-acceleration-box/active-acceleration-box.component.html
+++ b/frontend/src/app/components/acceleration/active-acceleration-box/active-acceleration-box.component.html
@@ -4,8 +4,11 @@
 <table>
   <tbody>
     <tr>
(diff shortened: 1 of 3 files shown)


softsimon (approved):
utACK

mononaut:
retroactive NACK:
 - the correct breakpoint for the columns on the transaction page is 850px, not 768px.
 - this `isMobile()` function doesn't get recomputed on window resize.
 - it would be good to keep the layout logic slightly abstracted from these page-specific responsiveness concerns, in case we want to use this component in other places.
 
I'd suggest adding an `@Input()` flag to the `active-acceleration-box` component to control which side the pie chart goes, and then set that flag using the `isMobile` field in the transaction component, which is already updated on resize.

e.g.
```
<app-active-acceleration-box [tx]="tx" [accelerationInfo]="accelerationInfo" [miningStats]="miningStats" [side]="isMobile ? 'left' : 'right'"></app-active-acceleration-box>
```

PR #5218, opened 2024-06-26
[accelerator] accelerate with lightning payments

nymkappa:


frontend/src/app/components/bitcoin-invoice/bitcoin-invoice.component.html
```diff
@@ -0,0 +1,89 @@
+<div class="wrapper">
```
nymkappa, line 1: This is mostly a copy/paste from the existing `bitcoin-method` component. But unfortunately it depends on too many things to simple move it in this repo. So copy paste is the way for now.

wiz:
Replaced by #5233

PR #5219, opened 2024-06-26
Show accelerator estimates on local instances

softsimon:
* The Accelerator is now visible on local instances if ACCELERATOR is `true` in the frontend config. previously it also required OFFICIAL_MEMPOOL_SPACE.
* Routes the estimate preview to mempool.space services

Screenshot 2024-06-26 at 21 43 28


nymkappa (approved):
Works as described @ b6d03953b9900a320564a12e6b6d2dd8b5346761

PR #5220, opened 2024-06-27
simple acceleration mode on main tx page

mononaut:
_(PR onto #5218 branch)_

refactors the accelerate checkout slightly, and integrates it into the main tx page for non-logged-in users.

Screenshot 2024-06-27 at 2 03 56 AM

the `/tracker` page implementation should be unchanged.

### DIFF
--- a/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.html
+++ b/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.html
@@ -18,7 +18,7 @@
       <div class="row">
         <div class="col-md">
           <div class="form-group form-check mb-2">
-            <input type="radio" [checked]="choosenOption === 'wait'"  class="form-check-input" id="wait" name="accelerate" (change)="selectedOptionChanged($event)">
+            <input type="radio" [checked]="choosenOption === 'wait'"  class="form-check-input" id="wait" name="accel" (change)="selectedOptionChanged($event)">
             <label class="form-check-label d-flex flex-column" for="wait">
               <span class="font-weight-bold">Wait</span>
               @if (eta.blocks < 7) {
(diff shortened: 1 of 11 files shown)


nymkappa:
Following comments are just UI/UX feedback. The integration itself works fine 👍🏻.

I think in that case, if the user click on the `Accelerate` button, we might not need to show the `Wait` option?

Or should selecting the `Wait` option close the modal on desktop only maybe?

Another suggestion would be to select the `Accelerate` option by default, again because the user already suggested he wants to accelerate by clicking on the `Accelerate` button in the transaction summary box.

Need to re-center the payment options as well _(I currently use flex justify around)_

image

mononaut:
> I think in that case, if the user click on the Accelerate button, we might not need to show the Wait option? Or should selecting the Wait option close the modal on desktop only maybe?

I think we still want to show that explicit choice with the pricing & timing info before moving on to the payment step, but we can iterate on this later.

> Another suggestion would be to select the Accelerate option by default, again because the user already suggested he wants to accelerate by clicking on the Accelerate button in the transaction summary box.

probably yes, but on the other hand maybe that defeats the point of making the user "opt in" with the radio button.

mononaut:
> Need to re-center the payment options as well (I currently use flex justify around)

I just pushed up a fix for this, alongside some other small UI changes:

https://github.com/mempool/mempool/assets/83316221/ce3abcf5-cbc4-4688-b991-87aba925893e

nymkappa:
> probably yes, but on the other hand maybe that defeats the point of making the user "opt in" with the radio button.

Well the user still have to actively click on the `Accelerate` button, alongside selecting a payment method and actually make the payment. I think that's enough "opt in" steps already.

> but we can iterate on this later

but yeah same vibe for my comment above 👍🏻

wiz:
Replaced by #5233

PR #5225, opened 2024-06-27
Fix acceleration preview showing with fragment on accel txs

### LINKED ISSUE #5221: [accelerator] ignore #accelerate fragment if transaction is already accelerated
If you reload the page after accelerating a transaction, the transaction preview is automatically shown again because the `#accelerate` fragment is in the url.
But it should be ignored if the transaction is already being accelerated.

image

natsoni:
Fixes #5221

https://github.com/mempool/mempool/assets/46578910/0eed68ed-cceb-485e-b084-ed23402e17ce


softsimon (approved):
tested ACK @ [c3963d6a0]

PR #5226, opened 2024-06-27
Fix logic for pool pie chart position

natsoni:
Implements @mononaut's suggestion from https://github.com/mempool/mempool/pull/5216#issuecomment-2193949639


softsimon (approved):
utACK

PR #5234, opened 2024-06-29
[btcpay] fix displayed amount

nymkappa:


frontend/src/app/components/bitcoin-invoice/bitcoin-invoice.component.ts
```diff
@@ -65,9 +65,7 @@ export class BitcoinInvoiceComponent implements OnInit, OnDestroy {
             this.paymentStatusSubscription = this.apiService.retreiveInvoice$(invoiceId).pipe(
               tap((invoice: any) => {
                 this.invoice = invoice;
-                this.invoice.amount = invoice.btcDue ?? (invoice.cryptoInfo.length ? parseFloat(invoice.cryptoInfo[0].totalDue) : 0) ?? 0;
```
nymkappa, line 68: This only happens if amount is dust, which should never happen in practice

PR #5236, opened 2024-06-30
Fix dust limit for undefined witness program outputs

mononaut:
Fixes a bug in the non-standard transaction detection which failed to apply the witness discount to the dust limit for unknown type witness program outputs.

According to https://github.com/bitcoin/bitcoin/blob/2f6dca4d1c01ee47275a4292f128d714736837a1/src/policy/policy.cpp#L54-L60
```
// Note this computation is for spending a Segwit v0 P2WPKH output (a 33 bytes
// public key + an ECDSA signature). For Segwit v1 Taproot outputs the minimum
// satisfaction is lower (a single BIP340 signature) but this computation was
// kept to not further reduce the dust level.
// See discussion in https://github.com/bitcoin/bitcoin/pull/22779 for details.
if (txout.scriptPubKey.IsWitnessProgram(witnessversion, witnessprogram)) {
    // sum the sizes of the parts of a transaction input
    // with 75% segwit discount applied to the script size.
    nSize += (32 + 4 + 1 + (107 / WITNESS_SCALE_FACTOR) + 4);
} else {
    nSize += (32 + 4 + 1 + 107 + 4); // the 148 mentioned above
}
```
All witness program outputs receive the segwit discount, even if they do not correspond to any known address type.

e.g. http://localhost:4200/tx/b10c0000004da5a9d1d9b4ae32e09f0b3e62d21a5cce5428d4ad714fb444eb5d#vout=7 was incorrectly labelled as non-standard because the dust limit was calculated *without* the segwit discount.

After applying this PR, the dust limit for this output is correctly calculated as 240 sats, and the transaction is therefore fully standard.

### DIFF
--- a/backend/src/__tests__/api/test-data/transactions-random.json
+++ b/backend/src/__tests__/api/test-data/transactions-random.json
@@ -331,5 +331,270 @@
             "block_hash": "00000000000000000002c69c7a3010fcd596c0c7451c23e7cd1f5e19ebf8ee6d",
             "block_time": 1718517071
         }
+    },
+    {
+        "txid": "b10c0000004da5a9d1d9b4ae32e09f0b3e62d21a5cce5428d4ad714fb444eb5d",
+        "version": 1,
+        "locktime": 1231006505,
+        "vin": [
+            {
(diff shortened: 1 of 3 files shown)


softsimon (approved):
tested ACK @ [eb3c248acd]

PR #5237, opened 2024-07-01
Fix errors caused by P2TR inputs without witness data

vostrnad:
The code incorrectly assumes all P2TR inputs have witness data, however transactions created before taproot activation don't need to have any. As a result, this transaction can't be displayed:  
https://mempool.space/tx/b10c007c60e14f9d087e0291d4d0c7869697c6681d979c6639dbd960792b4d41

FYI, there is code with similar assumptions in the backend folder as well, so maybe someone should look into that, but changing just the frontend code seems to be enough to fix the issue.


mononaut:
Screenshot 2024-07-01 at 1 11 11 AM 

he can't keep getting away with this

mononaut (approved):
Tested ACK @ [27c70bd9]

PR #5241, opened 2024-07-01
Fix missing bare multisig labels

vostrnad:
Since commit 7dfdb5553e5b06bf9c65add5cd8cada6c0d61790, bare multisig inputs and outputs have been missing the "Multisig m of n" label.

Happy to accept suggestions for a better approach, I'm not sure changing the interface of `AddressTypeInfo` this way is the best way to go about it.

Example transaction: https://mempool.space/tx/b10c0000004da5a9d1d9b4ae32e09f0b3e62d21a5cce5428d4ad714fb444eb5d

Current master:
multisigbefore

This PR:
multisigafter


mononaut (approved):
Tested ACK @ [1c89a1a4]

This is a good solution, thanks!

PR #5242, opened 2024-07-01
[logout] fix broken redirections

nymkappa:
To test alongside https://github.com/mempool/mempool.space/pull/1091. Also refresh footer auth state in real time


nymkappa:
Might wanna to just deploy https://github.com/mempool/mempool/pull/5244 directly

PR #5249, opened 2024-07-02
[auth] catch auth error and return null

nymkappa:
Sister pull request to https://github.com/mempool/mempool.space/pull/1094


softsimon (approved):
utACK

PR #5252, opened 2024-07-02
Fix bech32 regex and adapt tests

natsoni:
This PR fixes the regex detection for bech32 addresses ([minimum 6 characters of data](https://github.com/bitcoin/bips/blob/4f5a081d82323df5a7b0ff37e7e8d902837e2d13/bip-0173.mediawiki?plain=1#L89)) and adapts some of the frontend tests to accommodate this change.

Screenshot 2024-07-02 at 13 11 44

### DIFF
--- a/frontend/cypress/e2e/mainnet/mainnet.spec.ts
+++ b/frontend/cypress/e2e/mainnet/mainnet.spec.ts
@@ -144,13 +144,13 @@ describe('Mainnet', () => {
         });
       });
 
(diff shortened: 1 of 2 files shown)


mononaut (approved):
Tested ACK @ [7d7f9b16]

Maybe someday we should implement actual validation for the full base58 and bech32(m) specs, but in the meantime this nicely fixes the immediate issue.

PR #5253, opened 2024-07-02
Fix key navigation bug in accelerator dashboard

natsoni:
This PR fixes an issue where typing a random key on the accelerator dashboard would inadvertently navigate to the accelerations list page: 

https://github.com/mempool/mempool/assets/46578910/4ed1e6dc-4261-46d1-a04c-119ef428354c

### DIFF
--- a/frontend/src/app/components/acceleration/accelerations-list/accelerations-list.component.ts
+++ b/frontend/src/app/components/acceleration/accelerations-list/accelerations-list.component.ts
@@ -50,17 +50,38 @@ export class AccelerationsListComponent implements OnInit, OnDestroy {
     if (!this.widget) {
       this.websocketService.want(['blocks']);
       this.seoService.setTitle($localize`:@@02573b6980a2d611b4361a2595a4447e390058cd:Accelerations`);
+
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/acceleration/accelerations-list/accelerations-list.component.ts
```diff
@@ -50,17 +50,38 @@ export class AccelerationsListComponent implements OnInit, OnDestroy {
...
+      this.paramSubscription = this.route.params.pipe(
+        tap(params => {
+          this.page = +params['page'] || 1;
+          this.pageSubject.next(this.page);
+        })
+      ).subscribe();
+
+
+      this.keyNavigationSubscription = this.stateService.keyNavigation$.pipe(
+        tap((event) => {
```
softsimon, line 63: The subscription and `pageChange()` is triggered on every key press,

Solution:

`filter(() => event.key === nextKey || event.key === prevKey)`

softsimon (changes requested):
See comment

softsimon (approved):
tested ACK @ [6b0496029]

PR #5263, opened 2024-07-03
New ETA label

softsimon:
Displayed when transaction is at the 8th block or more.

Screenshot 2024-07-03 at 15 56 31

Screenshot 2024-07-03 at 15 56 41


mononaut (approved):
utACK

PR #5267, opened 2024-07-03
fix ln invoice flicker

mononaut:
long term fix for #5258


softsimon (approved):
tested ACK @ [1ce72e23a3]

PR #5274, opened 2024-07-04
Fix stale cpfp bug

mononaut:
Fixes an issue where stale CPFP ancestor & descendant info might not be correctly reset/updated for some transactions.

e.g. https://mempool.space/tx/eab2f16c89a47fbfd4f6cfb1a5e2fe83ae703034e84e9cec2818c341cab45566 shows two lower-fee-rate CPFP ancestors, which have actually already been confirmed.
Screenshot 2024-07-04 at 8 54 56 AM
Screenshot 2024-07-04 at 8 55 10 AM


softsimon (approved):
tested ACK @ [c37858fa5]

PR #5275, opened 2024-07-04
Remove Layer 2 divider

softsimon:
* Removes the "Layer 2" divider in the network dropdown
* Removes Liquid & Liquid Testnet from the dropdown (and footer) in production


wiz (approved):
utACK

PR #5278, opened 2024-07-04
Fix issue on key navigation logic

natsoni:
This PR applies this fix [6b0496029c1aaf54edd7c203e61f8e0eeeac5551] to blocks list and recent pegs list.


wiz (approved):
utACK

PR #5279, opened 2024-07-04
Add more context to error messages

natsoni:
Before: 
Screenshot 2024-07-04 at 19 48 38

After: 
Screenshot 2024-07-04 at 19 44 57
Screenshot 2024-07-04 at 19 45 09

Also added an error message to the pool page which would load indefinitely if an error occurred:

Screenshot 2024-07-04 at 19 45 25


wiz (approved):
utACK

PR #5280, opened 2024-07-05
Add first seen data to confirmed transactions

natsoni:
This PR adds first seen data to mined transactions, retrieved from the fetched block audit: 

Screenshot 2024-07-05 at 11 47 00


softsimon (approved):
tested ACK @ [c85e7b08]

PR #5285, opened 2024-07-06
[accelerator] error message after timeout

mononaut:
Shows an additional message if the user has been waiting for a transaction to be accelerated for >20 seconds.

Screenshot 2024-07-06 at 8 31 48 AM

Switches to the error screen after two minutes:

Screenshot 2024-07-06 at 8 15 00 AM


wiz:
This failure case sucks for the user, since they paid and it didn't get accelerated - I guess it would make sense to tell them to contact support with a request ID (or simply the TXID)

softsimon (approved):
utACK

PR #5287, opened 2024-07-07
Acceleration timeline polishing

natsoni:
This PR brings some polishing to the acceleration timeline:
- it now includes the _unaccelerated ETA_, computed from transaction in-band fee and eventual CPFP data
- greys out badges of steps that have not occured yet (mined, accelerated)
- adds missing i18n

Timeline of an accelerated transaction in the mempool: 

Screenshot 2024-07-07 at 16 50 57

-----------------------------------
[Not included in this PR]
It is possible to show the timeline on not yet accelerated transactions, like this for example: 

Screenshot 2024-07-07 at 16 55 33

But I'm not sure if/where we want to put the timeline here. So **for now it's only displayed on already accelerated transactions**.

### DIFF
--- a/frontend/src/app/components/acceleration-timeline/acceleration-timeline.component.html
+++ b/frontend/src/app/components/acceleration-timeline/acceleration-timeline.component.html
@@ -1,3 +1,4 @@
+@if (tx.status.confirmed) {
 <div class="acceleration-timeline box">
   <div class="timeline-wrapper">
     <div class="timeline">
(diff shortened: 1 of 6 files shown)


mononaut:
looks good!

the way the timeline branches *before* the "accelerated" node is slightly confusing.

perhaps something like this would be easier to understand?
Screenshot 2024-07-08 at 6 17 42 AM

natsoni:
Taking into account feedback: 
- The two timelines don't merge at the end separated 
- The accelerated timeline is the path that forks from the regular timeline
- The accelerated fork starts from the 'accelerated' point, not after

Screenshot 2024-07-08 at 21 38 09

Feels like the point at the end of the regular timeline is a little off. Maybe we could display something like this if appropriate: 

Screenshot 2024-07-08 at 21 20 21

wiz (approved):
Tested ACK @ v3.0.0-dev [6f31fb2a0][ea6f40d8]

PR #5288, opened 2024-07-07
Quick fix for statistics replication

natsoni:
The statistics replication service currently has two bugs that need fixing:

1. The service may add statistics that are already present in the table.
2. The service does not sanitize the data before adding it to the table, which can lead to problems if some of the trusted servers send wrong data.

This PR should mitigate the first bug by filtering more aggressively on what data we actually need to fill. 

Since the second bug is not fixed, this PR also disables the service on prod until I come up with a decent way to sanitizing the data.


wiz (approved):
utACK

PR #5291, opened 2024-07-08
Fix accelerator error message positioning

softsimon:
Made sure the error text can handle 2 rows of overflowing text with centered fixed width, and not overflow the UI. 

| Before | After |
|-|-|
| Screenshot 2024-07-08 at 16 34 22 | Screenshot 2024-07-08 at 16 38 24 |
| Screenshot 2024-07-08 at 16 34 26 | Screenshot 2024-07-08 at 16 42 03 |


wiz (approved):
utACK

PR #5292, opened 2024-07-08
[accelerator] hide modal for transactions near the top of the mempool

mononaut:
Minimizes the accelerator modal for transactions in the top 75% of the next projected block (i.e. that don't need to be accelerated)


wiz (approved):
utACK

PR #5294, opened 2024-07-08
[accelerator] improve rendering of acceleration fee rate graph

mononaut:
Enforces a minimum height for the bars on the acceleration fee rate graph, so that the text is always legible.

| Before | After |
|-|-|
| Screenshot 2024-07-08 at 2 45 03 PM | Screenshot 2024-07-08 at 2 45 14 PM |
| Screenshot 2024-07-08 at 2 48 33 PM | Screenshot 2024-07-08 at 2 48 43 PM |


softsimon (approved):
tested ACK @ [9396a4bba]

PR #5298, opened 2024-07-09
add seconds to address and transaction views

svrgnty:
Added seconds to the timestamps for the transaction and address frontend pages

Screenshot 2024-07-09 at 13 09 48
Screenshot 2024-07-09 at 13 09 58


softsimon (approved):
utACK

PR #5299, opened 2024-07-09
services api endpoint config

mononaut:
Adds a frontend configuration option to choose the URL prefix to use for services API requests, and changes the default to call `https://mempool.space/api/v1/services` by absolute URL rather than by relative path.

To override that default, set `SERVICES_API` to your preferred path in `mempool-frontend-config.json`.


wiz (approved):
utACK

PR #5301, opened 2024-07-09
Remove standard ETA from timeline

natsoni:
_builds on #5297_

This PR:
- removes the standard ETA from the timeline until we implement a more realistic ETA calculation method
- updates the timeline colors to be more similar to the RBF timeline

https://github.com/mempool/mempool/assets/46578910/57c4167a-5887-421d-87da-5df1bf16d26d

Screenshot 2024-07-09 at 21 57 22


wiz (approved):
utACK

PR #5302, opened 2024-07-09
Smaller block arrow

softsimon:
* The dashboard is moved 10px down from the blockchain blocks
* The arrow is smaller on the blocks to not cover the miner tags, but still bigger on the mempool blocks

Screenshot 2024-07-10 at 01 42 45


wiz (approved):
utACK

PR #5303, opened 2024-07-10
always show out-of-band block fees to 8 decimal places

mononaut:
Before:
Screenshot 2024-07-10 at 3 54 37 AM

After:

Screenshot 2024-07-10 at 3 53 54 AM


softsimon (approved):
utACK

PR #5304, opened 2024-07-10
Possibly fix miner tag loading on tracked transactions

natsoni:
This PR is an attempt to fix the intermittent bug where the miner tag would load forever on tracked transactions that just confirmed.


wiz (approved):
utACK

PR #5305, opened 2024-07-10
Avoid fetching full audit on tx page

natsoni:
This PR allows the frontend to load only the audit data it needs for the displayed transaction instead of the full block audit, whose loading on  which can be quite noticeable on slow connections.

To do so, it creates a new endpoint `api/v1/block/:hash/tx/:txid/audit` that allows to only fetch the necessary audit data for a particular transaction. When the block audit has already been loaded (if the user clicks on a transaction from the block page for example), we rely on the full cached audit and don't fetch from the new endpoint.

### DIFF
--- a/backend/src/repositories/BlocksAuditsRepository.ts
+++ b/backend/src/repositories/BlocksAuditsRepository.ts
@@ -1,7 +1,7 @@
 import blocks from '../api/blocks';
 import DB from '../database';
 import logger from '../logger';
-import { BlockAudit, AuditScore } from '../mempool.interfaces';
+import { BlockAudit, AuditScore, TransactionAudit } from '../mempool.interfaces';
 
 class BlocksAuditRepositories {
   public async $saveAudit(audit: BlockAudit): Promise<void> {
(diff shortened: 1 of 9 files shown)


backend/src/repositories/BlocksAuditsRepository.ts
```diff
@@ -98,6 +98,35 @@ class BlocksAuditRepositories {
...
+    try {
+      const blockAudit = await this.$getBlockAudit(hash);
+      
+      if (blockAudit) {
+        const isAdded = blockAudit.addedTxs.includes(txid);
+        const isPrioritized = blockAudit.prioritizedTxs.includes(txid);
+        const isAccelerated = blockAudit.acceleratedTxs.includes(txid);
+        const isConflict = blockAudit.fullrbfTxs.includes(txid);
+        const isExpected = blockAudit.template?.some(tx => tx.txid === txid);
+        const firstSeen = blockAudit.template?.find(tx => tx.txid === txid)?.time;
```
mononaut, line 111: nit: you probably don't need to search the whole template twice here.

softsimon (approved):
tested ACK @ [21a0218df5]

frontend/src/app/components/block/block.component.ts
```diff
@@ -295,6 +295,7 @@ export class BlockComponent implements OnInit, OnDestroy {
             ),
           !this.isAuditAvailableFromBlockHeight(block.height) ? of(null) : this.apiService.getBlockAudit$(block.id)
             .pipe(
+              tap(() => this.cacheService.setBlockAuditLoaded(block.id)),
```
mononaut, line 298: let's move this cache tracking stuff inside the `api.getBlockAudit$()` method so we don't need to remember to add this `tap()` section every time we use the API endpoint.

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -364,33 +364,41 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
...
-          retry({ count: 3, delay: 2000 }),
-          catchError(() => {
-            return of(null);
-          })
-        ) : of(isCoinbase ? { coinbase: true } : null);
+        if (fetchAudit) {
+        // If block audit is already cached, use it to get transaction audit
+          const blockAuditLoaded = this.cacheService.getBlockAuditLoaded(hash);
+          if (blockAuditLoaded) {
+            return this.apiService.getBlockAudit$(hash).pipe(
```
mononaut, line 371: are we using `blockAuditLoaded` here as a heuristic to guess whether the resource is already in the browser's disk cache?

I'm not sure how I feel about that 😆

mononaut (commented):
Concept ACK, but I have a few comments and questions re. approach

frontend/src/app/components/block/block.component.ts
natsoni: I did this initially, but ran into circular dependancy issue (`apiService` is imported in `cacheService`)
Not sure how to get around it 😅

mononaut: hmm, I guess either move the entry point into `cacheService`, and fetch audits like `cacheService.getBlockAudit$()`, or move the *cache* part into `apiService`.

softsimon (approved):
tested ACK @ [c3b9828d]

PR #5306, opened 2024-07-10
align block arrows & reposition selected block pool tag

mononaut:
possible alternative to #5302.

the smaller block arrow is used for both mempool & mined blocks, and positioned immediately under the blocks. To make room for the arrow, the pool tag shifts down for the active block.

the PR also adds a slight fade in/out effect to the mined block arrow to reduce flicker on page transitions.

https://github.com/mempool/mempool/assets/83316221/360af3eb-cc0d-4abe-a253-678c51dda3e0


wiz (approved):
utACK

PR #5309, opened 2024-07-10
Refactor pool logo component & fallback logic

mononaut:
consolidates the various separate implementations of the pool logo image into a single component, with more robust fallback logic to avoid showing flashes of alt-text or the broken image icon.


softsimon:
Looks nice but I don't like the new logo blinking on page switching :(

PR #5311, opened 2024-07-10
Update tx acceleration state on confirmation

natsoni:
[not tested on actual prod accelerations, only locally]

This PR fixes an issue where a pending accelerated tx would not properly update its state when getting included in a block: 

- The `Accelerated fee rate` field is not updated to the fair acceleration rate 
- If for some reason the tx is mined by a non-participating pool, the `Accelerated` tag stills appears until reload


softsimon (approved):
utACK

PR #5322, opened 2024-07-12
Fix accelerations graph view more

natsoni:
Fixes unclickable 'VIew more' button on acceleration dashboard


softsimon (approved):
tested ACK @ [5665c6e6e]

PR #5325, opened 2024-07-12
Restrict accelerator routes to mainnet

mononaut:
Restricts the accelerator dashboard, accelerator fees graph, and accelerations list pages to Bitcoin Mainnet only


wiz (approved):
utACK

PR #5330, opened 2024-07-13
Block preview new miner tag design

softsimon:
| Before | After |
|-|-|
| Screenshot 2024-07-13 at 16 48 08 | Screenshot 2024-07-13 at 16 47 58 |


wiz (approved):
utACK

PR #5332, opened 2024-07-13
Html quickfix

hans-crypto:
The `tx-skeleton` class overrides the `text-center mb-4` classes due to being specified last. This change removes the unused classes, that were never effective.


softsimon (approved):
utACK

PR #5334, opened 2024-07-13
Allow to click on bid boost graph to go on block page

natsoni:
Only possible on 24h and 3d views (where block height is not averaged)

https://github.com/user-attachments/assets/de079b1b-9833-442d-aea4-b491e46d193e


wiz (approved):
utACK

PR #5347, opened 2024-07-17
Bump Rust version to 1.79

junderw:
Maintaining an old MSRV is not a priority for this project. If you would like to keep an old MSRV active, please maintain your own patch/fork.

There have been repeated instances of the mempool team being stuck on Rust MSRV problems (transient dependencies suddenly break MSRV and cause issues on specific targets during docker builds etc.)...

I think as a project that uses `rustup` during the build process, trying to maintain an old MSRV is not good for the project.


wiz (approved):
utACK

PR #5351, opened 2024-07-19
Fix recursion loop in search bar

### LINKED ISSUE #5247: "InternalError: too much recursion" when searching for a future SegWit version address
### Description

Searching for a future SegWit version address and clicking it causes this error log in the console:

```
ERROR InternalError: too much recursion
```

#### Version

3dbbc830770231815e804ca147d636d59b77e30c

### Steps to reproduce

1. Start typing "bc1pfeessrawgf" into the search box.
2. Click on the address when it appears as a result.

### Expected behaviour

The address's page should appear.

### Actual behaviour

Nothing happens when the address is clicked.

### Screenshots

bc1pfees

natsoni:
Fixes #5247

The recursion was caused by the following: 

Clicking on any item in the search results that does not match any regex, for example the future SegWit version address `bc1pfeessrawgf` (prior commit 7d7f9b16654215a97a69226c3ebbf0d86bab1ec6) would start a loop that never ends. 

https://github.com/mempool/mempool/blob/0a116804e86bcf8de20f69e52bb0ede262fdf7d5/frontend/src/app/components/search-form/search-results/search-results.component.ts#L70-L71

triggers
https://github.com/mempool/mempool/blob/0a116804e86bcf8de20f69e52bb0ede262fdf7d5/frontend/src/app/components/search-form/search-form.component.ts#L240-L242

triggers (because no regex match)
https://github.com/mempool/mempool/blob/0a116804e86bcf8de20f69e52bb0ede262fdf7d5/frontend/src/app/components/search-form/search-form.component.ts#L305-L307

triggers 
https://github.com/mempool/mempool/blob/0a116804e86bcf8de20f69e52bb0ede262fdf7d5/frontend/src/app/components/search-form/search-results/search-results.component.ts#L38-L40

...etc 
 **************************
This PR removes the line that triggers the search back when no regex matches.

### DIFF
--- a/frontend/src/app/components/search-form/search-form.component.ts
+++ b/frontend/src/app/components/search-form/search-form.component.ts
@@ -303,7 +303,6 @@ export class SearchFormComponent implements OnInit {
           (error) => { console.log(error); this.isSearching = false; }
         );
       } else {
-        this.searchResults.searchButtonClick();
         this.isSearching = false;
       }
     }


softsimon (approved):
tested ACK @ [8c07e3c31a]

PR #5352, opened 2024-07-21
Fix crypto lib call crash with custom function

softsimon:


softsimon:
> maybe call the function insecureRandomUUID

done

knorrium (approved):
utACK

PR #5354, opened 2024-07-22
v1 audits

mononaut:
This draft PR:
 - adds versioning to the block audits table
 - implements an improved "v1" block audit
 - prepares migration code to convert existing block audits (but this is not activated yet)
 
The new "v1" audit:
 - records whether transactions were seen in our mempool before the block was mined (this data was not previously collected).
 - changes the definition of the "prioritized" category to include transactions that were mined with a higher priority than indicated by in-band fee rate alone, whether or not they were expected in the block.
    - this allows us to distinguish between transactions *submitted* out-of-band (i.e. "added" transactions), and those with additional fees paid out-of-band ("prioritized").
 - reduces the false positive rate for "added" and "prioritized" transactions near the end of blocks.
 
The new approach to detecting prioritized transactions is tightly coupled to the specific transaction selection algorithm used to construct blocks. This means it detects prioritized transaction much more accurately when miners use the exact same GBT algorithm as us, but may be less tolerant of other template algorithms.

In practice, almost all blocks are generated using standard GBT, but I want to monitor this running locally for a while to confirm that it's a strict improvement before taking the PR out of draft.

e.g. all prioritized transactions are now correctly identified at the start of this [ViaBTC block](https://mempool.space/block/00000000000000000000228fdfa72f5e82346574a7386f5bd86bdc910e51c498):

| Before | After |
|-|-|
| Screenshot 2024-07-22 at 9 58 15 AM | Screenshot 2024-07-22 at 9 58 30 AM |

### DIFF
--- a/backend/src/api/audit.ts
+++ b/backend/src/api/audit.ts
@@ -6,15 +6,16 @@ import rbfCache from './rbf-cache';
 const PROPAGATION_MARGIN = 180; // in seconds, time since a transaction is first seen after which it is assumed to have propagated to all miners
 
 class Audit {
-  auditBlock(transactions: MempoolTransactionExtended[], projectedBlocks: MempoolBlockWithTransactions[], mempool: { [txId: string]: MempoolTransactionExtended }, useAccelerations: boolean = false)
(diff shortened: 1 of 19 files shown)


softsimon (approved):
tested ACK @ [d2b57c8d]

PR #5356, opened 2024-07-22
fix btc payment waiting reply

softsimon:


wiz (approved):
YoloACK
@nymkappa please review

PR #5358, opened 2024-07-23
Fix miner loading forever

### LINKED ISSUE #5102: Miner loading forever on confirmed transaction
Tracked a accelerated transaction. It got confirmed and the miner is loading like this

Screenshot 2024-05-25 at 22 17 17

natsoni:
Fixes #5102

You can reproduce this intermittent issue by doing the following:
- Go on the next block page at https://mempool.space/mempool-block/0
- Click on a random transaction to start tracking it
- Go back to the next block page
- Go back to the same tracked transaction
- The miner tag (as well as the audit data) won't appear when the transaction gets mined

Closing the txConfirmed subscription when leaving the tx page fixes the issue.


softsimon (approved):
utACK

PR #5359, opened 2024-07-23
Prevent never ending loop of calls to transactionTimes

natsoni:
Sometimes when a tracked tx confirms, the frontend gets stuck in a never ending loop of calls to /api/v1/transaction-times.


softsimon (changes requested):
Use this solution to remove the setTimeout logic completely https://github.com/mempool/mempool/pull/5356/files ... to avoid future bugs/race conditions

softsimon (changes requested):
Convert the transactionTimes function into a active Subject observable just how we do with the `fetchCpfp$` and call it `this.transactionTimes$.next(this.tx.txid);`.. And unsubscribe onDestroy

softsimon (approved):
tested ACK @ [ca4b1943a]

PR #5361, opened 2024-07-23
On-demand acceleration polling

mononaut:
This PR is a first pass at on-demand acceleration polling & assorted other improvements for interacting with the accelerator from a personal/local instance:

Backend changes:
- Official Mempool backends now serve a copy of their latest active accelerations list at `/api/v1/accelerations`, so that non-official/FOSS builds can poll that instead of hitting the services backend directly.
- Official instances poll the services backend (`config.MEMPOOL_SERVICES.API`) for accelerations, while non-official instances poll general mempool servers (`config.EXTERNAL_DATA_SERVER.MEMPOOL_API`).
- Implemented on-demand polling, which is used when the backend `config.MEMPOOL_SERVICES.ACCELERATIONS` config setting is disabled:
    - after successfully paying for an acceleration, the frontend registers it with the local backend via a new `/acceleration/request/:txid` endpoint.
    - while there are any pending requests, the backend polls for relevant accelerations on every mempool loop.
    - once all pending requests are acknowledged, the backend continues to poll for updates 
        - every 10 minutes while any of the user's accelerations are in progress.
        - whenever an active acceleration is mined.
    - unacknowledged/completed/failed accelerations expire after 10 minutes.

Frontend changes:
- Added an `ACCELERATOR_BUTTON` config flag which controls the visibility of the acceleration button separately from the `ACCELERATOR` setting.
- Modified `AccelerateCheckout` to call the new `logAccelerationRequest$` API method when an acceleration is successfully paid.

Note that in "on-demand" mode, the backend only tracks accelerations initiated by users of that instance, and only while those transactions are in the mempool. Third-party accelerations and confirmed/historic accelerations are not displayed.

The list of on-demand user accelerations is only stored in memory, not persisted to the database, disk cache or redis.


wiz:
> Official Mempool backends now serve a copy of their latest active accelerations list at /api/v1/accelerations, so that non-official/FOSS builds can poll that instead of hitting the services backend directly.

the mempool backend shouldn't be used as an HTTP proxy, that's what nginx is for - in this case nginx already caches the existing accelerations endpoint for 1-2 seconds

mononaut:
it's not really a proxy, it's serving a separate copy of the data.

obviously that's very tightly coupled right now, but in future the mempool backend might receive acceleration data over websocket, or want to do some additional processing before serving it, so it feels like having a separate endpoint for non-official backends gives us more flexibility.

but I'll revert the change for now.

wiz:
ah ok fair point, i forgot about the websocket transition you had planned

mononaut:
removed that last commit and rebased on current master

softsimon (approved):
partially tested ACK @ [af7a962a0]

wiz (approved):

PR #5362, opened 2024-07-23
Add fee delta and pool name to acceleration list

natsoni:
This PR adds two rows to the accelerations list: the fee delta and the pool that mined the acceleration.


softsimon (changes requested):
Remove FeeDelta column

natsoni:
Done: 
Screenshot 2024-07-25 at 00 25 22

softsimon (approved):
tested ACK @ [aee2454a9]

PR #5365, opened 2024-07-25
Get all pools in accelerations list

natsoni:
Currently, we only fetch the pools that mined a block in the last month in the accelerations list. In the future, this might prevent the pools from appearing in the table. 
This PR fixes this issue by fetching all the pools instead of the last month data.


softsimon (approved):
tested ACK @ [90331e2c]

PR #5371, opened 2024-07-26
Add tooltip to acceleration timeline

natsoni:
This PR adds acceleration data to the timeline: 

Screenshot 2024-07-26 at 11 24 33

Screenshot 2024-07-26 at 11 27 03


wiz (approved):
Tested ACK @ 5d37e08

frontend/src/app/components/acceleration-timeline/acceleration-timeline-tooltip.component.html
```diff
@@ -45,7 +45,7 @@
         }
       </tr>
       <tr *ngIf="['accelerated', 'mined'].includes(accelerationInfo.status) && hasPoolsData()">
-        <td class="label" i18n="transaction.accelerated-by-pool|Accelerated by pool">Accelerated by</td>
```
natsoni: Note: I initially added a new i18n entry
`accelerated-by-pool|Accelerated by pool` 
and did not use 
`accelerated-by-hashrate|Accelerated to hashrate` 
because I wasn't sure that `Accelerated by` would translate correctly when referring to pools instead of a hashrate value.

PR #5383, opened 2024-07-29
[accelerator] use "mined" when accelerated tx is mined by non participating pool

nymkappa:
We know show `Mined` if included in a non participating pool block.

Screenshot 2024-07-29 at 11 06 16


softsimon (approved):
tested ACK @ [41441bb9]

PR #5385, opened 2024-07-30
Fix effective fee rate detection on tx page

mononaut:
Fixes the frontend logic used to detect when to display an effective fee rate row on the transaction page.

Previously we allowed the `effectiveFeePerVsize` to differ from the regular fee rate by up to 0.1 or 0.01 sats/vb before displaying the effective fee rate row, to allow for differences due to vsize rounding. However, for very small transactions this is insufficient.

This PR checks instead whether the `effectiveFeePerVsize` exactly matches the simple rate calculated with and without vsize rounding, which more accurately identifies the relevant transactions.

Before: (effective fee rate row incorrectly shown due to large vsize rounding error)
Screenshot 2024-07-30 at 11 51 30 AM

After:
Screenshot 2024-07-30 at 11 51 21 AM


softsimon (approved):
utACK

PR #5386, opened 2024-07-30
Fix missing accelerated fee rate in timeline tooltip

natsoni:
On accelerations with a bid boost of 0, the fee rate line was missing.

Before:
Screenshot 2024-07-30 at 17 24 31

After: 
Screenshot 2024-07-30 at 17 25 06


softsimon (approved):
utACK

PR #5387, opened 2024-07-30
Fix accelerated time centering

natsoni:
The relative time of the acceleration request had an incorrect offset on mined transaction: 

Screenshot 2024-07-30 at 17 31 50

This PR also fixes a small capitalization issue (capitalize `Just now` if the transaction is mined)


softsimon (approved):
utACK

PR #5389, opened 2024-07-30
fix: acceleration rtl issues

TechMiX:
Tried to fix the RTL issues related to the acceleration panel


softsimon (approved):
tested ACK @ [fa4b445d]

PR #5391, opened 2024-07-31
Set pool logos opacity to 0.3

natsoni:
Lowered the opacity of the pool logos that did not mine the acceleration and removed the pulsating effect:

Screenshot 2024-07-31 at 12 23 23


softsimon (approved):
utACK

PR #5399, opened 2024-08-02
Remove query param redirect on tx page

natsoni:
I often experience the following issue on mobile: 
- While browsing on mempool.space, I click on a transaction
- To return to the previous page, I need to click the browser's back button twice

https://github.com/user-attachments/assets/d0e52638-649d-4bf0-84a8-310248adaf9c

Not redirecting to `mode=details` fixes the issue. I also checked that this does not break the pizza tracker redirection logic.

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.ts
+++ b/frontend/src/app/components/transaction/transaction.component.ts
@@ -523,6 +523,7 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
               queryParamsHandling: 'merge',
               preserveFragment: true,
               queryParams: { mode: 'details' },
+              replaceUrl: true,
             });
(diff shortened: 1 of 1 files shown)


mononaut (changes requested):
We do still need to add the `?mode=details` query param to the URL as a routing hint, otherwise if you navigate to a `/tx` page on mobile and then reload the page you unexpectedly get routed to the tracker instead:


https://github.com/user-attachments/assets/ea5c6cc3-bb4e-41e3-a8ee-fb619d01e03b



The extra browser history entry is definitely a bug though. IIRC you can add `replaceUrl: true` to the navigation options to prevent that.

mononaut (approved):
Tested ACK @ [2489fc490]

PR #5401, opened 2024-08-02
adjust pie colors to handle more pools

mononaut:
pick accelerator hashrate pie chart colors from a fixed range instead of lightening by a constant increment (avoids washing out the colors when there are more pools)

| Before | After |
|-|-|
| Screenshot 2024-08-02 at 4 53 32 PM | Screenshot 2024-08-02 at 4 53 48 PM |

(extreme example, but you get the idea)


softsimon (approved):
utACK

wiz:
can we try swapping it so it fades from white to purple

wiz:
Screenshot 2024-08-03 at 12 00 43
ideally it would be white in the upper right side and fade to dark purple like the mempool accelerator logo itself

PR #5405, opened 2024-08-03
fix: acceleration rtl layout issues

TechMiX:
The third and hopefully last effort to fix all acceleration related layout issues in RTL mode.


softsimon (approved):
tested ACK @ [6f0a5c9b]

PR #5408, opened 2024-08-04
Pizza tracker: handle transaction not yet in mempool

natsoni:
The pizza tracker currently crashes if there is no transaction with given txid in mempool.

Before:

https://github.com/user-attachments/assets/6eb09f79-1d56-4aba-b978-48a91777650c

After: 

https://github.com/user-attachments/assets/f3e5a76c-7540-41cf-9024-b384cd1d4977


softsimon (approved):
tested ACK @ [e378df41]

wiz:
good looks

PR #5409, opened 2024-08-04
Fix truncated link to not refresh full window

natsoni:
Clicking on a link using the component `app-truncate` reloads the full page, which can trigger a redirection to the pizza tracker in some cases, for example on cpfp: 

https://github.com/user-attachments/assets/8d9a386d-2082-4fcc-8b41-d5f21aed5af5

With this fix: 

https://github.com/user-attachments/assets/230e5a81-9c39-4022-9ce3-4ab1e6e9232b


softsimon (approved):
tested ACK @ [439c52af3]

PR #5410, opened 2024-08-04
Show oob fees on tx details

natsoni:
https://github.com/user-attachments/assets/296b6654-2e71-423a-8589-e42653226c03


softsimon (approved):
tested ACK @ [de9fae5cd]

PR #5419, opened 2024-08-05
sats -> sat

natsoni:
When amounts are toggled to sats, use `sat` instead of `sats` for consistency.


softsimon:
Closed due to https://github.com/mempool/mempool/pull/5437

PR #5422, opened 2024-08-05
Purple accelerated fee rate

natsoni:
Pending acceleration: 

Screenshot 2024-08-05 at 12 16 42

Mined acceleration: 
Screenshot 2024-08-05 at 12 47 58


mononaut:
the fiat amount is a bit confusing, but it looks good 👍

wiz:
🤔

wiz (approved):
Let's try it out and see how it feels

PR #5423, opened 2024-08-05
Hide accelerator box if transaction is accelerated

natsoni:
Should prevent the acceleration box to sometimes appear on pending accelerated transactions.


wiz:
looks like something to fix in 3.1 and maybe 3.0.1

natsoni:
Solved by #5471

PR #5425, opened 2024-08-05
Clear mining service cache on network change

natsoni:
This PR fixes the mining pool pie chart not updating on network change by clearing the mining service cache.

There is still some weird subscription behaviour, particularly visible on network change, causing duplicated API calls. This will need more investigation.


wiz (approved):
utACK

PR #5429, opened 2024-08-05
Fix stray loading spinner in block unfurl

mononaut:
fixes broken logic to hide the loading spinner on the block preview block visualization component, which sometimes caused the spinner to be visible in unfurler images, e.g:

loadingunfurl


wiz (approved):
utACK

PR #5430, opened 2024-08-05
Immediately show oob fee on accelerated transaction

natsoni:
Currently we need to refresh the page to see the the out-of-band fee paid in transaction details.

With this PR: 

https://github.com/user-attachments/assets/6f5e1e24-728c-4b88-ac0b-d5bc8ed659be


wiz (approved):
utACK

PR #5437, opened 2024-08-07
only use sats, not sat

softsimon:
This is an alternative to https://github.com/mempool/mempool/pull/5419 that changes sats to sat.

Changed to use "sats" consistently across all values, instead of sat.
Sats is the plural form which makes more sense.

Before
Screenshot 2024-08-07 at 12 06 26

After
Screenshot 2024-08-07 at 12 06 16

After
Screenshot 2024-08-07 at 11 53 04


softsimon:
self ack, we should use _sats_

PR #5438, opened 2024-08-07
Hide fee delta on accelerated tx if bid boost is 0

natsoni:
This PR fixes the following issue: 
- track an accelerated transaction that has a sufficient in-band fee to be mined anyway
- when the transaction gets mined, it will have a bid boost of 0
- the fee delta is still shown on the mined transaction

With this PR, the out-of-band fee fee delta is hidden if the bid boost is 0 after the acceleration gets mined.

for example:

https://github.com/user-attachments/assets/127ec359-ebe9-48fb-a1cc-0ff98dcd16e5


wiz (approved):
utACK - @knorrium need to test this on device

PR #5444, opened 2024-08-08
Fix pool pie position on safari

natsoni:
On Safari the pool pie was incorrectly positioned when the chart was on the left (on smaller screens).

Before: 

Screenshot 2024-08-08 at 11 24 32

After:

Screenshot 2024-08-08 at 11 24 46


softsimon (approved):
tested ACK @ [132d6204c3]

PR #5445, opened 2024-08-08
Persist mempool block visualization between pages

mononaut:
This PR persists the mempool block visualization when possible, so that the it's displayed immediately when switching between e.g. the main dashboard, acceleration dashboard, and first mempool block page, instead of completely reloading the visualization and waiting for another copy of the data.

Before:

https://github.com/user-attachments/assets/d477f531-7c49-4c8c-9f14-6e519d73838b

After:

https://github.com/user-attachments/assets/f13a9b66-9c5b-4d96-8d5d-ed14b1fba8b6


softsimon (approved):
tested ACK @ [009fba3d]

PR #5449, opened 2024-08-09
fix mined acceleration detection logic on tx pages

mononaut:
Avoids showing unnecessary "effective fee rate" row on the transaction page when an accelerated transaction was mined by a non-participating pool, e.g. https://mempool.space/tx/6763a1ca42fcde0af2ad46d5c2b8c62cd20b10e0e2fd5e36c56bb6769a1bc4bc

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.ts
+++ b/frontend/src/app/components/transaction/transaction.component.ts
@@ -359,12 +359,16 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
     ).subscribe((accelerationHistory) => {
       for (const acceleration of accelerationHistory) {
         if (acceleration.txid === this.txId) {
-          if ((acceleration.status === 'completed' || acceleration.status === 'completed_provisional') && acceleration.pools.includes(acceleration.minedByPoolUniqueId)) {
-            const boostCost = acceleration.boostCost || acceleration.bidBoost;
-            acceleration.acceleratedFeeRate = Math.max(acceleration.effectiveFee, acceleration.effectiveFee + boostCost) / acceleration.effectiveVsize;
-            acceleration.boost = boostCost;
-            this.tx.acceleratedAt = acceleration.added;
-            this.accelerationInfo = acceleration;
+          if (acceleration.status === 'completed' || acceleration.status === 'completed_provisional') {
+            if (acceleration.pools.includes(acceleration.minedByPoolUniqueId)) {
+              const boostCost = acceleration.boostCost || acceleration.bidBoost;
+              acceleration.acceleratedFeeRate = Math.max(acceleration.effectiveFee, acceleration.effectiveFee + boostCost) / acceleration.effectiveVsize;
+              acceleration.boost = boostCost;
+              this.tx.acceleratedAt = acceleration.added;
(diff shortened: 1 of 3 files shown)


frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -358,7 +358,7 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
       }),
     ).subscribe((accelerationHistory) => {
       for (const acceleration of accelerationHistory) {
-        if (acceleration.txid === this.txId && (acceleration.status === 'completed' || acceleration.status === 'completed_provisional')) {
```
natsoni, line 361: Haven't tested yet, but I think this might cause the out of band fee to still display the fee delta after the transaction gets mined by a non participating pool (because then, `accelerationInfo` is not defined upon confirmation)

natsoni:
Following my comment, maybe a solution is to move the check that the pool that mined the transaction is a partner before setting `acceleration.acceleratedFeeRate`:
```javascript
if (acceleration.txid === this.txId && (acceleration.status === 'completed' || acceleration.status === 'completed_provisional')) {
  const boostCost = acceleration.boostCost || acceleration.bidBoost;
  if (acceleration.pools.includes(acceleration.minedByPoolUniqueId)) {
    acceleration.acceleratedFeeRate = Math.max(acceleration.effectiveFee, acceleration.effectiveFee + boostCost) / acceleration.effectiveVsize;
  }
  acceleration.boost = boostCost;
  this.tx.acceleratedAt = acceleration.added;
  this.accelerationInfo = acceleration;
  this.waitingForAccelerationInfo = false;
  this.setIsAccelerated();
}
```

and checking for `accelerationInfo.acceleratedFeeRate` instead of only `accelerationInfo` in the template will fix the issue?

frontend/src/app/components/transaction/transaction.component.ts
mononaut: > might cause the out of band fee to still display the fee delta

display the fee delta where?

mononaut: ah, I see.

in that case, I'd rather fix the root cause (either clear the `tx.feeDelta` once we know a transaction was not accelerated by a participating pool, or tighten up the template logic)

natsoni:
Tested ACK @ [72bb92d]

This works well at hiding the effective fee rate row on accelerated transactions that got mined by non participating pools with 0 bid boost.

I can still see a few marginal issues that can happen when a tracked acceleration is mined with a 0 bid boost:
- if it is mined by a participating mining pool (Antpool in my local env), the fee delta does not disappear from the fee row.

https://github.com/user-attachments/assets/a3df218e-ec06-49cf-98e6-790677826915

- if it is mined by a non participating mining pool (Spiderpool in my local env), the fee delta does disappear from the fee row but the fiat value in the fee row still includes the fee delta. 
- 
https://github.com/user-attachments/assets/0c8ba158-7756-4d41-abb4-c761ef16b5c8

These issues are fixed by clearing `tx.feeDelta` once we know a transaction was not accelerated by a participating pool.

mononaut:
LGTM

PR #5450, opened 2024-08-09
Don't show "effective fee rate" on accelerated tx

natsoni:
Try to address the issue on acceleration detection discussed in #5449


natsoni:
Closing because fixed by #5449

PR #5451, opened 2024-08-10
optimize processNewBlocks

mononaut:
Optimizes the `processNewBlocks` function, which currently dominates the main mempool loop.

Most of the changes are micro-optimizations to reduce array/object allocations, or to avoid attempting to look up missing txids in the main mempool dictionary (in theory an O(1) operation, but in practice the constant factor is large in this specific case)


softsimon (approved):
tested ACK @ [367ee68fe0]

PR #5452, opened 2024-08-10
Implement v1 audit in tx audit API

mononaut:
Updates the new `/api/v1/block/:hash/tx/:txid/audit` endpoint to correctly handle v1 audits.

Before, "added & prioritized" transactions show correctly as `Not seen in Mempool` and `Prioritized` when interpreting a cached v1 block audit (e.g. after clicking through to the transaction from a block page), but incorrectly show `Seen in Mempool` when loading the audit status from the tx audit endpoint (e.g, after refreshing the transaction page).

https://mempool.space/tx/b0b03d5496f65244f2461ccfd7a9616dae430665fafdab60a107df013ba63618
Screenshot 2024-08-10 at 10 01 35 PM
Screenshot 2024-08-10 at 10 01 46 PM

After, the correct audit tags are displayed in both cases.

### DIFF
--- a/backend/src/repositories/BlocksAuditsRepository.ts
+++ b/backend/src/repositories/BlocksAuditsRepository.ts
@@ -132,11 +132,12 @@ class BlocksAuditRepositories {
             firstSeen = tx.time;
           }
         });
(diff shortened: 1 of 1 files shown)


softsimon (approved):
utACK

PR #5462, opened 2024-08-18
Fix accelerated arrow not appearing

natsoni:
Screenshot 2024-08-18 at 14 14 06

This console error happens when viewing unconfirmed accelerated transactions, resulting in the accelerating arrow not displaying:

Screenshot 2024-08-18 at 14 13 58


softsimon (approved):
utACK

PR #5463, opened 2024-08-18
Backend config generator script

1ma:
Counterpart to the frontend's [update-config.js](https://github.com/mempool/mempool/blob/master/frontend/update-config.js)/[generate-config.js](https://github.com/mempool/mempool/blob/master/frontend/generate-config.js) scripts, but for building the backend configuration file.

I've rewritten the script to not depend on any new dependency, and added a couple special `SOURCE_CONFIG_FILE` and `DEST_CONFIG_FILE` arguments to be able to specify custom source and destination files.

### DIFF
--- a/backend/README.md
+++ b/backend/README.md
@@ -106,6 +106,25 @@ In particular, make sure:
   - "esplora" if you're using [mempool/electrs](https://github.com/mempool/electrs)
   - "none" if you're not using any Electrum Server
 
+You can use the `generate-config.js` script to define your configuration in a non-interactive way.
+Basic usage:
+
+```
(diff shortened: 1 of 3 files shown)


knorrium:
Thanks for your contribution.

Can you make it work with config entries that are arrays (`MEMPOOL.EXTERNAL_ASSETS`, `ESPLORA.FALLBACK` and `REPLICATION.SERVERS`)?

1ma:
Array arguments should work already (see README example). The idea is that the cast function will try to JSON.parse the argument, and if it decodes to an Array object it will use it has that.

However I only tested it in bash. LMK if you hit any problem in another kind of shell.

1ma:
Rebased.

Despite the pipeline failure the PR already contains the CLA agreement, I guess the bot must be looking for it in master.

PR #5469, opened 2024-08-20
[mining] fix pools updater only running at start

nymkappa:
Looks like the pools updater was actually never running except upon restart of the mempool backend, since [March 3rd 2023](https://github.com/mempool/mempool/commit/d76d14253aece830caa9ac3ce7c8ef14012406c3). We never noticed because of course we have frequent backend restart during prod upgrade.

I've made the update frequency configurable (in seconds). Default is set to one week. Upon error, we reduce the next attempt delay to 10 minutes vs 24h currently.


nymkappa:
@softsimon can you review please :)

softsimon (approved):
tested ACK @ [0c87a4e7f6]

PR #5470, opened 2024-08-20
Show one decimal digit on fees estimation

natsoni:
This PR updates the recommended fees endpoint to return values with one decimal digit precision, instead of integers.

Screenshot 2024-08-20 at 17 29 46

For now the frontend has been updated to always display the .1 precision. Maybe we could add logic to only display the .1 precision if one of the two conditions are met:
- fee rate is below 10 sat/vB
- the fee market is very flat (meaning High priority and Medium priority have the same integer value or something similar)


softsimon:
I don't think this is a good idea. Because the fee box is what people use to enter in their wallets, which is always integer. It should be rounded up. 

What people requested and make sense is 1 decimal on the mempool blocks, and mined blocks.. Because the `3-3` on every block although it's somewhere around `2.3-2.6`

wiz:
I think this PR would go nicely with a new `/api/v2/fees/recommended` API endpoint that returns decimals instead of integers - would only want to show the decimals on the home page if it's less than 10 s/vB tho

natsoni:
> I don't think this is a good idea to make the fee estimations technical.

> I think this PR would go nicely with a new `/api/v2/fees/recommended` API endpoint 

Implemented wiz' suggestion, so you guys can decide if it's a good idea or not 😁
- Created new API version `api/v2` and proxied to backend on dev and prod environments
- Added `/api/v2/fees/recommended` to get decimal fee recommendations, which is used by the websocket and the dashboard where fees above 10 sat/vB are rounded up

wiz:
note to self, I need to add some nginx configuration in `/production/nginx/location-api.conf` for the `/api/v2` routes

natsoni:
I thought about that but with only 1 decimal it looks more like an exponent: 
Screenshot 2024-08-29 at 11 52 58

with two decimals for the lolz: 
Screenshot 2024-08-29 at 11 54 05

Another option is to keep the decimal down:
Screenshot 2024-08-29 at 11 38 17

softsimon:
Again.. I see now reason why the fee estimation numbers should contain decimals. They should be rounded up making sure people pay slightly more than required. Who enters a sat/vB fee with decimals when making a transaction? The fee estimation UX should be human readable. But the fee API could use decimals to let any GUI do their own rounding with the numbers.

wiz:
@softsimon any objection to closing this? seems like it could be re-approached later on

wiz:
@mononaut now that we've shipped the precision fees recommendation API do we want to re-approach this? it still looks ugly but maybe you have some fresh new ideas?

PR #5471, opened 2024-08-22
Avoid brief display of accelerator checkout on already accelerated txs

natsoni:
Replaces #5423

When browsing to an accelerated transaction, the accelerator checkout panel being visible until acceleration data is finally loaded can be confusing. This PR makes sure that the accelerator checkout is displayed only once we know that the transaction hasn't already been accelerated.

Before: 

https://github.com/user-attachments/assets/02877bf4-49c7-4dc0-95ad-169f579941a5

After: 

https://github.com/user-attachments/assets/ed4646d5-8494-43a3-bd6b-d177cba383c9


softsimon (approved):
tested ACK @ [9457032a]

PR #5472, opened 2024-08-22
respect json Accept header in API error responses

mononaut:
Upgrades all of our API error responses to use a new error handler, which checks the request `Accept` header and sends the error in the appropriate format (only supports json & plain text for now).


softsimon (approved):
tested ACK @ [75316e60]

PR #5477, opened 2024-08-26
[docs] update READMEs to newer node version

mononaut:
Update recommended/required node versions from 16.10 to 20.x


softsimon (approved):
utACK

Issue #5046, opened 2024-05-08
RBF Trees don't always combine, subtree left unconfirmed until expiry


kiwidream:
### Description

In some cases there are RBF trees which do not properly combine to make a 'full tree'. When this happens, one tree is left permanently unmined / unconfirmed and hangs around until the RBF tree expiry comes into effect (24H in default config).

In our testing we had two instances which share an electrs backend and the same bitcoind instance. One instance successfully combines everything, and the other has two distinct RBF trees.

The two instances don't have any difference in config.

#### Version

a66920c51e2d5407f1508ecd8476f8b0ebde69a5

### Steps to reproduce

Not easy to intentionally reproduce, but can be observed in a running mempool instance on mainnet

### Expected behaviour

RBF trees should always combine such that at least one TX will be confirmed

### Actual behaviour

RBF trees can sometimes be split, and one tree is never mined

### Screenshots

<img width="831" alt="Screenshot 2024-05-08 at 12 24 31 PM" src="https://github.com/mempool/mempool/assets/1680799/ed8f374a-6172-4eac-b2a0-9a01c235aa6a">

PR #5479, opened 2024-08-27
RBF tracking fixes

mononaut:
_resolves #5046_

This PR fixes a couple of long-standing issues with the RBF feature:

#### "Delayed" replacements

Some replacements were not detected because they did not occur atomically within a single mempool update, leading to "split" RBF trees/timelines as described in issue #5046.

The original implementation of RBF detection assumed that "replacements" must always occur within a single mempool polling update.

However, it appears that sometimes a replaced transaction can be evicted from the mempool in one update, but the replacement not actually received until later. (I'm not sure if this is expected behavior of `getrawmempool`, or if it's due to esplora's mempool polling implementation).

This PR adds a trailing buffer of transactions removed from the mempool in the prior 10 updates, and uses it to detect these "delayed" RBF events.

#### Inconsistent RBF tree IDs

Some parts of the codebase identified RBF trees using the txid of the first leaf transaction, while others assumed that trees were identified by the txid of the *root* transaction. This could cause some inconsistencies, especially when loading the RBF cache on startup.

This PR standardizes on the latter approach, using the txid of the *root* transaction everywhere.

#### Misc fixes, comments & refactors

The PR clarifies the data structure invariants required to maintain consistency in the RBF module, and refactors a bit to ensure these are upheld (especially while importing/exporting RBF data to the cache).

This should result in fewer "broken" RBF trees.


softsimon (approved):
tested ACK @ [fc5af24b68]

PR #5481, opened 2024-08-27
Reset acceleration flow state when leaving transaction

natsoni:
Using the search bar after accelerating a transaction to display another transaction will cause the accelerator checkout to be hidden while it should be visible by default. See: 

https://github.com/user-attachments/assets/edd29ef5-381a-4647-ad98-36bdd078e2f2

This is fixed by reinitialising `accelerationFlowCompleted` when leaving the transaction.


mononaut (approved):
utACK

PR #5482, opened 2024-08-27
Hide accelerator panel if tx gets accelerated on another session

natsoni:
Currently the accelerator checkout is not hidden on a transaction that gets externally accelerated: 

https://github.com/user-attachments/assets/c3ad9fc3-4dd7-4ea6-9cfe-d033390fdbfa

With this PR: 

https://github.com/user-attachments/assets/e241de29-bc86-4cf3-8db8-3201f3b7d79d

### DIFF
--- a/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
+++ b/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
@@ -196,9 +196,11 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
     if (changes.scrollEvent && this.scrollEvent) {
       this.scrollToElement('acceleratePreviewAnchor', 'start');
     }
-    if (changes.accelerating) {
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
```diff
@@ -197,8 +197,12 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
       this.scrollToElement('acceleratePreviewAnchor', 'start');
     }
     if (changes.accelerating) {
-      if ((this.step === 'processing' || this.step === 'paid') && this.accelerating) {
-        this.moveToStep('success');
+      if (this.accelerating) {
```
softsimon, line 200: `changes.accelerating && this.accelerating` would remove one if-statement

softsimon (approved):
utACK

PR #5483, opened 2024-08-27
Handle city-states in geolocation component

natsoni:
| Before  | After |
| Screenshot 2024-08-27 at 14 51 01  | Screenshot 2024-08-27 at 14 51 06  |


softsimon (approved):
utACK

PR #5486, opened 2024-08-28
Remove difficulty adjustment block offset

natsoni:
There is a one block offset in the difficulty adjustment widget:

Screenshot 2024-08-28 at 16 35 54

`Remaining: 0 blocks (In ~9 minutes)` looks weird, I would expect this instead (the green mempool block being the block where the adjustment occurs): 

Screenshot 2024-08-28 at 16 35 14

It was introduced in 39051e94e3b66850263682555e3d3d7b3302faa2: 

https://github.com/mempool/mempool/blob/39051e94e3b66850263682555e3d3d7b3302faa2/frontend/src/app/components/difficulty/difficulty.component.ts#L141

but I'm not sure why it was added initially. I'm guessing it has something to do with the new progress bar but I didn't notice anything wrong with it by removing the offset.


softsimon:
Again.. I see now reason why the fee estimation numbers should contain decimals. They should be rounded up making sure people pay slightly more than required. Who enters a sat/vB fee with decimals when making a transaction? The fee estimation UX should be human readable. But the fee API could use decimals to let any GUI do their own rounding with the numbers.

mononaut (commented):
It's a rather philosophical question: does the adjustment occur on the last block of the period (the moment at which miners start working on templates with the new difficulty), or the start of the next period (when the first such block is found)?

Personally I'd say the former (the difficulty adjusts on the last block, when the *target* changes), but yeah I agree `Remaining: 0 blocks (In ~9 minutes)` looks weird - the number-of-blocks-remaining and the time-until-adjustment countdowns should at least match up!

natsoni:
> Personally I'd say the former (the difficulty adjusts on the last block, when the _target_ changes)

Oh I see why the offset was introduced now. From a miner point of view it makes sense because the difficulty adjustment already occurred. 
But if we were to think this way, shouldn't we also do the same for halvings, i.e. consider last halving to occur at block 839,999, when the _block reward_ changes, instead of 840,000?

Screenshot 2024-08-30 at 12 31 24

I feel like the above makes more sense from a user point of view: we want the halving to happen on block 840,000 when we **see** the rewards divided by 2 (even if it kind of happened on the previous block for miners). IMO I would expect the same to happen with difficulty adjustments?

mononaut:
yeah that's a good point.

ok, let's ditch the `da.remainingBlocks - 1` and count down to the first block of the new epoch/era everywhere.

softsimon (approved):
utACK

PR #5487, opened 2024-08-29
Update about page enterprise sponsors

orangesurf:
Add logos for bitkey and leather

image

localhost_4200_about(iPhone 14 Pro Max)


orangesurf:
Also tweaks foundry logo viewbox to fix cropping

hunicus:
ACK fad39e0bea1fce20e22ad200cdc2b6fee0bd69fe

- Links open in new tabs to the correct websites
- Layouts look good (different widths, browsers, and physical devices tested)
  - As you noted, the Leather log hangs off on its own at certain widths but this already happens with Exodus on the live site.
- Icon sizes look fine

Screenshot from 2024-08-29 12-20-53

Screenshot from 2024-08-29 12-20-37

wiz:
can you use the other leather logo, it looks like you used a light mode logo - it should look like bitkey's logo which is the correct dark mode theme

wiz:
> the Leather log hangs off on its own at certain widths but this already happens with Exodus 

Yeah please clean the logos up to all line up

362792617-867dbec9-988d-40d4-beaa-3ac67cf000aa
362792609-de936282-af5c-41e9-b23e-dc05da7c9200

hunicus:
> please clean up the logos up to all line up

That's a strange way of lining up the icons. If you just look at the logos and ignore the circles then all your logos will be different sizes and look worse IMO.

orangesurf:
Switch to alternate logo

image

The current page design forces a circular crop on all images. All circular logos are the same size, the non circular logos are as large as they can be without getting cropped. 

image

wiz (approved):
utACK

PR #5494, opened 2024-08-31
Allow OP_0 in multisig scripts

vostrnad:
Because the opcode `0x00` is decoded as `OP_0` and not `OP_PUSHNUM_0`, multisig scripts with zero keys and/or signatures are currently not labeled as such, even though they are perfectly valid. They are of course an extremely uncommon edge case and not very useful, but I think it's still better to label them than not.

Example transaction: [c3e384db67470346df163a2fa50024d674ef1b3e75aa97ec6534d806c82fee7e](https://mempool.space/tx/c3e384db67470346df163a2fa50024d674ef1b3e75aa97ec6534d806c82fee7e)

0of0

### DIFF
--- a/backend/src/utils/bitcoin-script.ts
+++ b/backend/src/utils/bitcoin-script.ts
@@ -158,7 +158,7 @@ export function parseMultisigScript(script: string): void | { m: number, n: numb
   if (!opN) {
     return;
   }
-  if (!opN.startsWith('OP_PUSHNUM_')) {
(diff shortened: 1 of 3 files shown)


mononaut:
more like a no-sig than a multi-sig :P

but concept ACK, I'll test & review properly soon.

mononaut (approved):
Tested ACK @ [e44f30d7]

PR #5499, opened 2024-09-05
api-docs-data.ts:replace slushpool with antpool data

RandyMcMillan:


RandyMcMillan:
i will push the release soon.

Issue #65, opened 2020-03-23
Use incoming tx weight per second to predict the future


softsimon:
Transactions ETA are now estimated based on blocks in queue * 10, which doesn't give an accurate time prediction due to constantly incoming transactions.

Make an algorithm that takes tx weight/second and their fees into account to predict the state of the future mempool for both transaction ETA and fee estimation.

wiz:
As a temporary fix to improve fee estimation, how about we just average the low + medium + high estimates together for a new blended estimate rate?

wiz:
@natsoni can you take a look at this?

orangesurf:
# Block confirmation time estimator

Given a Bitcoin transaction at a known position in the mempool, estimate how long until it receives its first confirmation, accounting for the probabilistic nature of block arrival and competition from higher-fee transactions.

## Parameters

| Variable | Description | Default |
|----------|-------------|---------|
| `block_interval` | Mean time between blocks (minutes) | 10 |
| `block_position` | How many blocks deep your transaction is in the mempool (e.g. 3 means three blocks must be mined for your tx to be included) | 3 |
| `confidence` | Probability threshold (0–1) | 0.95 |
| `incoming_rate` | Rate of new transactions entering the mempool ahead of yours (vB/s) | 500 |
| `max_block_rate` | Maximum block capacity as a rate: 1,000,000 vB / 600s | 1666.6 |

## Simple model

Blocks arrive as a Poisson process. If your transaction sits in the 3rd mempool block, three blocks need to be mined before it confirms. We want the shortest wait time where we're confident enough blocks have arrived.

This assumes every block fully advances your transaction's position. In practice new transactions paying higher fees arrive between blocks and consume capacity that would otherwise clear the queue ahead of you.

## Block discount factor

Each block has ~1,000,000 vB of capacity. During the `block_interval` between blocks, new higher-fee transactions consume `incoming_rate × block_interval × 60` vB of that space. The fraction of each block's capacity that actually advances your position is:

$$\text{discount} = 1 - \frac{\text{incomingRate}}{\text{maxBlockRate}}$$

When `discount = 1`, no new transactions enter ahead of you and each block brings you 1MvB closer to confirmation. When `discount → 0`, new transactions fill blocks as fast as they are produced and you make no progress.

This stretches the apparent block interval:

$$\text{effectiveInterval} = \frac{\text{blockInterval}}{\text{discount}}$$

## Finding the minimum wait time

[...]

PR #5501, opened 2024-09-07
Consider incoming transactions flow in ETA calculation

natsoni:
This PR is a first draft for implementing #65.

### Context
The ETA shown on the transaction page is always underestimated because it doesn't account for the flow of incoming transactions. For example, buried transactions that have been waiting for days, yet our estimate still suggests they will likely be confirmed in less than an hour. We also display a very optimistic ETA of ~10 minutes on transactions that are very likely to be pushed to the next mempool block.

### The algorithm 
Pseudo-code of an ETA algorithm that takes into account the estimated incoming transactions rate:

```js
blocksUntilMined = 0;
mined = false;

while (!mined) {
  vsizeBeforeTx = vsizeBeforeTx + vsizePerSecond * avgBlockTime - blockVSize
  if (vsizeBeforeTx < 0) { // Means that our tx fits in expected next block
    mined = true;
  }
  blocksUntilMined++;
}

eta = blocksUntilMined * avgBlockTime;
```
 
- `vsizeBeforeTx` is the total vsize in front of the transaction of interest
- `vsizePerSecond` is the estimated incoming transactions rate. If it's greater than the blockchain throughput, the loop never ends. In practice we cap `vsizePerSecond < 0.95 * throughput`.
- if `vsizePerSecond = 0` the algorithm is equivalent to the current one 

### What the PR does
The above algorithm needs a few things that are implemented in this PR. 
- Store the fee rate of incoming transactions similarly to the mempool statistics (vsize per fee rate steps)
Screenshot 2024-09-06 at 01 10 22

- Estimate the future incoming transaction rate. This first draft simply takes the median of the last 10 minutes as proof of concept

- Apply the above algorithm using `vsizePerSecond` computed in the previous step

TODO: 
- [ ] Find the best way of predicting the future incoming transaction rate 
- [ ] Apply the algorithm to other parts of the code like on the accelerator checkout and the pizza tracker


frontend/src/app/services/eta.service.ts
```diff
@@ -143,7 +145,28 @@ export class EtaService {
...
+      // This loop will always terminate because we cap vsizePerSecond to 0.95 * maxCapacity
+      while (!mined) {
+        vsize = vsize + vsizePerSecond * da.adjustedTimeAvg / 1000 - this.stateService.blockVSize;
+        if (vsize + tx.weight / 8 < 0) { // Means that our tx fits in expected next block
+          mined = true;
+        }
+        blocksUntilMined++;
+      }
+
+      const wait = blocksUntilMined * da.adjustedTimeAvg;
```
mononaut, line 168: if possible, it would be nice to use an analytic solution for this.

maybe something like:
```typescript
let blocksUntilMined = Infinity;
const vsizeAhead = mempoolPosition.vsize + (this.stateService.blockVSize * mempoolPosition.block);
const incomingVsizePerBlock = vsizePerSecond * da.adjustedTimeAvg / 1000;
const vsizeConsumedPerBlock = incomingVsizePerBlock - this.stateService.blockVSize;
if (vsizeConsumedPerBlock > 0) {
  blocksUntilMined = Math.ceil(vsizeAhead / vsizeConsumedPerBlock);
}
```

frontend/src/app/services/eta.service.ts
```diff
@@ -143,7 +145,28 @@ export class EtaService {
...
+      );
+
+      // Count the number of blocks until we expect this tx to be mined
+      let blocksUntilMined = 0;
+      let mined = false;
+      let vsize = mempoolPosition.vsize + this.stateService.blockVSize * mempoolPosition.block;
+      // This loop will always terminate because we cap vsizePerSecond to 0.95 * maxCapacity
+      while (!mined) {
+        vsize = vsize + vsizePerSecond * da.adjustedTimeAvg / 1000 - this.stateService.blockVSize;
+        if (vsize + tx.weight / 8 < 0) { // Means that our tx fits in expected next block
```
mononaut, line 162: why `tx.weight / 8`?

mononaut (commented):
tentative concept ACK

I spent some time last year exploring a similar idea with the hope of improving our fee estimations, and eventually came to the conclusion that predicting the future is hard :P

A few thoughts:

a) For ETAs, what we're actually interested in is the *net change* in vsize at each fee rate, rather than the raw incoming rate.

b) As we get closer to the top of the mempool, past vbyte inflows at specific fee rates become less and less predictive of the future (since most people use mempool-based fee estimation now, and increase their fee rates in response to congestion).

Ideally I think we'd track and project inflows at different percentiles of the mempool. My intuition is that this better captures the behavior of more price-insensitive users - e.g. some people will always pay whatever is required to get into the next block. Unfortunately that becomes intractable pretty quickly.

c) It would be good to have a framework for testing & evaluating our ETA projections, so we can at least tell if we're moving in the right direction.

natsoni:
> a) For ETAs, what we're actually interested in is the _net change_ in vsize at each fee rate, rather than the gross incoming rate.

Right now we simply compute the sum of the vsizes at each fee rate above the transaction of interest. Are you suggesting to not sum it up?

> c) It would be good to have a framework for testing & evaluating our ETA projections, so we can at least tell if we're moving in the right direction.

I'm working on a simple script that selects random transactions and tracks both legacy and new ETAs until it gets mined. But it needs to run quite a long time and process a lot of transactions to reduce mempool conditions variance.

backend/src/api/mempool.ts
```diff
@@ -270,6 +276,7 @@ class Mempool {
             this.vBytesPerSecondArray.push({
               unixTime: new Date().getTime(),
               vSize: transaction.vsize,
+              effectiveFeePerVsize: transaction.effectiveFeePerVsize
```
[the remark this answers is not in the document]
natsoni: You're right, actually I think the only case where the `effectiveFeePerVsize` is different from the raw fee rate is for high sigops transactions where `effectiveFeePerVsize` is the adjusted fee rate

frontend/src/app/services/eta.service.ts
natsoni: Yes, that's better than the loop for sure 😁

frontend/src/app/services/eta.service.ts
natsoni: It seems `mempoolPosition.vsize` is the sum of the vsizes of the transactions ahead, plus half of the transaction vsize itself. 
So to check that the transaction fits in a block with `mempoolPosition.vsize` ahead, we check that `mempoolPosition.vsize` plus the other half of the transaction vsize `(tx.weight / 4) / 2` is lower than `blockVsize`... it's probably nitpicking since this value is at most 50 kvB in our mempool 😅

mononaut: ah ok, that makes sense

mononaut:
> > a) For ETAs, what we're actually interested in is the net change in vsize at each fee rate, rather than the gross incoming rate.

> Right now we simply compute the sum of the vsizes at each fee rate above the transaction of interest. Are you suggesting to not sum it up?

I just mean we should subtract any replaced/evicted transactions from the totals

mononaut:
> > c) It would be good to have a framework for testing & evaluating our ETA projections, so we can at least tell if we're moving in the right direction.

> I'm working on a simple script that selects random transactions and tracks both legacy and new ETAs until it gets mined. But it needs to run quite a long time and process a lot of transactions to reduce mempool conditions variance.

Awesome.

I was wondering if we could backtest using the historic data we've already collected, since we've been recording first seen transaction times for a while now.

Obviously we don't have the incoming transaction data broken down by fee rate, but maybe we could approximate that by looking at how the mempool composition changes in our existing statistics?

But I suppose we might as well just start recording the actual data and start testing once we have enough.

PR #5507, opened 2024-09-10
[faucet] show unverified warning if no email provided

nymkappa:
Show a warning when user logged in with Twitter, but no email was provided. In such case, the account is not marked as `verified` unless the user verify his account with a valid email. 

Currently in the faucet, we show no warning of such thing and account are automatically disabled if user does not verify his account, which is horrible UX

Screenshot 2024-09-10 at 12 04 53


wiz (approved):

wiz:
utACK

PR #5508, opened 2024-09-10
stratum job visualization

mononaut:
This draft PR adds a new `/stratum` page with a simple visualization of (the similarity between) the latest stratum jobs received from different mining pools.

stratumjobs

---

### Visualization

The visualization consists of a table with one row per job from each pool.

The `merkle_branches` section displays one cell per branch, in order from left to right (i.e. the leftmost cell represents the first non-coinbase transaction in the block template, and the rightmost cell represents the branch immediately below the merkle root of the template).

The cells are colored by interpreting the first 6 hex characters of the branch as an HTML color code, so that identical branches have identical colors.

To make it easier to see similarities between templates from different pools, the jobs are sorted by merkle branch with a dendrogram drawn on top (the tree of white lines).

The white lines of the dendrogram branch out at the last common merkle branch between pools.

---

### Data

The feature requires a websocket connection to a stratum-relay server, which parses and forwards the latest jobs from different pools. (see the config section).

The backend caches and serves those jobs to its own websocket clients via a new websocket subscription `{ 'track-stratum': pool }`, where `pool` may either be a numeric pool id (to receive new jobs from only that pool), or the string 'all' to subscribe to all new jobs.

New jobs are returned in the "stratumJobs" field of the websocket response, with the following schema:
```typescript
{
  "stratumJobs": {
    [unique_pool_id]: {
      pool: number; // unique_pool_id
      height: number;
      coinbase: string; // the entire coinbase transaction, in hex
      scriptsig: string; // the scriptsig of the coinbase input, in hex
      reward: number; // total output value of the coinbase, in sats
      jobId: string;
      extraNonce: string;
      extraNonce2Size: number;
      prevHash: string;
      coinbase1: string;
      coinbase2: string;
      merkleBranches: string[];
      version: string;
      bits: string;
      time: string; // block header time field, in hex
      timestamp: number; // block header time field, in unix seconds
      cleanJobs: boolean;
      received: number; // time the template was received, in unix seconds
    }
  }
}
```

---

#### Backend

The PR adds a new "STRATUM" section to the backend config (defaults to disabled):
```json
"STRATUM": {
  "ENABLED": true,
  "API": "http://127.0.0.1:1234"
},
```
which can be used to enable the feature and specify a stratum-relay server to connect to.

#### Frontend

The PR adds a new "STRATUM_ENABLED" (default false) field to the frontend config.

PR #5509, opened 2024-09-10
mining pool next block info

mononaut:
_(builds on #5508)_

This PR uses the new stratum data subscriptions added in #5508 to display information about the current work template on the Pool page:

braiinsstratum

The section is only displayed if `STRATUM_ENABLED` is set to `true` in the frontend config, and stratum data is available over websocket for this pool.

The "expected" field shows when this pool is next expected to mine a block, considering current difficulty and share of hashrate.

The "prevhash" field links to the corresponding block page. The "merkle branches" link goes to the new `/stratum` page.

The merkle branches section itself isn't very informative yet since we're only displaying the colors, but I think it's helpful as a visual cue to connect this info to the main stratum page.


mononaut:
rebased

wiz (approved):
Tested ACK @ v3.1.0-dev [b254be2f4]

PR #5513, opened 2024-09-12
[accelerator] avoid duplicated accel request with double click

nymkappa:
Add a basic boolean flag to avoid sending multiple acceleration request by clicking multiple times on the payment button.

Backend already protects against this right now using the `accelerationUUID`, but I'll remove it in another PR in favor on that frontend only fix.

It is very easy to reproduce the issue by enabling the debugger throttling. By setting yourself in GPRS mode before requesting a pro acceleration, without this fix, you can effectively click on the `Pay` button as many times as you want.


mononaut (approved):
utACK

seems like a good solution, and the code looks fine.

PR #5525, opened 2024-09-13
Add utxo chart to address page

mononaut:
This PR adds a new UTXO chart to the address page:

Screenshot 2024-09-13 at 5 07 30 PM

The utxo data is fetched from the existing `/api/address/:address/utxo` / `/api/scripthash/:scripthash/utxo` endpoints.

To avoid wasting resources fetching incomplete data, the chart is disabled and hidden (and no API requests made) if an address has fewer than 2 or more than 500 utxos.

---

The chart is rendered using the CustomChart echarts type with a custom layout algorithm, since there isn't a built-in chart that does quite what I wanted.

The algorithm optimizes for nice layouts rather than efficiency, but should be fast enough for <=500 utxos.

Each UTXO is represented by a circle, with area proportional to value. The circles are greedily packed as close to the center of mass of the chart as possible, in descending order of value. The result is a ~roughly circular cluster of utxos with the largest in the middle, which is more or less maximally compact.

The layout is then scaled and centered to fit the chart area.

---

Clicking on a UTXO in the chart navigates to the transaction that created it, with the corresponding output highlighted.

### DIFF
--- a/frontend/src/app/components/address/address.component.html
+++ b/frontend/src/app/components/address/address.component.html
@@ -94,6 +94,20 @@
       </div>
     </ng-container>
 
+    <ng-container *ngIf="(stateService.backend$ | async) === 'esplora' && address && utxos && utxos.length > 2">
+      <br>
+      <div class="title-tx">
+        <h2 class="text-left" i18n="address.unspent-outputs">Unspent Outputs</h2>
+      </div>
+      <div class="box">
+        <div class="row">
+          <div class="col-md">
+            <app-utxo-graph [utxos]="utxos" left="80" />
+          </div>
+        </div>
+      </div>
+    </ng-container>
+
     <br>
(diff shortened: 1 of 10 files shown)


mononaut:
> * similar to how block overview uses different color tints depending on the age of transactions, it would be nice to adapt the color of circles based on the age of the outputs, since the output age is already available.

yup, that's the plan :) I just need to choose a nice color gradient and figure out the best scale.

> * sometimes, some outputs are hidden because they are relatively small, but it would be informative if they were still displayed. For example, on El Salvador’s address with a lot of 1 BTC outputs:

yeah good point, we could definitely allow slightly smaller outputs.

softsimon (approved):
Nathans suggestions sounds good .. for next PR .. :) this works great for now

tested ACK @ [a1968e01e5]

PR #5526, opened 2024-09-17
Fix mobile routing to tx push and test pages

natsoni:
This PR fixes an issue where browsing to https://mempool.space/tx/push or https://mempool.space/tx/test on mobile would show a pizza tracker page instead of the broacast / test transaction page.


softsimon (approved):
tested ACK @ [a76d6c29]

PR #5527, opened 2024-09-17
Pizza tracker: don't show ETA on replaced tx

natsoni:
| Before  | After |
| Screenshot 2024-09-17 at 14 58 04 | Screenshot 2024-09-17 at 14 57 57 |


softsimon (approved):
utACK

PR #5528, opened 2024-09-17
Show http error in pizza tracker

natsoni:
A wrong input pasted in URL does not raise any error on the pizza tracker:

| Before | After |
|  Screenshot 2024-09-17 at 15 17 41 | Screenshot 2024-09-17 at 15 17 27  |


mononaut (approved):
utACK

good catch

PR #5529, opened 2024-09-17
"Be your own explorer" on non official mempool instance

natsoni:
On non-official mempool instances: 

Screenshot 2024-09-17 at 15 51 44

Screenshot 2024-09-17 at 16 10 16


softsimon (approved):
tested ACK @ [0b29b61e9]

PR #5530, opened 2024-09-17
Fix off-by-one error in multi-pool eta calculation

mononaut:


softsimon (approved):
tested ACK @ [b64caf8f4]

Before and after
40 min seems correct, not 20 min for a 18% available hash rate.

Screenshot 2024-09-19 at 01 27 00

PR #5532, opened 2024-09-18
Fix accelerations list page navigation on first load

natsoni:
This PR fixes an issue where directly loading https://mempool.space/acceleration/list/10 from the URL would correctly display page 10, but the page selector stays stuck at page 1:

Screenshot 2024-09-18 at 16 59 55


softsimon (approved):
tested ACK @ [556313a6]

PR #5534, opened 2024-09-19
Only fetch 1m mining stats

natsoni:
This PR allows to only fetches 1 month mining stats on the transaction page instead of both 1w and 1m stats.


softsimon (approved):
tested ACK @ [bb06a66a0]

PR #5535, opened 2024-09-19
Fix ETA calculation error

natsoni:
This PR fixes the error occuring in ETA calculation that occured when a participating mining pool does not appear in the fetch mining statistics.


softsimon (approved):
tested ACK @ [68a58046]

PR #5536, opened 2024-09-19
Wrap pool logos in timeline tooltip

natsoni:
Screenshot 2024-09-19 at 16 09 54


softsimon (approved):
noticed a flicker that happens sometimes when hovering in and out .. but it seems like it was there before

tested ACK @ [c6b0e5ff0e]

PR #5539, opened 2024-09-20
Show miner name on block timeline

BitcoinMechanic:
This will display the name of a miner who creates their own templates. 
image


softsimon (changes requested):
It looks OK but we need to remove the "on" text label.

A text label may or may not work on all 30+ languages due to various reasons. Just the fact that it needs to be translated is a hurdle.

So try removing the label completely. On the block page it might look good adding / or @ or something. Like. Steve @ OCEAN

mononaut (approved):
Tested ACK @ [81ddce2]

softsimon (approved):
partial tested ACK

wiz:
Is there a testcase for this? i.e. please tell me a block height that has the extra data in it

BitcoinMechanic:
https://mempool.space/testnet/mining/pool/ocean

PR #5541, opened 2024-09-22
Update accelerating pie chart in real time

mononaut:
Fixes an issue where the accelerating pool pie chart would not update when the set of accelerating pools changed after page load.


softsimon (approved):
utACK

Issue #5544, opened 2024-09-23
Negative acceleration time


orangesurf:
Negative acceleration time should not be shown.

To address properly requires [storing block arrival times.](https://github.com/mempool/mempool/issues/5543)

Example of issue

https://mempool.space/tx/6f1ed999edf679edf20cc8b11feff696a478647435f2e3c9ee6d7a989d32ddfd

<img width="1161" alt="image" src="https://github.com/user-attachments/assets/3948aa2b-ccff-41bf-ab53-bdd800dbc7cb">

softsimon:
Basically fixed with https://github.com/mempool/mempool/pull/5545 There's a separate issue open for using "arrival time".

Issue #5543, opened 2024-09-23
Store block arrival times


orangesurf:
Previously we have manually extracted block arrival times from [debug logs](https://github.com/bitcoin-data/block-arrival-times/tree/main) for the mempool research empty block report, and limited block time auditing. 

We should store this in a db to enable
- More accurate acceleration timeline information (avoiding negative time errors) 
- Integrated block time auditing on the block audit page
- Further research applications

PR #5545, opened 2024-09-23
Don't show negative timespans on timeline

natsoni:
Partially addresses #5544

This PR avoids showing negative timespans on the acceleration timeline (example https://mempool.space/tx/6f1ed999edf679edf20cc8b11feff696a478647435f2e3c9ee6d7a989d32ddfd)

The issue should be definitely fixed once #5543 is implemented, by using first seen time instead of block time.


softsimon (approved):
tested ACK @ [2a9346f6]

PR #5547, opened 2024-09-24
Add testnet4 to install script

Emzy:
Add testnet4 as on option to the production install script. For Ubuntu and FreeBSD.


wiz (approved):
utACK

PR #5540, opened 2024-09-22
address utxo chart color by age & updates

mononaut:
Updates the new address UTXO chart with changes suggested in https://github.com/mempool/mempool/pull/5525#issuecomment-2354961855, and some other fixes and improvements:
 - UTXOs colored by age
     - gold for pending.
     - gradient from cyan to purple for confirmed UTXOs, on a logarithmic scale.
 - adds age/confirmation status to the tooltip.
 - updates the chart as new transactions arrive.
 - better amount labels.

https://github.com/user-attachments/assets/2fe9d25f-0fee-4c79-808f-d45803dca82f

Screenshot 2024-09-22 at 5 41 29 PM
Screenshot 2024-09-22 at 5 42 00 PM

There are some scary looking changes to the Time component, but it should mostly be superficial refactoring to allow us to use the time formatting functions outside of component templates (required for nice localized "X time ago" strings in the chart tooltip).

PR #5548, opened 2024-09-25
utxo chart optimization

mononaut:
_(builds on #5540)_

 - speeds up the utxo chart layout algorithm from $O(n^4)$ to $O(n)$.
 - flattens the layout's target shape into an ellipse, so that more of the available space is filled.
 - enables chart transition animations when the layout changes.

https://github.com/user-attachments/assets/e5f5973c-b4ec-4195-98d1-12219ff23904


softsimon (approved):
tested ACK @ [83b60941]

PR #5550, opened 2024-09-26
refactor miner name truncation

mononaut:
a light refactor of a few parts of the miner name tag feature:
 - moves the ASCII filtering into the backend to reduce code repetition.
 - simplifies the HTML templates.
 - changes the miner name truncation on the blockchain blocks to use CSS `text-overflow: ellipsis`, to take better advantage of the limited space available.
 - adds JS truncation to the names on the block and transaction pages.
     - limit of 16 characters, *including* any ellipsis.
     
The `text-overflow` CSS property isn't compatible with the baseline vertical alignment we were previously using on those badges, so for consistency this PR applies the overflow ellipsis to *all* badges, not just the new miner name ones.

This means some particularly long historical pool names are now truncated where they weren't before, although I think this is an improvement. To avoid changing the appearance of active pools with borderline length names like "Carbon Negative", the width limit is slightly more generous for existing pool names than it is for the new miner names. (145px vs 125px)

| Before | After |
|-|-|
| Screenshot 2024-09-26 at 5 23 03 PM | Screenshot 2024-09-26 at 5 23 10 PM |

### DIFF
--- a/backend/src/utils/bitcoin-script.ts
+++ b/backend/src/utils/bitcoin-script.ts
@@ -223,5 +223,5 @@ export function parseDATUMTemplateCreator(coinbaseRaw: string): string[] | null 
   let tagString = String.fromCharCode(...tags);
   tagString = tagString.replace('\x00', '');
 
-  return tagString.split('\x0f');
+  return tagString.split('\x0f').map((name) => name.replace(/[^a-zA-Z0-9 ]/g, ''));
 }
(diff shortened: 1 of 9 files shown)


softsimon (approved):
tested ACK @ [b29c4cf228]

Screenshot 2024-09-26 at 23 58 42

Screenshot 2024-09-26 at 23 59 02

PR #5551, opened 2024-09-26
fix utxo chart on-click navigation

mononaut:
restores the on-click behavior of the new UTXO chart, so that clicking on a UTXO navigates to the corresponding output on the transaction page.


wiz (approved):
utACK

PR #5552, opened 2024-09-27
export bitcoinsatoshis pipe module, allow custom class for first part

nymkappa:


softsimon (approved):
tested ACK @ [2d7316942]

PR #5554, opened 2024-09-27
fix acceleration history paging w/ undefined total

mononaut:
Fixes a bug where `getAllAccelerationHistory` would try to load infinite pages if the `X-Total-Count` header was missing or not parseable.


softsimon (approved):
utACK

PR #5555, opened 2024-09-27
Fix acceleration timeline tooltip on mobile

natsoni:
This PR improves the acceleration timeline tooltip UX on touchscreens (one had to keep their finger on the screen to view the tooltip). Now, it behaves like the block overview graph tooltip, where a single tap shows the tooltip, and tapping elsewhere hides it.

https://github.com/user-attachments/assets/ede182b3-ee3a-48b0-b9c7-37f8307155a2


softsimon (changes requested):
It seems to not be so consistent.

Often when clicking on the circle it responds but the tooltip is not displayed.

PR #5558, opened 2024-09-29
handle /utxos error on address page

mononaut:
Fixes a bug introduced with the new UTXO chart where a non-200 response from the `/utxos` API would error out the whole address page.

Screenshot 2024-09-29 at 9 39 36 AM

This could happen when an address doesn't currently have too many utxos (so that we do attempt to load them), but did have too many at some point in the past (so the API throws a `Too many unspent transaction outputs` error).

(see https://github.com/mempool/electrs/blob/055aba1e8d0d8f4cf7080014c89e3b3055e3d0b7/src/new_index/schema.rs#L790-L793)

Example (assuming default `utxos_limit`):
https://mempool.space/address/bc1qjr2zgtqq6z0q05xt48f8scf78udlnpp45a6gf5


softsimon (approved):
tested ACK @ [60adad8db]

PR #5562, opened 2024-10-04
Wallet page transactions

mononaut:
_(builds on #4831)_
_(requires https://github.com/mempool/electrs/pull/82)_

Follow up to #4831, using the draft `mempool/electrs` address group endpoints to restore the transactions list on the named wallet page.

- Adds a link to the `/wallet` page from the custom dashboard wallet transactions widget
- Restores the transactions list, including live websocket updates & infinite scrolling functionality.

#### Limitations
- We can't know how many transactions exist for a set of addresses without fetching them all (each individual `/address` call reports the total tx count for that address, but some transactions may involve multiple addresses), therefore we:
    - can't display the total number
    - need to make an extra request when scrolling to confirm we got everything.
- The page only recognizes *confirmed* transactions/balances/utxos. Partly to simplify the implementation, and partly to match the desired behavior of the corresponding custom dashboard widgets.

### DIFF
--- a/frontend/src/app/components/custom-dashboard/custom-dashboard.component.html
+++ b/frontend/src/app/components/custom-dashboard/custom-dashboard.component.html
@@ -281,9 +281,11 @@
           <div class="col" style="max-height: 410px" [style.order]="isMobile && widget.mobileOrder || 8">
             <div class="card">
               <div class="card-body">
-                <span class="title-link">
+                <a class="title-link mb-0" style="margin-top: -2px" href="" [routerLink]="['/wallet/' + widget.props.wallet | relativeUrl]">
                   <h5 class="card-title d-inline" i18n="dashboard.treasury-transactions">Treasury Transactions</h5>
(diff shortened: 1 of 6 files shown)


mononaut:
rebased

wiz:
please rebase

mononaut:
rebased on current master

mononaut:
rebased

wiz (approved):
utACK

NullandBlock861222:
Why are u all still trying to steal?

NullandBlock861222:
And why have u went and used the Unknown ID as part of ur gist Wiz I sent that to u to prove I had the info for the so called Unknown miner with in mempool space as well as Null ID text as well as the so called Golden Partial

NullandBlock861222:
[{ _id__.txt _Null_.txt](https://github.com/user-attachments/files/18432882/_id__.txt._Null_.txt)

PR #5563, opened 2024-10-05
Reset block audit on block navigation

natsoni:
This PR fixes an issue that sometimes happen when browsing through blocks, where previous block's health is still displayed while current audit is loading.


softsimon (approved):
utACK

PR #5566, opened 2024-10-05
Avoid briefly showing incorrect health value

natsoni:
Reverts the change made by #5563, and actually solves the issue: when quickly browsing through blocks, previous block's health is still displayed for a short time:

https://github.com/user-attachments/assets/5b87d6e7-3957-4104-b940-20187d2c92f1


softsimon (approved):
utACK

PR #5567, opened 2024-10-05
Store first seen time in block audit

natsoni:
Partially implements #5543

This PR introduces a new `first_seen` field to the block audit, which records the timestamp of when a block was first observed. The timestamp is extracted from the `Saw new header ...` entries in Core's `debug.log`, with support for both second and microsecond precision.

To enable this functionality:
- add `"DEBUG_LOG_PATH": "/path/to/bitcoin/debug.log"` to the backend config
- add `-logtimemicros=1` to `bitcoin.conf` for microsecond precision (optional)

I will make a follow-up PR to use the first seen data in the frontend later on when this has been running for a few days.


backend/src/repositories/BlocksAuditsRepository.ts
```diff
@@ -78,6 +78,7 @@ class BlocksAuditRepositories {
           blocks_audits.height,
           blocks_audits.hash as id,
           UNIX_TIMESTAMP(blocks_audits.time) as timestamp,
+          UNIX_TIMESTAMP(blocks_audits.first_seen) as firstSeen,
```
mononaut, line 81: `firstSeen` comes out as a string here. I think we're safe to convert to a javascript number before returning the final object, even with microsecond precision.

backend/src/repositories/BlocksAuditsRepository.ts
```diff
@@ -106,6 +108,10 @@ class BlocksAuditRepositories {
         rows[0].acceleratedTxs = JSON.parse(rows[0].acceleratedTxs);
         rows[0].template = JSON.parse(rows[0].template);
 
+        if (!rows[0].firstSeen) {
```
mononaut, line 111: since `firstSeen` is a string here, a null timestamp of "0.000000" is still truthy and never gets deleted.

mononaut (changes requested):
Tested @ [1a75e3e3]

A couple of minor issues in the comments, but otherwise looks good!

(1) I ended up with some very late `firstSeen` times on a bunch of recent blocks while Core was catching up after a restart.

I guess that's *technically* correct since those were the times my node first saw the block headers, but it might be worth applying some sanity checks to the timestamps we choose to index.

(2) It would be neat to backfill the firstSeen timestamps for older audits as far as we have data available in the `debug.log`, perhaps in a new indexing task. We can save that for another PR though.

mononaut:
also, are we sure we want to store the timestamp in the `blocks_audits` table rather than the regular blocks table?

softsimon:
> also, are we sure we want to store the timestamp in the `blocks_audits` table rather than the regular blocks table?

Isn't the long term plan to integrate the `block` table into electrs itself, which makes it a bad option to store dynamic data that can't be re-indexed from the blockchain. Or what's your take on this.

mononaut:
> > also, are we sure we want to store the timestamp in the `blocks_audits` table rather than the regular blocks table?
> 
> Isn't the long term plan to integrate the `block` table into electrs itself, which makes it a bad option to store dynamic data that can't be re-indexed from the blockchain, although we already store pool-Ids. Or what's your take on this.

yeah that's a good point. I was just thinking that this feature doesn't necessarily overlap with the rest of the block audit table

e.g. I could imagine wanting to index block first seen times even with audits disabled, or to index that data for blocks we didn't/couldn't audit, like those added during a reorg.

natsoni:
> also, are we sure we want to store the timestamp in the `blocks_audits` table rather than the regular blocks table?

Right, the rationale for storing the timestamp in`blocks_audit` is that it's kind of an audit feature, but as you said one might want to store first seen data without having audits enabled. Also storing first seen times for reorg'd blocks is a nice use case, so I do think it makes sense to move the first seen field to the `blocks` table.

natsoni:
> (1) I ended up with some very late firstSeen times on a bunch of recent blocks while Core was catching up after a restart.
> I guess that's technically correct since those were the times my node first saw the block headers, but it might be worth applying some sanity checks to the timestamps we choose to index.

Good point, but how strict should we be with the sanitizing? Even if we require the first seen time to be at most 1-2 hours after the block timestamp, we could end up with some false positives. But with such a loose interval I’m wondering if the sanitizing would be useful at all.

mononaut:
fair enough. in that case let's index the data as it appears in the logs, and we can always clean it up on the client side later.

natsoni:
Added `first_seen` to the `blocks` table as datetime(6)

backend/src/api/blocks.ts
```diff
@@ -362,6 +362,14 @@ class Blocks {
           extras.expectedWeight = auditScore.expectedWeight;
         }
       }
+
+      extras.firstSeen = null;
+      if (config.CORE_RPC.DEBUG_LOG_PATH) {
+        const blockFirstSeen = await blocksRepository.$getFirstSeenTime(block.id);
```
mononaut, line 368: will a firstSeen time have been saved to the DB yet, when we make this lookup?

backend/src/mempool.interfaces.ts
```diff
@@ -320,6 +320,7 @@ export interface BlockExtension {
   segwitTotalSize: number;
   segwitTotalWeight: number;
   header: string;
+  firstSeen: string | null;
```
mononaut, line 323: I'd prefer `firstSeen` to be returned as a number, in unix seconds, for consistency with all of our other datetime fields (and to reduce ambiguity with timezones etc)

mononaut (commented):
looks good! just a few small issues left

backend/src/api/blocks.ts
natsoni: Good catch, this is a remaining of me trying to debug while blocks were cached in redis 😅

mononaut (approved):
Tested ACK @ [198d79f1]

softsimon (approved):
tested ACK @ [198d79f1492]

PR #5569, opened 2024-10-05
Add option to display runestones and inscriptions metadata

natsoni:
This PR adds basic, optional support for displaying runestones and inscriptions details within a mempool instance.

*****************

- show runestone message:

https://github.com/user-attachments/assets/48d01daa-8959-4ea2-b327-907f70a9c398

- show inscription metadata:

https://github.com/user-attachments/assets/605d1f0d-ad31-4c7a-b2da-2d10d2458680

- for text inscriptions smaller than 150 bytes, the content is rendered (we probably want to add some sanitizing here)

https://github.com/user-attachments/assets/950e640e-809d-4623-aee7-2b41c2ece0ca

No ord-related queries will be made unless the user specifically clicks on the 'Runestone' or 'Inscription' button.

**********************

This PR needs a `ord` server running with `index_runes` flag enabled.

The mempool backend config should include the following, assuming the ord server runs on port 4321:

```json
"ORD": {
  "ENABLED": true,
  "API": "http://localhost:4321"
}
```

The frontend config needs the flag `"ORD": true`.


mononaut:
tentative concept ACK.

I'll review the actual code in more depth later, but before we get too carried away with the implementation it might be worth discussing our end goals and the tradeoffs of different approaches:

---

it seems to me there are four main options for getting and serving inscription/rune data:
 1. Call out to an external API
 2. Run Ord
 3. Build our own indexer
 4. Decode transaction data on-the-fly

calling out to an API wouldn't be very cypherpunk of us, and building our own would be a huge waste of effort, unless we're only interested in a very limited subset of data that wouldn't justify running a full Ord node.

so we can probably rule out options 1 and 3.

then the question is whether we expect to need anything that we can't easily get without Ord.

---

off the top of my head, that probably includes:
- anything involving ordinal sats (except the creation of "rare" sats)
- rendering (some) recursive media inscriptions
	- (several of the recursive endpoints reference stuff by ordinal sat numbers or things that depend on sat numbers, which require an index)
- inscription parent/child provenance
- inscription numbers (different from inscription ids)
- inscription transfers/trades
- non-edict rune transfers/trades
- current locations of inscriptions and runes
- anything involving metaprotocol rules
 	- e.g. whether a Rune/BRC-20/etc mint or transfer is invalid or unsuccessful
 	
---

stuff we *can* do without an index
- display some inscription creation metadata (content type/size/etc)
- render simple inscriptions (anything without sat-based recursion)
- identify & display runestone messages, including:
	- info about rune etchings and mint attempts
	- edict-based rune transfers
	- rune names & etching inscriptions
	
---

stuff we *can't* do with Ord alone (unless it's changed recently)
- get inscription/rune data for unconfirmed transactions
- track non-rune metaprotocol token balances (for BRC-20s etc)

---

so far the features from this PR seem achievable without Ord (and for unconfirmed transactions, maybe *can't* be done with Ord alone).

but obviously if we plan to implement any of the features that do strictly require Ord in future, then we may as well start using it now.

mononaut:
sorry for the essay :P

natsoni:
Removed the need for the ord server: parsing of inscriptions and runestones is done client-side

softsimon (approved):
Very cool to see everything client side with existing. Subtle and informative.

tested ACK @ [8b6db768]

I'll not merge before @mononaut review and final discussions on if this is the correct approach and shippable with no frontend setting.

softsimon (approved):
tested ACK @ [65f080d5]

frontend/src/app/shared/ord/inscription.utils.ts
hans-crypto, line 1: [](url)

hans-crypto:
Regarding https://github.com/mempool/mempool/blob/d31c2665eea4b287a85f4f8127e3be6782459944/frontend/src/app/shared/ord/inscription.utils.ts#L1

Please add a proper copyright notice, because you copied "substantial portions of the Software".
> MIT License
>
> Copyright (c) 2024 HAUS HOPPE
>
> |...]
>
> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

You are also invited to add the `ordpool-parser` as a dependency. This way, you won’t have to address issues or bugs that are still present in your current adaptation:

```
npm install ordpool-parser
```

This package has zero dependencies and should work without any additional configuration.

PR #5571, opened 2024-10-06
Fix block page spinner loader

natsoni:
https://github.com/user-attachments/assets/e451bb98-2af6-4dfd-b695-62caa37b68a3


softsimon (approved):
tested ACK @ [6fb68203bc]

PR #5573, opened 2024-10-06
never show a utxo chart with missing data

mononaut:
Fixes a bug with the address page where a utxo chart could be displayed with partial data (recieved via websocket some time after page load) even if the initial `/utxos` request failed or was skipped due to too many or too few utxos.

e.g:
mempool space_address_bc1qmcdp5999hswqmdpkzk93kf788xj8sn5g7qj3gp

This PR disables the utxo chart for the rest of the page visit if we didn't obtain the initial `/utxos` state.


softsimon (approved):
utACK

PR #5577, opened 2024-10-08
Fix rune premine amount

natsoni:
Fixes an issue with premined runes by taking into account the rune's decimal number into amounts calculation


softsimon (approved):
utACK

PR #5578, opened 2024-10-09
Fix wrong timespan on acceleration timeline

natsoni:
Fixes the `55 years` issue on the acceleration timeline


softsimon (approved):
utACK

PR #5579, opened 2024-10-09
Use adjusted block time for difficulty and ETA calculation

natsoni:
It seems that difficulty widget and acceleration estimates are not using the adjusted block time using sliding window that was implemented in #4680: 

Screenshot 2024-10-09 at 16 50 57

Screenshot 2024-10-09 at 17 00 39

Using adjusted block time: 

Screenshot 2024-10-09 at 17 11 31

Screenshot 2024-10-09 at 17 10 47


mononaut (commented):
yes, good catch, we should be using the adjusted figure for ETA projections.

but `timeAvg` is correct for the difficulty widget, since that field is supposed to display the actual average time between blocks so far this epoch.

softsimon (approved):
tested ACK @ [24a76cafa]

PR #5581, opened 2024-10-09
Fix timespan on acceleration timeline

natsoni:
Second try at fixing the displaying of timespans on acceleration timeline


softsimon (approved):
tested ACK

Issue #5495, opened 2024-09-03
Add package broadcaster


glozow:
Add a package broadcasting page similar to https://mempool.space/tx/push, calling Bitcoin Core `submitpackage` RPC. It could take a list of hex transactions (which is what `submitpackage` accepts), or it could contain 2 fields for parent hex and child hex.

It seems #3859 tried to do this, but the RPC was not available in a released version at the time. It is available now (v26.0 and later). Additionally, the submitted packages may propagate soon!

Assuming adoption, from v28.0, users can submit a 1-parent-1-child CPFPing package and, even if the parent is below mempool minimum feerate, it will be accepted and relayed. The child's fees can even be used to replace stuff that the parent conflicts with. TRUC packages can even have a 0-fee parent (i.e. be below min relay feerate). See for example https://mempool.space/testnet4/tx/8e3f38bf6854dd3c358be8d4f9a40a6dccc50de49616125d27af9fdbe65287eb

I think there are 2 ways you could do this:

(1) Have 1 field, allowing a list of hex transactions (which is what `submitpackage` accepts). This is the more future-proof way to do it, but may be confusing since what `submitpackage` allows is different from what p2p transaction relay supports. TLDR while `submitpackage` works for multi-parent-1-child package, the p2p bit only works for 1p1c right now. In the future, we hope to enable both to support generic packages.

If you want to give users a warning about the limitations, this kind of text could be appropriate: "Warning: this feature is experimental and still a work in progress. Some features may be unavailable and transactions may not necessarily be accepted by miners and the rest of the network. See Bitcoin Core's [documentation](https://github.com/bitcoin/bitcoin/blob/master/doc/policy/packages.md) for more details." And maybe link to your docs about "there's no such thing as The mempool".

(2) Have 2 fields, one for parent hex and one for child hex. This would require updating later if/when more package support is available, but is more closely aligned with what the feature can support now.

glozow:
Just tested it, looks like it works :)

PR #5585, opened 2024-10-12
Add package broadcaster to tx push page

natsoni:
Implements #5495

This PR adds a `POST /api/txs/package` endpoint for non-esplora backends to use the new core rpc `submitpackage`. Note that this RPC needs to be added to the mempool/electrs API before we can use this on prod. 

Broadcast of a 1p1c with 0 fee parent `791e9f9e98563d99e1fa511c166d233e4d2789140555112345a532180e0aa195`

https://github.com/user-attachments/assets/0f85dd10-7c7f-4d0f-8d2e-d49797c85028

If only one transaction is passed, `sendrawtransaction` rpc is called instead: 

https://github.com/user-attachments/assets/021ccc68-9874-47f5-89df-82e68e82f91a

Non allowed package: 

https://github.com/user-attachments/assets/31b0d623-53da-4433-a6c7-5b3693806321


mononaut:
looks good!

I was imagining this would be a separate page rather than a second form on the existing `/tx/push`.

Would that complicate the implementation at all?

softsimon (approved):
concept ACK

I think it should be on the same page

PR #5586, opened 2024-10-13
Include optional seconds in search bar date

natsoni:
https://github.com/user-attachments/assets/bcfa0011-91e4-4028-aa26-cf1c9db204ae


softsimon (approved):
tested ACK @ [071e9b6c2]

PR #5599, opened 2024-10-17
multi-block views

mononaut:
Adds new secret pages at `/view/mempool-blocks` and `/view/blocks` to render arbitrary numbers of mempool or mined blocks simultaneously.

By default the pages will try to fill the screen with 320x320px blocks with 4px padding, or specific aspects of the layout can be configured via query params.

mempool
Screenshot 2024-10-17 at 9 48 30 AM
mined
Screenshot 2024-10-17 at 10 06 15 AM

To make this work, the PR includes a fairly deep rewrite/refactor of the block visualization components & shaders to allow rendering multiple `BlockScene`s on a single webgl canvas.

The PR also adds backend support for a new websocket subscription `track-mempool-blocks`, which accepts an array of mempool-block indices and then sends delta-encoded updates for multiple projected blocks.

To support more than the default 8 mempool blocks, `config.MEMPOOL.MEMPOOL_BLOCKS_AMOUNT` should be increased in the backend config (in theory this could make the mempool loop a little slower, as it requires applying the full GBT algorithm to more of the mempool, but in practice that's rarely the most significant part of the loop).


mononaut:
rebased on master

wiz (changes requested):
Tested but GBT loop can get quite slow, it needs some more work in terms of optimization - let's revisit this in a bit

PR #5600, opened 2024-10-17
Fix input/output overflow in transaction list

natsoni:
This PR fixes a CSS issue in the transaction list tables where the `large` class, was not being applied correctly if other rows in the table did not also have the class.

Before: 
Screenshot 2024-10-17 at 16 34 19

After:
Screenshot 2024-10-17 at 16 34 22


softsimon (approved):
tested ACK @ [67750bd]

PR #5611, opened 2024-10-24
[demo] better? demo mode

nymkappa:
Enable with eg: [https://mempool.space?next=/acceleration](https://mempool.space?next=/acceleration)

### DIFF
--- a/frontend/src/app/shared/common.utils.ts
+++ b/frontend/src/app/shared/common.utils.ts
@@ -239,7 +239,7 @@ export function handleDemoRedirect(route: ActivatedRoute, router: Router) {
   route.queryParams
     .subscribe(params => {
       if (params.next) {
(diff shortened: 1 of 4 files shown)


frontend/src/app/shared/common.utils.ts
```diff
@@ -234,6 +235,21 @@ export function sleep$(ms: number): Promise<void> {
   });
 }
 
+export function handleDemoRedirect(route: ActivatedRoute, router: Router) {
+  route.queryParams
+    .subscribe(params => {
+      if (params.next) {
+        const path = ['/', '/acceleration', '/mining', '/lightning', '/graphs'];
```
wiz, line 242:
```suggestion
        const path = ['/', '/acceleration', '/mining', '/lightning'];
```

wiz (approved):
We'll see if it's better or not

PR #5612, opened 2024-10-25
[core routes] /api/internal -> /api/v1/internal

nymkappa:


wiz (approved):
Tested ACK @ c950e3d0

PR #5613, opened 2024-10-25
fix purple pie chart with single pool

mononaut:
Before
Screenshot 2024-10-25 at 5 41 54 AM

After
Screenshot 2024-10-25 at 5 40 29 AM


wiz (approved):
utACK

PR #5617, opened 2024-10-29
Add missing fake_scripthash to the data filter

vostrnad:
The "Data" filter is currently missing "Fake scripthash", even though it's listed under the "Data" category.

mempoolscripthash1
mempoolscripthash2


mononaut (approved):
utACK

looks good, thanks!

PR #5621, opened 2024-10-30
Show tx first seen time with audit disabled

### LINKED ISSUE #5564: First seen data can be missing on mined transactions
First seen data is missing on mined transactions that are not expected in the block.

Examples:
https://mempool.space/tx/b881839c94d8476bb61e016f2c120be3c724032c9a14852b8e03162ef1eab42e
https://mempool.space/tx/c0fbe7fae2e98b34222ceff1e70b2bcba8cba39f67781a39a741eded1ebe257e

natsoni:
Fixes #5564

Currently, the first seen time is not displayed on the transaction page if: 
- the transaction is not in the audit template (it was not expected in the block) it is missing
- if the audit feature is disabled

We go around this by getting the first seen time from the block summary if the above situation happens. 

Block summaries need to be indexed (`MEMPOOL.BLOCKS_SUMMARIES_INDEXING = true` in the backend config) and a new frontend flag `MAINNET_TX_FIRST_SEEN_START_HEIGHT = <blockheight>` needs to be set in the frontend config.

It seems we started recording first seen times on block [838316](https://mempool.space/block/000000000000000000030cc61bbbb0252275acd421257649e409371cd48cf3ee) on prod, so I added the corresponding entry to the frontend mainnet prod config. 

### Test cases

- `267f88557ecd4c7fc2e7b91e38531fdcb2dd16fe5db58f18528586e12196ce17`

| Before  | After |
| Screenshot 2024-10-30 at 15 59 54  |  Screenshot 2024-10-30 at 16 01 48 |

- `c0fbe7fae2e98b34222ceff1e70b2bcba8cba39f67781a39a741eded1ebe257e`

Before (missing timeline because no first seen data):

Screenshot 2024-10-30 at 16 06 03

After: 

Screenshot 2024-10-30 at 16 06 11

This PR also ninja added a frontend flag that was missing `TESTNET4_BLOCK_AUDIT_START_HEIGHT`.


mononaut (approved):
Tested ACK @ [db321c3f]

PR #5625, opened 2024-11-02
Fix navigation to use relative paths

natsoni:
Mempool blocks at `/mempool-blocks/1` and blocks list at `/blocks/1` were broken on non-root network. This is fixed by using relative URL paths.


softsimon (approved):
tested ACK @ [b3aed2f5]

PR #5626, opened 2024-11-03
[accelerator] remove tx restriction for cashapp payments

nymkappa:

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -138,7 +138,6 @@
 
       <app-accelerate-checkout
         *ngIf="(da$ | async) as da;"
-        [cashappEnabled]="cashappEligible"
         [advancedEnabled]="true"
         [tx]="tx"
         [accelerating]="isAcceleration"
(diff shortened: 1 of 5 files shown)


frontend/src/app/components/transaction/transaction.component.html
```diff
@@ -137,7 +137,6 @@ <h2><a [href]="[ isMempoolSpaceBuild ? '/accelerator' : 'https://mempool.space/a
 
       <app-accelerate-checkout
         *ngIf="(da$ | async) as da;"
-        [cashappEnabled]="cashappEligible"
```
nymkappa, line 140: `cashappEnabled` default is `true` in `frontend/projects/mempool/src/app/components/accelerate-checkout/accelerate-checkout.component.ts`

mononaut:
do we want to remove the restriction in the `tracker` component as well? if so, I think we can get rid of the `cashappEnabled` input in `AccelerateCheckout` altogether.

wiz (approved):

vladislav957:
Yes

сб, 25 янв. 2025 г., 10:54 wiz ***@***.***>:

PR #5632, opened 2024-11-11
regularly ping acceleration websocket server

mononaut:
explicitly pings the acceleration websocket server every 30 seconds, and adds log prints for ping/pong messages.

if a pong is not received within ~5s, the connection is terminated and we'll retry in *another* 5s.


backend/src/api/services/acceleration.ts
```diff
@@ -267,12 +269,32 @@ class AccelerationApi {
...
+        this.ws.on('ping', () => {
+          logger.debug('received ping from acceleration websocket server');
+        });
+
+        this.ws.on('pong', () => {
+          logger.debug('received pong from acceleration websocket server');
+        });
+      } else {
+        if (this.lastPing > this.lastPong && Date.now() - this.lastPing > 5000) {
+          logger.warn('No pong received within 5 seconds, terminating connection');
```
wiz, line 289:
```suggestion
        if (this.lastPing > this.lastPong && Date.now() - this.lastPing > 10000) {
          logger.warn('No pong received within 10 seconds, terminating connection');
```

wiz (approved):

PR #5634, opened 2024-11-11
fix broken cpfp button

softsimon:
Was only broken on active accelerations with CPFPs


mononaut (approved):
utACK

looks good to me

PR #5635, opened 2024-11-12
Refactor clipboard component

natsoni:
This PR fixes the clipboard success message not displaying on Safari. We don't use `clipboard` and `tlite` dependencies anymore (they can be removed in a separate PR). The standard clipboard API is used instead, but it's only available on secure context. We fallback on a 'out of viewport hidden text area' hack on non-secure contexts. 

Tested on Chrome, Firefox, and Safari desktop and iOS


wiz (approved):
utACK

PR #5398, opened 2024-08-02
Fix accelerator skeleton loader

natsoni:
On dashboard: 

- Before: 

https://github.com/user-attachments/assets/218d59d7-f908-4f08-aa69-a4c168e3ac0a

- After: 

https://github.com/user-attachments/assets/c7f491a6-eb54-4040-8bfa-355406839231

On accelerations list: 

- Before: 

https://github.com/user-attachments/assets/bc26213e-cf1c-462d-99f3-f5b3b0e8fff4

- After: 

https://github.com/user-attachments/assets/ba1da78d-6b23-4738-b0d5-9ffdc4bda23f

Kept the spinner for clarity while switching pages in the accelerations list.

PR #5638, opened 2024-11-14
Fix acceleration list observable subscription logic

mononaut:
Fixes a bug that could cause the recent accelerations list widget not to update properly when new accelerations arrive.

#5398 moved the implicit async pipe `accelerationList$` subscription inside another element, conditional on a simple `nonEmptyAccelerations` variable.

This means that when `nonEmptyAccelerations` turns false (i.e. whenever the accelerationsList$ observable emits an empty list), that inner part of the template is removed, Angular automatically unsubscribes from the observable, and we get stuck in the empty state forever.

This PR moves the `accelerationList$` async pipe back to the top level so that the subscription is maintained for the full lifetime of the component.

### DIFF
--- a/frontend/src/app/components/acceleration/accelerations-list/accelerations-list.component.html
+++ b/frontend/src/app/components/acceleration/accelerations-list/accelerations-list.component.html
@@ -4,7 +4,7 @@
 
   <div class="clearfix"></div>
 
-  <div class="acceleration-list">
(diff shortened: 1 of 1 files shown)


natsoni (approved):
Tested ACK @ [fd7f3408]

When accelerations come in while on the acceleration dashboard, the recent accelerations list correctly fills up from being empty vs stays empty without this PR. 



After:

wiz (approved):

Issue #5639, opened 2024-11-14
[accelerator] display square payment error properly instead of generic "something went wrong"


nymkappa:
```
      } else if (e.errors?.length && ['GENERIC_DECLINE', 'PAYMENT_LIMIT_EXCEEDED', 'INVALID_CARD_DATA', 'INSUFFICIENT_FUNDS', 'TRANSACTION_LIMIT', 'PAN_FAILURE', 'INVALID_POSTAL_CODE', 'SOURCE_USED', 'INVALID_ACCOUNT', 'UNAUTHORIZED'].includes(e.errors[0].code)) {
        res.status(400).send(`Payment declined (${e.errors[0].code})`);
      } else {
```

PR #5640, opened 2024-11-14
[accelerator] display payment errors, auto reload after 10 secs instead of 3 secs

nymkappa:
Fixes https://github.com/mempool/mempool/issues/5639

Currently we always show a generic message if we fail to accelerate a tx, regardless of the reason. However the frontend knows if it's a payment issue or not, therefore I display the Square payment error when appropriate instead of the generic message.

Screenshot 2024-11-14 at 16 42 14


wiz (approved):
utACK

PR #5642, opened 2024-11-15
Add missing db indexes

mononaut:
Adds a bunch of new database indexes for columns that are used for filters/sorting/aggregation.


wiz (changes requested):
please don't have more than one alter statement per migration, otherwise it can timeout and fail and get into a broken state - @knorrium we should even have a test case for this

mononaut:
fixed

softsimon (approved):
tested ACK @ [d9f3611da34]

PR #5659, opened 2024-11-21
[internal] provide internal rest api to retreive btcusd price history

nymkappa:
I've added a new internal rest api to allow our services backend to retreive our complete btcusd price history in one go (needed for stats).


wiz (approved):
utACK

PR #5666, opened 2024-12-02
fix undefined mempool tx errors

mononaut:
adds some missing checks and guards to prevent undefined property accesses while processing mempool blocks in rare edge cases.


wiz (approved):
utACK

PR #5668, opened 2024-12-04
Fix tx frontend issues after acceleration cancellation

natsoni:
Fixes a few issues that occur when cancelling an acceleration: 

- ETA was loading forever 
- Fee delta was still being displayed in transaction details
- acceleration timeline was not removed upon cancellation

https://github.com/user-attachments/assets/79a30c13-7d39-47f3-83c9-4cdf13dca309

I will include some visualization of the cancellation in the timeline in a separate PR.


mononaut (approved):
Tested ACK @ [0310452d]

I could reliably reproduce the issues described without the PR, and confirmed that the fix resolves them without breaking anything else, afaict.

wiz:
Okay so I guess we need to ship v3.1 for Umbrel users which would contain this important bug fix to properly visualize canceled accelerations

PR #5669, opened 2024-12-04
Fix liquid database index

natsoni:
Fixes Liquid database migration that adds an index to the `balance` column in `federation_txos` table (there is no such column in this table)


softsimon (approved):
tested ACK @ [ebd4170d]

PR #5671, opened 2024-12-05
Canceled acceleration on timeline

natsoni:
Builds on #5668

Acceleration timeline on canceled accelerations: 

https://github.com/user-attachments/assets/cf687695-f1e0-420e-bc85-9e7cfc974732

If / when the canceled acceleration is confirmed: 

Screenshot 2024-12-05 at 13 05 04


wiz (approved):
utACK

wiz:
```
     AssertionError: Timed out retrying after 4000ms: Expected <div.spinner-border.text-light> not to exist in the DOM, but it was continuously found.
```
looks like a failing test, but will merge anyway

PR #5676, opened 2024-12-07
Fix fiat tick precision on address balance chart

mononaut:
Adds a new `sigfigs` boolean param to the `AmountShortenerPipe`, which formats numbers with a certain number of significant figures instead of a fixed number of decimal places, and uses it to fix the USD y-axis ticks on the address balance chart.

| Before | After |
|-|-|
| Screenshot 2024-12-07 at 8 32 45 PM | Screenshot 2024-12-07 at 8 31 47 PM |


wiz (approved):
utACK

PR #5680, opened 2024-12-09
Unify database schema for all backend types

### LINKED ISSUE #5142: [migration] unify all database migration regardless of the network
Currently the database migration script is filled with `isBitcoin` and other special conditions. This leads to inconsistent database migration between networks, making it much more prone to bugs we only spot in staging or prod.

We need to have one migration version that unifies all database schema.

natsoni:
Fixes #5142

Adds a migration to make database structure across all mempool backends the same.

Tested upgrading from version 93 as well as clean install on:
- mainnet
- liquid
- testnet4

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -7,7 +7,7 @@ import cpfpRepository from '../repositories/CpfpRepository';
 import { RowDataPacket } from 'mysql2';
 
 class DatabaseMigration {
-  private static currentVersion = 93;
+  private static currentVersion = 94;
   private queryTimeout = 3600_000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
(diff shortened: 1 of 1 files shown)


softsimon (approved):
tested ACK @ [2372d8cff]

Tried upgrading from master state on Liquid and Mainnet.
Also made a clean install migration on Liquid and Mainnet.
Diff checks out and the database state is now the same.

Would be good if @wiz or @mononaut also acknowledge that this is the correct approach before merging.

All new migrations from the PR must not contain the IfBitcoin or IfNetwork checks.

backend/src/api/database-migration.ts
```diff
@@ -801,6 +801,323 @@ class DatabaseMigration {
       `);
       await this.updateToSchemaVersion(93);
     }
+
+    // Unify database schema for all mempool netwoks
+    // versions above 94 should not use network-specific flags
+    if (databaseSchemaVersion < 94) {
+
+      if (!isBitcoin) {
+        // Apply all the bitcoin specific migrations to non-bitcoin networks: liquid, liquidtestnet and testnet4 (!)
```
mononaut, line 810: for obscure reasons, `config.MEMPOOL.NETWORK` ought be set to "testnet" in the backend for both legacy testnet and testnet4, so both should have been treated correctly as bitcoin networks.

mononaut (approved):
Tested ACK @ [2372d8cf]

Approach seems sensible.

I Tested on both bitcoin and liquid networks, and confirmed that the final DB schemas are identical afterwards.

PR #5679, opened 2024-12-09
update unfurler and build config

nymkappa:
cherry picked https://github.com/mempool/mempool/pull/5376/commits/79e2883ebe5ff02e7f1cf759ef3fb56e640bdb21 from PR https://github.com/mempool/mempool/pull/5376

PR #5681, opened 2024-12-10
wallet unfurler preview

mononaut:
_(builds on #5679)_

Adds an unfurler preview for named wallet pages:

meta-unfurl


wiz (approved):
utACK

PR #5702, opened 2024-12-22
fix duplicated response header

nymkappa:
Make sure we always `return` when calling the `handleError` function

### DIFF
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -229,6 +229,7 @@ class BitcoinRoutes {
       if (e instanceof Error && e instanceof Error && e.message && e.message.indexOf('No such mempool or blockchain transaction') > -1) {
         statusCode = 404;
         handleError(req, res, statusCode, 'No such mempool or blockchain transaction');
+        return;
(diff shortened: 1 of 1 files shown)


mononaut (approved):
Tested ACK @ [1e3926c8]

possibly a little overzealous - only `getTransaction`, `getRawTransaction` and `getTransactionStatus` are missing early returns after handling 404s, but the other `handleError` usages already return early when necessary.

if we're going to standardize on *always* explicitly returning `handleError`, then we should probably extend that convention to the other `.routes` files too (lightning, mining etc).

nymkappa:
> Tested ACK @ [[1e3926c](https://github.com/mempool/mempool/commit/1e3926c8df451696ab3694fda61178fa92f71fed)]
> 
> possibly a little overzealous - only `getTransaction`, `getRawTransaction` and `getTransactionStatus` were missing early returns after handling 404s, but the other `handleError` usages already return early when necessary.
> 
> if we're going to standardize on _always_ explicitly returning `handleError`, then we should probably extend that convention to the other `.routes` files too (lightning, mining etc).

You're right, let me rebase to only fix the actually issue and we can keep a potential refactor for another time

PR #5703, opened 2024-12-22
revert difficulty widget to true avg block time

mononaut:
Reverts the change to the difficulty widget made in #5579.


softsimon (approved):
utACK

PR #5704, opened 2024-12-22
[accelerator] improve SCA UX

mononaut:
In the accelerator checkout component, while the square payment & SCA modals are open, the PR:
 - disables pointer events in the checkout widget and grays out the interface
 - shows a loading spinner
 

https://github.com/user-attachments/assets/dbf815c7-2122-4faf-abe2-51a2c1c072fa


nymkappa:
Much better, thanks. I've noticed that for a split second between the time the modal closes and the time the tx gets accelerated, I can actually click on the Google Pay button again and go through the payment process again.

Also, a 🐵 could press tab and enter to click the Google Pay button as well even tho the UI is greyed out. Might wanna lock this too with a simple `return` in code while the current payment process is in progress. This also applies for the `go back button`

mononaut:
updated to fix the split second of interactivity between the two modals, and to fully disable the payment and "go back" buttons in javascript instead of only disabling pointer events.

nymkappa (approved):
Tested ACK @ f49152d09dc46a05d7bf122356bdbe2ed58dad66

wiz (approved):

PR #5706, opened 2024-12-22
Add websocket commands doc

natsoni:
This PR updates the docs' WebSocket section to mention available commands, following the same approach as the REST section:

Screenshot 2024-12-22 at 20 25 11


mononaut (approved):
Tested ACK @ [a5c67b5c]

Nice! this is a big improvement over the current websocket docs

PR #5713, opened 2025-01-06
clean up subscriptions & component references

mononaut:
Profiling revealed various memory issues involving the block overview component, which could prevent old copies of data used for the block visualization from being properly garbage collected after use:
 - missing rxjs unsubscriptions
     - where rxjs pipelines and subscription callbacks use arrow functions (or bound function expressions), those callbacks capture the lexical scope (including references to the component instance via `this`).
     - if we fail to unsubscribe in the `ngOnDestroy` handler, those subscription callbacks are never cleaned up and the references to the component instance are retained indefinitely, preventing the component object from being properly garbage collected after the end of its natural lifecycle.
 - `shareReplay`:
     - the default behavior of the `shareReplay` rxjs operator prevents the event source from being automatically unsubscribed, even if all of the downstream subscriptions are correctly cleaned up.
     - if the pipeline up to that point includes any arrow functions or bound function expression callbacks, those may be retained and prevent anything captured in their closures from being properly garbage collected as described above.
     - in those cases, the `refCount` parameter should be used to ensure the source is unsubscribed at the end of its lifetime, like `shareReplay({ bufferSize: N, refCount: true })`.
 - implicit block scene cleanup:
     - in various places we relied on the `BlockOverviewGraphComponent` being correctly garbage collected to clean up the memory intensive block visualization objects instead of proactively clearing this data at the end of its Angular lifecycle.
     
This PR attempts to resolve these issues by:
 - adding the missing rxjs subscriptions.
 - changing relevant uses of `shareReplay` to use `refCount`.
 - explicitly calling `destroy()` on the `BlockScene` and `FastVertexArray` instances in the `BlockOverviewGraphComponent.ngOnDestroy` handler to clean up that data.

---

### Testing

- click around between different pages with block visualizations (`/mempool-block/...`,`/block/...`, dashboard, acceleration dashboard) for a while.
- take a heap snapshot.
- verify that the appropriate number of `BlockOverviewGraphComponent` and/or `BlockScene` objects have been retained.

### DIFF
--- a/frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.ts
+++ b/frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.ts
@@ -46,6 +46,8 @@ export class AccelerationFeesGraphComponent implements OnInit, OnChanges, OnDest
 
   aggregatedHistory$: Observable<any>;
   statsSubscription: Subscription;
+  aggregatedHistorySubscription: Subscription;
+  fragmentSubscription: Subscription;
   isLoading = true;
   formatNumber = formatNumber;
   timespan = '';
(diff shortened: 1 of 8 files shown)


softsimon (changes requested):
getting a couple of new errors when switching between views.

```
Uncaught TypeError: Cannot read properties of null (reading 'dirty')
(block-overview-graph.component.ts:463:28)
```

```
Uncaught TypeError: Cannot read properties of null (reading 'fill')
 at FastVertexArray.clearData (fast-vertex-array.ts:69:15)
```

mononaut:
pushed a fix

softsimon (approved):
tested ACK @ [7740908]

no crashes and the heap doesn't seem to grow uncontrollable after several hours

Screenshot 2025-01-09 at 00 26 34

PR #5723, opened 2025-01-13
Fix unconfirmed badge on broken RBF txs

mononaut:
This PR fixes an edge case where replaced/evicted transactions can be labeled "unconfirmed" instead of "replaced" or "removed" if they get evicted from the mempool indirectly, making it unclear whether the transaction is still in the mempool or not.

### Testing

To reproduce (best on regtest):

- submit a parent transaction
- submit a child transaction
- RBF the child
- RBF the parent
- visit either of the child txids
- observe that the transaction is labeled "unconfirmed", despite being evicted from the mempool (due to the replaced parent)

Screenshot 2025-01-14 at 6 30 43 AM

With this PR, the transaction will be correctly labeled "removed" instead.

Screenshot 2025-01-14 at 6 31 47 AM

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -24,6 +24,7 @@
           [height]="tx?.status?.block_height"
           [replaced]="replaced"
           [removed]="this.rbfInfo?.mined && !this.tx?.status?.confirmed"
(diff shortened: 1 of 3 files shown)


softsimon (approved):
partially tested ACK @ [6a4aeaf7e]

PR #5724, opened 2025-01-14
Fix textarea keyboard navigation

natsoni:
This PR fixes left / right keyboard navigation in `textarea` elements (such as on the push transaction page).


softsimon (approved):
utACK

PR #5725, opened 2025-01-14
avoid creating incomplete RBF trees

mononaut:
Fixes a bug where partial RBF trees could be created in the case where all of the replaced transactions already belonged to different trees.

Screenshot 2025-01-14 at 6 36 48 AM

This can occur where two separate new transactions txB and txC each spend a different input of the same evicted transaction txA, in the same mempool update (or within a short period).

This PR fixes the bug by checking if all of the "replaced" transactions in an identified RBF event already belong to other trees before proceeding to create a new RBF tree.


softsimon (approved):
partially tested ACK @ [4dcbccd]

PR #5726, opened 2025-01-14
Fix time traveling balance charts

mononaut:
Balance charts rely on block timestamps for the time axis. However, since block timestamps are not reliable, consecutive blocks are not necessarily timestamped in chronological order.

Before:
Screenshot 2025-01-14 at 9 05 30 AM

Transactions must still be ordered in correct logical order (by block height & intra-block transaction index) to avoid negative or incorrect balances on the chart, so this PR simply adjusts any out-of-order timestamps to ensure they are always monotonically increasing.

After:
Screenshot 2025-01-14 at 9 05 59 AM


softsimon (approved):
tested ACK @ [cd02d8923]

PR #5728, opened 2025-01-16
Update node matrix to v20 and v22

knorrium:
First PR to test node compatibility on CI


knorrium:
@wiz I don't know why it's still checking for 21, maybe conflicting with the definition in master?

wiz:
Can you please also bump the versions in the install script and start scripts etc.

PR #5734, opened 2025-01-19
Update staging hosts

knorrium:


wiz (approved):

knorrium:
@wiz the e2e job was split into 3 (mempool, liquid and testnet4). Rerouting happens only on the testnet4 tests.

We can merge this mess now 🤣

PR #5665, opened 2024-11-29
[blocks] fetch list of block hash filtered by pools-v2.json sha

nymkappa:
Store used pools-v2.json hash into the blocks table
Add 3 new internal endpoints:
- `/internal/blocks/definition/list`: retrieve list of all distinct pools-v2.json hash present in the blocks table
- `/internal/blocks/definition/current`: retrieve latest pools-v2.json hash
- `/internal/blocks/:definitionHash`: retrieve list of block hash which were indexed using pools-v2.json :definitionHash. Note: this endpoint has not pagination at all therefore it can return hundreds thousands rows at once (which is why I've added it in the `/internal` namespace)

```
[17:53:22] nymkappa @ /Users/nymkappa $ curl localhost:8999/api/v1/internal/blocks/definition/current
32a2d23379609b9841f8c3eb0d12d06cecd1ce9d%
[17:54:03] nymkappa @ /Users/nymkappa $ curl localhost:8999/api/v1/internal/blocks/definition/list
["32a2d23379609b9841f8c3eb0d12d06cecd1ce9d","cd604489f715e255aed21ae22721475f56b1b067"]%
[17:54:17] nymkappa @ /Users/nymkappa $ curl localhost:8999/api/v1/internal/blocks/32a2d23379609b9841f8c3eb0d12d06cecd1ce9d
["00000000000000000002115c4cb1223b89284b0ee03abdd4c44b6af756d1962f"]%
```

@sha2fiddy

PR #5758, opened 2025-01-28
[indexing] make `blocks.definition_hash` nullable and exit early if appropriate

nymkappa:
Fix a bug introduced by https://github.com/mempool/mempool/pull/5665 where blocks indexing would get stuck forever if we're unable to set a valid `pools-v2.json` sha

The new column is now null-able, but we exit early in case the user actually want to index that data and the `pools-v2.json` sha cannot be retrieved.

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -7,7 +7,7 @@ import cpfpRepository from '../repositories/CpfpRepository';
 import { RowDataPacket } from 'mysql2';
 
 class DatabaseMigration {
-  private static currentVersion = 96;
+  private static currentVersion = 97;
   private queryTimeout = 3600_000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
@@ -1121,7 +1121,7 @@ class DatabaseMigration {
 
     // blocks pools-v2.json hash
     if (databaseSchemaVersion < 95) {
-      let poolJsonSha = 'f737d86571d190cf1a1a3cf5fd86b33ba9624254';
+      let poolJsonSha = 'f737d86571d190cf1a1a3cf5fd86b33ba9624254'; // https://github.com/mempool/mining-pools/commit/f737d86571d190cf1a1a3cf5fd86b33ba9624254
       const [poolJsonShaDb]: any[] = await DB.query(`SELECT string FROM state WHERE name = 'pools_json_sha'`);
       if (poolJsonShaDb?.length > 0) {
         poolJsonSha = poolJsonShaDb[0].string;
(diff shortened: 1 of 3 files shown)


backend/src/api/database-migration.ts
```diff
@@ -1130,6 +1130,17 @@ class DatabaseMigration {
...
     }
+
+    // Make definition_hash nullable
+    if (databaseSchemaVersion < 96) {
+      let poolJsonSha = '0138019f39e0f6033db576be15a4880a216c1dd0';
+      const [poolJsonShaDb]: any[] = await DB.query(`SELECT string FROM state WHERE name = 'pools_json_sha'`);
+      if (poolJsonShaDb?.length > 0) {
+        poolJsonSha = poolJsonShaDb[0].string;
+      }
+      await this.$executeQuery(`ALTER TABLE blocks ADD definition_hash varchar(255) NULL DEFAULT "${poolJsonSha}"`);
```
mononaut, line 1141: doesn't the `definition_hash` column already exist from the previous migration?

mononaut: I'm also slightly confused about the default value here

nymkappa: Yes sorry I've changed it to `MODIFY COLUMN`.
Default value is whatever is the current latest commit hash in https://github.com/mempool/mining-pools

mononaut: if we have (and set) a default, does it still need to be nullable? or is that used somewhere else?

nymkappa: Originally the nullable fix was to make sure the block indexing does not get stuck if the `pools-v2.json` sha cannot be fetched from github.

I've then added the early exit check to make sure we don't even try in such case _(no sha in db AND no sha from github, so a first run without any pool definition)_.
https://github.com/mempool/mempool/blob/b5d5231bc1745eb834a0d4fbfe6799043152bc14/backend/src/index.ts#L154-L158

So in theory I don't think the nullable is still needed but I'm thinking there might be a weird config combination where this could actually be needed so I left it to make sure it does get the main loop stuck.

nymkappa: Also maybe we should flip those two blocks actually to ensure we initialize `currentSha` before attempting to hit github

https://github.com/mempool/mempool/blob/f909d6bca52138131274b4dc3d1db28143514f3e/backend/src/tasks/pools-updater.ts#L48-L55

JSRAInmob:
I hereby accept the terms of the Contributor License Agreement in the
CONTRIBUTING.md file of the mempool/mempool git repository as of January
25, 2022.

Signed: @JSRAInmob
@JSRAInmob

mononaut:
rebased on current master

wiz (approved):

PR #5766, opened 2025-02-03
[accelerator] fix card on file confirmation button html

nymkappa:
The confirmation step was still using the old broken html, so I fixed it

Screenshot 2025-02-03 at 11 15 42


nymkappa:
Replaced by https://github.com/mempool/mempool/pull/5862

PR #5767, opened 2025-02-04
Fix for windows

cpuchainorg:
Replaced bash script for building gbt package to more readable TypeScript script ( because, we are typescript people though ).

This PR should fix windows support as a whole.


softsimon (changes requested):
We can't allow adding of a new NPM dep, `shx`.

PR #5769, opened 2025-02-04
detect and warn about address poisoning attacks

### LINKED ISSUE #5756: Address Poisoning Warning
Our [resident eagle eyed capuchin](https://x.com/mononautical/status/1883802593677435065) spotted that these addresses are not the same. They are in fact likely part of an address poisoning attack in which a user is tricked into sending bitcoin to the duster if they copy this address. 

We could help signal this to users of mempool.space, with a link to a new FAQ article on what address poisoning is and how to avoid falling victim to it. 

Image

We could take this opportunity to consider whether using visual separation for blocks of addresses would be useful, such as illustrated in the bitcoin design guide: https://bitcoin.design/guide/glossary/address/

Image

mononaut:
_(resolves #5756)_

This PR adds some basic protections against address poisoning attacks.

First, it implements a simple detection algorithm which looks for pairs of addresses within the same transaction with suspiciously matching prefixes and/or postfixes.

Then, for matches above a certain improbability threshold:
 - highlights the similarities and differences
 - adds inline warning icons with explanatory tooltips
 - adds a warning banner to the top of the transaction box
 
Once we add an FAQ entry for address poisoning, we can link to it from both the banner and warning icons.
 
Screenshot 2025-02-04 at 11 57 47 AM

Screenshot 2025-02-04 at 12 00 38 PM

This is performed at the level of the `TransactionList` component, so applies to both the transaction page and anywhere else we show lists of transactions (address, wallet, block pages etc).

Screenshot 2025-02-04 at 11 57 33 AM

---

#### Address poisoning detection algorithm
This is a simple, fast algorithm that counts matching sequences of characters at the start and end of addresses.

Within each sequence (prefix and postfix) we tolerate up to one edit (a substitution, insertion or deletion in either address), which makes the detection slightly more robust.

Common prefixes for each address type are ignored, since those are always expected to match (i.e. the human-readable part + witness version of a bech32(m) address, or the first character of a legacy base58check address).

Once we have a count of suspiciously matching characters, we assign a score based on the probability of that number of matches considering the size of the alphabet of the address encoding scheme used.

I have somewhat arbitrarily set the threshold for warnings such that less than 1 in every 10 million comparisons will generate a false positive, which corresponds to 4+ matching base58 characters or 5+ matching bech32(m) (excluding common prefixes).

This sounds like quite a high bar, but any lower threshold generates a surprising number of false positives in larger transactions due to the birthday paradox effect, and any higher misses some genuinely confusing matches.

We could potentially be a bit smarter here by e.g. considering visually similar characters or more general similarity metrics, but the tradeoff between sensitivity and false positive rate gets worse as we generalize away from the most obvious attacks.

#### Application

This implementation primarily focuses on visible outputs, since I assume attackers can't generate transactions spending inputs controlled by a victim's real address, and that for an address poisoning attack to be effective at least one of the addresses needs to be "above the fold" and immediately visible on the page. (although perhaps these assumptions don't hold true for all threat models?)

So for the first ~20 output addresses of each transaction, we check for confusingly similar addresses in each of the first ~20 inputs and other outputs.

On address and wallet pages, we additionally check those first 20 inputs and outputs against the page's relevant address(es).

### DIFF
--- a/frontend/src/app/components/transactions-list/transactions-list.component.html
+++ b/frontend/src/app/components/transactions-list/transactions-list.component.html
@@ -16,6 +16,11 @@
   <div class="header-bg box" [attr.data-cy]="'tx-' + i">
 
     <div *ngIf="errorUnblinded" class="error-unblinded">{{ errorUnblinded }}</div>
+    @if (similarityMatches.get(tx.txid)?.size) {
+      <div class="alert alert-mempool" role="alert">
+        <span i18n="transaction.poison.warning">Warning! This transaction involves deceptively similar addresses. It may be an address poisoning attack.</span>
+      </div>
+    }
     <div class="row">
       <div class="col">
         <table class="table table-fixed table-borderless smaller-text table-sm table-tx-vin">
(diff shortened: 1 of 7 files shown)


orangesurf (approved):
utACK

mononaut:
for testing, here's an example of a recent mass address poisoning attack:
https://mempool.space/tx/7e9fc6393f24e6d574bfce8c19aac20c1be3c5171a818afb9701483cd3eb5c30

follow any of the outputs forward, and you should see the warning & address highlighting.

PR #5771, opened 2025-02-04
Visual address hashes

mononaut:
_(builds on #5769)_

_**STATUS:** seeking input on concept_

This experimental draft PR implements an idea for adding simple visual hashes to address strings:

Screenshot 2025-02-04 at 3 42 02 PM

A visual hash (or identicon) is a simple graphic generated deterministically to allow humans to more easily identify and distinguish pieces of arbitrary data.

e.g. in the screenshot above, the pie-shaped visual hashes help to more easily see that the first three addresses are the same, and the last is different, without the user needing to compare random strings of characters.

identicons for addresses are a fairly common UX pattern on shitcoin chains where address reuse is more normalized, to improve recognition and protect against impersonation.

There are a ton of [existing identicon standards](https://github.com/drhus/awesome-identicons), but they all look terrible and most don't work at the small sizes required for input/output lists etc.

This draft PR implements an ultra simple, lower-entropy scheme that's more suitable for our use cases:
- hash the address string once with sha-256
- take the first 12 bytes
- split into 3-byte chunks
- interpret each chunk as an html hex color
- display the four colors as quarter-slices of a circle

The scheme could be made more aesthetically pleasing without losing too much entropy by using a perceptually uniform color-space and restricting the palette to a smaller range, at the cost of making the spec and implementation quite a bit more complex.

Since these simpler identicons compress addresses into a much smaller perceptual space they'll collide much more often, but should still be sufficient to aid casual address recognition, and to foil most address poisoning attacks (in combination with displaying the actual address strings).

Screenshot 2025-02-04 at 3 41 23 PM


wiz:
cool, I suppose if you take the idea to the maximum you would implement clustering analysis and assign identities as well

softsimon:
Concept ACK.

I think it's a good important addition.

Design wise I think going forward we should try again to switch to the monospace font (which I believe we already use in some instances) for all addresses display. Then the positioning of the visual hash and address ellipsis will be aligned and easier to read.

PR #5778, opened 2025-02-06
set content limit to 5mb

nymkappa:
This is needed to call `/api/v1/internal/bitcoin-core/decode-raw-transaction` with a large hex transaction

I've set it to 5mb but I'm not sure about this value, @mononaut do you have a suggestion? @wiz can this create some ops issue?


nymkappa:
> 5mb wouldn't be enough for the largest possible valid transactions, considering the 2x hex-encoding overhead, so I'd set it to at least 8mb

Rounded up to 10mb 🤣

mononaut (approved):
utACK

PR #5780, opened 2025-02-07
Fix left-to-right scrollable blockchain 

natsoni:
This PR is a small patch to fix the scrollable blockchain that is broken in left-to-right mode, see:

https://github.com/user-attachments/assets/63684fba-9815-425a-9960-3b5723ad4e4a


mononaut (approved):
Tested ACK @ [1779c672]

I tested with all combinations of RTL and LTR locales, chain direction and mobile vs desktop layout, and it seems to fix the bug without breaking any other scroll behavior.

PR #5791, opened 2025-02-20
Removing the (broken) tv view

softsimon:
The legacy TV view page has been broken for some time and it's not longer relevant.


wiz (approved):

wiz:
We should redirect /tv to the new clock view

PR #5797, opened 2025-02-28
Switch tagline in global footer to byoe

hunicus:
Displays "Be your own explorer" tagline in global footer everywhere regardless of `officialMempoolSpace` value.

Screenshot from 2025-02-27 19-13-56


wiz (commented):
Instead of changing this, perhaps we can add this either in the footer or somewhere else?

Be your own explorer(tm) with The Mempool Open Source Project(R)

hunicus:
Added to `/about` and footer FAQ list.

PR #5798, opened 2025-03-05
disabled ON UPDATE for blocks_audits time field

mononaut:
similar fix to #2743, but for the `blocks_audits` table


softsimon (approved):
tested ACK @ [c4e22a622]

PR #5799, opened 2025-03-05
PSBT support in transaction preview

natsoni:
builds on #5664

This PR adds support for visualizing PSBTs ([version 0](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki) for now) in the transaction preview page. From a base64 or hex serialized PSBT data, we use its global map and inputs map to:
- extract the raw unsigned transaction
- populate inputs’ prevouts when available
- if offline mode is disabled and some prevouts are still missing, fetch the missing ones as usual.
- for each input, fill up signatures on a best-effort basis:
  - check if the input has a finalized signature
  - if not, look for redeem/witness scripts and partial signatures

Support for verifying the presence of a signature in each input, and adapting the transaction's metrics (vsize/feerate) will be added in a subsequent pull request.

### DIFF
--- a/frontend/src/app/components/transaction/transaction-raw.component.html
+++ b/frontend/src/app/components/transaction/transaction-raw.component.html
@@ -6,7 +6,7 @@
 
     <form [formGroup]="pushTxForm" (submit)="decodeTransaction()" novalidate>
       <div class="mb-3">
-        <textarea formControlName="txRaw" class="form-control" rows="5" i18n-placeholder="transaction.hex" placeholder="Transaction hex"></textarea>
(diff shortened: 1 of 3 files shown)


softsimon (approved):
tested ACK @ [c01e11899c]

softsimon:
> After broadcast should it redirect to the tx page?

Probably should but there is also a delay from broadcast until it's processed through the mempool backend so it will display "waiting for transaction to appear" for a few seconds before showing the TX. But I guess that's fine. Perhaps it should redirect to the published transaction after 1-2 seconds.

PR #5800, opened 2025-03-05
Tapscript multisig parsing

natsoni:
This PR adds support for parsing and labelling multisig in taproot script path using `CHECKSIGADD`.

Test cases: 

https://mempool.space/tx/7393096d97bfee8660f4100ffd61874d62f9a65de9fb6acf740c4c386990ef73
https://mempool.space/tx/b10c0000004da5a9d1d9b4ae32e09f0b3e62d21a5cce5428d4ad714fb444eb5d
https://mempool.space/tx/2eb8dbaa346d4be4e82fe444c2f0be00654d8cfd8c4a9a61b11aeaab8c00b272

Screenshot 2025-03-05 at 15 35 51

### DIFF
--- a/frontend/src/app/shared/script.utils.ts
+++ b/frontend/src/app/shared/script.utils.ts
@@ -251,6 +251,11 @@ export function detectScriptTemplate(type: ScriptType, script_asm: string, witne
     return ScriptTemplates.multisig(multisig.m, multisig.n);
   }
 
+  const tapscriptMultisig = parseTapscriptMultisig(script_asm);
+  if (tapscriptMultisig) {
+    return ScriptTemplates.multisig(tapscriptMultisig.m, tapscriptMultisig.n);
+  }
+
   return;
 }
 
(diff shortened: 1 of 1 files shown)


frontend/src/app/shared/script.utils.ts
```diff
@@ -299,6 +304,62 @@ export function parseMultisigScript(script: string): undefined | { m: number, n:
...
 }
 
+export function parseTapscriptMultisig(script: string): undefined | { m: number, n: number } {
+  if (!script) {
+    return;
+  }
+
+  const ops = script.split(' ');
+  // At minimum, one pubkey group (3 tokens) + m push + final opcode = 5 tokens
+  if (ops.length < 5) return;
```
mononaut, line 314: should the minimum size be *two* pubkey groups? otherwise it's just a single sig

frontend/src/app/shared/script.utils.ts
```diff
@@ -299,6 +304,62 @@ export function parseMultisigScript(script: string): undefined | { m: number, n:
...
+
+  const finalOp = ops.pop();
+  if (finalOp !== 'OP_NUMEQUAL' && finalOp !== 'OP_GREATERTHANOREQUAL') {
+    return;
+  }
```
mononaut, line 319: are there any other commonly used constructions for enforcing the minimum number of valid signatures?

output descriptor and miniscript specifications only use `OP_NUMEQUAL`, but if we're going off-spec for custom multisigs maybe we also need to consider `OP_NUMEQUALVERIFY`, `OP_GREATERTHAN`, `OP_EQUAL(_VERIFY)` etc.

mononaut (approved):
Tested ACK @ [ad140dc6]

I suspect others will have *Opinions* about labeling taproot scripts as multisig due to the ambiguity and unclear scope (is every taproot output with both a multisig script and an enabled keypath actually at least a `1-or-m-of-n+1`? if there are multiple tapleaves, could `n` actually be arbitrarily larger than implied by the revealed script? what about non-standard or more complex multisig constructions? etc etc).

but personally I think it's still worth labeling standard taproot multisigs, and fine to say that the label is for the *specific revealed script* rather than the input in general.

frontend/src/app/shared/script.utils.ts
natsoni: Right, but since we also label these [weird 1-of-1 multisig](http://localhost:4200/tx/675b106c71f110ebca9e6435f49508a07b45d40d38e237d02b89e5506f3b136b), I thought we should do the same for taproot multisig as well?
Although, at least in these cases, there's an `OP_CHECKMULTISIG` that justifies the multisig label unlike in the taproot 1-of-1.

frontend/src/app/shared/script.utils.ts
natsoni: I included `OP_GREATERTHANOREQUAL` in the check just to catch that cool 998-of-999 multisig that used it. 😁 Haven't found any multisigs using other opcodes besides this one tho

natsoni:
> but personally I think it's still worth labeling standard taproot multisigs, and fine to say that the label is for the _specific revealed script_ rather than the input in general.

Yep, showing the multisig label on the taproot input is definitely worth it (implying that this particular input is concerned by the label). IMO the issue is when displaying labels the address page, in that case we associate the full output to this label which can be wrong. Maybe the best approach is to simply avoid showing the label on the address page if it's a P2TR.

frontend/src/app/shared/script.utils.ts
mononaut: yeah, I think the legacy version is less ambiguous because of the explicit multisig opcode, whereas this is plausibly a legitimate way to check a single signature in tapscript.

although maybe `<pubkey> OP_CHECKSIG 1 OP_NUMEQUAL` is weird enough that in practice you'd only see it as a result of `multi_a(1, ...) output descriptors 🤷

mononaut:
I think my original plan was to show a list of all known labels for P2TR addresses as the different script paths are revealed.

PR #5813, opened 2025-03-12
[accelerator] show square receipt if available

nymkappa:
Requires https://github.com/mempool/mempool.space/pull/1582

In order to always show the `success` state, I've tweaked the copy.

### Tab that made the payment

Screenshot 2025-03-12 at 15 16 25

### Tab that got front run

Screenshot 2025-03-12 at 15 16 21


mononaut (approved):
utACK @ [1121377c]

wiz:
Let's avoid having link text that says "click here", i.e instead of "click here to get a receipt" you should just say "download receipt" or something

PR #5815, opened 2025-03-12
Tx preview: trim input and redirect to tx page

natsoni:
- input is trimmed before parsing
- user is redirected to the tx page 2 seconds after broadcast

https://github.com/user-attachments/assets/6c78ba20-4541-457f-9382-fe963feece72


mononaut (approved):
utACK @ [17249838]

PR #5838, opened 2025-04-01
polish pool updates

mononaut:
Fixes some minor issues related to the automatic pool updates:
 - on failure, retry in 10 minutes (as the logs claim) instead of 10 minutes before the next scheduled check
 - refresh the in-memory and persistent recent block caches with the new data if anything changed

### DIFF
--- a/backend/src/tasks/pools-updater.ts
+++ b/backend/src/tasks/pools-updater.ts
@@ -99,7 +99,7 @@ class PoolsUpdater {
 
     } catch (e) {
       // fast-forward lastRun to 10 minutes before the next scheduled update
-      this.lastRun = now - (config.MEMPOOL.POOLS_UPDATE_DELAY - 600);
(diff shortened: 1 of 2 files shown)


backend/src/tasks/pools-updater.ts
```diff
@@ -98,7 +98,8 @@ class PoolsUpdater {
       logger.info(`Mining pools-v2.json (${githubSha}) import completed`, this.tag);
 
     } catch (e) {
-      this.lastRun = now - 600; // Try again in 10 minutes
+      // fast-forward lastRun to 10 minutes before the next scheduled update
+      this.lastRun = now - (config.MEMPOOL.POOLS_UPDATE_DELAY - 600);
```
nymkappa, line 102: We might want to clamp `config.MEMPOOL.POOLS_UPDATE_DELAY - 600` to a minimum value to avoid continuous attempt without delay

EG: `Math.max(config.MEMPOOL.POOLS_UPDATE_DELAY - 600, 600)`

wiz (approved):

PR #5848, opened 2025-04-03
Fix false positive address poisoning warnings on wallet page

mononaut:
The address poisoning warning implementation on the wallet page currently checks for similarity matches against all of a wallets addresses (even if they aren't involved in a particular transaction).

If a wallet has many addresses, this is almost guaranteed to produce a huge number of false positives due to the birthday paradox:

Screenshot 2025-04-03 at 7 50 40 AM

This PR mitigates the issue by limiting comparisons to addresses within the same transaction.


knorrium (approved):
utACK

wiz (approved):

PR #5856, opened 2025-04-06
Set automatic pool updates on by default

knorrium:
```
Apr  6 14:42:52 [147] WARN: <lightning> [PoolsUpdater] pools-v2.json is outdated, fetching latest from https://raw.githubusercontent.com/mempool/mining-pools/master/pools-v2.json over clearnet
Apr  6 14:42:53 [147] NOTICE: <lightning> Updating addresses and/or coinbase tags for Luxor mining pool.
Apr  6 14:42:54 [147] INFO: <lightning> [Mining] 201 blocks assigned to a new pool
Apr  6 14:42:54 [147] INFO: <lightning> [PoolsUpdater] Mining pools-v2.json (14edbf14be9ae39f6c60daeaef93dacb8b5aa502) import completed

```


nymkappa:
This impacts on-click installs right?

I can't even recall how self hosted instances were suppose to get updated definitions. AFAIK we turned this off by default because we don't want umbrel&co instances to get stuck into some re-indexing without the user explicitly asking for it (UX would be similar to random Windows updates lol).

Did we change our mind on this?

mononaut:
pool reindexing is efficient now, so it should be fine

mononaut (approved):
utACK

PR #5858, opened 2025-04-07
Switch hammer to pickaxe

hunicus:
## Current

Screenshot from 2025-04-07 15-51-06

## Proposed

Screenshot from 2025-04-07 15-50-59


softsimon:
The pick makes more sense than a hammer for mining, but doesn't it feel more "thinner" than the rest of the menu icons slightly breaking the consistency. Not sure. I'm fine with both.

hunicus:
Over time it would be nice to replace all the other icons with ones that appear more cohesive and less generic. For now it looks fine to me.

wiz (approved):
ok let's ship this for now and improve later

pedromvpg:
pickaxe-mining-icon_revised

PR #5860, opened 2025-04-07
portlandhodl/ckpool-install-support

russeree:
Updated installer that support CKSolo and fixes various issues building on FreeBSD


wiz (approved):
utACK

PR #5864, opened 2025-04-09
Fix network change bug on stratum pages

mononaut:
Fixes a bug where stratum data from mainnet would persist after changing networks


wiz (approved):
utACK

PR #5865, opened 2025-04-09
preview stratum coinbase transactions

mononaut:
This PR adds support for quickly previewing the coinbase transactions from stratum jobs in the new /tx/preview page.

https://github.com/user-attachments/assets/5d44066d-0ba1-4a22-84cd-12c9fb0da913

To support this, the PR also updates the transaction preview page:
 - allow transaction raw hex & offline mode preference to be provided in URL fragment params
 - some UI changes when the transaction is a coinbase
     - (disable broadcast button & "stored in your browser" alert message)
     - (different title)
 - proactively updating URL fragment params when new raw transactions are submitted and/or offline mode preference changes.

These changes have the nice side effect of allowing the user's current previewed transaction to persist across page reloads, and to allow sharing a transaction preview by URL.

### DIFF
--- a/frontend/src/app/components/pool/pool.component.html
+++ b/frontend/src/app/components/pool/pool.component.html
@@ -219,7 +219,12 @@
                   <table class="job-table table table-xs table-borderless table-fixed table-data">
                     <thead>
                       <tr>
-                        <th class="data-title clip text-center coinbase" i18n="latest-blocks.coinbasetag">Coinbase tag</th>
+                        <th class="data-title clip text-center coinbase" i18n="latest-blocks.coinbasetag">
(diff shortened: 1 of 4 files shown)


natsoni (approved):
Tested ACK @ [17cc2ded4ced3903fb5bd062fa8ee10fb2235ec0]

Very cool! I just fixed a small issue with how offline mode was handled in the fragments. 

I also enabled offline mode when previewing coinbase transactions from stratum jobs, so that we don't fetch cpfp unnecessarily, and the previewed coinbase is displayed instantly. 

Before:

https://github.com/user-attachments/assets/a5cb22f2-d631-4b95-a3ae-0b6126e6be64


After: 

https://github.com/user-attachments/assets/f6996cfc-c341-4ec8-bf52-41d73b8a1b36

PR #5875, opened 2025-04-14
Don't zoom when double click on amount button (mobile)

natsoni:
Some buttons in the app triggers a zoom on mobile devices when double clicked. See:

https://github.com/user-attachments/assets/a267d539-4f14-4443-9319-9728b6d434de

This PR disables this behavior by adding the css property `touch-action: manipulation` to button components.


wiz (approved):
utACK

PR #5876, opened 2025-04-14
Fix cropped difficulty series in hashrate chart

### LINKED ISSUE #5868: Hashrate/difficulty chart overflow issue
While https://github.com/mempool/mempool/pull/4824 made the hashrate and difficulty lines match nicely, it also broke the chart Y-axis in case those two lines diverge too much like on testnet4.

Also I'm not sure why there is a line without legend at the top of the chart at all time.

Image

natsoni:
Fixes #5868

If the difficulty exceeds the scaled y-axis maximum, update its y-axis to match the difficulty max value. This fixes the overflow while keeping the scaled axis when there is no overflow issues.

testnet4: 
| Before  | After |
| Screenshot 2025-04-14 at 17 45 49 | Screenshot 2025-04-14 at 17 44 57 |

### DIFF
--- a/frontend/src/app/components/hashrate-chart/hashrate-chart.component.ts
+++ b/frontend/src/app/components/hashrate-chart/hashrate-chart.component.ts
@@ -370,9 +370,10 @@ export class HashrateChartComponent implements OnInit {
             const newMin = Math.floor(firstYAxisMin / selectedPowerOfTen.divider / 10)
             return 600 / 2 ** 32 * newMin * selectedPowerOfTen.divider * 10;
           },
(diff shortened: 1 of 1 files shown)


nymkappa (changes requested):
It breaks if you zoom in, the difficulty scaling is not dynamic anymore

### This PR
image

### Prod

image

natsoni:
Thanks, should be fixed now

nymkappa (approved):
Works great - Tested ACK @ 411beb1c9569e7621e1e5502212d3819e0af997e

PR #5880, opened 2025-04-15
Reset taptree widget on address change

natsoni:
`hasTaptree` variable was not reset, causing an empty Taptree widget to be visible for a second on address change:  

https://github.com/user-attachments/assets/dcc47e73-2e58-4fa7-9149-a8e3ab40650b


softsimon (approved):
utACK

PR #5881, opened 2025-04-16
experimental sighash highlighting

mononaut:
Draft PR for an experimental sighash highlighting/visualization feature.

Color-coded icons are displayed to show how many signatures were used to spend each input, and with which sighash modes.

Hovering over the icons highlights which inputs and outputs that signature committed

Screenshot 2025-04-16 at 2 56 50 AM
Screenshot 2025-04-16 at 2 55 10 AM

In the expanded transaction details view, individual signatures (and their sighash flags) in the witness data are similarly identified.

Screenshot 2025-04-16 at 2 55 39 AM

### DIFF
--- a/frontend/src/app/components/transaction/transaction-raw.component.html
+++ b/frontend/src/app/components/transaction/transaction-raw.component.html
@@ -147,7 +147,7 @@
     </div>
 
 
-    <app-transactions-list #txList [transactions]="[transaction]" [transactionPage]="true" [txPreview]="true"></app-transactions-list>
+    <app-transactions-list #txList [transactions]="[transaction]" [transactionPage]="true" [txPreview]="true" [signatures]="true"></app-transactions-list>
 
     <div class="title text-left">
       <h2 i18n="transaction.details|Transaction Details">Details</h2>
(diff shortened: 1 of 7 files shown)


natsoni:
Looks great! I assume the signature processing part in this PR will come in handy later when adding support for unsigned transactions on the transaction preview page (notably to include missing signatures data to weight and feerate calculation)

mononaut:
Updated to better handle very large multisigs, and to support difference display preferences:

| preference | description |
|-|-|
| all | Display all sighash icons everywhere |
| interesting | Only show for "interesting" sighashes (i.e. not for sighash_all, sighash_default, or inputs without signatures) |
| none | Don't show any sighash icons at the top level |

I've set the default to "interesting", which I think is a good balance between highlighting the flags where they're relevant and overloading users with too much info.

Regardless of this display preference, the signature annotations inside the expanded transaction details are always displayed, since I think its always helpful to see which witness items are signatures vs other data for example.

In future, perhaps we can add an actual toggle or dropdown for this preference (or maybe a more general "advanced mode" setting), but for now it can only be temporarily overridden via query param (e.g. `?sigs=all`) to allow a bit more control when sharing links.

wiz (changes requested):
please fix failing CI tests, looks like something related to liquid

wiz (approved):
Tested ACK @ v3.3-dev [bce25d082]

Note that I had to add `?sigs=all` to see the padlocks, seems default is not to show anything despite PR description saying default would show some?

PR #5882, opened 2025-04-18
Switch all nav icons to svg

hunicus:
Currently the pickaxe is a custom SVG and other nav icons are Fork Awesome icons. This makes aligning them annoying and inconsistent. Switching all icons to SVG fixes this.

Tested Mempool and Liquid builds across browsers and across devices with conditions (mining off, faucet on, etc).


wiz:
I think this is a bad idea because the SVG elements often have CSS namespace collisions
@mononaut wdyt

softsimon:
concept ACK for making it more unified. It was a bad idea to mix given the inconsistencies.

hunicus:
Aren't namespace issues caused by classes and IDs in the SVG documents? These don't have any.

nymkappa:
The pickaxes could be made a bit bigger. It looks smaller than other icons. But at least it's properly centered now.

Screenshot 2025-04-21 at 15 12 36

nymkappa (changes requested):
Tested NACK - Removing `fa` icons from `shared.module.ts` breaks the menu.

Here is what we used so far for the menu icons:

```
faIcon: 'money-bill-trend-up',
faIcon: 'rocket',
faIcon: 'wand-magic-sparkles',
faIcon: 'circle-check',
faIcon: 'database',
faIcon: 'list',
faIcon: 'id-card-alt',
faIcon: 'check-circle',
faIcon: 'share-nodes',
faIcon: 'line-chart',
faIcon: 'th-list',
faIcon: 'fast-forward',
faIcon: 'robot',
faIcon: 'microscope',
faIcon: 'history',
faIcon: 'tshirt',
faIcon: 'credit-card',
faIcon: 'user-clock',
faIcon: 'wrench',
faIcon: 'key',
faIcon: 'question-circle',
faIcon: 'sign-out-alt',
```

hunicus:
Screenshot from 2025-04-21 08-23-11

Screenshot from 2025-04-21 08-28-34

nymkappa:
Tested ACK @ b16cc36ffa9a100c5ca4a4fadffb0edfbe76132c
Thank you for fixing, it was so triggering my OCD

Not merging as @mononaut did not get a chance to answer wiz's question

> I think this is a bad idea because the SVG elements often have CSS namespace collisions @mononaut wdyt

nymkappa (approved):
_redacted_

nymkappa (changes requested):
There is still something wrong with the pickaxe
- the width of the button is not the same as the others (60.58px vs 57.58px) making the whole thing imbalanced
- it is still not properly centered. in red you can see the center of the svg image, in green the center of the html element which contains the svg

image

hunicus:
Button width should be the same now.

What tool are you using to evaluate centering? I removed some spacing on the edges of the SVG which may have helped but I can't verify.

nymkappa:
> What tool are you using to evaluate centering

My eyes. But I simply used Gimp measure tool to illustrate the issue. The centering seems good now.

But I've noticed that the color is wrong, it's not pure white.

image

hunicus:
Fixed.

nymkappa (approved):
Running out of idea to troll @hunicus ACK @b1ea5bdb5a61f21bfb618eed24401344d185eddf

Thanks for fixing this

PR #5901, opened 2025-05-08
[mining] fix potential bug where we try to index hashrate even if there are no blocks

nymkappa:


backend/src/api/mining/mining.ts
```diff
@@ -256,31 +256,36 @@ class Mining {
...
-            hashrates.push({
-              hashrateTimestamp: toTimestamp / 1000,
-              avgHashrate: pool['hashrate'] ,
-              poolId: pool.poolId,
-              share: pool['share'],
-              type: 'weekly',
+        if (blockStats.blockCount <= 0) {
+          logger.debug(`No block found between ${fromTimestamp / 1000} and ${toTimestamp / 1000}, skipping hashrate indexing for this period`, logger.tags.mining);
+        } else {
+          const lastBlockHashrate = await bitcoinClient.getNetworkHashPs(blockStats.blockCount,
```
nymkappa, line 263: code is unchanged here, just indentation because of the `else {}` block

backend/src/api/mining/mining.ts
```diff
@@ -256,31 +256,36 @@ class Mining {
...
-          });
 
-          for (const pool of pools) {
-            hashrates.push({
-              hashrateTimestamp: toTimestamp / 1000,
-              avgHashrate: pool['hashrate'] ,
-              poolId: pool.poolId,
-              share: pool['share'],
-              type: 'weekly',
+        if (blockStats.blockCount <= 0) {
```
nymkappa, line 260: that's the fix™️

wiz (approved):
Tested ACK @ e5082863de6b713613dea8e80d613f519fe299ff

PR #5902, opened 2025-05-08
[lightning] make sure channel id has valid format before caching its funding transaction

nymkappa:


wiz (changes requested):
Tested NACK @ 13b4126
```
May  8 11:01:32 [58600] NOTICE: <testnet-lightning> Starting Mempool Server... (13b4126)
...
May  8 11:01:52 [58600] ERR: <testnet-lightning> Exception in LightningStatsImporter::$run(). Error: Wrong type passed:
{
    "Position 1 (txid)": "JSON value of type null is not of expected type string"
}
```

wiz (approved):
Tested ACK @ c67c257
```
May  8 11:14:01 [6563] DEBUG: <testnet-lightning> [Lightning] Cannot cache 3614171x3148x0 funding tx. TX index 3148 does not exist in block 00000000bde3105bd66c732bee025aac92844f671c95616805895c575605782c
```

PR #5896, opened 2025-05-02
Fix get-block api doc

hunicus:
Fixes #2818 
Also fixes #3192

Issue #3192, opened 2023-03-02
Update block(s) APIs documentation


nymkappa:
Currently the REST API documentation for `GET Block` is wrong and `Get Blocks` is incomplete. We need to update it with the following info.

### GET Block

There are two `GET Block` API. Somehow we need to differentiate them since one goes through esplora and the other goes through our nodejs backend

- `/api/block/:hash` - Esplora backend - Returns the following response format:

Example for `/api/block/0000000000000000000538f443ed8d2a058b7cde674900eeb5b1473b268b57b4`

```
{
  "id": "0000000000000000000538f443ed8d2a058b7cde674900eeb5b1473b268b57b4",
  "height": 778903,
  "version": 551550976,
  "timestamp": 1677720918,
  "tx_count": 3120,
  "size": 1702978,
  "weight": 3992935,
  "merkle_root": "6b8b6133296758d4b691a6af54c833469a50eeeb471ae2766bc1c6679bdd70c2",
  "previousblockhash": "000000000000000000040a74ae8ce9e91803ba654318a12f29e37b18f48f88fe",
  "mediantime": 1677717969,
  "nonce": 2337141100,
  "bits": 386304419,
  "difficulty": 43053844193928
}
```

- `/api/v1/block/:hash` - Nodejs backend - Returns the following response format:

Example for `/api/v1/block/0000000000000000000538f443ed8d2a058b7cde674900eeb5b1473b268b57b4`

```
{
  "id": "0000000000000000000538f443ed8d2a058b7cde674900eeb5b1473b268b57b4",
  "height": 778903,
  "version": 551550976,
  "timestamp": 1677720918,
  "bits": 386304419,
  "nonce": 2337141100,
  "difficulty": 43053844193928.45,
  "merkle_root": "6b8b6133296758d4b691a6af54c833469a50eeeb471ae2766bc1c6679bdd70c2",
  "tx_count": 3120,
  "size": 1702978,
  "weight": 3992935,
  "previousblockhash": "000000000000000000040a74ae8ce9e91803ba654318a12f29e37b18f48f88fe",
  "mediantime": 1677717969,
  "extras": {
    "reward": 641681192,
    "coinbaseRaw": "0397e20b0457fdff632f466f756e6472792055534120506f6f6c202364726f70676f6c642f13bdcc3f00001951183d0300",
    "orphans": [],
    "medianFee": 15,
    "feeRange": [
      1,
      3,
      5,
      15,
      20,
      27,
      616
    ],
[...]

PR #5909, opened 2025-05-14
Add get-block v1 and get-blocks v1

### LINKED ISSUE #2818: getBlock API doc not accurate
The getBlock API docs is showing the the block `extras` properties like feeRange. But clicking on the link to the API the extras doesn't exist there.

This is because we have 2 block APIs
`/api/block` that goes to esplora and
`/api/v1/block` that goes to NodeJS and our database with the extras is present

Maybe both should be documented separately or just fix the current one.

hunicus:
Succeeds #5896. Fixes #2818 and #3192.


frontend/src/app/docs/api-docs/api-docs-nav.component.ts
```diff
@@ -25,6 +29,9 @@ export class ApiDocsNavComponent implements OnInit {
   ngOnInit(): void {
     this.env = this.stateService.env;
     this.officialMempoolInstance = this.env.OFFICIAL_MEMPOOL_SPACE;
+    this.stateService.backend$.pipe(takeUntil(this.destroy$)).subscribe((backend) => {
```
nymkappa, line 32: How about something simpler like this

```
ngOnInit(): void {
    ...
    this.backendInfoSub = this.stateService.backendInfo$.subscribe((backend) => { ...})
    ...
}

ngOnDestroy(): void {
    this.backendInfoSub.unsubscribe();
}
```

nymkappa (approved):
mempool bitcoin backend detection works well and the distinction between both block(s) APIs is correct.

I've only checked locally using mainnet and I did not carefully reviewed other network examples, but further correction can be made if needed.

This PR do what was needed, and does not seem have any ninja css that breaks the whole site.

### `"BACKEND": "esplora"` (production)
Screenshot 2025-05-14 at 2 58 41 PM

### `"BACKEND": "none"`
Screenshot 2025-05-14 at 2 58 00 PM

PR #5916, opened 2025-05-20
Add USD series to Liquid reserves graph

natsoni:
To make this work I had to add a historical price endpoint to the Liquid backend because the existing endpoint we use for historical prices is part of the mining routes which aren't included in Liquid backend.

https://github.com/user-attachments/assets/0096900b-1d31-46f7-9878-8996669f88e5


mononaut (approved):
Tested ACK @ [a87cdff3]

PR #5918, opened 2025-05-21
Fix coinbase tx details toggle

natsoni:
Toggling the details of a coinbase transaction was broken because we don't process signature data on coinbase transactions.

Screenshot 2025-05-21 at 13 36 18


mononaut (approved):
Tested ACK @ [4d6935cc]

good catch, thanks!

PR #5919, opened 2025-05-21
Fix colspan value

natsoni:
Toggling transaction details when signatures mode is on breaks table alignment: 

https://github.com/user-attachments/assets/9f107314-b43f-42ff-88f4-e71c8241aa24


mononaut (approved):
Tested ACK @ [b3db3ce0]

PR #5920, opened 2025-05-22
Crop large OP_RETURN data in tx details section

natsoni:
Currently, we don't crop OP_RETURN data in the transaction details, which can cause the page to freeze on some transactions:

https://mempool.space/tx/b47eba144b4b9e9ee4e99f1db2081986f5ac59a7944780125e8c3360fec659c7

https://mempool.space/tx/516c63376556d87c4779033327184ee00a08c4c14498e14673357ce4a791406b

This PR fixes the issue by truncating large `asm`, `hex`, and `OP_RETURN data` fields and adding a "Show all" button to expand if needed, similarly to how we crop large witnesses in inputs.

Screenshot 2025-05-22 at 17 28 27


frontend/src/app/components/transactions-list/transactions-list.component.html
```diff
@@ -384,15 +384,58 @@
...
+                        <td style="text-align: left;">
+                          @if (vout.scriptpubkey.length > 1000) {
+                            @if (!showFullScriptPubkeyHex[vindex]) {
+                              {{ vout.scriptpubkey | slice:0:1000 }}
+                            } @else {
+                              {{ vout.scriptpubkey }}
+                            }
+                          } @else {
+                            {{ vout.scriptpubkey }}
+                          }
```
mononaut: the outer `@if ... @else` here is redundant (if the scriptpubkey is less than 1000 characters, the `slice` pipe will have no effect anyway)

frontend/src/app/components/transactions-list/transactions-list.component.html
```diff
@@ -384,15 +384,58 @@
...
+                          </div>
+                        </td>
                       </tr>
                       <tr *ngIf="vout.scriptpubkey_type == 'op_return'">
                         <td>OP_RETURN <span>data</span></td>
-                        <td style="text-align: left;">{{ vout.scriptpubkey_asm | hex2ascii }}</td>
+                        <td style="text-align: left; word-break: break-word;">
+                          @if (vout.scriptpubkey_asm.length > 1000) {
+                            @if (!showFullOpReturnData[vindex]) {
+                              {{ (vout.scriptpubkey_asm | slice:0:1000 | hex2ascii) }}
```
mononaut: better to swap the order of the pipes here, so that we're truncating the *displayed* text instead of the unconverted hex data - this will also make the size of the cropped fields more consistent.

so:
```
{{ (vout.scriptpubkey_asm | hex2ascii | slice:0:1000) }}
```

frontend/src/app/components/transactions-list/transactions-list.component.html
```diff
@@ -384,15 +384,58 @@
...
+                              <span *ngIf="showFullScriptPubkeyHex[vindex]" i18n="show-less">Show less</span>
+                            </label>
+                          </div>
+                        </td>
                       </tr>
                       <tr *ngIf="vout.scriptpubkey_type == 'op_return'">
                         <td>OP_RETURN <span>data</span></td>
-                        <td style="text-align: left;">{{ vout.scriptpubkey_asm | hex2ascii }}</td>
+                        <td style="text-align: left; word-break: break-word;">
+                          @if (vout.scriptpubkey_asm.length > 1000) {
```
mononaut: redundant outer if statement again

frontend/src/app/shared/components/asm/asm.component.ts
```diff
@@ -82,6 +82,7 @@ export class AsmComponent {
     ['ELSE', 'control'],
     ['ENDIF', 'control'],
     ['VERIFY', 'control'],
+    ['RETURN', 'control'],
```
mononaut, line 106: good catch, I'm not sure how that line got lost 😅

frontend/src/app/components/transactions-list/transactions-list.component.html
```diff
@@ -384,15 +384,58 @@
                     <tbody>
                       <tr>
                         <td i18n="transactions-list.scriptpubkey.asm|ScriptPubKey (ASM)">ScriptPubKey (ASM)</td>
-                        <td style="text-align: left;" [innerHTML]="vout.scriptpubkey_asm | asmStyler"></td>
+                        <td style="text-align: left;">
+                          <app-asm [asm]="vout.scriptpubkey_asm" [crop]="showFullScriptPubkeyAsm[vindex] ? 0 : 1000"></app-asm>
```
mononaut, line 388: The default behavior of the `crop` property on the ASM component can look a bit odd here with large op_returns, since if there's a data push over 500 bytes it gets hidden entirely instead of just truncated to the first ~1000 hex characters.

Screenshot 2025-06-05 at 12 10 09 AM


I don't think this feature of the component is actually used anywhere else yet, so it's probably fine to just change the behavior to blindly truncate based on character count for now, although it would be even nicer to attempt to always truncate either in whitespace or in the middle of a sufficiently large data item (rather than halfway through an opcode, for example).

mononaut (changes requested):
implementation looks good, just needs a few minor tweaks mentioned in the inline comments before it's ready to merge

frontend/src/app/components/transactions-list/transactions-list.component.html
mononaut: (you can use the hex2ascii pipe inside the `*ngIf` condition, to make the "show all"/"show less" toggle work correctly too, like `(vout.scriptpubkey_asm | hex2ascii).length > 1000`)

frontend/src/app/components/transactions-list/transactions-list.component.html
natsoni: I updated the crop logic in the ASM component to fill up to the limit, provided that at least the opcode + a few more chars fits within the limit so that only data item is cropped:
 
Screenshot 2025-06-18 at 21 30 40

natsoni:
I think I addressed the issues you brought up, thanks for the feedback!

PR #5921, opened 2025-05-22
Add some checks to the production install script

knorrium:
The installer currently assumes a clean server state to run as it creates zfs datasets, users and groups.

Due to some issues installing node on arm64 FreeBSD, the install script fails and leaves the server in a dirty state.

These checks ensure the datasets are recreated, and only adds the user and groups if they don't currently exist.

This has only been tested on FreeBSD and a separate PR should add the same checks to Linux based servers.

### DIFF
--- a/production/install
+++ b/production/install
@@ -499,15 +499,21 @@ osUserCreate()
 {
     case $OS in
         FreeBSD)
-            # pw useradd -d /mempool -g mempool [-G additional_group] -n mampool
-            if [ $# -eq 3 ] ; then
-                osSudo "${ROOT_USER}" pw useradd -d "$2" -g "$3" -n "$1"
-            elif [ $# -eq 4 ]; then
-                osSudo "${ROOT_USER}" pw useradd -d "$2" -g "$3" -G "$4" -n "$1"
+          if id "$USERNAME" >/dev/null 2>&1; then
+            echo "User '$USERNAME' already exists."
             else
-                echo "Illegal number of parameters"
-                exit 1
-           fi
+                echo "Creating user '$USERNAME'..."
+                # pw useradd -d /mempool -g mempool [-G additional_group] -n mampool
+                if [ $# -eq 3 ] ; then
+                    osSudo "${ROOT_USER}" pw useradd -d "$2" -g "$3" -n "$1"
+                elif [ $# -eq 4 ]; then
+                    osSudo "${ROOT_USER}" pw useradd -d "$2" -g "$3" -G "$4" -n "$1"
+                else
+                    echo "Illegal number of parameters"
+                    exit 1
+            fi
+        fi
+
         ;;
         Debian)
             # useradd -d /mempool -g mempool [-G additional_group] mempool
@@ -527,7 +533,12 @@ osGroupCreate()
 {
     case $OS in
         FreeBSD)
-            osSudo "${ROOT_USER}" pw groupadd $*
+            if pw groupshow $1 >/dev/null 2>&1; then
+                echo "Group already exists"
+            else
+                echo "Creating "${$1}" group..."
+                osSudo "${ROOT_USER}" pw groupadd $*
+            fi
         ;;
         Debian)
             osSudo "${ROOT_USER}" groupadd $*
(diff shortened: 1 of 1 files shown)


production/install
```diff
@@ -551,119 +562,169 @@ osCertbotDryRun()
...
+    zfsCleanupDataset "${ZPOOL}/minfee"
     zfs create -o "mountpoint=${MINFEE_HOME}" "${ZPOOL}/minfee"
+
+    zfsCleanupDataset "${ZPOOL}/electrs"
     zfs create -o "mountpoint=${ELECTRS_HOME}" "${ZPOOL}/electrs"
+
+    zfsCleanupDataset "${ZPOOL}/mempool"
     zfs create -o "mountpoint=${MEMPOOL_HOME}" "${ZPOOL}/mempool"
+
+    zfsCleanupDataset "${ZPOOL}/mysql"
```
wiz, line 598: You can't just yolo nuke all of these filesystems if they already exist, what happened to the upgrade codepath that @portlandhodl made? If the filesystems already exist then we want to upgrade things on the server, not delete everything.

production/install
```diff
@@ -1101,12 +1162,17 @@ esac
...
+  echo "[*] Installing nvm.sh from GitHub"
+  osSudo "${MEMPOOL_USER}" sh -c 'curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.2/install.sh | bash'
 
-echo "[*] Building NodeJS via nvm.sh"
-osSudo "${MEMPOOL_USER}" zsh -c 'source ~/.zshrc ; CC=gcc CXX=g++ nvm install v22.14.0 --shared-zlib'
-osSudo "${MEMPOOL_USER}" zsh -c 'source ~/.zshrc ; nvm alias default 22.14.0'
+  echo "[*] Building NodeJS via nvm.sh"
+  osSudo "${MEMPOOL_USER}" zsh -c 'source ~/.zshrc ; CC=gcc CXX=g++ nvm install v22.14.0 --shared-zlib'
+  osSudo "${MEMPOOL_USER}" zsh -c 'source ~/.zshrc ; nvm alias default 22.14.0'
+fi
```
wiz, line 1175: Please try to match the current indentation

production/install
```diff
@@ -1855,7 +1921,12 @@ echo "[*] Adding MySQL configuration"
 case $OS in
 
     FreeBSD)
-        osSudo "${ROOT_USER}" service mysql-server onestart
+        if ! osSudo "${ROOT_USER}" service mysql-server status >/dev/null 2>&1; then
+            echo "MySQL is not running. Starting..."
+            osSudo "${ROOT_USER}" service mysql-server onestart
+        else
+            echo "MySQL is already running."
+        fi
```
wiz, line 1929: It might make more sense to change this to a loop (including replacing the `sleep 10` below), where you run the onestart command with `|| true` at the end of the commend to not exit this script if it fails for some reason (like if mysql is already running), and then `sleep 1` inside the loop checking the status and continuing after you verify that mysql has indeed started

production/install
```diff
@@ -551,119 +562,169 @@ osCertbotDryRun()
...
         zfs create -o "mountpoint=${ELECTRS_LIQUIDTESTNET_DATA}" "${ELECTRS_LIQUIDTESTNET_ZPOOL}/electrs/liquidtestnet"
         for folder in cache history txstore
         do
+            zfsCleanupDataset "${ELECTRS_LIQUIDTESTNET_ZPOOL}/electrs/liquidtestnet/${folder}"
             zfs create -o "mountpoint=${ELECTRS_LIQUIDTESTNET_DATA}/newindex/${folder}" "${ELECTRS_LIQUIDTESTNET_ZPOOL}/electrs/liquidtestnet/${folder}"
         done
     fi
 
     if [ "${CLN_INSTALL}" = ON ];then
+        zfsCleanupDataset "${ZPOOL}/cln"
```
wiz, line 727: No, you cannot just yolo nuke this filesystem, we need to preserve the data and continue with an upgrade. I believe @portlandhodl already worked on this, so maybe you can check his WIP work branch and base on top of that

production/install
```diff
@@ -551,119 +562,169 @@ osCertbotDryRun()
...
+{
+  DATASET="$1"
+  echo "Cleaning up dataset: $DATASET"
+  if zfs list -H -o name "$DATASET" >/dev/null 2>&1; then
+    echo "Destroying existing dataset: $DATASET"
+    zfs unmount -f "$DATASET"
+    zfs destroy -r "$DATASET"
+  fi
+}
+
```
wiz, line 575: "cleanup" is a hell of a way of saying delete all of our filesystems without any confirmation or anything lol

wiz (changes requested):
I would check @portlandhodl WIP work branch - you might want to make a new PR based on that

PR #5929, opened 2025-06-03
Show decimal in block fee range

natsoni:
Mempool blocks: 

Screenshot 2025-06-03 at 11 45 41

Also shown on mined blocks for consistency:

Screenshot 2025-06-03 at 11 46 01


orangesurf:
I wonder if we should make the numbers after the decimal visually distinct to make it easier to differentiate e.g. 1.22 vs 201 in the screenshot. Perhaps a slightly different color or smaller size?

image

natsoni:
A smaller size for decimal part would look like this:

Screenshot 2025-06-03 at 16 25 08

wiz (approved):
utACK - curious to hear @softsimon thoughts before merging

softsimon (changes requested):
Use the new fee range also on the Block component for consistency

MiguelMedeiros:
Would it also be possible to add decimals in this "transactions fees" section?
image

softsimon:
> Would it also be possible to add decimals in this "transactions fees" section? 

No, because those are more user friendly numbers rounded up for people to input in wallets

MiguelMedeiros:
Got it. But the UX ends up being quite poor when we have identical fees like in this screenshot. Maybe once most wallets support fees below 1 sat/vbyte, it’ll be worth updating this section as well.

image

PR #5933, opened 2025-06-15
[accelerator] show square payment receipt on pizza tracker

nymkappa:


wiz (approved):
"Download Receipt" would be better

PR #5936, opened 2025-06-21
[accelerator] fix top-up link

nymkappa:
Fix broken link introduced in https://github.com/mempool/mempool.space/pull/1618/files


mononaut (approved):
utACK

PR #5939, opened 2025-06-22
fix sub-1 sat/vb colors

mononaut:
before:
Screenshot 2025-06-22 at 12 46 16 PM

after:
Screenshot 2025-06-22 at 12 46 07 PM


wiz:
Isn't this non-trivial to fix, i.e. requiring a database migration to adjust all the historical fee rate tiers etc.

softsimon:
> Isn't this non-trivial to fix, i.e. requiring a database migration to adjust all the historical fee rate tiers etc.

This PR only fix the cosmetic frontend color by changing the fee range from 1 to <2 to 0 to <2.

Properly adding support for 0.1 sat/vB logging requires new backend and frontend changes.

wiz:
Yeah we need the full fixes

mononaut:
updated with full changes to log the 0-1 s/vb in the statistics db and display on the mempool chart:

Screenshot 2025-06-23 at 5 59 40 AM

mononaut:
I'm open to other suggestions on the mempool statistics chart color.

also, do we want to differentiate between 0-1 and 1-2 s/vb in the block visualizations / mempool blocks?

e.g:
Screenshot 2025-06-23 at 6 03 11 AM

vs
Screenshot 2025-06-23 at 6 05 10 AM

mononaut:
note - due to the change in statistics schema, it's important that un-upgraded nodes don't attempt to replicate statistics from nodes running this PR. (but the PR should handle the reverse case fine).

wiz:
how about blue-green for <1

mononaut:
for the chart or the blocks?

wiz (approved):
utACK

PR #5940, opened 2025-06-22
Use actual block weight in median fee calculation

### LINKED ISSUE #5908: Median fee rate incorrectly reported as 0 for underfilled blocks
### Description

Median fee rate is 0 for blocks less than half full.

I think that it counted the empty space as if it was filled with zero fee transactions.

#### Version

node204.fra.mempool.space (v3.3-dev) [4f2a0bd1][e175fd60]

### Steps to reproduce

Example of block: https://mempool.space/block/0000000000000000000190b9fb7d401c47056f45cb0d3db88e3e75b23322a402

### Expected behaviour

It should tell that median feerate is 1 sat/vB

### Actual behaviour

It says:
```
Fee span	1 - 15 sat/vB
Median fee	~0 sat/vB $0.00 
```

### Screenshots

Image

natsoni:
Fixes #5908 by using actual transaction weight instead of full block weight in median fee rate calculation (this meant that medianFee was returning 0 for blocks less than half full).

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -900,9 +900,11 @@ export class Common {
     let medianWeight = 0;
 
     // calculate the "medianFee" as the average fee rate of the middle 0.25% weight units of transactions
-    const halfWidth = config.MEMPOOL.BLOCK_WEIGHT_UNITS / 800;
-    const leftBound = Math.floor((config.MEMPOOL.BLOCK_WEIGHT_UNITS / 2) - halfWidth);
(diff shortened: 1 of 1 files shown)


mononaut:
Concept NACK - the existing behavior is intended.

In my opinion, the most consistent and informative definition of "median fee" for a block is as a percentile of the *available* block space.

Otherwise you get bizarre situations like where this block
Screenshot 2025-06-23 at 2 27 06 AM
has a "median fee" of 4 sats/vb while fuller neighboring blocks appear to be much lower fee.
Screenshot 2025-06-23 at 2 33 28 AM

natsoni:
Yeah that does make sense, an almost empty block should not have a higher median fee rate than a full one next to it; closing this

PR #5947, opened 2025-06-29
bump package-lock.json version (mismatch with package.json)

nymkappa:
We've bumped our package.json version but not the one in the lock file, causing dependabot to automatically bump it for us in its PRs


mononaut (approved):
utACK - yeah seems fine

PR #5951, opened 2025-07-01
Improve PSBT handling in transaction preview

natsoni:
This PR brings some changes to how we handle PSBT data on the transaction preview page, notably:

- if previewed transaction was passed as PSBT, the url fragment now stores it instead of the raw hex so that prevouts and other data is keeped on page refresh
- added support for tapscript and signature pre-filling in P2TR inputs, with the following logic
  - If no signature is provided (only tapscripts): assume a key spend and ignore the tapscripts, unless the internal key is unspendable: in that case, display the first tapscript given
  - If one or more signatures are provided: 
     - count the number of signatures associated with each tapscript
     - select the tapscript with the highest number of matching signatures

These changes are required for adding signature weight to not-fully-signed transactions on the preview page and a follow-up PR is coming to implement that part.


wiz:
Oh I see you guys are working on #5953 so I'll close this

PR #5952, opened 2025-07-02
[accelerator] accelerator referral code rev share

nymkappa:
Required companion PR to https://github.com/mempool/mempool.space/pull/1728

### DIFF
--- a/frontend/src/app/components/tracker/tracker.component.ts
+++ b/frontend/src/app/components/tracker/tracker.component.ts
@@ -160,6 +160,8 @@ export class TrackerComponent implements OnInit, OnDestroy {
     this.referralCode = fragmentsParams.get('referralCode') ?? this.storageService.getValue('referralCode') ?? undefined;
     if (this.referralCode) {
       this.storageService.setValue('referralCode', this.referralCode);
+      // Cleanup referralCode from url after storing it in localStorage to avoid re-use upon page reload
+      window.location.hash = window.location.hash.replace(`&referralCode=${this.referralCode}`, '').replace(`#referralCode=${this.referralCode}`, '');
     }
 
     this.acceleratorAvailable = this.stateService.env.OFFICIAL_MEMPOOL_SPACE && this.stateService.env.ACCELERATOR && this.stateService.network === '';
--- a/frontend/src/app/components/transaction/transaction.component.ts
+++ b/frontend/src/app/components/transaction/transaction.component.ts
@@ -215,6 +215,8 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
     this.referralCode = fragmentsParams.get('referralCode') ?? this.storageService.getValue('referralCode') ?? undefined;
     if (this.referralCode) {
       this.storageService.setValue('referralCode', this.referralCode);
+      // Cleanup referralCode from url after storing it in localStorage to avoid re-use upon page reload
+      window.location.hash = window.location.hash.replace(`&referralCode=${this.referralCode}`, '').replace(`#referralCode=${this.referralCode}`, '');
     }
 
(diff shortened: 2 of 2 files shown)


frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -209,6 +210,13 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
     const urlParams = new URLSearchParams(window.location.search);
     this.forceAccelerationSummary = !!urlParams.get('cash_request_id');
 
+    // Accelerator referral code
+    const fragmentsParams = new URLSearchParams(window.location.hash);
```
mononaut: we typically access URL fragments via the route.fragment observable (and there's already a subscription which calls an `updateFragmentParams` method for handling in the transaction component)

otherwise if you want to use `window.location.hash` directly you need to remember to strip the leading '#', otherwise the first param ends up like `#referralCode` instead of `referralCode` and you won't get the behavior you expect.

mononaut: oh, lol, copilot beat me to it 😅

frontend/src/app/components/tracker/tracker.component.ts
```diff
@@ -147,12 +148,20 @@ export class TrackerComponent implements OnInit, OnDestroy {
...
     private zone: NgZone,
+    private storageService: StorageService,
     @Inject(ZONE_SERVICE) private zoneService: any,
   ) {}
 
   ngOnInit() {
     this.onResize();
 
+    // Accelerator referral code in fragment
+    const fragmentsParams = new URLSearchParams(window.location.hash);
```
mononaut: same issue again, either strip the leading '#' or use the route.fragment observable which handles it for you.

frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
```diff
@@ -531,9 +534,11 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
                   tokenResult.token,
                   cardTag,
                   `accelerator-${this.tx.txid.substring(0, 15)}-${Math.round(new Date().getTime() / 1000)}`,
-                  costUSD
+                  costUSD,
+                  this.referralCode
                 ).subscribe({
                   next: (response) => {
+                    this.storageService.removeItem('referralCode'); // Consume localStorage referralCode
```
mononaut, line 541: we remove the referral code from localStorage when an acceleration request successfully goes through, but since the fragment is still in the URL it gets added right back (and then probably not cleared) if the user refreshes the page.

I'd suggest stripping the `referralCode` fragment out of the URL as soon as we parse it and add it to local storage.

mononaut (approved):
Tested ACK @ [77cdb5fc]

Works nicely!

frontend/src/app/components/transaction/transaction.component.ts
nymkappa: yeah I noticed the comment but assumed we will always have the #accelerate fist, which of course is a buggy assumption so I'll fix lol

nymkappa: Regarding the use of observables all the time... while it makes sense to use it if you need real time update upon change, I don't see the point of using rxjs aside over-complicating a thing that should stay extremely simple _(and it's clearly not simple to integrate properly as shown by the good old `setTimeout(()=>{}, 0)` in the `updateFragmentParams` function)_. Vanilla javascript still works even in an Angular framework 😄.

frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
nymkappa: [This feels hacky](https://github.com/mempool/mempool/pull/5952/changes/1a136aa59e9587272aa08bcae39f4c96f93723ae) but this works well. I'm not sure how Angular would like that though lol

I could not find a way to do this using the router class

PR #5953, opened 2025-07-02
Fill signatures data on transaction preview page

natsoni:
Builds on #5951

This PR adds support to the `tx/preview` page for pre-filling missing signature data in partially signed transactions.

Signatures weight are calculated for the following input formats: 
- P2PKH
- P2SH with wrapped multisig, P2WPKH and P2WSH
- P2WPKH
- P2WSH with wrapped multisig
- P2TR key spend and script spend (simple `CHECKSIG` script or multisig)
- P2PK
- P2MS (bare multisig)

Weight and size of the fully signed transaction are calculated and shown:

Screenshot 2025-07-02 at 15 32 02

Also, broadcast button is disabled if at least one signature is missing, see the signing process of [this transaction](https://mempool.space/testnet4/tx/335dac852bc354ae6edbe41f2c26a61963d1563388bba5363d6a9c8e2455fee2):

signing-process

<details>
<summary>With signature details</summary>

signing_details

</details>


frontend/src/app/components/transaction/transaction-raw.component.ts
```diff
@@ -156,18 +170,52 @@ export class TransactionRawComponent implements OnInit, OnDestroy {
...
+  checkSignatures(transaction: Transaction, hex: string): void {
+    let missingWitnessBytes = 0;
+    let missingNonWitnessBytes = 0;
+
+    let isSegwitTransaction = hex.substring(8, 10) === '00' && hex.substring(10, 12) === '01';
+    transaction.vin.forEach(vin => {
+      addInnerScriptsToVin(vin);
+      const result = fillUnsignedInput(vin);
+      vin['_missingSigs'] = result.missingSigs;
+      result.addToWitness ? missingWitnessBytes += result.bytes : missingNonWitnessBytes += result.bytes;
```
mononaut: this side-effect-only ternary expression feels a bit gross - I think it would be much more readable as a simple if-statement.

frontend/src/app/components/transaction/transaction-raw.component.ts
```diff
@@ -156,18 +170,52 @@ export class TransactionRawComponent implements OnInit, OnDestroy {
...
+      if (result.addToWitness && !isSegwitTransaction) {
+        missingWitnessBytes += 2; // marker and flag
+        missingWitnessBytes += transaction.vin.length; // 1 byte compact size per input (assume witness stack count < 253)
+        isSegwitTransaction = true;
+      }
+    });
```
mononaut: shouldn't checking for the witness flag above (line 179) be sufficient? If we're seeing surprise witness inputs while the flag is missing I think the transaction is probably malformed?

mononaut: oh, I see that this code *is* triggered for unsigned transactions with witness inputs even if the witness flag was set 🤔

I guess the hex string being passed in here isn't the raw user input, but something re-serialized after parsing it into a `Transaction`? (and the flag gets lost along the way if there's no witness data).

it's beyond the scope of this PR, but it would be good to make sure that we *don't* unexpectedly mutate the user-inputted hex or psbt, especially if it's re-exposed to the user as a URL parameter or copy raw hex button etc.

mononaut: final note - just for readability I'd update the `isSegwitTransaction` flag here inside the loop, but wait to add the missing witness bytes later outside the loop if the flag was set.

it took me a minute to spot the `!isSegwitTransaction` part of the condition and realize this block can only run once, so it looked like `+= transaction.vin.length` might happen for every segwit input.

mononaut (dismissed):
a couple of nits in the comments, but otherwise looks good!

I tested with a bunch of PSBTs from https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki#user-content-Test_Vectors, as well as some random raw txs in various states of signing and everything seems to work well.

frontend/src/app/components/transaction/transaction-raw.component.ts
natsoni: > I guess the hex string being passed in here isn't the raw user input, but something re-serialized after parsing it into a Transaction?
> it's beyond the scope of this PR, but it would be good to make sure that we don't unexpectedly mutate the user-inputted hex or psbt, especially if it's re-exposed to the user as a URL parameter or copy raw hex button etc.

Correct, and your comment makes me notice that the round-trip of de-serialize and re-serialize back to hex is not always needed. it's only needed if user input is psbt data, to add the signatures to the raw hex transaction.
I'll make another PR to update the `fromBuffer` function to return the re-serialized data only if necessary (i.e. not if user input is raw hex tx)

> final note - just for readability I'd update the isSegwitTransaction flag here inside the loop, but wait to add the missing witness bytes later outside the loop if the flag was set.
> it took me a minute to spot the !isSegwitTransaction part of the condition and realize this block can only run once, so it looked like += transaction.vin.length might happen for every segwit input.

added `segwitFlagSet` variable so that `+= transaction.vin.length` is done after the loop

PR #5960, opened 2025-07-12
handle sub-1-sat-vb in audits

mononaut:
This PR adds special handling for <1 sat/vb transactions to avoid marking transactions as "removed" which are more likely simply below the default `minTxRelayFee`.

Before:
sub-1 sat/vb transactions in the local mempool marked as "removed" and reduce block health score
Screenshot 2025-07-12 at 4 03 24 AM

After:
sub-1 sat/vb transactions marked as "marginal feerate" and excluded from the health score calculation
Screenshot 2025-07-12 at 4 02 56 AM


wiz (approved):

Evoneem:
Merge

PR #5966, opened 2025-07-15
bluer <1sat color

mononaut:
Screenshot 2025-07-15 at 2 33 35 PM


wiz (approved):
this is not blue at all, but ok

PR #5986, opened 2025-07-23
Liquid quick patch for simplicity taproot script spends

mononaut:
Ahead of mainnet activation of Simplicity on Liquid, this PR detects taproot scriptpath spends using Simplicity tapleaf script versions (`0xbe` or `0xbf`) in the frontend, and displays them as raw hex instead of the garbage assembly instructions currently provided by esplora (which assumes all taproot scripts are [Bitcoin/Elements tapscript](https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki)).

There doesn't seem to be a easy way to decode Simplicity into anything more informative than raw hex yet, but we can upgrade `mempool/electrs` to do that later.

| Before | After |
|-|-|
| Nonsense instructions in transaction input details: Screenshot 2025-07-23 at 10 30 05 AM | Labeled as Simplicity, and displayed as raw hex: Screenshot 2025-07-24 at 8 12 17 AM |
| Screenshot 2025-07-23 at 10 33 04 AM | Screenshot 2025-07-24 at 8 13 11 AM |

### DIFF
--- a/frontend/src/app/components/transactions-list/transactions-list.component.ts
+++ b/frontend/src/app/components/transactions-list/transactions-list.component.ts
@@ -338,14 +338,15 @@ export class TransactionsListComponent implements OnInit, OnChanges, OnDestroy {
         } else { // check for simplicity script spends
           for (const vin of tx.vin) {
             if (vin.prevout?.scriptpubkey_type === 'v1_p2tr' && vin.inner_witnessscript_asm) {
-              const hasAnnex = vin.witness?.[vin.witness.length - 1].startsWith('50');
-              // script spend
-              if (vin.witness.length > (hasAnnex ? 2 : 1)) {
-                const controlBlock = vin.witness[vin.witness.length - (hasAnnex ? 2 : 1)];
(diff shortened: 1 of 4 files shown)


frontend/src/app/components/transactions-list/transactions-list.component.ts
```diff
@@ -331,6 +335,21 @@ export class TransactionsListComponent implements OnInit, OnChanges, OnDestroy {
             }
           }
           tx['_showSignatures'] = this.shouldShowSignatures(tx);
+        } else { // check for simplicity script spends
+          for (const vin of tx.vin) {
+            if (vin.prevout.scriptpubkey_type === 'v1_p2tr' && vin.inner_witnessscript_asm) {
```
natsoni: tests are failing probably because of missing optional chaining here: 

```suggestion
            if (vin.prevout?.scriptpubkey_type === 'v1_p2tr' && vin.inner_witnessscript_asm) {
```

wiz (approved):

PR #5994, opened 2025-08-01
handle missing price data

mononaut:
Much of the backend price & exchange rate functionality assumes that `prices` table includes only valid & complete USD price data.

However, some older instances may still have `USD = -1` rows in their DB from earlier versions of the backend.

This PR:
 - Fixes some DB queries to avoid returning bad data from e.g. the `historical-price` API.
 - Adds a DB migration to explicitly delete bad price rows from the database.
 - Fixes some issues with the price indexing code to allow updating/correcting bad data in the `prices` and `block_prices` tables.
 
Where bad price rows are referenced in the `blocks_prices` table, deleting these rows will cause those blocks to be omitted from some of the historical block fees/rewards APIs.

This probably doesn't actually affect any `blocks_prices` entries in prod, but in any case the broken foreign keys will be fixed during the next blocks prices indexing task.

### DIFF
--- a/backend/src/api/database-migration.ts
+++ b/backend/src/api/database-migration.ts
@@ -7,7 +7,7 @@ import cpfpRepository from '../repositories/CpfpRepository';
 import { RowDataPacket } from 'mysql2';
 
 class DatabaseMigration {
-  private static currentVersion = 100;
+  private static currentVersion = 101;
   private queryTimeout = 3600_000;
   private statisticsAddedIndexed = false;
   private uniqueLogs: string[] = [];
@@ -1297,6 +1297,10 @@ class DatabaseMigration {
     if (version < 58) {
       queries.push(`DELETE FROM state WHERE name = 'last_hashrates_indexing'`);
       queries.push(`DELETE FROM state WHERE name = 'last_weekly_hashrates_indexing'`);
+    }
+
+    if (version < 101) {
+      queries.push(`DELETE FROM prices WHERE USD = -1`);
     }
 
     return queries;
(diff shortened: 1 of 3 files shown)


nymkappa:
Feel free to truncate `blocks_prices`, it takes only a few seconds to rebuild

mononaut:
I don't think a full truncation is necessary. with these changes:
https://github.com/mempool/mempool/blob/66eb31bb4625e8c1e77426abbef8a3f2bb235286/backend/src/repositories/BlocksRepository.ts#L884-L900
https://github.com/mempool/mempool/blob/66eb31bb4625e8c1e77426abbef8a3f2bb235286/backend/src/repositories/BlocksRepository.ts#L912
it should fix just the affected rows just as quickly

nymkappa (approved):
That's a lot of reviewers 😆

I've tested it by manually setting one price entry USD to `-1`. This PR indeed fixes the entry and properly re-link the block to the next available price.

Diff looks fine.

Remember that the historical price API has a 5 minutes cache so the fix would take a few minutes to be reflected for all users.

Tested ACK @ 66eb31bb4625e8c1e77426abbef8a3f2bb235286

natsoni (approved):
Tested ACK @ [66eb31bb4625e8c1e77426abbef8a3f2bb235286]

I checked that a faulty entry `USD = -1` for the row with timestamp `2024-12-31 23:00:00` in the `prices` table was causing the historical price to be 0 on data relying on this row: 

https://github.com/user-attachments/assets/3a903ff0-8524-491a-bdc6-eed14b98ec02

This PR properly deleted that faulty row and now the historical block price fallbacks to correct data.

However we should note that on prod, most price data from March to June 2022 is missing, resulting in the price during that period is not accurate (price went from $40k to $20k on this period), and this PR will probably not fix this inaccuracy? Maybe we should wipe the price table and re-index too

mononaut:
I think the regular price updater task should fill in that missing data automatically now that there aren't a bunch of -1 entries in the way (previously those missing prices weren't automatically fixed due to duplicate timestamps and a missing `...ON DUPLICATE UPDATE...` in the insert query)

PR #5995, opened 2025-08-01
Add 'None' filter mode to mempool goggles

natsoni:
This PR introduces a `None` filtering mode to Mempool Goggles, which can be nice to have in some scenarios. 

For example with current filters we can't highlight non data-carrying transactions: we have to manually toggle the data-carrying flags on `Any` mode and look for the non-highlighted transactions, which can be tricky if there is only a few such transactions in the block.

find transaction with neither ancestors nor descendants:
| `Any`: need to look for darkened txs  | `None`: target txs are highlighted |
| Screenshot 2025-08-01 at 15 08 25 | Screenshot 2025-08-01 at 15 08 23 |

Or if we want to find transactions only using P2WSH: 
| `Any`  | `None` |
| Screenshot 2025-08-01 at 14 46 23 |Screenshot 2025-08-01 at 14 46 13 |

### DIFF
--- a/frontend/src/app/components/block-overview-graph/block-overview-graph.component.ts
+++ b/frontend/src/app/components/block-overview-graph/block-overview-graph.component.ts
@@ -665,7 +665,7 @@ export class BlockOverviewGraphComponent implements AfterViewInit, OnDestroy, On
           matches = flags === 0n || (tx.bigintFlags & flags) > 0n;
           break;
         case 'nor':
-          matches = flags === 0n || (tx.bigintFlags & flags) === 0n;
+          matches = (tx.bigintFlags & flags) === 0n;
           break;
       }
       if (matches) {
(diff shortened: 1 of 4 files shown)


frontend/src/app/components/block-overview-graph/block-overview-graph.component.ts
```diff
@@ -656,7 +656,19 @@ export class BlockOverviewGraphComponent implements AfterViewInit, OnDestroy, On
...
+      let matches = false;
+      switch (this.filterMode) {
+        case 'and':
+          matches = (tx.bigintFlags & flags) === flags;
+          break;
+        case 'or':
+          matches = flags === 0n || (tx.bigintFlags & flags) > 0n;
+          break;
+        case 'nor':
+          matches = flags === 0n || (tx.bigintFlags & flags) === 0n;
```
mononaut: the first `flags === 0n` is kinda redundant (if true, then the following clause will also always be true)
```suggestion
          matches = (tx.bigintFlags & flags) === 0n;
```

mononaut (approved):
Tested ACK @ [252b7961]

PR #6008, opened 2025-08-12
Expose chain tips in public API

mononaut:
This PR exposes the cached "chain tips" object that we already periodically fetch from the local Bitcoin node in a new API at `/api/v1/chain-tips`.


mononaut:
yeah, that sounds sensible

PR #6021, opened 2025-08-27
always serve latest good prices

mononaut:
This PR extends the price updater service to track the latest known good price for each currency, so that we can avoid ever serving responses with bad price data.

### DIFF
--- a/backend/src/tasks/price-updater.ts
+++ b/backend/src/tasks/price-updater.ts
@@ -56,6 +56,7 @@ class PriceUpdater {
   private feeds: PriceFeed[] = [];
   private currencies: string[] = ['USD', 'EUR', 'GBP', 'CAD', 'CHF', 'AUD', 'JPY'];
   private latestPrices: ApiPrice;
+  private latestGoodPrices: ApiPrice;
   private currencyConversionFeed: ConversionFeed | undefined;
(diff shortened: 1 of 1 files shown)


wiz:
guys?

backend/src/tasks/price-updater.ts
```diff
@@ -179,6 +182,14 @@ class PriceUpdater {
     this.running = false;
   }
 
+  private setLatestPrice(currency, price): void {
+    this.latestPrices[currency] = price;
+    if (price ?? price > -1) {
```
natsoni: this condition does not return false for negative prices, meaning that latestGoodPrices is still being updated to -1 when connection drops.

Replacing it with `(price ?? -1) > -1` seems to work as intended: the `api/v1/prices` endpoint correctly returns latest correct data

natsoni (changes requested):

backend/src/tasks/price-updater.ts
mononaut: oops, good catch!

I think that was just supposed to be `if (price > 0) { ...`

natsoni (approved):
Tested ACK @ [53f8665]

I checked that on network interruption, both the `api/v1/prices` endpoint and the websocket return last known prices.

PR #6023, opened 2025-08-29
Add FAQ link for block child earlier than parent

natsoni:
This adds an icon info between two blocks with unordered timestamps, linking to corresponding FAQ entry:

Screenshot 2025-08-29 at 16 19 57

Note that currently, the icon still appears on older blocks where it's not obvious that timestamps are unordered: 

Screenshot 2025-08-29 at 16 20 50


softsimon:
IMO the info icon placement isn't that beautiful or useful.

A FAQ item is enough to show people that wonder.

wiz:
> A FAQ item is enough to show people that wonder

if that were true we wouldn't have 9000 people reporting it as a "bug" whenever block timestamps are out of order

mononaut:
can we put the icon *inside* the block next to the time field? (making it a bit smaller if necessary to fit)

natsoni:
Done: 

Screenshot 2025-09-08 at 11 46 48

PR #6029, opened 2025-09-07
[accelerator] we don't need to start a new price subscription at checkout

nymkappa:
Fixes https://github.com/mempool/mempool.space/issues/1879

I did not re-indent the code backward to keep the github diff clean, can be done in another PR

### DIFF
--- a/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
+++ b/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
@@ -474,9 +474,6 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
     if (this.processing) {
       return;
     }
-    if (this.conversionsSubscription) {
-      this.conversionsSubscription.unsubscribe();
-    }
 
     this.processing = true;
 
@@ -591,9 +588,6 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
   async requestGooglePayPayment(): Promise<void> {
     if (this.processing) {
       return;
-    }
-    if (this.conversionsSubscription) {
-      this.conversionsSubscription.unsubscribe();
     }
     
     this.processing = true;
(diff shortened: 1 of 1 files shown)


frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
```diff
@@ -479,14 +479,12 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
     }
```
mononaut: do we still want to unsubscribe from conversions here?

frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
```diff
@@ -601,14 +597,12 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
...
```
mononaut: and here

frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
```diff
@@ -726,11 +718,8 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
...
```
mononaut: etc

frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
```diff
@@ -831,15 +818,13 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
...
```
mononaut: etc

frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
nymkappa: indeed not!

mononaut: uh, this bit I mean:
https://github.com/mempool/mempool/blob/0c0e6c298159bfc236bfe09069d7dc6a4a68ae7a/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts#L477-L479

mononaut (approved):
semi-tested ACK @ [27e88920]

nymkappa:
Re-tested ACK @ 27e88920c7e60609cee41b8a6e4ab29b4b82c989

PR #6030, opened 2025-09-08
index & serve stale blocks

mononaut:
This draft PR is a fairly large refactor of how we handle indexing blocks to support properly indexing and serving _stale_ blocks.

Screenshot 2025-09-08 at 9 33 20 AM

Main changes:

#### DB
 - changes `blocks` table primary key from `height` to `hash` (to allow multiple blocks at the same height)
 - adds a new indexed boolean column `stale`

#### Repository modules
 - updates most blocks-related queries to exclude stale blocks in order to preserve existing behavior
 
#### Indexing
 - indexes stale blocks lazily the first time they are requested, using the same processes as regular blocks
   - (adapting some indexing methods to make this possible)
   - only attempts to index known stale blocks, as determined by the chain-tips module
 - adapts various indexing methods to fetch data relating to stale blocks from Core instead of esplora when necessary
 - stale block indexing includes block summaries
 - block audits continue to work automatically, if they were performed before a block went stale
 
#### Reorgs
 - rewrites both the real-time reorg handling in `updateBlocks` & the `validateChain` method to  avoid ever deleting indexed block data
 - instead, simply update which blocks are marked `stale` and re-index derived data like hashrates from the earliest affected height
 - also fixes the handling in `updateBlocks` to work with any length reorg, instead of always rolling back a fixed 10 blocks
 
---
 
Next steps (maybe for separate PRs)
 - [x] Test reorg handling
 - [x] Dedicated stale block visualization page, comparing the stale and winning blocks in an audit-like view (see #6038)
 - [x] Display/link to stale blocks from the blockchain bar or the main blocks page (see also #6054)
 - [ ] New stale blocks page listing/visualizing all detected stale blocks
 - [ ] Replicate stale blocks between different servers
     - send the raw block to the local core node
     - record & share which servers saw each stale block

PR #6038, opened 2025-09-15
Compare stale blocks

mononaut:
_(builds on #6030)_

This PR extends #6030 to add a feature to the `/block` page to visualize the differences between a stale block and the winning canonical block:

localhost_4200_block_000000000000000000000cdf5d7f499305261f6fadab837f94a3e5acb28795b8 (2)

If an audit is available for the stale block, the audit toggle still works as usual, with some minor UI changes to make it clear the block is stale (although for stale blocks, the page defaults to the new comparison view):

localhost_4200_block_000000000000000000000cdf5d7f499305261f6fadab837f94a3e5acb28795b8 (1)


natsoni:
Awesome! It seems that it fails to load the stale block summary on esplora backend. I managed to fix it here: https://github.com/mempool/mempool/pull/6048

PR #6040, opened 2025-09-18
Unfurler canvas: add address component

natsoni:
A few samples of the address preview on canvas unfurler:

bc1pxpwv7vh7529ga29ejnjrl2cg0qqmq066p8yd494mfw6uhqms9jxqt6kg7x

<br>
<br>

04678afdb0fe5548271967f1a67130b7105cd6a828e03909a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112de5c384df7ba0b8d578a4c702b6bf11d5f

<br>
<br>

1wizSAYSbuyXbt9d8JV8ytm5acqq2TorC (1)

### DIFF
--- a/unfurler/src/canvas/components.ts
+++ b/unfurler/src/canvas/components.ts
@@ -176,9 +176,22 @@ export const components: Record<string, (...args: any[]) => Component> = {
       data: [
         dataRequirements.address(id),
       ],
+      props: (data) => {
+        const a = data[`address_${id}`];
+        const received = a.chain_stats.funded_txo_sum / 1e8;
+        const sent = a.chain_stats.spent_txo_sum / 1e8;
+        const rows = [
+          { label: 'Total received', value: { num: formatNumber(received, '1.8-8'), unit: 'BTC' } },
+          { label: 'Total sent', value: { num: formatNumber(sent, '1.8-8'), unit: 'BTC' } },
+          { label: 'Balance', value: { num: formatNumber(received - sent, '1.8-8'), unit: 'BTC' } },
+          { label: 'Transactions', value: formatNumber(a.chain_stats.tx_count + a.mempool_stats.tx_count, '1.0-0') },
+          { label: 'Unspent TXOs', value: formatNumber(a.chain_stats.funded_txo_count - a.chain_stats.spent_txo_count, '1.0-0') },
+        ];
+        return { tableRows: rows };
+      },
       children: [
         components.qrcode(id, { x: bounds.x + bounds.w - 48 - 480, y: bounds.y + bounds.h - 16 - 480, w: 480, h: 480 }),
-        components.table(id, { x: bounds.x + 48, y: bounds.y + 129 }, 'address'),
+        components.table({ x: bounds.x + 48, y: bounds.y + 129 }),
       ],
       render: async (ctx: CanvasRenderingContext2D, data: any, props: any = {}): Promise<void> => {
         const address: string = id;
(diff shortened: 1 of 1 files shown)


unfurler/src/canvas/components.ts
```diff
@@ -312,5 +377,14 @@ export const views: Record<string, (...args: any[]) => Component> = {
...
       components.blockContent(hash, { x: 0, y: 80, w: 1200, h: 520 })
     ]
+  }),
+  address: (hash: string): Component => ({
+    type: 'address',
+    data: [],
+    children: [
+      components.background({ x: 0, y: 0 }),
+      components.header('Address', { x: 0, y: 0 }),
+      components.addressContent(hash, { x: 0, y: 80, w: 1200, h: 520 })
```
mononaut:
```suggestion
  address: (address: string): Component => ({
    type: 'address',
    data: [],
    children: [
      components.background({ x: 0, y: 0 }),
      components.header('Address', { x: 0, y: 0 }),
      components.addressContent(address, { x: 0, y: 80, w: 1200, h: 520 })
```

unfurler/src/canvas/components.ts
```diff
@@ -212,6 +257,17 @@ export const components: Record<string, (...args: any[]) => Component> = {
...
+        const addressData = data[`address_${id}`];
+        rows = [
+          { label: 'Total received', value: { num: formatNumber((addressData.chain_stats.funded_txo_sum) / 100000000, '1.8-8'), unit: 'BTC' } },
+          { label: 'Total sent', value: { num: formatNumber((addressData.chain_stats.spent_txo_sum) / 100000000, '1.8-8'), unit: 'BTC' } },
+          { label: 'Balance', value: { num: formatNumber((addressData.chain_stats.funded_txo_sum - addressData.chain_stats.spent_txo_sum) / 100000000, '1.8-8'), unit: 'BTC' } },
+          { label: 'Transactions', value: formatNumber(addressData.chain_stats.tx_count + addressData.mempool_stats.tx_count, '1.0-0') },
+          { label: 'Unspent TXOs', value: formatNumber(addressData.chain_stats.funded_txo_count - addressData.chain_stats.spent_txo_count, '1.0-0') },
+        ];
+      }
+
```
mononaut: The table component is starting to get a bit unwieldy already with the row data defined inside the render function depending on the `type` like this.

Ideally I think we'd keep the component as abstract as possible and pass in the row data from the parent somehow, but that's a little bit awkward with the current rendering architecture. Perhaps we need way to pass computed data from parent to child, a bit like an Angular Input, alongside our existing component arguments & data requirements?

It's fine for now, but maybe worth considering a refactor when we add the next preview with tabular data.

natsoni: Right, it's a bit awkward how data is managed in the table component currently, I'll figure something out

unfurler/src/canvas/renderer.ts
```diff
@@ -85,14 +85,19 @@ class CanvasRenderer implements Renderer {
...
       if (next?.render) {
         await next.render(ctx, data, params);
       }
       if (next?.children) {
-        renderStack.push(...[...next.children].reverse());
+        for (let i = next.children.length - 1; i >= 0; i--) {
+          renderStack.push({ next: next.children[i], props: params });
+        }
       }
     }
```
mononaut, line 102: with this approach, the parent passes the same set of props to every child.

but what happens if you want two components of the same type (two tables with different sets of rows, for example) as children of the same parent?

more generally, if every child also inherits all of the parents props (and their parents' parents and so on), you'll need to be very careful about namespace pollution/collisions when defining views and components.

some possible alternatives (in no particular order):
 - define separate props callbacks passed to each child, instead of a single callback on the parent (bonus - you can get component-specific type-checking on the props callback and results for free).
 - give each component instance a unique id, and prefix the props fields accordingly (doesn't really solve the underlying architecture problem, but it's a systematic way of dealing with it).
 - return a list of separate props objects from the parent's single props callback, one for each child.

when thinking about the architecture, it might be helpful to consider how the rendering pipeline would handle simpler, more generic components (like text fields) as well as these complex higher-level views.

natsoni: First alternative sounds good, yeah this first shot was terrible eh 😅

PR #6045, opened 2025-09-21
stricter async/await linting

mononaut:
This draft PR adds two stricter linting rules for the backend related to use of async/await, and then a bunch of jsdoc comments and explicit `void`s to satisfy those rules in the current codebase:
 - [@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises/):
     - built-in typescript linting rule
     - reports Promise-valued statements that are not treated in one of the following ways:
        - Calling its .then() with two arguments
        - Calling its .catch() with one argument
        - awaiting it
        - returning it
        - voiding it
 - `no-unhandled-await`:
     - A custom rule
     - reports `await`ed async function calls which are not wrapped in a try-catch unless:
         - the _callee_ is annotated with `@asyncSafe` to indicate that it *cannot* throw an error or reject
         - the _caller_ is annotated with `@asyncUnsafe` to indicate that errors/rejections are intended to bubble up and be handled at a higher level.
         
In theory, the two rules together make it impossible to _accidentally_ leave a potential promise rejection unhandled without seeing linting errors.

I already found a number of potential crash bugs and other more minor issues in the process of applying these linting rules & working through all of the missing annotations and `void`s etc, which I think demonstrates the need for something like this.

(fixes for those bugs in a separate PR #6044)

---

The diff touches a lot of files, but it's almost entirely the initial setup of those `@asyncSafe`/`@asyncUnsafe` annotations on async functions and some missing `void`s on intentional floating promises, which should be fairly safe.

The linting WOMM, but I'm leaving this in draft for now until I get some concept ACKs or confirmations that it works/doesn't break anything in other environments.


mononaut:
Rebased on master, and fixed the linting errors on all of the new code introduced since this PR was originally opened.

Also tightened up the rules to disallow using `void` to dispatch floating promises unless they're known to be safe, and fixed various instances where that was happening.

mononaut:
onceagain

PR #6048, opened 2025-09-23
Fallback to core for stale blocks on $getBlock calls

natsoni:
Builds on #6038

When we call `$getStrippedBlockTransactions` for a stale block, `const block = await bitcoinApi.$getBlock(hash);` is called. but because `hash` is a stale block it returns a 404 which makes the request fail.

This PR updates the logic to fallback to Core for those cases similarly to how it's done in `$getTxsForBlock`.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
+++ b/backend/src/api/bitcoin/bitcoin-api-abstract-factory.ts
@@ -15,7 +15,7 @@ export interface AbstractBitcoinApi {
   $getTxsForBlock(hash: string, fallbackToCore?: boolean): Promise<IEsploraApi.Transaction[]>;
   $getBlockHash(height: number): Promise<string>;
   $getBlockHeader(hash: string): Promise<string>;
-  $getBlock(hash: string): Promise<IEsploraApi.Block>;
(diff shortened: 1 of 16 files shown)


mononaut:
thanks! does this only reproduce when querying the `/api/v1/block/:hash/summary` and `/api/v1/block/:hash/tx/:txid/summary` endpoints directly?

my intention with #6030, once it's finished, is to proactively index known stale blocks and only serve this data from the DB to avoid adding a codepath for those public APIs to directly trigger requests to Core.

natsoni:
> thanks! does this only reproduce when querying the /api/v1/block/:hash/summary and /api/v1/block/:hash/tx/:txid/summary endpoints directly?

It reproduces both when querying the backend directly, and on the stale block frontend page: 

Screenshot 2025-09-24 at 12 31 55

> my intention with https://github.com/mempool/mempool/pull/6030, once it's finished, is to proactively index known stale blocks and only serve this data from the DB to avoid adding a codepath for those public APIs to directly trigger requests to Core.

That makes sense, it’s better not to expose requests to Core from the frontend. The error only occurs when the stale block summary hasn’t been indexed, which I assume won't happen anymore for future stale blocks with #6030 (we keep the stale block summaries in the DB)

mononaut:
resolved directly in #6030

PR #6055, opened 2025-10-04
address poisoning detection fixes

mononaut:
This PR adjusts the address poisoning similarity detection heuristics to improve the false positive & false negative rates:
 - exclude the witness version character from bech32(m) address prefix matches (which was artificially inflating the similarity score of bech32(m) addresses)
 - dynamically adjust the score threshold based on the number of comparisons to account for the birthday problem effect, so that each transaction has a similar chance of producing a false positive result.
 - slightly discount the contribution of imperfectly matching prefixes/postfixes to the similarity score (so that e.g. `bc1qxxyx... <-> bc1qxxzx...` is considered less similar than `bc1qxxx... <-> bc1qxxx...`)
 - change the similarity threshold from 1-in-10-million *comparisons* to a 1-in-1-million *transactions* random false positive rate.
 
---
 
Test transactions:
| | Type | Before | After |
|-|-|-|-|
| [tx](https://mempool.space/tx/26d146d63a834f3108f52b6fed1090a8adc84c152465494e9ed13736ecc5c6d0) | not very similar bech32(m) addresses | false positive | no warning |
| [tx](https://mempool.space/tx/777b7c7bae8489668244ef20d9842c449c8b44f5f5e89748cbd2c132046dc74a) | many alphabetically sorted outputs | many false positives | no warning |
| [tx](https://mempool.space/tx/036641c1461928126815b001b536c1cb8fa702a1156e159295ff420355f72e8b) | large batch withdrawal | false positive | no warning |
| [tx](https://mempool.space/tx/eb692afae43f8246f1b60657271d37e4a4ab183ae542aa7452fa49b2359cf839) | actual address poisoning attack | warning displayed | warning displayed |
| all spends of outputs created in [tx](https://mempool.space/tx/7e9fc6393f24e6d574bfce8c19aac20c1be3c5171a818afb9701483cd3eb5c30) | actual address poisoning attacks | warning displayed for every child transaction | warning displayed for every child transaction |

---
 
The PR also fixes a bug in the similar address group construction, which prevented matching addresses from being correctly color-coded, e.g:
https://mempool.space/tx/4d2127a5c020569648912e1fc192e09822b24369b57081dd50bcb0f6ca71c159 (false positives due to vanity addresses, but this illustrates the point)

| Before | After |
|-|-|
| Screenshot 2025-10-04 at 2 10 28 PM | Screenshot 2025-10-04 at 2 10 08 PM |
| random colors | similar addresses in matching colors |

### DIFF
--- a/frontend/src/app/components/transactions-list/transactions-list.component.ts
+++ b/frontend/src/app/components/transactions-list/transactions-list.component.ts
@@ -404,8 +404,38 @@ export class TransactionsListComponent implements OnInit, OnChanges, OnDestroy {
         ]) {
           const similarity = checkedCompareAddressStrings(address, compareAddr.scriptpubkey_address, addressType as AddressType, this.stateService.network);
           if (similarity?.status === 'comparable' && similarity.score > ADDRESS_SIMILARITY_THRESHOLD) {
-            let group = similarityGroups.get(address) || lastGroup++;
+            // Get or create group numbers for both addresses
+            let group1 = similarityGroups.get(address);
+            let group2 = similarityGroups.get(compareAddr.scriptpubkey_address);
+
+            let group: number;
+            if (group1 !== undefined && group2 !== undefined) {
+              // Both have groups - merge by using the lower group number
+              group = Math.min(group1, group2);
+              // Update all addresses with the higher group number to use the lower one
+              if (group1 !== group2) {
+                const higherGroup = Math.max(group1, group2);
+                for (const [addr, g] of similarityGroups.entries()) {
+                  if (g === higherGroup) {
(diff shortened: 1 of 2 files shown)


mononaut:
also fixes overflow issues when we match very long address prefixes:
(in this case, false positives due to fake scripthashes

before:
Screenshot 2025-10-04 at 2 43 25 PM

after:
Screenshot 2025-10-04 at 2 43 11 PM

mononaut:
also exclude the most obvious fake scripthashes from the address similarity checks to reduce false positives due to trivially similar fake addresses:

Screenshot 2025-10-04 at 2 48 45 PM

PR #6073, opened 2025-10-14
fix block first seen time extraction

mononaut:
this PR fixes some issues with the block first seen time extraction process from #5567 which caused first seen times to be missed for many blocks:
 - looks back twice as far into the `debug.log` file
   - sometimes the logs can be noisy, so by the time we read the file to extract the first seen timestamp the relevant lines not be so close to the tail.
 - also matches "Saw new cmpctblock header hash=<hash>" logs, and falls back to the " UpdateTip: new best=<hash>" log if neither a regular nor compactblock "new header" line is found.
   - currently we only look for "Saw new header hash=<hash>", which isn't always printed if we received a compact block header instead.
   - the fallback to "UpdateTip..." is probably unnecessary since I think one of the header logs should always be present, but I think we can afford the extra string comparison.

### DIFF
--- a/backend/src/utils/file-read.ts
+++ b/backend/src/utils/file-read.ts
@@ -24,7 +24,7 @@ function extractDateFromLogLine(line: string): number | undefined {
 
   const dateStr = dateMatch[0];
   const date = new Date(dateStr);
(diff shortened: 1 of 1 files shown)


mononaut:
thanks!

PR #6074, opened 2025-10-14
fix debug.log permissions

mononaut:


production/freebsd/rc.d/bitcoin
```diff
@@ -88,6 +88,10 @@ bitcoin_precmd()
 {
   bitcoin_checkconfig
 
+  # ensure debug.log exists & override default permissions
+  touch "${bitcoin_data_dir}/debug.log"
+  chmod 644 "${bitcoin_data_dir}/debug.log"
+
```
wiz: please move this into the `bitcoin_checkconfig` function before the `chown -R "${bitcoin_user}:${bitcoin_group}" "${bitcoin_data_dir}"` line so that the touched file will become owned by bitcoin - currently as it's implemented the touched file will be owned by root and unwritable by bitcoind (possibly even causing bitcoind to fail to start)

mononaut: done

PR #5959, opened 2025-07-11
Add /api/address/:address/utxo endpoint

adambor:
This PR adds the `/api/address/:address/utxo` endpoint as specified in the official mempool docs [here](https://mempool.space/docs/api/rest#get-address-utxo) when running with electrum as a backend.

This fixes the known issues some people have been having:
- https://community.umbrel.com/t/mempool-unable-to-get-utxo-of-an-address-endpoint-returning-cannot-get-error/22401
- https://github.com/romanz/electrs/issues/1057

PR #6077, opened 2025-10-14
Support utxo bubble graph on electrum backends

natsoni:
This PR uses the newly added `/api/address/:address/utxo` in electrum backend to display the utxo bubble graph on the address page:

Screenshot 2025-10-14 at 12 31 07

I didn't set an upper limit on the UTXO count for fetching from the endpoint because on electrum backend we can't know the UTXO count before actually querying them, and there is no point trying to predict it based on transaction count. Also if there are too many utxos for a given address, there's a good chance that the /api/address/:address call will fail anyway on electrum backend.

### DIFF
--- a/frontend/src/app/components/address/address.component.html
+++ b/frontend/src/app/components/address/address.component.html
@@ -108,7 +108,7 @@
       </div>
     </ng-container>
 
-    <ng-container *ngIf="(stateService.backend$ | async) === 'esplora' && address && utxos && utxos.length > 2">
+    <ng-container *ngIf="address && utxos && utxos.length > 2">
       <br>
       <div class="title-tx">
         <h2 class="text-left" i18n="address.unspent-outputs">Unspent Outputs</h2>
--- a/frontend/src/app/components/address/address.component.ts
+++ b/frontend/src/app/components/address/address.component.ts
@@ -221,7 +221,7 @@ export class AddressComponent implements OnInit, OnDestroy {
             address.is_pubkey
               ? this.electrsApiService.getScriptHashTransactions$((address.address.length === 66 ? '21' : '41') + address.address + 'ac')
               : this.electrsApiService.getAddressTransactions$(address.address),
-            (utxoCount > 2 && utxoCount <= 500 ? (address.is_pubkey
+            ((utxoCount > 2 && utxoCount <= 500) || this.address.electrum ? (address.is_pubkey
               ? this.electrsApiService.getScriptHashUtxos$((address.address.length === 66 ? '21' : '41') + address.address + 'ac')
               : this.electrsApiService.getAddressUtxos$(address.address)) : of(null)).pipe(
                 catchError(() => {


frontend/src/app/components/address/address.component.html
```diff
@@ -108,7 +108,7 @@ <h2 class="text-left" i18n="address.balance-history">Balance History</h2>
       </div>
     </ng-container>
 
-    <ng-container *ngIf="(stateService.backend$ | async) === 'esplora' && address && utxos && utxos.length > 2">
+    <ng-container *ngIf="address && utxos && utxos.length > 2">
```
softsimon, line 111: Perhaps need keep the backend check for "esplora" and "electrum" to make sure backend isn't `"none"` where no electrum backend is used at all.

mononaut:
In #5959 @softsimon [reported that requesting too many UTXOs can crash some electrum implementations](https://github.com/mempool/mempool/pull/5959#issuecomment-3235852581).

which I thought was borderline acceptable if you have to explicitly make an API call to trigger it, but not great if you can crash your home node server just by browsing random addresses...

especially considering
> I didn't set an upper limit on the UTXO count for fetching from the endpoint because on electrum backend we can't know the UTXO count before actually querying them

mononaut:
on the other hand, it _would_ be cool for people running their own instances to have the UTXO chart where possible, and it's the electrum server's bug not ours 🤷

frontend/src/app/components/address/address.component.html
natsoni: If the backend is `none` then the first query to `/api/address/:address` will fail so I figured the checks would be redundant?

natsoni:
> In #5959 @softsimon [reported that requesting too many UTXOs can crash some electrum implementations](https://github.com/mempool/mempool/pull/5959#issuecomment-3235852581).
> 
> which I thought was borderline acceptable if you have to explicitly make an API call to trigger it, but not great if you can crash your home node server just by browsing random addresses...

It seems that it's already possible on master to crash electrs just by browsing addresses, when querying `blockchain.scripthash.get_balance`:
```
WARN  electrs::electrum] RPC blockchain.scripthash.get_balance failed: failed to get block 00000000000000000237d3d85123ad85014aca18ce0565359b3a96a146de7c29: receiving on an empty and disconnected channel
INFO  electrs::db] closing DB at ./db/bitcoin
Error: electrs failed
Caused by:
    0: sync failed
    1: sending on a disconnected channel
```

And since the call to `listunspent` happens only once `get_balance` returned, it might be reasonable to assume that it won't crash electrs either... 😅

mononaut:
lol

I guess it depends whether `listunspent` is more expensive in electrs than `getbalance` and `gethistory` or not?

wiz:
> it's the electrum server's bug not ours

i would add that it's likely not a bug in the electrum code but instead an ops fail, i.e. insufficient hardware without corresponding limits on pending requests or whatever so it exhausts the limited resources and gets killed by the OS

PR #6081, opened 2025-10-19
Refactor block first seen extraction from logs

natsoni:
This draft PR refactors the logic for extracting a block’s first seen time from Core logs, since current implementation does not index the first seen time for stale blocks, or any block mined while the mempool backend is offline.

To fix that the first seen time is extracted on-the-fly when the block gets indexed.
- if the block is recent we just scan the tail of the logs
- if not, we binary search the relevant log window (takes ~1ms)

We filter out clearly invalid first seen times (for blocks mined while the Core node was offline) by scanning log entries within a time window that is plausible according to consensus: we require the block timestamp to be no more than one hour earlier than the first seen time (an approximation of the median past time) and no more than two hours later. Note that this still produces false positives if the core node is frequently offline, but if this is the case the feature should not be enabled anyway.

To do next:
- [x] Implement background indexing of first seen times to get already indexed blocks' first seen times.

### DIFF
--- a/backend/src/__tests__/api/common.ts
+++ b/backend/src/__tests__/api/common.ts
@@ -37,4 +37,27 @@ describe('Common', () => {
       });
     });
   });
+
+  describe('Effective Fee Statistics', () => {
+    test('returns safe defaults for blocks with only coinbase', () => {
+      const coinbaseTx = { weight: 1000, fee: 0, txid: 'coinbase0' };
+      const result = Common.calcEffectiveFeeStatistics([coinbaseTx]);
+
+      expect(result.medianFee).toBe(0);
+      expect(result.feeRange).toEqual([0, 0, 0, 0, 0, 0, 0]);
+    });
+
+    test('excludes coinbase from fee stats when multiple txs', () => {
+      const coinbaseTx = { weight: 1000, fee: 0, txid: 'coinbase0' };
+      const tx1 = { weight: 400, fee: 100, txid: 'tx1' }; // vsize 100, rate 1 sat/vB
+      const tx2 = { weight: 400, fee: 250, txid: 'tx2' }; // vsize 100, rate 2.5 sat/vB
+
+      const result = Common.calcEffectiveFeeStatistics([coinbaseTx, tx1, tx2]);
+
+      // Verify that coinbase (fee 0) was excluded from stats
+      // Fee range min/max should be > 0 (not affected by coinbase's 0 fee)
+      expect(result.feeRange[0]).toBeGreaterThan(0); // min fee
+      expect(result.feeRange[6]).toBeGreaterThan(0); // max fee
+    });
+  });
 });
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -1030,8 +1030,18 @@ export class Common {
   }
 
   static calcEffectiveFeeStatistics(transactions: { weight: number, fee?: number, effectiveFeePerVsize?: number, txid: string, acceleration?: boolean }[]): EffectiveFeeStats {
-    const sortedTxs = transactions.map(tx => { return { txid: tx.txid, weight: tx.weight, rate: tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4)) }; }).sort((a, b) => a.rate - b.rate);
-    const totalWeight = transactions.reduce((acc, tx) => acc + tx.weight, 0);
+    // return early with safe default values
+    if (transactions.length <= 1) {
+      return {
+        medianFee: 0,
+        feeRange: [0, 0, 0, 0, 0, 0, 0],
+      };
+    }
+    // assume the first transaction is a coinbase if the fee is falsy (0 or undefined)
+    const nonCoinbaseTransactions = transactions[0].fee ? transactions : transactions.slice(1);
+
+    const sortedTxs = nonCoinbaseTransactions.map(tx => { return { txid: tx.txid, weight: tx.weight, rate: tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4)) }; }).sort((a, b) => a.rate - b.rate);
+    const totalWeight = nonCoinbaseTransactions.reduce((acc, tx) => acc + tx.weight, 0);
 
     // include any unused space
     let weightCount = config.MEMPOOL.BLOCK_WEIGHT_UNITS - totalWeight;
@@ -1060,7 +1070,7 @@ export class Common {
     // b) the minimum effective fee rate in the last 2% of transactions (in block order)
     const minFee = Math.min(
       Common.getNthPercentile(1, sortedTxs).rate,
-      transactions.slice(-transactions.length / 50).reduce((min, tx) => { return Math.min(min, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, Infinity)
+      nonCoinbaseTransactions.slice(Math.ceil(nonCoinbaseTransactions.length * 49 / 50)).reduce((min, tx) => { return Math.min(min, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, Infinity)
     );
 
     // maximum effective fee heuristic:
@@ -1069,7 +1079,7 @@ export class Common {
     // b) the maximum effective fee rate in the first 2% of transactions (in block order)
     const maxFee = Math.max(
       Common.getNthPercentile(99, sortedTxs).rate,
-      transactions.slice(0, transactions.length / 50).reduce((max, tx) => { return Math.max(max, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, 0)
+      nonCoinbaseTransactions.slice(0, nonCoinbaseTransactions.length / 50).reduce((max, tx) => { return Math.max(max, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, 0)
     );
 
     return {
@@ -1083,6 +1093,9 @@ export class Common {
   }
 
   static getNthPercentile(n: number, sortedDistribution: any[]): any {
+    if (sortedDistribution.length === 0) {
+      return { rate: 0 };
+    }
     return sortedDistribution[Math.floor((sortedDistribution.length - 1) * (n / 100))];
   }
 
--- a/contributors/JMoises-XCode.txt
+++ b/contributors/JMoises-XCode.txt
@@ -1,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of February 13, 2026.
+
+Signed: JMoises-XCode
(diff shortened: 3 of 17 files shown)


backend/src/utils/file-read.ts
```diff
@@ -36,27 +27,249 @@ function extractDateFromLogLine(line: string): number | undefined {
...
+  const fd = fs.openSync(filePath, 'r');
+  try {
+    const stat = fs.fstatSync(fd);
+    let pos = startOffset;
+    let logTimestamp: number | undefined = undefined;
+
+    const needles = [
+      `Saw new header hash=${hash}`,
+      `Saw new cmpctblock header hash=${hash}`,
+      `Initialized PartiallyDownloadedBlock for block ${hash}`,
```
mononaut: does the "Initialized PartiallyDownloadedBlock for block..." pattern catch any additional blocks missed by "Saw new cmpctblock header..."?

natsoni: I think I've seen new blocks in my debug.log where neither 'Saw new header' nor 'Saw new cmpctblock header' were here, so 'Initialized PartiallyDownloadedBlock for block' was the first log entry for the block. But of course I can't find an example now 😅

mononaut:
one issue I'm seeing testing locally is that this misses perfectly good timestamps for a few hours after a node restart, because:
 - the logs are very noisy immediately after startup.
 - for the next ~two hours, the binary search sets the start of the window *before* the startup spam.
 - the startup spam consumes the entire `MAX_WINDOW_SIZE` before we reach the lines we're actually interested in.

this isn't necessarily a big issue in prod because those nodes aren't restarted too often, but it's a bit of a shame for dev & home node deployments that might stop and start more often.

so idk if it's worth the effort to fix, but since we expect most blocks to arrive close to their reported timestamp, you could do something like:
 - binary search for the `blockTimestamp` instead of the start of the window
 - scan the file "middle-out" (i.e. simultaneously forward & backward) from there until
    - you find the firstSeen
    - hit the start/end of the window
    - exceed MAX_WINDOW_SIZE bytes

natsoni:
Pushed a fix that implements middle out search of the first seen time around the log line closest to the `blockTimestamp`, feel free to test again!

natsoni:
[8e9a26b](https://github.com/mempool/mempool/pull/6081/commits/8e9a26bf465917aa4c6b4124573dc241429d35f5) fixes an unrelated issue where the firstSeen field in `/api/v1/block/hash` is returned as string when the requested block is not in the backend cache, and 

[666239a](https://github.com/mempool/mempool/pull/6081/commits/666239a7a52c68373d493a4aef7023cdf1e2af3f) adds background indexing of block first-seen times:
 - check whether the debug log file has changed since the previous run
   - if it has not changed, only index blocks that have never been processed before (their `first_seen` is `NULL` in the database)
   - if it has changed, also reindex blocks that were previously attempted but failed to resolve a timestamp (marked with a sentinel `first_seen` value)

backend/src/utils/file-read.ts
```diff
@@ -36,27 +27,274 @@ function extractDateFromLogLine(line: string): number | undefined {
...
+    const end = blockTimestamp + 3600; // block time can be up to 1 hour in the past (approximating median past time)
+    const anchor = findTimestampPosition(fd, blockTimestamp);
+    return searchForFirstSeen(fd, anchor, start, end, hash);
+  } catch (e) {
+    logger.debug(`Cannot parse block first seen time from Core logs. Reason: ${e instanceof Error ? e.message : e}`);
+    return 1;
+  } finally {
+    fs.closeSync(fd);
+  }
+}
```
mononaut: it's reasonable to use a magic/sentinel value to represent "explicitly set to null" in the DB instead of adding a new flag column for this, but we should try to avoid letting the magic leak any further than necessary.

e.g. if I were reviewing or changing this snippet
https://github.com/mempool/mempool/blob/666239a7a52c68373d493a4aef7023cdf1e2af3f/backend/src/api/blocks.ts#L942-L943

I'd need to dig into this method and already know we're using magic DB values to a) realize that this method might return `1` instead of a legit timestamp, and b) understand what `1` even means in this context.

Ideally we'd use something native to represent the possible states of this variable (e.g. `undefined | null | number`), and convert to/from the mysql representation as a (de)serialization step as close to the DB as possible.

backend/src/api/blocks.ts
```diff
@@ -907,6 +916,50 @@ class Blocks {
     return await BlocksRepository.$validateChain();
   }
 
+  /**
+   * [INDEXING] Index all blocks first seen time from Bitcoin Core debug logs
+   */
+  public async $indexBlocksFirstSeen(): Promise<void> {
+    const old = this.oldestCoreLogTimestamp;
+    const current = this.getOldestCoreLogTimestamp(true);
+    const hasLogFileChanged = old !== null && current !== old;
```
mononaut: I imagine the two situations where the oldest log timestamp might change are:
1) the log is rotated or truncated
2) the user manually imported an older log file from another node

in the first case the log is unlikely to contain anything we haven't already seen, so maybe it makes more sense to only re-try failed blocks if the oldest timestamp is *older* than we saw previously?

or did you have some other case in mind?

backend/src/api/blocks.ts
```diff
@@ -907,6 +916,50 @@ class Blocks {
...
+    let indexedTotal = 0;
+    logger.debug(`Indexing ${blocks.length} block first seen times${hasLogFileChanged ? ' (log file changed since last run)' : ''}`);
+    for (const block of blocks) {
+      const firstSeen = getBlockFirstSeenFromLogs(block.hash, block.timestamp, current);
+      await BlocksRepository.$saveFirstSeenTime(block.hash, firstSeen);
+
+      const cachedBlock = this.blocks.find(blk => blk.id === block.hash);
+      if (cachedBlock) {
+        cachedBlock.extras.firstSeen = firstSeen;
+      }
```
mononaut: instead of binary-searching every block separately, would it make sense to just scan linearly through the whole log file (or whatever portion is relevant) and backfill first seen times that way?

backend/src/api/blocks.ts
natsoni: There is the edge case of a user who has an old but incomplete log due to the node often being offline, and imports a new log from a fully online node. Even if this new log is *not* older than the previous one we would want to retry all the blocks to fulfill the holes 😅

natsoni: In any case it's quite rare that the logs will change so I suppose we can afford to re-check everything again

backend/src/api/blocks.ts
natsoni: The current implementation is the easy way to re-use the already lengthy code to find the first seen for a given block, but yes I think ideally the logic would be: 
- store all the relevant block hashes in a set or map `missing`
- scan the logs chronologically
- when we see a first seen, check that it's in `missing`
  - if it is, update the DB and remove the hash from `missing`
  
 
 _Edit: the above is probably more efficient that binary-searching every block separately only when a large number of blocks is missing_

mononaut: re. the edit:

yeah I'm assuming that a backfill like this will pretty much always be for a big chunk of contiguous blocks, in which case scanning linearly should always be more efficient (maybe after binary-searching to skip ahead to the start of the relevant region of logs, and then returning once you've passed the end)

natsoni: > maybe after binary-searching to skip ahead to the start of the relevant region of logs, and then returning once you've passed the end

Say the lowest block height we need to index is 1000: I don't think we should assume that starting the scan from block 1000 will include following contiguous blocks (1001, 1002, etc. could have shown up earlier in the logs and were orphaned, so starting at 1000 would skip them)

So to avoid these annoying edge cases I think it's easier and safer to just bulk read the full log in the case we have more than an arbitrary number of first seen times to index (say 10k).

mononaut: you could binary-search for a point ~2 hours prior to the first block header timestamp and start reading from there? if later blocks were seen any earlier than that we'd consider their log times implausible anyway.

natsoni: Yes good point, we can just start the linear scan from the earliest missing timestamp

orangesurf:
+1 for this, could have done with it today

orangesurf:
+1 for this, could have done with it today

mononaut:
linear scan change looks good.

@natsoni if you're happy with this please rebase on master and mark it ready for review

backend/src/utils/file-read.ts
```diff
@@ -36,27 +27,370 @@ function extractDateFromLogLine(line: string): number | undefined {
...
-      logger.err(`Cannot parse block first seen time from Core logs. Reason: ` + (e instanceof Error ? e.message : e));
+        const window = missing.get(hash);
+        if (!window) {
+          continue;
+        }
+
+        missing.delete(hash);
+        if (logTimestamp >= window.start && logTimestamp <= window.end) {
+          firstSeenMap.set(hash, logTimestamp);
+        }
```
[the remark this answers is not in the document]
natsoni: That's because one function scans the logs linearly so the first entry found is guaranteed to be the earliest.

backend/src/repositories/BlocksRepository.ts
```diff
@@ -59,7 +59,7 @@ interface DatabaseBlock {
   utxoSetChange: number;
   utxoSetSize: number;
   totalInputAmt: number;
-  firstSeen: number;
+  firstSeen: string; // UNIX_TIMESTAMP() returns a string when applied to datetime(6)
```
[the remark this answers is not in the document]
natsoni: Should be fine, `UNIX_TIMESTAMP(datetime(6))` is returned as a DECIMAL-like value i.e. a string

backend/src/utils/file-read.ts
```diff
@@ -36,27 +27,370 @@ function extractDateFromLogLine(line: string): number | undefined {
...
+          searchPos -= chunkSize;
+          const slice = buf.subarray(0, bytesRead);
+          chunks.unshift(slice);
+          length += slice.length;
+          count++;
+        }
+      }
+      if (!found && chunks.length > 0) {
+        startPos = searchPos;
+      }
```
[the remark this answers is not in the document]
natsoni: I'm pretty confident that no log line is longer than 20KB

backend/src/utils/file-read.ts
```diff
@@ -36,27 +27,370 @@ function extractDateFromLogLine(line: string): number | undefined {
...
+    }
+
+    if (ts < targetTimestamp) {
+      low = record.nextPos;
+    } else {
+      high = record.startPos;
+    }
+  }
+
+  return Math.min(low, size);
```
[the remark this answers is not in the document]
natsoni: All the code assumes the file being read is a legitimate `debug.log` from Core so this is not an issue (and the loop is bounded by `MAX_ITERATIONS` anyway)

natsoni:
Rebased and addressed copilot suggestions.

I noticed that even with https://github.com/mempool/mempool/pull/6074 merged, block first seen times are still missing in production, so there might another issue preventing the backend to properly read the `debug.log` file or the timestamps. We’ll know if this PR helps once it’s deployed by checking the backend logs for `Unable to read Core debug log file at`.

PR #6089, opened 2025-10-29
fractional fee API

### LINKED ISSUE #6085: [Feature] Add optional sub-sat fee recommendations via query parameter
### Problem
While sub-1 sat/vB fee recommendations may be [inappropriate for normal users](https://x.com/mempool/status/1945694681234604257), many [advanced users](https://x.com/callebtc/status/1980968616402080179) want access to these recommendations without manually checking mempool.space to determine appropriate sub-sat rates.

### Proposed Solution
Add an optional `subsat` query parameter to `/api/v1/fees/recommended` that enables sub-1 sat/vB fee recommendations.

This would allow wallet developers to:
- Give advanced users a toggle to enable sub-sat recommendations
- Provide sub-sat recommendations to all users if they choose
- Keep the current conservative behavior by default

### API Behavior

**Default (current behavior):**
```
GET https://mempool.space/api/v1/fees/recommended
```
```json
{
  "fastestFee": 2,
  "halfHourFee": 1,
  "hourFee": 1,
  "economyFee": 1,
  "minimumFee": 1
}
```

**With sub-sat enabled:**
```
GET https://mempool.space/api/v1/fees/recommended?subsat=1
```
```json
{
  "fastestFee": 2,
  "halfHourFee": 0.5,
  "hourFee": 0.5,
  "economyFee": 0.1,
  "minimumFee": 0.1
}
```

mononaut:
This draft PR resolves #6085 by adding a new API endpoint at `/api/v1/fees/precise` with an optional `?min=<sats/vB>` parameter (default 1).

This new endpoint returns fee recommendations with up to 3.d.p of decimal precision (equivalent to the highest precision possible with integer sat/kvB units), after applying a floor of the `min` query parameter (or 1 sat/vB if min is not specified) and of a new backend config option `config.MEMPOOL.MIN_RECOMMENDED_FEE` (also default 1, but set to 0.1 in prod for bitcoin backends).

The frontend & websocket fee recommendations are unchanged.

#### Comparison
`/api/v1/fees/precise`
```
{
  "fastestFee": 1.203,
  "halfHourFee": 0.697,
  "hourFee": 0.441,
  "economyFee": 0.2,
  "minimumFee": 0.1
}
```

`/api/v1/fees/precise?min=1`
```
{
  "fastestFee": 1.203,
  "halfHourFee": 1,
  "hourFee": 1,
  "economyFee": 1,
  "minimumFee": 1
}
```

`/api/v1/fees/recommended`
```
{
  "fastestFee": 2,
  "halfHourFee": 1,
  "hourFee": 1,
  "economyFee": 1,
  "minimumFee": 1
}
```


wiz:
we should get this into v3.3.0

mononaut:
any preferences re. the API path or config options? cc @orangesurf

orangesurf:
Happy with the path.

It could be worth including the servers minimum fee rate recommendation in the response, so users can know when configuring their fee API what the min they will get is. 

Is it fair to assume the recommended fee rate will always be sufficient for that node to accept for broadcast?

orangesurf (approved):
utACK

mononaut:
> Is it fair to assume the recommended fee rate will always be sufficient for that node to accept for broadcast?

yes, we also enforce a minimum of at least the current `mempoolminfee`

mononaut:
> It could be worth including the servers minimum fee rate recommendation in the response, so users can know when configuring their fee API what the min they will get is.

I need to think this one through a bit more, but might add it as a response header later in a separate PR if that still seems like a good idea.

mononaut:
considering it a bit more, I've changed the default behavior of the `/api/v1/fees/precise` to *not* enforce any additional floor on recommendations unless the `?min` query param is explicitly set. 

this brings the default REST API recommendations in line with the values returned by the websocket in #6090, and I think makes the query param usage more intuitive.

I've also removed the `MIN_RECOMMENDED_FEE` config setting since it now seems unnecessary.

This means the new `precise` endpoint will include sub-sat recommendations in prod by default.  Which should be fine, since users are "opting-in" by using the new endpoint, and better than having to change the default behavior later if market rates are still frequently <1 sat/vB after subsat relay policies are widely adopted.

orangesurf:
Great, that makes more sense to me

wiz (approved):
utACK

PR #6090, opened 2025-10-29
decimal fee recommendations

mononaut:
This draft PR extends #6089 by:
 - sending the new more precise fee recommendations for the websocket `want-data` subscription.
 - increasing the maximum precision of fee recommendations rendered in the UI (up to 2.d.p).
 - adding a new `MIN_RECOMMENDED_FEE` frontend config option (default 1), and overriding recommendations received from the websocket with that minimum value.
 
Even with the default minimum of 1 sat/vB, this has the effect of lowering our recommendations by up to 1 s/vB, since we were rounding up to the next integer before.

before: 
Screenshot 2025-10-29 at 11 20 18 AM
 
after (with default `MIN_RECOMMENDED_FEE = 1`):
Screenshot 2025-10-29 at 11 11 03 AM

after (with `MIN_RECOMMENDED_FEE = 0.1`):
Screenshot 2025-10-29 at 11 10 26 AM

### DIFF
--- a/frontend/src/app/components/fees-box/fees-box.component.html
+++ b/frontend/src/app/components/fees-box/fees-box.component.html
@@ -14,23 +14,23 @@
   <div class="fee-estimation-container">
     <div class="item">
       <div class="card-text">
-        <div class="fee-text"><app-fee-rate [fee]="recommendedFees.economyFee" [dp]="2"></app-fee-rate></div> <span class="fiat"><app-fiat i18n-ngbTooltip="Transaction fee tooltip" ngbTooltip="Based on average native segwit transaction of 140 vBytes" placement="bottom" [value]="recommendedFees.economyFee * 140" ></app-fiat></span>
+        <div class="fee-text"><app-fee-rate [fee]="recommendedFees.economyFee" [dp]="1"></app-fee-rate></div> <span class="fiat"><app-fiat i18n-ngbTooltip="Transaction fee tooltip" ngbTooltip="Based on average native segwit transaction of 140 vBytes" placement="bottom" [value]="recommendedFees.economyFee * 140" ></app-fiat></span>
       </div>
     </div>
     <div class="band-separator"></div>
     <div class="item">
       <div class="card-text">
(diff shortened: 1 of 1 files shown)


softsimon:
It's good that it's configurable, but can we keep current behaviour on mempool.space with full integers and round up? I find that most useful and intuitive. I can't see how less than 99% users want to and will put 2 sat/vB as fee instead of 1.23 in their wallet app (do most wallet support decimals?).

With that said the fee API could be more precise but the frontend how we display it should be more human readable (rounded).

As soon as the 0.1 sat summer is over and we start getting over 10+ sat/vB again I don't see much point in showing decimals. Asking users to write 10.43 sat instead of 11 sat/vB, although I don't know how much decimal points are kept as the value gets higher.

mononaut:
decimal fee recommendations (not just subsat) are quite frequently requested, especially when the fee market is very flat and e.g. 2.1 vs 2.9 sats/vb makes a big difference in placement.

but maybe it should be an advanced preference toggle?

softsimon:
> but maybe it should be an advanced preference toggle?

Not a fan of more toggles. But what about a filter that reduces the decimals the higher the number like 1.24 -> 5.2 and then 10 .. I think we have something like that already somewhere

mononaut:
something like this?
Screenshot 2025-10-31 at 8 27 51 AM

wiz:
concept ACK, please rebase, and let's make the decimals darker so the integers standout, perhaps like how the calculator does it

mononaut:
rebased

mononaut:
and pushed this change to render decimals a bit darker so that the integers stand out, but I'm not sure it's necessarily an improvement

Screenshot 2025-12-02 at 7 37 08 AM

mononaut:
(by default, we won't recommend anything below 1.0 sats/vB even with this PR.

to allow sub-sat recommendations, set e.g. `"MIN_RECOMMENDED_FEE": 0.1` in the `mempool-frontend-config.json`)

mononaut:
pushed changes:
 - removed the minFee query parameter on the precise fees API
 - removed the MIN_RECOMMENDED_FEE frontend config option
 - added a hard floor of
     - 1.0 sats/vB for High Priority
     - 0.5 sats/vB for Medium Priority
 - added a constant factor of
     - 0.5 sats/vB for High Priority
     - 0.25 sats/vB for Medium Priority
     - (brings the higher priority recommendations closer to the old rounded-up integer recommendations on average)
 - changed the decimal rendering again
 
Screenshot 2025-12-02 at 9 04 23 AM

mononaut:
reverted small decimals
Screenshot 2025-12-02 at 9 25 25 AM

mononaut:
only 1 significant figure
Screenshot 2025-12-02 at 9 38 03 AM

PR #6101, opened 2025-11-04
[accelerator] use /accelerator/accelerations/:txid in transaction component

nymkappa:
Replace the call to services API `/accelerator/accelerations/history` with `/accelerator/accelerations/:txid` in the transaction component for better performance


nymkappa:
With @mononaut additional fixes it looks good to me, tested with mempool.space and self hosted and the accelerator flow is not affected.

I'll let you test the PR as well and ACK/merge it if you agree with my self ACK.

mononaut (approved):
Tested ACK @ [512a7f07]

PR #6104, opened 2025-11-06
Update default theme to wiz and rename classic to softsimon

natsoni:
Requires companion PR https://github.com/mempool/mempool.space/pull/1910


softsimon:
concept nACK

wiz theme is definitely too dark to be the new default theme

PR #6105, opened 2025-11-06
Add database integration tests

knorrium:
First stab at database integration tests using a mariadb Docker container and real queries.

`npm run test:integration` only checks the database integration

`npm run test:with-db` builds and runs the backend to verify it can start 

Tests run the migration script to create and populate the database. The tests in `database-migration.test.ts` can be used the validate the end state of the migration.

Future improvements:

- Extract hardcoded sql commands to a shared module so migrations can be composed and tested in isolation
- Load SQL dumps for more realistic validations pre and post migrations
- Support for liquid only migrations
- Support for non-mainnet only migrations

### DIFF
--- a/backend/mempool-config.test.json
+++ b/backend/mempool-config.test.json
@@ -4,7 +4,7 @@
     "NETWORK": "mainnet",
     "BACKEND": "none",
     "ENABLED": true,
-    "HTTP_PORT": 8999,
+    "HTTP_PORT": 8998,
     "SPAWN_CLUSTER_PROCS": 0,
     "API_URL_PREFIX": "/api/v1/",
     "POLL_RATE_MS": 2000,


backend/mempool-config.test.json
```diff
@@ -0,0 +1,163 @@
+{
+  "MEMPOOL": {
+    "OFFICIAL": false,
+    "NETWORK": "mainnet",
+    "BACKEND": "none",
+    "ENABLED": true,
+    "HTTP_PORT": 8999,
```
mononaut: I'd suggest setting a different port here than the default, otherwise I have to stop my local mempool backend when running these tests to avoid
```
uncaughtException: Error: listen EADDRINUSE: address already in use :::8999
```

PR #6111, opened 2025-11-15
Docker PR workflow

knorrium:
Builds, tags and pushes images for PRs with the `docker` label using the version defined in `package.json` and the latest commit short sha (i.e, `v3.3-dev-45047c6f`)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -84,10 +84,14 @@ jobs:
         uses: actions/checkout@v4
 
       # For PRs: use package.json version + short sha as TAG
-      - name: Set TAG from package.json for pull requests
+      - name: Set TAG from service package.json for pull requests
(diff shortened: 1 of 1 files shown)


mononaut (approved):
utACK

nymkappa (approved):
Tested ACK by removing and re-adding the docker tag which properly triggered the build

nymkappa (changes requested):
Partially tested ACK by removing and re-adding the docker tag which properly triggered the build, however the sha seemed incorrect.

When I re-triggered the build, it published 18d3f2ea (https://hub.docker.com/layers/mempool/backend/v3.3-dev-18d3f2ea/images/sha256-634cf517950b0a792d1118b5c542807f5ba7d6f3640a4f6c97602cb2dc6ea13b, commit not attached to the PR) instead of 713f416 (merge commit)

PR #6113, opened 2025-11-17
Prevent accumulation of scheduled indexing runs

natsoni:
I've noticed that after the backend has been running for a few weeks, the indexing task restarts immediately after it finished, instead of waiting the intended one hour:

```
Nov 9 00:36:38 [1098747] DEBUG: Running mining indexer
...
Nov 9 00:38:19 [1098747] DEBUG: Indexing completed. Next run planned at Sun, 09 Nov 2025 01:38:19 GMT
...
Nov 9 00:38:40 [1098747] DEBUG: Running mining indexer
```

It's probably because of uncleared timers accumulating over time for some reason. This PR clears any potential timer when a reindex is scheduled.

### DIFF
--- a/backend/src/indexer.ts
+++ b/backend/src/indexer.ts
@@ -170,6 +170,9 @@ class Indexer {
       return;
     }
 
+    this.runIndexer = false;
+    this.indexerRunning = true;
+
     if (config.FIAT_PRICE.ENABLED) {
       try {
         await priceUpdater.$run();
@@ -183,9 +186,6 @@ class Indexer {
     if (blockchainInfo.blocks !== blockchainInfo.headers) {
       return;
     }
-
-    this.runIndexer = false;
-    this.indexerRunning = true;
 
     logger.debug(`Running mining indexer`);
 


mononaut:
I'm not sure it's possible for timers to accumulate in that way, since all the `indexer.reindex()` method does is flip the `runIndexer` flag to `true` ready for the next call to `indexer.$run()` in the main loop.

possibly what's happening is that we're managing to dispatch multiple overlapping runs of `indexer.$run()` early on because we don't update the state of those `runIndexer` and `indexerRunning` flags until *after* the `priceUpdater` finishes (by which time the main loop might have looped and called `$run` again a few times already)?

https://github.com/mempool/mempool/blob/5d0115593c2bad8da7d835b8b58a846550fb67a9/backend/src/indexer.ts#L152-L174

backend/src/indexer.ts
```diff
@@ -76,10 +77,24 @@ class Indexer {
...
+  private scheduleNextRun(timeout: number): void {
+    if (this.reindexTimeout) { // Only one future run should be planned, so always replace existing timer
+      clearTimeout(this.reindexTimeout);
+    }
+    this.reindexTimeout = setTimeout(() => {
+      this.reindexTimeout = undefined;
+      this.reindex();
+    }, timeout);
+  }
+
```
mononaut, line 96: rather than canceling the existing timeout and scheduling a new one, it's probably better to simply avoid scheduling *duplicate* timeouts.

otherwise if this method is called more frequently than every X seconds, the timeout could keep getting constantly reset before it can ever fire.

```suggestion
  private scheduleNextRun(timeout: number): void {
    if (!this.reindexTimeout) { // Only one future run should be planned, so always replace existing timer
      this.reindexTimeout = setTimeout(() => {
        this.reindexTimeout = undefined;
        this.reindex();
      }, timeout);
    }
  }

```

natsoni:
> I'm not sure it's possible for timers to accumulate in that way, since all the indexer.reindex() method does is flip the runIndexer flag to true ready for the next call to indexer.$run() in the main loop.

I think calling `indexer.reindex()` from outside source like we do during reorg or fast forwarding blocks can accumulate timers, since the call kicks off a run directly and schedules a new timeout, but the 'original' timeout is still waiting, does that make sense? 

What I don't really get is that these accumulated timeouts should disappear over time because we just ignore them if the indexer is already running... so I'm still not sure how the backend ended up in the state it was when I made the PR (the indexer just kept running over and over).

natsoni:
> possibly what's happening is that we're managing to dispatch multiple overlapping runs of `indexer.$run()` early on because we don't update the state of those `runIndexer` and `indexerRunning` flags until _after_ the `priceUpdater` finishes (by which time the main loop might have looped and called `$run` again a few times already)?
> 
> https://github.com/mempool/mempool/blob/5d0115593c2bad8da7d835b8b58a846550fb67a9/backend/src/indexer.ts#L152-L174

Yes good point, I guess updating the `runIndexer` and `indexerRunning` at the beginning of `$run` should fix this.

PR #6122, opened 2025-11-20
Fix bowtie tooltip output on electrum backend

natsoni:
Electrum doesn’t expose the spending outpoint’s block height, so the bowtie tooltip shows wrong data ("spent 33 blocks later" which is just the distance to the tip):

Screenshot 2025-11-20 at 13 58 54

This PR simply hides that block-age detail on the output tooltip when electrum is used as backend.


mononaut (approved):
Tested ACK @ [65dce1f9]

PR #6145, opened 2025-12-04
[docs] fix enterprise cta

nymkappa:
Fixes https://github.com/mempool/mempool.space/issues/1964#event-21353604208


wiz (changes requested):
It should only be hard-coded to mempool.space if it's not isMempoolSpaceBuild - can you check one of these conditions

nymkappa:
> Shouldn't the enterprise link always send the user to [mempool.space/enterprise](http://mempool.space/enterprise) like the about page
> this is also wrong, should be an internal angular route if hybrid build

Will redo

PR #6170, opened 2025-12-17
softsimon theme primary color fix

softsimon:
Correcting a bootstrap primary color in the softsimon theme that was making the toggle states difficult to interpret.

| Before | After |
|----------|----------|
|  Screenshot 2025-12-17 at 21 50 18 | Screenshot 2025-12-17 at 21 50 39  |

### DIFF
--- a/frontend/src/theme-softsimon.scss
+++ b/frontend/src/theme-softsimon.scss
@@ -18,7 +18,7 @@ $gray-700: $fg;
 
 $nav-tabs-link-active-bg: $active-bg;
 
-$primary: #105fb0;
+$primary: #007cfa;
 $secondary: #2d3348;
 $tertiary: #653b9c;
 $success: #1a9436;


frontend/src/theme-softsimon.scss
```diff
@@ -18,7 +18,7 @@ $gray-700: $fg;
 
 $nav-tabs-link-active-bg: $active-bg;
 
-$primary: #105fb0;
+$primary: #007cfa;
```
natsoni, line 21: Changing softsimon primary fixes the button contrast issue but I think the underlying bug is that bootstrap doesn't recompile with the softsimon colors, meaning buttons (and badges and everthing using bootstrap) keep using the wiz colors. 

I guess the proper fix would be to override bootstrap classes in each theme, I'll try it see if that works

softsimon: Yes I realised this is the wrong fix. The primary i correct, it's more the primary "active" that is wrong, but not sure how to override it with scss

softsimon:
Closing because this is the wrong solution

PR #6171, opened 2025-12-17
Accelerate button on unconfirmed transactions in address page

natsoni:
Adds an `Accelerate` button on each unconfirmed transaction in the address page:

https://github.com/user-attachments/assets/01988803-d9d2-47a2-97d8-6ae4a4c8a721

For now the button redirects to the transaction page with acceleration checkout opened


orangesurf:
LGTM 👍

natsoni:
Updated the animation to be softer and similar to the one in block overview tooltip:

https://github.com/user-attachments/assets/0085f335-76e4-412b-9427-bcb216f43313

I could reproduce the jump issue but it happens intermittently. maybe it's some race condition when the accelerator checkout is not loaded yet, but what's weird is that the bug doesn't happen when coming from the quick accelerator at `/accelerator`. I'll see what's going on

natsoni:
Refactored how fragments are handled in the transaction component to trigger the auto scroll once the accelerator checkout is loaded (tested on both builds, WOMM™️):

https://github.com/user-attachments/assets/233bb086-4f72-477b-9d5b-c7866007d0c8

natsoni:
UI issues fixed on small screens:
- the 3 buttons were wrapping, to fix that the accelerate sparkles is hidden and font is smaller
- the scroll offset was a bit off because the nav bar is thicker on mobile

https://github.com/user-attachments/assets/76111ece-03f9-42f9-ad86-cde35332c8a3

An issue on hybrid build, navigating from the address page to `/tx/...#accelerate` doesn’t scroll to the top as seen above. Setting `anchorScrolling: 'disabled'` fixes that but we should check that it doesn’t break other things elsewhere (I see that it's already disabled in the FOSS frontend).

PR #6180, opened 2025-12-23
Angular 21 upgrade experiment

knorrium:
Experimenting with vibe upgrading to Angular 21 using an AI agent nicknamed _Simeon_.

There were several breaking changes with styling, animations, the zone.js deprecation and event dispatching, but most seem to have been addressed.

A few issues encountered along the way:

- Dependencies had to be upgraded with `npm install --legacy-peer-deps`
- Blocks component styling was completely broken and all blocks were stacked together
- Tab selection (i.e, chart timespan, main Goggles filters) did not work
- Blocks keyboard navigation was broken
- Search was broken
- Hovering on the blocks did not show the tx summaries
- Address component did not remove the skeleton loaders after fetching data
- Liquid asset component did not remove the skeleton loaders after fetching data

### DIFF
--- a/frontend/src/styles.scss
+++ b/frontend/src/styles.scss
@@ -256,6 +256,12 @@ body {
   border-radius: 4px !important;
 }
 
+// Bootstrap 5 changed .card-header padding from 12px 20px to 0
+// Override to match Bootstrap 4 / production styling
+.card-header {
+  padding: 12px 20px !important;
+}
+
 // Bootstrap 5 changed .card-body padding from 1.25rem (20px) to 1rem (16px)
 // Override to match Bootstrap 4 / production styling
 // Use shorthand padding with !important to match production computed styles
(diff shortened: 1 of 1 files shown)


knorrium:
Obviously not production ready until the dependency issues are resolved, but worth checking out and running locally.

wiz:
I assume we don't want to use legacy peer deps?

knorrium:
> I assume we don't want to use legacy peer deps?

@wiz correct. this is needed mostly to keep bootstrap 4 as there are a lot of changes in bootstrap 5 which will need to be addressed

wiz:
i see, might as well go for bootstrap 5 then, since this is a longer term project to upgrade deps ?

natsoni:
There are still quite a few padding issues related to bootstrap 5, but the diff being huge with both Angular 21 and bootstrap 5 that it's hard to work on. Maybe it's easier to handle the migration to Bootstrap 5 first and once everything is nice and clean we can go on with Angular 21?

knorrium:
> There are still quite a few padding issues related to bootstrap 5, but the diff being huge with both Angular 21 and bootstrap 5 that it's hard to work on. Maybe it's easier to handle the migration to Bootstrap 5 first and once everything is nice and clean we can go on with Angular 21?

@natsoni that would be ideal but we would need to use the legacy dependencies which is another problem on its own

PR #6182, opened 2025-12-23
Fix frontend deprecation warnings

natsoni:
- 171df606463662dbd965875540ad8c27f5252b72 replaces `@import` with `@use`, switches `darken(...)` to `color.adjust(...)`
- 65fb3c25fcdb2f5b911b39558ab30f219b7505ef moves declarations ahead of `@media` blocks to fix nested-rules warnings
- e75db8d7451caa46fe2b8d81810497bc7b2c7d9d silences the Angular CommonJS warning about `qrcode` dependency. To actually fix the warning we would need to update the dependency we use for qrcodes to ESM instead of common JS. There are a few candidates, one is https://github.com/nuintun/qrcode. It has less dependency and can also decode a QR, which we might need in the future to import PSBTs from QR for example.

### DIFF
--- a/backend/testSetup.ts
+++ b/backend/testSetup.ts
@@ -1,5 +1,20 @@
 jest.mock('./mempool-config.json', () => ({}), { virtual: true });
-jest.mock('./src/logger.ts', () => ({}), { virtual: true });
+jest.mock('./src/logger.ts', () => ({
+  emerg: jest.fn(),
+  alert: jest.fn(),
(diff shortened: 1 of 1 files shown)


mononaut (approved):
Tested ACK @ [45938cd6]

wiz:
since we are about to ship v3.3 are we going to merge this now or later?

mononaut:
I'm happy to merge as soon as the liquid tests pass

PR #6183, opened 2025-12-23
update i18n

mononaut:


wiz:
"update i18n" is ambiguous, i can tell from the files change that it's an update from source, but in the future please specify if you're updating from source or pulling from transifex

PR #6190, opened 2026-01-04
Fix the lightning node rankings dashboard

knorrium:
Screenshot 2026-01-03 at 4 01 23 PM

Screenshot 2026-01-03 at 4 00 56 PM


wiz:
are you saying it's broken in production? because mempool.space/lightning looks fine, how to reproduce?

knorrium:
> are you saying it's broken in production? because mempool.space/lightning looks fine, how to reproduce?

@wiz yes it's broken in production https://mempool.space/lightning/nodes/rankings

wiz:
is it possible to navigate to this page? i think it might be dead code and the component can be deleted? @nymkappa @mononaut

knorrium:
> is it possible to navigate to this page? i think it might be dead code and the component can be deleted? @nymkappa @mononaut

Not directly but you can just take out `/liquidity` or `/connectivity` from the urls linked out in the main dashboard

nymkappa:
> is it possible to navigate to this page? i think it might be dead code and the component can be deleted? @nymkappa @mononaut

Yeah this page was more like a demo for future possible ranking widgets. It is only accessible by manually typing the url. This is also true for https://mempool.space/lightning/nodes/oldest.

Feel free to delete that code.

wiz:
@knorrium thanks for catching and fixing this but yeah, the reason nobody noticed or cared is because it's dead code, wanna use copilot to make a new PR to just delete this

PR #6209, opened 2026-01-13
Improve API Docs Nav

hunicus:
Part of https://github.com/mempool/mempool.space/issues/1945. Primary goals are to maximize the visibility of the full range of the API and emphasize accelerator docs.

- [x] Implement scroll-spy heading highlighting with intersection observer
- [x] Make nav position scroll automatically to keep highlighted endpoint in view
- [ ] Collapse all API Docs sections except currently-visible category (see https://github.com/mempool/mempool/pull/6209#issuecomment-3832113810)
- [x] Move accelerator endpoints toward top of the list

https://github.com/user-attachments/assets/e0de7dc7-152b-4e49-aa2b-258b738ea9d8


hunicus:
@orangesurf not sure we want to make the sidebar collapsible. Originally I thought it would be good to show only categories and expand as the user scrolls but doing so hides quite a bit.

It would also hide the Enterprise logos when we add them next to Enterprise items in the side nav area.

Accelerator endpoints are higher up and more visible, which I guess was the main goal anyway (see video in OP above...think that's sufficient?

[screenshot below is not from the changes in this pr; it's a rendering of how the page would look if categories were hidden by default and expanded on scroll]

Screenshot from 2026-02-01 15-59-28

orangesurf:
Having the collapsed sections gives a great overview of the sections - but I do see what you mean, it's not obvious that there's lots of content behind those section headings.

Perhaps we need could have an icon to indicate that there is content there, like a chevron which points right when the section is closed and down when it's open 

We could also put the number of entries in the section in brackets 

'> Addresses (6)'

hunicus:
I think it's nice to be able to easily take a glance over the full range of endpoints, but I do remember you saying something about people/prospects not knowing there was a side panel and/or that it could scroll (?).

Another approach could be to put "+" symbols in front of the categories with a button to expand/collapse all.
- Close them all by default
- Expand only the category that is currently visible
- Allow user to expand or collapse all items with a button

### Default / User Selects "Collapse All"

docs-expandable

### Expanded / User Selects "Expand All"

docs-expandable-2

orangesurf:
Looks great

PR #6210, opened 2026-01-13
Update electrum RPC responses

orangesurf:
Change docs to 
1. show actual responses 
2. Use a scripthash with fewer tx's as the example


mononaut (approved):
Tested ACK @ [36ffe1e5]

PR #6212, opened 2026-01-13
Bump node to v24.13

knorrium:


wiz (commented):
Yolo utACK - can revert if issues later

wiz (approved):

PR #6238, opened 2026-01-27
[liquid] use bitcoin esplora for peg parsing queries

mononaut:
Adds support for querying a bitcoin mempool/electrs instance from Liquid backends, and uses it for some requests during Liquid peg parsing.


mononaut:
closing this as not useful

PR #6244, opened 2026-01-28
Freeze skeleton loader while Liquid audit is in progress

natsoni:


mononaut (approved):
Tested ACK @ [3e22f853]

Issue #2225, opened 2022-07-31
The overpaid tooltip fee should be rounded


softsimon:
Example TX
f0e3dba19e481b35ade7a67c82b25ccd49ab5389926d28b95a7b8c64b2dcd269

<img width="285" alt="Screen Shot 2022-08-01 at 00 59 20" src="https://user-images.githubusercontent.com/8561090/182048728-e2ec4086-a4e6-4aad-a4ae-814fc16ead1d.png">

We only use single digits for fees on the block recommendations so this tooltip should show `~1 sat/vB` not `1.00 sat/vB`.

At most it should be rounded up to just 1 decimal like `~1.2 sat/vB` but I don't think it's required in this case.

PR #6261, opened 2026-02-06
fix(ui): round overpaid tooltip fee to integer / 1 decimal (#2225)

naomo123:
### **Summary**
This PR fixes the overpaid fee tooltip so it shows rounded values instead of unnecessary decimals (e.g. ~1 sat/vB instead of 1.00 sat/vB), matching the style used in block recommendations.

### **Results**

- Tooltip shows integer when appropriate (e.g. ~1 sat/vB, ~25 sat/vB)

- At most one decimal when needed (e.g. ~1.2 sat/vB)

- Aligns with the “single digits for fees” approach used elsewhere

### **Testing**

- Verified tooltip on transaction pages with "Overpaid" badge (example: tx f0e3dba19e481b35ade7a67c82b25ccd49ab5389926d28b95a7b8c64b2dcd269)

- Confirmed no visual regressions; only the tooltip text format changes

### DIFF
--- a/frontend/src/app/components/fees-box/fees-box.component.html
+++ b/frontend/src/app/components/fees-box/fees-box.component.html
@@ -14,23 +14,23 @@
   <div class="fee-estimation-container">
     <div class="item">
       <div class="card-text">
-        <div class="fee-text"><app-fee-rate [fee]="recommendedFees.economyFee" [dp]="1"></app-fee-rate></div> <span class="fiat"><app-fiat i18n-ngbTooltip="Transaction fee tooltip" ngbTooltip="Based on average native segwit transaction of 140 vBytes" placement="bottom" [value]="recommendedFees.economyFee * 140" ></app-fiat></span>
+        <div class="fee-text"><app-fee-rate [fee]="recommendedFees.economyFee" [dp]="2"></app-fee-rate></div> <span class="fiat"><app-fiat i18n-ngbTooltip="Transaction fee tooltip" ngbTooltip="Based on average native segwit transaction of 140 vBytes" placement="bottom" [value]="recommendedFees.economyFee * 140" ></app-fiat></span>
       </div>
     </div>
     <div class="band-separator"></div>
     <div class="item">
       <div class="card-text">
-        <div class="fee-text"><app-fee-rate [fee]="recommendedFees.hourFee" [dp]="1"></app-fee-rate></div> <span class="fiat"><app-fiat i18n-ngbTooltip="Transaction fee tooltip" ngbTooltip="Based on average native segwit transaction of 140 vBytes" placement="bottom" [value]="recommendedFees.hourFee * 140" ></app-fiat></span>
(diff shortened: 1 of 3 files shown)


nymkappa:
@mononaut while we're on this it looks like the overpaid tooltip is capped at min 1 sats/vb ? Shouldn't we update this to reflect miner's behavior accepting 0.1 sats/vb now ?

natsoni:
> @mononaut while we're on this it looks like the overpaid tooltip is capped at min 1 sats/vb ? Shouldn't we update this to reflect miner's behavior accepting 0.1 sats/vb now ?

Yes that was a temporary measure from 3 months ago until 0.1 sat/vB txs were not reliably relayed. It's easier to get them mined now but I think we should still wait a little more before updating the tooltip, since not all miners are mining these sub-sats txs yet

mononaut (changes requested):
concept ACK

I think `dp` should be set to 2 instead of 1 in the feeRounding pipe to achieve the desired behavior.

the parameter name is slightly misleading, but `dp=2` means:
| input | format string | decimals shown | e.g. |
|-|-|-|-|
| x < 1 | 1.2-2 | 2 | 0.31 |
| 1 <= x < 10 | 1.1-1 | 1 | 4.1 |
| x >= 10 | 1.0-0 | 0 | 15 |

frontend/src/app/shared/pipes/fee-rounding/fee-rounding.pipe.ts
```diff
@@ -10,14 +10,14 @@ export class FeeRoundingPipe implements PipeTransform {
...
+  transform(fee: number, rounding = null, dp = 2): string {
     if (rounding) {
       return formatNumber(fee, this.locale, rounding);
     }
 
-    if (fee >= Math.pow(10, (dp || 3) - 1)) {
+    if (fee >= Math.pow(10, (dp || 2) - 1)) {
       return formatNumber(fee, this.locale, '1.0-0');
-    } else if (fee < Math.pow(10, (dp || 3) - 2)) {
+    } else if (fee < Math.pow(10, (dp || 2) - 2)) {
```
mononaut, line 20: this PR should not change the default value for `dp` in the fee-rounding pipe. doing so breaks the intended precision of fee rates in various other places, such as:

the recent transactions widget:

| original (intended) | | with this PR |
|-|-|-|
| Image | ⮕ | Image |

and the block visualization tooltip:

| original (intended) | | with this PR |
|-|-|-|
| Image | ⮕ | Image |

frontend/src/app/components/fees-box/fees-box.component.html
```diff
@@ -14,23 +14,23 @@
   <div class="fee-estimation-container">
     <div class="item">
       <div class="card-text">
-        <div class="fee-text"><app-fee-rate [fee]="recommendedFees.economyFee" [dp]="1"></app-fee-rate></div> <span class="fiat"><app-fiat i18n-ngbTooltip="Transaction fee tooltip" ngbTooltip="Based on average native segwit transaction of 140 vBytes" placement="bottom" [value]="recommendedFees.economyFee * 140" ></app-fiat></span>
+        <div class="fee-text"><app-fee-rate [fee]="recommendedFees.economyFee" [dp]="2"></app-fee-rate></div> <span class="fiat"><app-fiat i18n-ngbTooltip="Transaction fee tooltip" ngbTooltip="Based on average native segwit transaction of 140 vBytes" placement="bottom" [value]="recommendedFees.economyFee * 140" ></app-fiat></span>
```
mononaut, line 17: changes to this component are also out-of-scope for this PR

mononaut:
closing this PR as it seems like the task turned out to be more complex than anticipated.

Issue #5757, opened 2025-01-27
[Bug] Lower fee rate range = 0


orangesurf:
Block [00000000000000000000deb062c30c0804fe14a0c6ab835ad48c753dd8d13a1e](https://mempool.space/block/00000000000000000000deb062c30c0804fe14a0c6ab835ad48c753dd8d13a1e) has a Fee span = 0 - 4 sat/vB

This is because minFee for the 1st percentile of effective fee rates in a block with 86 transactions is the feerate of the 1st transaction, which in this case is the coinbase. 

https://github.com/mempool/mempool/blob/073fe8e8cd6121d8710e076aafd195479e99b355/backend/src/api/common.ts#L882-L898

I guess we should eliminate the coinbase from the set of sorted transactions first?

mononaut:
resolved by #6263

PR #6263, opened 2026-02-06
[Bug] Lower fee rate range = 0 #5757

jramos0:
The main problem was that the “fee range” (the range of fees per transaction) showed 0 sat/vB as the starting point in specific blocks. This occurred because the system counted the coinbase transaction (the miner's reward, which has no fee) This is because of the set up of minFee for the 1st percentile of effective fee rates in a block with 86 transactions is the feerate of the 1st transaction

Thought process and logic Instead of simply hiding the zero, separate system transactions from user transactions and use logic that identifies and ignores the coinbase transaction before calculating any statistics.

I made sure that this change would not hide very low actual fees. As we can see on the live network today, fees can be as low as 0.13 sat/vB. My logic was to preserve these decimals and only remove the false 0 from the coinbase. I also made sure that if a block is completely empty (only the coinbase), the system does not crash and safely displays a default value.

Local verification I tested this in a local Regtest environment by forcing transactions with very low fees (e.g., 0.15 sat/vB). The result was satisfactory: blocks now show the actual range of user fees (e.g., 1.00 - 1.00 or 0.20 - 5.00 for times like now with low fees) instead of incorrectly starting at 0.00.

I look forward to your feedback on this matter.

Solves #5757

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -1030,8 +1030,18 @@ export class Common {
   }
 
   static calcEffectiveFeeStatistics(transactions: { weight: number, fee?: number, effectiveFeePerVsize?: number, txid: string, acceleration?: boolean }[]): EffectiveFeeStats {
-    const sortedTxs = transactions.map(tx => { return { txid: tx.txid, weight: tx.weight, rate: tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4)) }; }).sort((a, b) => a.rate - b.rate);
-    const totalWeight = transactions.reduce((acc, tx) => acc + tx.weight, 0);
+    // return early with safe default values
+    if (transactions.length <= 1) {
+      return {
+        medianFee: 0,
+        feeRange: [0, 0, 0, 0, 0, 0, 0],
+      };
+    }
+    // assume the first transaction is a coinbase if the fee is falsy (0 or undefined)
+    const nonCoinbaseTransactions = transactions[0].fee ? transactions : transactions.slice(1);
+
+    const sortedTxs = nonCoinbaseTransactions.map(tx => { return { txid: tx.txid, weight: tx.weight, rate: tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4)) }; }).sort((a, b) => a.rate - b.rate);
+    const totalWeight = nonCoinbaseTransactions.reduce((acc, tx) => acc + tx.weight, 0);
 
     // include any unused space
     let weightCount = config.MEMPOOL.BLOCK_WEIGHT_UNITS - totalWeight;
@@ -1060,7 +1070,7 @@ export class Common {
     // b) the minimum effective fee rate in the last 2% of transactions (in block order)
     const minFee = Math.min(
       Common.getNthPercentile(1, sortedTxs).rate,
-      transactions.slice(-transactions.length / 50).reduce((min, tx) => { return Math.min(min, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, Infinity)
+      nonCoinbaseTransactions.slice(Math.ceil(nonCoinbaseTransactions.length * 49 / 50)).reduce((min, tx) => { return Math.min(min, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, Infinity)
     );
 
     // maximum effective fee heuristic:
@@ -1069,7 +1079,7 @@ export class Common {
     // b) the maximum effective fee rate in the first 2% of transactions (in block order)
     const maxFee = Math.max(
       Common.getNthPercentile(99, sortedTxs).rate,
-      transactions.slice(0, transactions.length / 50).reduce((max, tx) => { return Math.max(max, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, 0)
+      nonCoinbaseTransactions.slice(0, nonCoinbaseTransactions.length / 50).reduce((max, tx) => { return Math.max(max, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, 0)
     );
 
     return {
@@ -1083,6 +1093,9 @@ export class Common {
   }
 
   static getNthPercentile(n: number, sortedDistribution: any[]): any {
+    if (sortedDistribution.length === 0) {
+      return { rate: 0 };
+    }
     return sortedDistribution[Math.floor((sortedDistribution.length - 1) * (n / 100))];
   }
 
--- a/backend/src/__tests__/api/common.ts
+++ b/backend/src/__tests__/api/common.ts
@@ -37,4 +37,27 @@ describe('Common', () => {
       });
     });
   });
+
+  describe('Effective Fee Statistics', () => {
+    test('returns safe defaults for blocks with only coinbase', () => {
+      const coinbaseTx = { weight: 1000, fee: 0, txid: 'coinbase0' };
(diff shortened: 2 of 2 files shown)


backend/src/api/common.ts
```diff
@@ -1029,9 +1029,31 @@ export class Common {
...
+      return true; // Include all others
+    });
+
+    // return safe default values to prevent NaN/undefined errors in empty array (e.g., block with only coinbase)
+    if (nonCoinbaseTransactions.length === 0) {
+      return {
+        medianFee: 0,
+        feeRange: [0, 0, 0, 0, 0, 0, 0],
+      };
+    }
```
mononaut: The `vin[0].is_coinbase` check seems unnecessary, since only the very first transaction could be a coinbase, and the second check should catch that case.

Then, we can simplify from a `filter` to a simple conditional slice:

```suggestion
    // return early with safe default values
    if (transactions.length <= 1) {
      return {
        medianFee: 0,
        feeRange: [0, 0, 0, 0, 0, 0, 0],
      };
    }

    // assume the first transaction is a coinbase if the fee is falsy (0 or undefined)
    const nonCoinbaseTransactions = transactions[0].fee ? transactions : transactions.slice(1);
```

mononaut (changes requested):
concept ACK, but I think we can simplify the implementation a bit (see inline comment)

backend/src/api/common.ts
jramos0: Thanks for the catch, @mononaut.
I overcomplicated the validation by doing a redundant check. I'll follow your recommendation, run the tests, and let you know if everything works out so you can review it with the changes you suggested.

jramos0: @mononaut , I've updated the logic like you suggested, taking advantage of the consensus rule by replacing .filter with an early return and a conditional segment to improve O(1) performance and readability. Everything has worked correctly locally, so I'm waiting for your new comments and a new review.
Thanks for the heads up!

backend/src/api/common.ts
```diff
@@ -1069,7 +1079,7 @@ export class Common {
     // b) the maximum effective fee rate in the first 2% of transactions (in block order)
     const maxFee = Math.max(
       Common.getNthPercentile(99, sortedTxs).rate,
-      transactions.slice(0, transactions.length / 50).reduce((max, tx) => { return Math.max(max, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, 0)
+      nonCoinbaseTransactions.slice(0, nonCoinbaseTransactions.length / 50).reduce((max, tx) => { return Math.max(max, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, 0)
     );
```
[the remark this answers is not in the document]
mononaut: ignore copilot here, the calculations are correct: in either case if the sliced array is empty, the outer min/max clause will resolve to the result of getNthPercentile, which is guaranteed to be a valid fee rate

mononaut:
deleted some copilot comments which made incorrect recommendations based on a misunderstanding of the codebase.

please address the remaining comments and re-request a review from me when you're ready

jramos0:
> deleted some copilot comments which made incorrect recommendations based on a misunderstanding of the codebase.
> 
> please address the remaining comments and re-request a review from me when you're ready

Thank you, because I found those comments a little confusing as they were somewhat contradictory. I will make those changes as soon as possible and let you know

backend/src/api/common.ts
```diff
@@ -1083,6 +1093,9 @@ export class Common {
   }
```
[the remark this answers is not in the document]
jramos0: Fix on the **backend/src/__tests__/api/common.ts** [commit ](https://github.com/mempool/mempool/pull/6263/changes/a3ef57c12e008641c9933efe116dd47e71a79d07), but I honestly tried to replicate the logic of the other tests run in that section, and I'm not sure about that part, so I would like your detailed opinion @mononaut on what I should add or remove, or if it was really necessary to follow the copilot review

backend/src/api/common.ts
```diff
@@ -1029,9 +1029,19 @@ export class Common {
     }
   }
 
-  static calcEffectiveFeeStatistics(transactions: { weight: number, fee?: number, effectiveFeePerVsize?: number, txid: string, acceleration?: boolean }[]): EffectiveFeeStats {
-    const sortedTxs = transactions.map(tx => { return { txid: tx.txid, weight: tx.weight, rate: tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4)) }; }).sort((a, b) => a.rate - b.rate);
-    const totalWeight = transactions.reduce((acc, tx) => acc + tx.weight, 0);
+  static calcEffectiveFeeStatistics(transactions: { weight: number, fee?: number, effectiveFeePerVsize?: number, txid: string, acceleration?: boolean, vin?: Array<{ is_coinbase: boolean }> }[]): EffectiveFeeStats {
```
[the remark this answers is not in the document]
jramos0: Fixed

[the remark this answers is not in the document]
jramos0: Fixed

jramos0:
I made the changes you approved on behalf of copilot, so I await your review
The part I have the most doubts about is the test, so I look forward to your comments.
Thank you.

backend/src/api/common.ts
```diff
@@ -1060,7 +1070,7 @@ export class Common {
     // b) the minimum effective fee rate in the last 2% of transactions (in block order)
     const minFee = Math.min(
       Common.getNthPercentile(1, sortedTxs).rate,
-      transactions.slice(-transactions.length / 50).reduce((min, tx) => { return Math.min(min, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, Infinity)
+      nonCoinbaseTransactions.slice(-nonCoinbaseTransactions.length / 50).reduce((min, tx) => { return Math.min(min, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, Infinity)
```
mononaut: I just noticed a subtle bug (mine, not yours 😅) that we should fix while we're in here - when there are fewer than 50 transactions, this slice returns the *whole* array instead of an empty array because the value here gets rounded to zero and loses the negative index behavior.

to correct this, we should look *forward* to the 98% percentile instead of *back* 2% from the end, like so:

```suggestion
      nonCoinbaseTransactions.slice(Math.ceil(nonCoinbaseTransactions.length * 49 / 50)).reduce((min, tx) => { return Math.min(min, tx.effectiveFeePerVsize || ((tx.fee || 0) / (tx.weight / 4))); }, Infinity)
```

jramos0: Ah, nice catch , that makes sense
I hadn’t considered the -0 → 0 case on small arrays, but your forward-slice suggestion fixes it cleanly. I’ll update the code accordingly 😉

jramos0:
Hey @mononaut 
I fixed the subtle bug , tested it, and it works great so let me know if the rest works or if I have to make more updates

mononaut:
looks good! let's just squash the commit history down to one commit for the CLA and one for the technical change 🙏

jramos0:
> looks good! let's just squash the commit history down to one commit for the CLA and one for the technical change 🙏

Done @mononaut ! Squashed the history into two clean commits (CLA and technical changes). Ready for another look! 🙏
thanks!

mononaut:
merged 🚀

congratulations, and thank you for the contribution!

jramos0:
> merged 🚀
> 
> congratulations, and thank you for the contribution!

Thanks a lot for the feedback and guidance, @mononaut. Glad to have my first contribution merged!
I'm excited to keep contributing to the repo. Thanks again! 🚀

PR #6264, opened 2026-02-06
Add PSBT and tx validation checks

natsoni:
This PR adds PSBT and transaction validation functions.

`transaction.utils.ts` is starting to get bloated, I'll move all the code related to transaction signing, PSBT, etc. to a new file in another PR later.

### DIFF
--- a/frontend/src/app/shared/transaction.utils.ts
+++ b/frontend/src/app/shared/transaction.utils.ts
@@ -1,6 +1,6 @@
 import { TransactionFlags } from '@app/shared/filters.utils';
 import { getVarIntLength, parseMultisigScript, isPoint, parseTapscriptMultisig, parseTapscriptUnanimousMultisig, ScriptInfo } from '@app/shared/script.utils';
-import { Transaction, Vin, Utxo } from '@interfaces/electrs.interface';
+import { Transaction, Vin, Vout } from '@interfaces/electrs.interface';
 import { CpfpInfo, RbfInfo, TransactionStripped } from '@interfaces/node-api.interface';
 import { StateService } from '@app/services/state.service';
 import { hash, Hash } from '@app/shared/sha256';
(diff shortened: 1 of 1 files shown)


frontend/src/app/shared/transaction.utils.ts
```diff
@@ -1643,6 +1643,112 @@ export function encodePsbt(rawTx: Uint8Array, inputs: PsbtKeyValueMap[], outputs
...
 }
 
+export interface PsbtCheck {
+  label: (params: any) => string;
+  run: (tx: Transaction, params: any) => boolean;
+}
+
+export interface PsbtCheckDefinitions {
+  [key: string]: PsbtCheck;
+}
```
mononaut: prefer
```suggestion
export type PsbtCheckDefinitions = Record<string, PsbtCheck>;
```

mononaut: (or, at that point, you could just inline the type in the `PSBT_CHECKS` declaration)

frontend/src/app/shared/transaction.utils.ts
```diff
@@ -1643,6 +1643,112 @@ export function encodePsbt(rawTx: Uint8Array, inputs: PsbtKeyValueMap[], outputs
   return new Uint8Array(result);
 }
 
+export interface PsbtCheck {
+  label: (params: any) => string;
+  run: (tx: Transaction, params: any) => boolean;
+}
```
mononaut: suggest using generics instead of `any`, for slightly tighter type-checking:
```suggestion
export interface PsbtCheck<T> {
  label: (params: T) => string;
  run: (tx: Transaction, params: T) => boolean;
}
```

and there's probably a clever way to avoid repeating yourself too much in the `PSBT_CHECKS` types too

frontend/src/app/shared/transaction.utils.ts
```diff
@@ -1643,6 +1643,112 @@ export function encodePsbt(rawTx: Uint8Array, inputs: PsbtKeyValueMap[], outputs
...
+
+export interface PsbtCheckDefinitions {
+  [key: string]: PsbtCheck;
+}
+
+export const PSBT_CHECKS: PsbtCheckDefinitions = {
+  feeRange: {
+    label: (params: { min: number; max: number }) => `Absolute fee between ${params.min} and ${params.max} sats`,
+    run: (tx: Transaction, params: { min: number; max: number }) => tx.fee >= params.min && tx.fee <= params.max,
+  },
```
mononaut: it would be cleaner (and give you better type checking) if these were _factories_ instead, e.g:

```
type CheckFn = (tx: Transaction) => PsbtCheckResult;

export const PSBT_CHECKS = {
  feeRange(p: { min: number; max: number }): CheckFn {
    return (tx) => ({
      label: `absolute fee between ${p.min} and ${p.max} sats`,
      passed: tx.fee >= p.min && tx.fee <= p.max,
    });
  },
  ...
}
```

then you can do

```
this.psbtChecks = checkPsbt(psbt, [
  PSBT_CHECKS.feeRange({ min: 50, max: 10_000 }),
  ...
]
```

executed in `checkPsbt` like:

```
const results = checks.map(c => c(tx));
```

and never have to worry about passing the wrong params anywhere.

natsoni:
Refactored the checks to factories as suggested, they can be used like this:

```ts
const { tx } = decodeRawTransaction(value, this.network);

const checks = [
  TX_CHECKS.fee(123),
  TX_CHECKS.input(0, { txid: '0000...', vout: 0 }),
  TX_CHECKS.outputValue(1, 12345),
  TX_CHECKS.outputAddress(1, 'bc1p...'),
  TX_CHECKS.locktime(0),
  TX_CHECKS.version(2),
  TX_CHECKS.inputSignature(0, [SighashFlag.ALL, SighashFlag.DEFAULT]),
];

checks.map((check) => check(tx));
```

Issue #5646, opened 2024-11-16
[UI] bad wrapping css on the address page


nymkappa:
It does not look clean when values are high

![Screenshot 2024-11-16 at 11 38 54](https://github.com/user-attachments/assets/0d15f061-9c86-462e-8d95-1a9e51eb6f9f)

most-improve123:
Thanks for reporting this, I agree that large fiat values can look cluttered and reduce readability on the address page.

Since fiat values on mempool.space are mainly informational, one possible improvement could be to switch to a compact representation when values exceed a certain threshold (for example using K / M / B notation, or powers of ten).

This would:
- Improve visual clarity for large balances
- Prevent layout issues caused by long numbers
- Keep the UI consistent without affecting underlying data

If this approach sounds reasonable, I’d be happy to work on a small UI-only improvement for this.

alemanjosh7:
Hi.
I have the following proposal to solve this issue, could you provide feedback if it would be an acceptable or should I change it for another one?

<img width="976" height="249" alt="Image" src="https://github.com/user-attachments/assets/42051797-29c7-461a-9692-466054564e95" />

PR #6268, opened 2026-02-06
fix(ui): prevent large fiat values from wrapping on address page

most-improve123:
This PR improves the readability of large fiat values on the address page.

When fiat amounts become large, they may wrap across multiple lines and
break the layout. This change keeps fiat values on a single line and
prevents visual clutter.

- UI-only change
- No logic or data changes

Related to #5646


nymkappa:
Not only the PR does not fix the issue, it actually makes things even more broken

Prod

Screenshot 2026-02-10 at 19 14 05


This PR

Screenshot 2026-02-10 at 19 14 14

PR #6272, opened 2026-02-06
feat: add showDetails query param and toggle for tx metadata

### LINKED ISSUE #6059: [Feature] Show Transaction Details with query param
like [showFlow](https://github.com/mempool/mempool/pull/2591/commits/619a6bd34d5fe4c088aea856cf11c0ec2ceeca1c) but for the transaction details section 

Image

jramos0:
Following the existing showFlow architectural pattern, I have implemented a way to toggle the visibility of the “Details” section:

URL parameter support: handling of the query parameter ?showDetails=true/false has been added. This allows direct control of the user interface state via the URL.

State management: a new hideDetails state has been integrated into StateService.

Persistence: The user preference is saved in localStorage under the key details-preference, ensuring that the user interface remembers your choice across sessions.

User interface integration: A Show Details/Hide Details toggle button has been added to the transaction header area (next to the Show Diagram button) for easy access.

Since there wasn't much information about the visual aspect of the issue, I am ready to make any necessary changes that may be required.

Closes #6059

### DIFF
--- a/frontend/src/app/components/transaction/transaction.component.html
+++ b/frontend/src/app/components/transaction/transaction.component.html
@@ -180,7 +180,7 @@
 
       <div class="title-buttons">
         <button *ngIf="!flowEnabled" type="button" class="btn btn-outline-info flow-toggle btn-sm" (click)="toggleGraph()" i18n="show-diagram">Show diagram</button>
-        <button type="button" class="btn btn-outline-info btn-sm" (click)="txList.toggleDetails()" i18n="transaction.details|Transaction Details">Details</button>
+        <button type="button" class="btn btn-outline-info btn-sm" (click)="toggleDetailsFromTxPage()" i18n="transaction.details|Transaction Details">Details</button>
       </div>
     </div>
 
--- a/frontend/src/app/components/transaction/transaction.component.ts
+++ b/frontend/src/app/components/transaction/transaction.component.ts
@@ -1,6 +1,7 @@
 import { Component, OnInit, AfterViewInit, OnDestroy, HostListener, ViewChild, ElementRef, Inject, ChangeDetectorRef } from '@angular/core';
 import { ElectrsApiService } from '@app/services/electrs-api.service';
 import { ActivatedRoute, ParamMap, Router } from '@angular/router';
+import { TransactionsListComponent } from '@components/transactions-list/transactions-list.component';
 import {
   switchMap,
   filter,
@@ -144,6 +145,7 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
   hideFlow: boolean = this.stateService.hideFlow.value;
   overrideFlowPreference: boolean = null;
   flowEnabled: boolean;
+  isDetailsOpen: boolean = false;
   tooltipPosition: { x: number, y: number };
   isMobile: boolean;
   firstLoad = true;
@@ -170,6 +172,16 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
   referralCode: string | undefined;
 
   graphContainer: ElementRef;
+  private txList: TransactionsListComponent;
+
+  @ViewChild('txList')
+  set txListSetter(component: TransactionsListComponent | undefined) {
+    if (component) {
+      this.txList = component;
+      this.txList.setDetailsOpen(this.isDetailsOpen);
+    }
+  }
+
   @ViewChild('graphContainer')
   set flowAnchor(element: ElementRef | null | undefined) {
     if (element) {
@@ -208,6 +220,7 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
 
   ngOnInit() {
     this.enterpriseService.page();
+    this.isDetailsOpen = this.route.snapshot.queryParams['showDetails'] === 'true';
 
     const urlParams = new URLSearchParams(window.location.search);
     this.forceAccelerationSummary = !!urlParams.get('cash_request_id');
@@ -621,6 +634,7 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
           const seoDescription = seoDescriptionNetwork(this.stateService.network);
           this.seoService.setDescription($localize`:@@meta.description.bitcoin.transaction:Get real-time status, addresses, fees, script info, and more for ${network}${seoDescription} transaction with txid ${this.txId}.`);
           this.resetTransaction();
+
           return merge(
             of(true),
             this.stateService.connectionState$.pipe(
@@ -860,6 +874,20 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
     this.setGraphSize();
   }
 
+  toggleDetailsFromTxPage(): void {
+    this.isDetailsOpen = !this.isDetailsOpen;
+    this.router.navigate([], {
+      relativeTo: this.route,
+      queryParams: { showDetails: this.isDetailsOpen ? 'true' : null },
+      queryParamsHandling: 'merge',
+      preserveFragment: true,
+      replaceUrl: true,
+    });
+    if (this.txList) {
+      this.txList.setDetailsOpen(this.isDetailsOpen);
+    }
+  }
+
   dismissAccelAlert(): void {
     this.storageService.setValue('accel-cta-type', 'button');
     this.accelerateCtaType = 'button';
@@ -1074,6 +1102,7 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
     this.pool = null;
     this.auditStatus = null;
     this.accelerationPositions = null;
+    this.isDetailsOpen = this.route.snapshot.queryParams['showDetails'] === 'true';
     document.body.scrollTo(0, 0);
     this.isAcceleration = false;
     this.isAccelerated$.next(this.isAcceleration);
@@ -1163,20 +1192,20 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   onAccelerationCompleted(): void {
-    this.router.navigate([], { fragment: null });
+    this.router.navigate([], { fragment: null, queryParamsHandling: 'merge' });
     this.accelerationFlowCompleted = true;
     this.forceAccelerationSummary = false;
   }
 
   closeAccelerator(): void {
-    this.router.navigate([], { fragment: null });
+    this.router.navigate([], { fragment: null, queryParamsHandling: 'merge' });
     this.hideAccelerationSummary = true;
     this.forceAccelerationSummary = false;
     this.storageService.setValue('hide-accelerator-pref', 'true');
   }
 
   openAccelerator(): void {
-    this.router.navigate([], { fragment: 'accelerate' });
+    this.router.navigate([], { fragment: 'accelerate', queryParamsHandling: 'merge' });
     this.accelerationFlowCompleted = false;
     this.hideAccelerationSummary = false;
     this.storageService.setValue('hide-accelerator-pref', 'false');
--- a/frontend/src/app/components/transactions-list/transactions-list.component.ts
+++ b/frontend/src/app/components/transactions-list/transactions-list.component.ts
@@ -532,6 +532,12 @@ export class TransactionsListComponent implements OnInit, OnChanges, OnDestroy {
     }
   }
 
+  setDetailsOpen(open: boolean): void {
+    if (open !== this.showDetails$.value) {
+      this.toggleDetails();
+    }
+  }
+
   loadMoreInputs(tx: Transaction): void {
     if (!tx['@vinLoaded'] && !this.txPreview) {
       this.electrsApiService.getTransaction$(tx.txid)


jramos0:
Thanks @natsoni , I really appreciate the feedback. Sorry for not understanding the issue correctly. I'll make the necessary changes to achieve the objective, so you can test it and let me know your thoughts.

jramos0:
I've pushed a new commit with the actual feature. I'm attaching some screenshots of it running locally to prove that the URL parameters are being set in the same way as in the 'showflow'. I look forward to your thoughts and any change requests you may have for the code.
 
 Thanks @natsoni
 
Screenshot 2026-02-10 105512

Screenshot 2026-02-10 110713f

image

natsoni:
@jramos0 can you address/resolve copilot comments? Thanks!

jramos0:
> @jramos0 can you address/resolve copilot comments? Thanks!

Hi @natsoni !

I’ve addressed all the comments raised by Copilot, and the code should now be ready for a fresh review from both you and Copilot.

I also noticed that the CI job “E2E tests for mempool (pull_request)” is currently failing in my pr. I’m investigating it on my side and will push a fix as soon as I understand the root cause — the failure is not entirely clear to me yet.

If you happen to have any hints or context about that test, I’d really appreciate it. Thanks!

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -1098,6 +1128,20 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
...
   }
 
+  toggleDetails() {
+    const showDetails = !this.detailsEnabled;
+    this.stateService.hideDetails.next(!showDetails);
+    this.router.navigate([], {
+      relativeTo: this.route,
+      queryParams: { showDetails: showDetails },
+      queryParamsHandling: 'merge'
+    });
```
mononaut: same-page router.navigate calls need `preserveFragment: true` to avoid overriding fragments/anchor tags set elsewhere on the page like `#accelerate`

```suggestion
    this.router.navigate([], {
      relativeTo: this.route,
      queryParams: { showDetails: showDetails },
      queryParamsHandling: 'merge',
      preserveFragment: true,
    });
```

mononaut: while we're at it, we should add `queryParamsHandling: 'merge'` to the `router.navigate([], {...})` params in `openAccelerator()` and `closeAccelerator()`, so that those navigation events don't override our new query params.

jramos0: Thanks for the feedback, @mononaut 
I overlooked these details, but they have now been corrected

I've implemented the suggested changes in transaction.component.ts and verified the changes. Everything is working as expected in my local environment 

Let me know if there's anything else that needs adjustment!

mononaut: thanks, but we're still missing the `queryParamsHandling: 'merge'` in `openAccelerator()`

frontend/src/app/components/transaction/transaction.component.ts
mononaut: down here ☝️

jramos0: My bad, I thought I had added it, but now I have done so in the new commit, so I look forward to your comments if anything else is missing or if it is already correct
And thank you for being specific.

frontend/src/app/components/transaction/transaction.component.html
```diff
@@ -180,12 +180,13 @@ <h2 i18n="transaction.inputs-and-outputs|Transaction inputs and outputs">Inputs
 
       <div class="title-buttons">
         <button *ngIf="!flowEnabled" type="button" class="btn btn-outline-info flow-toggle btn-sm" (click)="toggleGraph()" i18n="show-diagram">Show diagram</button>
-        <button type="button" class="btn btn-outline-info btn-sm" (click)="txList.toggleDetails()" i18n="transaction.details|Transaction Details">Details</button>
+        <button *ngIf="!detailsEnabled" type="button" class="btn btn-outline-info btn-sm" (click)="toggleDetails()" i18n="transaction.show-details|Show Transaction Details">Show Details</button>
+        <button *ngIf="detailsEnabled" type="button" class="btn btn-outline-info btn-sm" (click)="toggleDetails()" i18n="transaction.hide-details|Hide Transaction Details">Hide Details</button>
```
mononaut: please leave the Details button as it was for now

frontend/src/app/components/transaction/transaction.component.html
```diff
@@ -180,12 +180,13 @@ <h2 i18n="transaction.inputs-and-outputs|Transaction inputs and outputs">Inputs
...
       </div>
     </div>
 
 
-    <app-transactions-list #txList [transactions]="[tx]" [cached]="isCached" [errorUnblinded]="errorUnblinded" [inputIndex]="inputIndex" [outputIndex]="outputIndex" [transactionPage]="true"></app-transactions-list>
+    <app-transactions-list [transactions]="[tx]" [cached]="isCached" [errorUnblinded]="errorUnblinded" [inputIndex]="inputIndex" [outputIndex]="outputIndex" [transactionPage]="true" [detailsVisible]="detailsEnabled"></app-transactions-list>
```
mononaut: I want to leave the `(click)="txList.toggleDetails()` click handler in the original Details button, so we'll need to revert this change too (both keeping the `#txList` template reference, and invoking the `toggleDetails` method instead of binding to `detailsVisible`)

frontend/src/app/services/state.service.ts
```diff
@@ -213,6 +213,7 @@ export class StateService {
   resetScroll$: Subject<boolean> = new Subject<boolean>();
   timeLtr: BehaviorSubject<boolean>;
   hideFlow: BehaviorSubject<boolean>;
+  hideDetails: BehaviorSubject<boolean>;
```
mononaut:
```suggestion
```

it isn't necessary to involve the state service for this, please revert

frontend/src/app/services/state.service.ts
```diff
@@ -369,6 +370,16 @@ export class StateService {
       }
     });
 
+    // Always initialize Details as collapsed (hideDetails = true)
+    // Do NOT read from localStorage to prevent auto-opening on new page loads
+    // Only query param ?showDetails=true or explicit user click should open Details
+    this.hideDetails = new BehaviorSubject<boolean>(true);
```
mononaut:
```suggestion
```

it isn't necessary to involve the state service for this, please revert

frontend/src/app/services/state.service.ts
```diff
@@ -369,6 +370,16 @@ export class StateService {
...
+
+    // Clean up legacy localStorage key if it exists
+    if (this.storageService.getValue('details-preference')) {
+      this.storageService.removeItem('details-preference');
+    }
+
```
mononaut:
```suggestion
```

this localStorage key will only exist for you, from earlier versions of this PR - so the cleanup here doesn't belong in the master branch

frontend/src/app/components/transactions-list/transactions-list.component.ts
```diff
@@ -218,6 +219,13 @@ export class TransactionsListComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   ngOnChanges(changes): void {
+    // Sync detailsVisible input with internal showDetails$ state
+    if (changes.detailsVisible) {
+      if (this.detailsVisible !== this.showDetails$.value) {
+        this.toggleDetails();
+      }
+    }
+
```
mononaut: as mentioned above, I think we should simply invoke `toggleDetails()` directly from the `TransactionComponent` in the Details button click handler

mononaut (changes requested):
on second review, this approach feels overcomplicated:

`TransactionComponent.toggleDetails()` → `StateService.hideDetails` → `TransactionComponent.hideDetails` → `TransactionComponent.setDetailsEnabled()` → `TransactionComponent.detailsEnabled` → `TransactionsListComponent.detailsVisible` → `TransactionsListComponent.toggleDetails` → `TransactionsListComponent.showDetails$`

is an exceptionally roundabout route to toggle a single boolean state 😅

instead, I'd suggest keeping one boolean flag in the Transaction component to track the current state of the details toggle, and:
 - initialize it from the query param (if any) on init
 - and call `toggleDetails` on the TransactionsList component if necessary (you can access the component element via a `ViewChild` for the `#txList` template reference)
 - leave everything else unchanged

alternatively, you could leave the Transaction component itself completely unchanged and move the `showDetails` query param logic into the Transactions List component, since that's where the behavior logically belongs.

I appreciate that the `showFlow` query parameter implementation set a poor example, and these components carry a lot of tech debt that makes this less straightforward than it needs to be - sorry about that!

jramos0:
Thank you for the detailed feedback  @mononaut ! 
I see your point now—the current implementation is indeed a bit of a 'roundabout' for a single boolean toggle 😅.
Like you said I was following the showFlow pattern, I appreciate the apologies but this helped me to learn more so don't worry, and I agree that avoiding the StateService for a direct parent-child interaction is much cleaner.

I will refactor this to use @ViewChild to trigger the toggleDetails() directly on the TransactionsListComponent and handle the initial state from the query param in the parent. This should significantly reduce the complexity and remove the tech debt I was inadvertently adding. I'll ping you again once the simplified version is ready

jramos0:
refacto to avoid roundabout route to Details toggle on #
50b86d9

- Initialized isDetailsOpen from query params in ngOnInit.

- Synchronized state with child component using ViewChild setter to handle async loading.

- Removed indirect state management to ensure a single source of truth.

Ref: @mononaut feedback

frontend/src/app/services/state.service.ts
```diff
@@ -369,6 +369,7 @@ export class StateService {
       }
     });
 
+
```
mononaut: stray newline

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -206,11 +220,11 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
 
   ngOnInit() {
     this.enterpriseService.page();
+    this.isDetailsOpen = this.route.snapshot.queryParams['showDetails'] === 'true';
 
     const urlParams = new URLSearchParams(window.location.search);
     this.forceAccelerationSummary = !!urlParams.get('cash_request_id');
 
-    // Accelerator referral code
```
mononaut, line 215: why is this deleted?

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -852,6 +867,20 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
     this.setGraphSize();
   }
 
+  toggleDetailsFromTxPage(): void {
+    this.isDetailsOpen = !this.isDetailsOpen;
+    this.router.navigate([], {
+      relativeTo: this.route,
+      queryParams: { showDetails: this.isDetailsOpen ? 'true' : 'false' },
```
mononaut:
```suggestion
      queryParams: { showDetails: this.isDetailsOpen ? 'true' : null },
```

slightly cleaner to *unset* the param when it's false instead of cluttering the URL with `showDetails=false`.

(note that we couldn't do the same with the `showFlow` param because the default there is `true`)

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -168,6 +170,18 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
...
   graphContainer: ElementRef;
+  private txList: TransactionsListComponent;
+
+  @ViewChild('txList')
+  set txListSetter(component: TransactionsListComponent | undefined) {
+    if (component) {
+      this.txList = component;
+      if (this.isDetailsOpen) {
+        this.txList.toggleDetails();
+      }
```
mononaut: it would be slightly safer to add an explicit `setDetailsOpen` method on the transaction-list component, and use that instead of relying on an assumed starting position.

e.g. (in `transactions-list.component.ts`)

```
setDetailsOpen(open: boolean): void {
  if (open !== this.showDetails$.value) {
    this.toggleDetails();
  }
}
```

then 
```suggestion
      this.txList.setDetailsOpen(this.isDetailsOpen);
```

jramos0: Done, setDetailsOpen was added to transactions-list.component.ts and the txListSetter now uses it directly instead of the conditional toggleDetails() call. 
When I made this change, I had to fix a problem in resetTransaction(): it unconditionally set isDetailsOpen = false, which was executed after ngOnInit had already read ?showDetails=true
  from the URL, causing the parameter to be ignored when the page was first loaded, as the conditional toggleDetails() call was not being made. It has been replaced with this.isDetailsOpen = this.route.snapshot.queryParams[“showDetails”] === “true”, so that
   each transaction reset correctly reflects the current state of the URL. So you know why this addition was made.

frontend/src/app/components/transaction/transaction.component.ts
jramos0: fixed

jramos0:
Hey @mononaut I've fixed the minor issues and implemented the setDetailsOpen method. I also performed a squash to consolidate all changes into a single clean commit. After running the necessary tests, everything is working as expected. Please let me know if there are any further adjustments needed.

btw, the CLA-Signed test is coming up as incorrect, but I already have my contributor file. Is this normal, or am I doing something wrong?

mononaut (approved):
Tested ACK @ [eef0ef32]

PR #6273, opened 2026-02-07
fix(ui): position block overview tooltip above cursor when hovering tx

OscarG673:
### Summary
When hovering over transactions near the bottom area of the block overview graph, the tooltip was always rendered below the cursor
If there was not enough vertical space, the tooltip was partially cut off, forcing users to scroll to see the full transaction details

### Result
- Tooltip no longer gets cut off near the bottom edge
- No need to scroll to see full transaction details

### DIFF
--- a/contributors/OscarG673
+++ b/contributors/OscarG673
@@ -1,0 +1,3 @@
+I hereby accept the terms of the Contributor License Agreement in the CONTRIBUTING.md file of the mempool/mempool git repository as of February 7, 2026.
+
+Signed: OscarG673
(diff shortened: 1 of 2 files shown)


natsoni:
> When hovering over transactions near the bottom area of the block overview graph, the tooltip was always rendered below the cursor
> If there was not enough vertical space, the tooltip was partially cut off, forcing users to scroll to see the full transaction details

Could you share a screenshot of the issue or add context like device / browser?

nymkappa:
@natsoni you can repro by having a higher page zoom (or a small screen resolution)

### Prod

Screenshot 2026-02-10 at 23-28-06 mempool - Bitcoin Explorer


### This PR

Screenshot 2026-02-10 at 23-28-14 mempool - Bitcoin Explorer

nymkappa:
> concept NACK
> 
> as nymkappa mentions, the feature was originally designed so that the tooltip remains fully contained within the bounds of the visualization, so the current behavior is intended.

Well there is a middle ground where it should be bounded by the block visualization AND the window

nymkappa:
No feedback from the author since we've reviewed the PR, closing

Issue #6007, opened 2025-08-12
[Feature] RBF Diff


orangesurf:
Add a diff to compare RBF'd transaction with it's replacement

PR #6278, opened 2026-02-08
 [Feature] Draft testing  RBF Diff #6007

jramos0:
This is a PR Draft, since it is a new feature, I would like to receive feedback on the visual aspect, as I am not convinced that it follows the UX guidelines defined by the design team. The internal part of the code is already finished and functional.

image

Works #6007


natsoni:
On the UX:
To be more in line with the rest of the app, how about showing the changes in a table similarly to how it's done in the block audit, where the left side would the previous tx, and the right side the current tx:

Screenshot 2026-02-10 at 13 01 39

Some more general comments: 

- we should include the transaction version and locktime in the diff
- IMO it would be better to only show the metrics that got updated in the RBF, for example don't show the tx weight if it didn't change
- when the only change to an output is its value being reduced by the fee bump (see screenshot below), it looks weird to mark it as changed. IMO this should only show up when there’s a real change to the output address or value that isn’t just the fee delta.

Screenshot 2026-02-10 at 13 09 25

jramos0:
Thank you very much @natsoni for all the context and specific instructions cause I agree with you that specific data should only be displayed when the changes are significant. With this, I can continue working on this issue and deliver a high-quality result that meets your expectations.

I'll ping you when I have significant progress ready for you to review.

mononaut:
concept ACK - this is a cool idea!

I agree with natsoni's suggestions re. style and UX.

It's also worth considering how this will work with more complex RBF events and branching timelines:

208249921-2f25720c-78b9-41df-8345-ec8841e06e65

jramos0:
Hey @natsoni and @mononaut, I want to show you the progress I've made so far. In terms of appearance, I've managed to make it look like this. It already looks like a native mempool piece (except for the button I have to change it), and it's interactive. The data that has changed is what will be shown in the RBF diff, meaning that if the weight or locktime hasn't changed, it won't be shown in the table.

Here you can see how it would look with complete RBFs that change a lot of data:
image

image

Y este es un ejemplo de RBF más sencillo con un simple bumpfee
image

Having said that, and with some screenshots, I would like your review of the code and architecture, as well as a little more context and direction so that it can handle more complex RBF events and branching timelines. I will leave more comments about that below.

jramos0:
> concept ACK - this is a cool idea!
> 
> I agree with natsoni's suggestions re. style and UX.
> 
> It's also worth considering how this will work with more complex RBF events and branching timelines:
> 

Concept ACK to the idea of handling the RBF branches @mononaut , 
it's a fascinating challenge. However, after analyzing it, I find myself confused about the UI: If the ‘Tree’ already shows the branches, how should the ‘RBF Diff’ (which is a binary comparison) handle multiple parents or conflicting branches without overwhelming the user?

I have thought of a couple of options to address this, but none that are truly functional, so I would like to hear your thoughts before proceeding.

Thank you!

frontend/src/app/shared/rbf-diff.utils.ts
```diff
@@ -0,0 +1,156 @@
...
+    versionChanged: boolean;
+    locktimeChanged: boolean;
+    oldVersion: number;
+    newVersion: number;
+    oldLocktime: number;
+    newLocktime: number;
+  };
+  inputs: {
+    added: Vin[];
+    removed: Vin[];
```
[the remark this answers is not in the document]
jramos0: Instead of just adding a comment to explain why we store both old and new for unchanged inputs, I simplified unchanged to Vin[]. An input classified as "unchanged" means it spends the same UTXO in both transactions — storing a redundant pair just to document a maybe-someday use case adds noise without adding value. If we ever
  need to surface sequence or witness differences, the right approach is a proper modified inputs category (like outputs already have), not hiding that info inside unchanged. So let me know what do you think of this choice

frontend/src/app/components/rbf-timeline/rbf-timeline.component.html
```diff
@@ -62,6 +62,178 @@
...
+        <tbody>
+          <tr class="diff-head previous">
+            <td colspan="2">
+              <div class="diff-head-content">
+                <span i18n="rbf-diff.previous-tx">Previous Transaction</span>
+                <span class="txid-separator">:</span>
+                <code>{{ selectedOldTx.txid.substring(0, 12) }}...{{ selectedOldTx.txid.substring(selectedOldTx.txid.length - 8) }}</code>
+              </div>
+            </td>
+          </tr>
```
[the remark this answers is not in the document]
jramos0: There is no variation in column count anywhere in the table. No fix needed

frontend/src/app/components/rbf-timeline/rbf-timeline.component.ts
```diff
@@ -217,4 +262,318 @@ export class RbfTimelineComponent implements OnInit, OnChanges {
...
+        i18nKey: 'transaction.vsize',
+        previous: oldVsize,
+        current: newVsize,
+        changed: true,
+        changeType: this.rbfDiff.metrics.vsizeDelta > 0 ? 'negative' : 'positive',
+        percentage: vsizePercentage,
+        isAmount: false,
+        unit: 'vB'
+      });
+    }
```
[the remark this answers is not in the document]
jramos0: not a bug fix, and is not warranted here. No code change needed.

jramos0:
So I made like the mvp version of RBF Diff, and here is what it does.

When a transaction gets replaced, the RBF Timeline now has a button that opens a
side by side comparison of the old and the new version. It only shows the fields
that actually changed like @natsoni  asked, so a simple fee bump shows just the fee and the fee rate
instead of a long list of identical rows.

The important part is the outputs, because you can see the real addresses. If a
replacement sends the money to a different place, that row is marked as an
address change and painted in red. If an output only lost value because the fee
went up, it is marked as fee adjusted and it is not treated as a change, since
that is not a real edit to the output.

You can also compare any two transactions in the tree, not only two that follow
each other. Click one node, then click another one, and the diff opens for that
pair. If you only want two that are next to each other, you can click the line
that joins them and it compares those two directly this came from the request of @mononaut when we have a more complex timelines of rbf.

I think this is useful right now because of the Coldcard sweeps. A lot of those
transactions signal RBF, so people are watching the mempool to find out if a
replacement moved their coins to another address. This feature answers that in
one look call from @orangesurf 

The two videos show the whole flow.

https://github.com/user-attachments/assets/f61978ee-5661-4186-88c4-2e933ef2435c


https://github.com/user-attachments/assets/2e5f429a-3bbb-47cf-9f5d-d6669e65f19b

orangesurf:
Looking good! Some thoughts
- Fee increase is shown in red, I'd expect it to be green
- Transaction size decrease is shown in green, I'd expect it to be red 
- Output (Address changed) wrapping over two lines doesn't look great, perhaps you could just have the label `Output Address` which shows the address bc1qabcabc...xyz123 followed by ↪ or ⚠️ with tooltip saying `Output address changed`.

jramos0:
Here are the recommendations you asked for @orangesurf . I think they make more sense now, and I added the ⚠️ with tooltip saying Output address changed

https://github.com/user-attachments/assets/01d3592e-a97b-4a00-9f7f-c08943cfa8d3


https://github.com/user-attachments/assets/a275a71d-8569-4c0b-b8aa-f06a0c0c39cb

orangesurf:
Looking really clean!

A few suggestions
- It's a little unclear what is selected when either the first or last transaction are selected, is there much benefit in these always having outline colors? 
- The names "Previous" and "Current" transaction could be confused (particularly "Current"). 
- It's not easy to scan across rows due to the duplicate variable name column, please can you explore what it looks like if there is one column for the variable names and then a col for each of the selected transactions
- I feel the added/removed input rows could be removed and have inputs = 1 (-1) if one is removed for example
- I am unclear what the green dot means in the second video next to Total outputs

jramos0:
Thanks @orangesurf . I applied your suggestions and the comparison is much easier to scan now with one shared column for the variable names.

I also simplified the timeline colors. While the diff is open, the normal viewed and mined colors step aside, so the selected nodes only show which side of the comparison they belong to. When the diff is closed, the normal colors come back. The green dot you noticed is the existing mined indicator from the repo, which is why it was confusing together with the comparison colors.

I made the improvement into the added and removed inputs rows to the simpler format like `Inputs 1 (-1)`, which works better.

I'm still not sure about the labels. I changed "Current" to "New", but I couldn't find a good alternative for "Previous". Could you help me with that? Please

I updated the videos with these changes. Let me know what you think or if you see anything else to improve.

https://github.com/user-attachments/assets/3bf72010-180f-43b3-9dfb-da3bc7a2931d


https://github.com/user-attachments/assets/bae40f6e-3164-442e-958b-1faec783fd29


https://github.com/user-attachments/assets/5b3b6bb0-ed47-4587-bbcc-d9b3aa1f1fc7

orangesurf:
Looking great! A small further improvement idea - where the address hasn't changed show the address in the middle (horizontally) between the two columns (as if the columns were merged and the text centered)

jramos0:
@orangesurf I make the change of the address in the middle and look :
image

orangesurf:
I think it makes sense for the output address amounts to be aligned horizontally in a given column

frontend/src/app/shared/rbf-diff.utils.ts
```diff
@@ -0,0 +1,214 @@
...
+    index,
+    matched: false,
+    outputId: out.scriptpubkey_address ?? `scriptpubkey:${out.scriptpubkey}`,
+  }));
+  // Candidate indexes, so each pass looks up its matches instead of rescanning
+  // every remaining output for every output it has to place
+  const byIdAndValue = indexBy(newOutputs, (entry) => `${entry.outputId}|${entry.out.value}`);
+  const byId = indexBy(newOutputs, (entry) => entry.outputId);
+  const byValue = indexBy(newOutputs, (entry) => `${entry.out.value}`);
+  const byIndex = indexBy(newOutputs, (entry) => `${entry.index}`);
```
mononaut: do we need this index? a list indexed by `index` is just an array, and we already have one of those.

you could use `newOutputs[index]` plus a guard on `!matched` to achieve the same thing.

frontend/src/app/shared/rbf-diff.utils.ts
```diff
@@ -0,0 +1,214 @@
...
+
+/**
+ * A FIFO of candidate outputs sharing a key. Keeping the original order and a
+ * cursor means popping returns the first still-unmatched candidate, exactly as a
+ * linear scan would, without re-walking the whole list for every output.
+ */
+interface CandidateQueue {
+  entries: OutputEntry[];
+  cursor: number;
+}
```
mononaut: since the cursor never backtracks, this might be expressed more cleanly as just a plain array where you `shift()` off the consumed elements instead of incrementing a cursor.

frontend/src/app/shared/rbf-diff.utils.ts
```diff
@@ -0,0 +1,214 @@
...
+    index,
+    matched: false,
+    outputId: out.scriptpubkey_address ?? `scriptpubkey:${out.scriptpubkey}`,
+  }));
+  const newOutputs = newTx.vout.map((out, index) => ({
+    out,
+    index,
+    matched: false,
+    outputId: out.scriptpubkey_address ?? `scriptpubkey:${out.scriptpubkey}`,
+  }));
```
mononaut: both of these map functions are identical - you could extract them out to avoid repeating yourself:


```suggestion
  const annotate = (vouts: Vout[]): OutputEntry[] => vouts.map((out, index) => ({
    out,
    index,
    matched: false,
    outputId: out.scriptpubkey_address ?? `scriptpubkey:${out.scriptpubkey}`,
  }));
  const oldOutputs = annotate(oldTx.out);
  const newOutputs = annotate(newTx.vout);
```

frontend/src/app/shared/rbf-diff.utils.ts
```diff
@@ -0,0 +1,214 @@
...
+      addedInputs.push(newInput.vin);
+    }
+  });
+
+  const removedInputs: Vin[] = [];
+  oldInputs.forEach((oldInput, key) => {
+    if (!newInputs.has(key)) {
+      removedInputs.push(oldInput.vin);
+    }
+  });
```
mononaut: I think you can simplify:


```suggestion
  // INPUT COMPARISON
  const keyOf = (vin: Vin) => `${vin.txid}:${vin.vout}`;
  const oldKeys = new Set(oldTx.vin.map(keyOf));
  const newKeys = new Set(newTx.vin.map(keyOf));
  
  const unchangedInputs = newTx.vin.filter((vin) => oldKeys.has(keyOf(vin)));
  const addedInputs = newTx.vin.filter((vin) => !oldKeys.has(keyOf(vin)));
  const removedInputs = oldTx.vin.filter((vin) => !newKeys.has(keyOf(vin)));
```

frontend/src/app/shared/rbf-diff.utils.ts
```diff
@@ -0,0 +1,214 @@
...
+    locktimeChanged: boolean;
+    oldVersion: number;
+    newVersion: number;
+    oldLocktime: number;
+    newLocktime: number;
+  };
+  inputs: {
+    added: Vin[];
+    removed: Vin[];
+    unchanged: Vin[];
```
mononaut: is this ever used?

frontend/src/app/shared/rbf-diff.utils.ts
```diff
@@ -0,0 +1,214 @@
...
+  };
+  outputs: {
+    added: Vout[];
+    removed: Vout[];
+    modified: Array<{ old: Vout; new: Vout; index: number; changeType: 'address' | 'value' | 'both' }>;
+    unchanged: Array<{ old: Vout; new: Vout; index: number }>;
```
mononaut: this one too

mononaut (dismissed):
The feature works great locally, and the code mostly looks good.

I just have a few inline comments to address, plus a broader point that I think a lot of the diff-related logic in rbf-timeline could be extracted into the rbf-diff.utils.ts file instead.

mononaut (changes requested):
.

jramos0:
Hi @mononaut  , thanks for the feedback I'm glad you liked it how it's going the feature. 

As for the code, I do need to clean it up because it's messy at this point sorry for that. Thank you very much for the detailed review I'll fix everything

frontend/src/app/shared/rbf-diff.utils.ts
jramos0: No we didn't neeed it  now is Removed

frontend/src/app/shared/rbf-diff.utils.ts
jramos0: Done
The interface is gone and the indexes are plain arrays now that shift off
whatever is already matched. I left the head in place instead of shifting it out
on return, since an entry sits in several indexes at once and only the caller
setting `matched` retires it.

frontend/src/app/shared/rbf-diff.utils.ts
jramos0: Applied as written

frontend/src/app/shared/rbf-diff.utils.ts
jramos0: Applied

jramos0:
Hey @mononaut all  your comments are done and fix some specific replies are on each thread.,  at the same time I tried to clean even more little parts of my code cause weren't being used

The same commit also carries the output row layout @orangesurf  sketched, with the
shared address centered across both columns and every amount aligned on its own
column cause for me it's also the best desing.

I I'm attaching a video and an image of how it looks in the end so that both of you can give me your final feedback now that the feature is almost ready 

image


https://github.com/user-attachments/assets/708d545e-1da9-480f-997f-6e8389772a04

orangesurf:
LGTM

mononaut (approved):
Tested ACK @ [ce636d0b]

Looks and works great, nice work!

Issue #5660, opened 2024-11-22
Add true min / max feerate values to API


orangesurf:
"feeRange" in the extras field corresponds to 

[minFee, 10th percentile, ,25th percentile, 50th percentile, 75th percentile, 90th percentile, maxFee]

However for blocks with > 1 transaction, the minFee and maxFee used are [calculated](https://github.com/mempool/mempool/blob/073fe8e8cd6121d8710e076aafd195479e99b355/backend/src/api/blocks.ts#L276) to eliminate extreme outliers. 

https://github.com/mempool/mempool/blob/073fe8e8cd6121d8710e076aafd195479e99b355/backend/src/api/common.ts#L882-L898

This can lead to unexpected results with blocks containing large transactions at the extremes:

For example[ block 871466 ](https://mempool.space/api/v1/block/0000000000000000000259b2120b042193eee3090384ce60ba04f507389eb8cd) has the following feeRange:
 
[11, 11, 11.0521327014218, 14.1086261980831, 20.4759615384615, 30, 1010.41666666667]

But more than half of the block (‎2.05 MWU) consists of a transaction with a feeRate of 4.0 sat/vB
 
 Perhaps we could add the following non modified min & max feerate values to the API 
 - minfeerate (matching bitcoin core's getblockstats minfeerate)
 - maxfeerate (matching bitcoin core's getblockstats maxfeerate)
 - effective_minfeerate (use effective fee rates)
 - effective_maxfeerate (use effective fee rates)

orangesurf:
Note: Mononaut is [not convinced](https://github.com/mempool/mempool/pull/6279#issuecomment-3919826211) the spec described is worth the additional complexity.

PR #6279, opened 2026-02-08
Add raw and effective min/max fee stats to block extras

Truja503:
EDIT by nymkappa: Fixes (?) https://github.com/mempool/mempool/issues/5660

This PR improves transparency in block fee data. Previously, feeRange showed only percentiles, which hides extreme transactions. By adding minfeerate, maxfeerate, effective_minfeerate, and effective_maxfeerate:

- True min/max fee rates are preserved alongside percentile metrics.
- Developers can detect anomalies or extreme fees that would otherwise be hidden.
- Aligns with Bitcoin Core’s getblockstats for consistency.

How it works:

- Computes raw and effective fee rates for each transaction.
- Tracks the minimum and maximum for both, skipping invalid or zero-fee txs.
- Returns { minfeerate, maxfeerate, effective_minfeerate, effective_maxfeerate }.

Integration:

- Included in extras in block.ts.
- Works for single or multi-transaction blocks.
- Percentiles remain; now true extremes coexist for accurate analysis.

Fixes:

- Exposes extreme fees that were hidden in blocks like #871466, without changing existing percentile logic.

Screenshot of passing tests:

Screen Shot 2026-02-08 at 01 26 51

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -1081,24 +1081,22 @@ export class Common {
       ].flat(),
     };
   }
-  
- static calcMinMaxFeeRates(
-  transactions: TransactionExtended[]
-): { minfeerate: number; maxfeerate: number; effective_minfeerate: number; effective_maxfeerate: number; } {
-  const minmax = transactions.reduce((a: { min?: number; max?: number; effMin?: number; effMax?: number; }, tx) => ({
-    min: tx.feePerVsize ? Math.min(tx.feePerVsize, a.min ?? tx.feePerVsize) : a.min,
-    max: tx.feePerVsize ? Math.max(tx.feePerVsize, a.max ?? tx.feePerVsize) : a.max,
-    effMin: tx.effectiveFeePerVsize ? Math.min(tx.effectiveFeePerVsize, a.effMin ?? tx.effectiveFeePerVsize) : a.effMin,
-    effMax: tx.effectiveFeePerVsize ? Math.max(tx.effectiveFeePerVsize, a.effMax ?? tx.effectiveFeePerVsize) : a.effMax,
-  }), {});
 
-  return {
-    minfeerate: minmax.min ?? 0,
-    maxfeerate: minmax.max ?? 0,
-    effective_minfeerate: minmax.effMin ?? 0,
-    effective_maxfeerate: minmax.effMax ?? 0,
-  };
-}
+  static calcMinMaxFeeRates(transactions: TransactionExtended[]): 
+    {minFeeRate: number; maxFeeRate: number; effectiveMinFeeRate: number; effectiveMaxFeeRate: number; } {
+      const minmax = transactions.reduce((a: { min?: number; max?: number; effMin?: number; effMax?: number; }, tx) => ({
+      min: tx.feePerVsize ? Math.min(tx.feePerVsize, a.min ?? tx.feePerVsize) : a.min,
+      max: tx.feePerVsize ? Math.max(tx.feePerVsize, a.max ?? tx.feePerVsize) : a.max,
+      effMin: tx.effectiveFeePerVsize ? Math.min(tx.effectiveFeePerVsize, a.effMin ?? tx.effectiveFeePerVsize) : a.effMin,
+      effMax: tx.effectiveFeePerVsize ? Math.max(tx.effectiveFeePerVsize, a.effMax ?? tx.effectiveFeePerVsize) : a.effMax,
+    }), {});
+    return {
+      minFeeRate: minmax.min ?? 0,
+      maxFeeRate: minmax.max ?? 0,
+      effectiveMinFeeRate: minmax.effMin ?? 0,
+      effectiveMaxFeeRate: minmax.effMax ?? 0,
+    };
+  }
   
   static getNthPercentile(n: number, sortedDistribution: any[]): any {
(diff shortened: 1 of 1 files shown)


orangesurf:
[Solves](https://github.com/mempool/mempool/issues/5660)

backend/src/api/common.ts
```diff
@@ -1081,7 +1080,47 @@ export class Common {
...
+      }
+    }
+
+    return {
+      minfeerate: rawMin === Infinity ? 0 : rawMin,
+      maxfeerate: rawMax,
+      effective_minfeerate: effMin === Infinity ? 0 : effMin,
+      effective_maxfeerate: effMax
+    };
+  }
```
mononaut: in the only place `calcMinMaxFeeRates` is currently used, the transactions are of type `TransactionExtended`, which is guaranteed to have the raw fee rate already set as `tx.feePerVsize`.

I think you can then simplify your logic down to a single clean `.reduce(...)`:

```suggestion
  static calcMinMaxFeeRates(
    transactions: TransactionExtended[]
  ): { minfeerate: number; maxfeerate: number; effective_minfeerate: number; effective_maxfeerate: number; } {
    const minmax = transactions.reduce((a: { min?: number; max?: number; effMin?: number; effMax?: number; }, tx) => ({
      min: tx.feePerVsize ? Math.min(tx.feePerVsize, a.min ?? tx.feePerVsize) : a.min,
      max: tx.feePerVsize ? Math.max(tx.feePerVsize, a.max ?? tx.feePerVsize) : a.max,
      effMin: tx.effectiveFeePerVsize ? Math.min(tx.effectiveFeePerVsize, a.effMin ?? tx.effectiveFeePerVsize) : a.effMin,
      effMax: tx.effectiveFeePerVsize ? Math.max(tx.effectiveFeePerVsize, a.effMax ?? tx.effectiveFeePerVsize) : a.effMax,
    }), {});

    return {
      minfeerate: minmax.min ?? 0,
      maxfeerate: minmax.max ?? 0,
      effective_minfeerate: minmax.effMin ?? 0,
      effective_maxfeerate: minmax.effMax ?? 0,
    };
  }
```

backend/src/api/blocks.ts
```diff
@@ -274,14 +270,26 @@ class Blocks {
       extras.segwitTotalTxs = 0;
       extras.segwitTotalSize = 0;
       extras.segwitTotalWeight = 0;
+      extras.minfeerate = 0;
+      extras.maxfeerate = 0;
+      extras.effective_minfeerate = 0;
+      extras.effective_maxfeerate = 0;
```
mononaut: prefer camelCase

mononaut (changes requested):
concept ACK, but I've left some comments inline, and I think this might also need some changes to `BlocksRepository.ts` (in `$saveBlockInDatabase` and `$formatDbBlockIntoExtendedBlock`, for example) to properly handle the new fields.

Truja503:
Refactored and optimized the fee rate calculation logic as suggested by @mononaut, I've just changed some part of the code to make it more optimized. But always having the reduce. 

To ensure consistency and avoid mismatches, I also updated the database layer and the related interfaces in $saveBlockInDatabase and $formatDbBlockIntoExtendedBlock, so the new fields are properly persisted and formatted throughout the pipeline.

backend/src/api/database-migration.ts
```diff
@@ -559,7 +561,7 @@ class DatabaseMigration {
       await this.updateToSchemaVersion(66);
     }
 
-    if (databaseSchemaVersion < 67  && isBitcoin === true) {
+    if (databaseSchemaVersion < 67 && isBitcoin === true) {
```
mononaut: out of scope for this PR

backend/src/api/database-migration.ts
```diff
@@ -1222,6 +1222,16 @@ class DatabaseMigration {
...
+      if (databaseSchemaVersion < 107) {
+          await this.$executeQuery(`
+          ALTER TABLE blocks
+            ADD COLUMN IF NOT EXISTS min_fee_rate BIGINT UNSIGNED NOT NULL DEFAULT 0,
+            ADD COLUMN IF NOT EXISTS max_fee_rate BIGINT UNSIGNED NOT NULL DEFAULT 0,
+            ADD COLUMN IF NOT EXISTS effective_min_fee_rate BIGINT UNSIGNED NOT NULL DEFAULT 0,
+            ADD COLUMN IF NOT EXISTS effective_max_fee_rate BIGINT UNSIGNED NOT NULL DEFAULT 0;
+        `);
+       await this.updateToSchemaVersion(107);
+      }
```
mononaut: fee rate columns should be floats or doubles

mononaut: also the indentation here is wonky, and we usually leave a blank line between migrations for clarity

Truja503:
I've just made the changes, and the previous issues related to whitespace, out-of-scope columns, and type mismatches have been resolved.

I've also reverted the calcMinMaxFeeRates function to the version by @mononaut. After reconsidering carefully, I think this version is more precise and the optimization I attempted wasn’t significant. This implementation works better and provides more accurate values.

mononaut:
this was a solid PR tackling an ambitious task, but on reflection I'm not convinced the spec described in #5660 is worth the additional complexity.

we might revisit the idea in future with a different approach, but for now I'm going to close this PR as it's unlikely to be merged.

nevertheless, I appreciate the effort and the contribution - thank you!

Issue #5709, opened 2024-12-26
[UI] search result show twice the same link (duplicated result)


nymkappa:
![Screenshot 2024-12-26 at 11 05 47](https://github.com/user-attachments/assets/c28d3229-d19b-468e-b030-12b35abbe378)

nymkappa:
Actually this is the correct UI behavior. I got confused because of this other bug https://github.com/mempool/mempool/issues/6296

PR #6283, opened 2026-02-08
[UI] search result show twice the same link (duplicated result) 

JMoises-XCode:
## Description
Fixes duplicate address display in search results when searching by address prefix.

## Problem
When searching for an address by prefix, the search dropdown was showing the same address twice:
1. Once under "[Network] Address" (exact match section)
2. Again under "[Network] Addresses" (prefix search results section)

This occurred because the existing logic only checked for exact string equality (`searchText === addressPrefixSearchResults[0]`), which failed when the search text was a prefix of the full address.

## Solution
Updated the duplicate detection logic to use `startsWith()` instead of strict equality:
```typescript
// Before
addresses: matchesAddress && addressPrefixSearchResults.length === 1 && searchText === addressPrefixSearchResults[0] ? [] : addressPrefixSearchResults

// After
addresses: matchesAddress && addressPrefixSearchResults.length === 1 && addressPrefixSearchResults[0].startsWith(searchText) ? [] : addressPrefixSearchResults
```

Now when there's only one address result that starts with the search text, it's only shown once in the "Address" section, and the "Addresses" section is hidden.

<table>
<tr>
<td width="50%" valign="top">

**Before**

Screenshot 2026-02-09 234430

Duplicate address shown in both sections

</td>
<td width="50%" valign="top">

**After**

Screenshot 2026-02-09 234456

Single address shown only once

</td>
</tr>
</table>

fix: https://github.com/mempool/mempool/issues/5709

### DIFF
--- a/frontend/src/app/components/search-form/search-form.component.ts
+++ b/frontend/src/app/components/search-form/search-form.component.ts
@@ -219,7 +219,7 @@ export class SearchFormComponent implements OnInit {
             blockHash: matchesBlockHash,
             address: matchesAddress,
             publicKey: publicKey,
-            addresses: matchesAddress && addressPrefixSearchResults.length === 1 && searchText === addressPrefixSearchResults[0] ? [] : addressPrefixSearchResults, // If there is only one address and it matches the search text, don't show it in the dropdown
+            addresses: matchesAddress && addressPrefixSearchResults.length === 1 && addressPrefixSearchResults[0].startsWith(searchText) ? [] : addressPrefixSearchResults, // If there is only one address and it matches the search text, don't show it in the dropdown
             otherNetworks: otherNetworks,
             nodes: lightningResults.nodes,
             channels: lightningResults.channels,


nymkappa:
Thank you for the PR. The issue you're trying to fix was opened more than one year ago. I cannot figure out how to reproduce the bug on https://mempool.space/testnet4.

Can you share how you've managed to reproduce this bug? It seems to be is has already been fixed.

Screenshot 2026-02-10 at 09 20 08

JMoises-XCode:
With pleasure. Here I show how to reproduce the problem. 


https://github.com/user-attachments/assets/e4f47fd1-99f3-4411-a658-31fb3d5613d3

nymkappa:
Sorry but actually I realized thanks to your video that if you type a full address, only one link is shown (correct behavior).

If the address is incomplete, then in this case it will show the two sections, which is also the correct behavior.

So I don't think there is actually a bug to fix on that part. Do you agree ?

nymkappa:
@JMoises-XCode see https://github.com/mempool/mempool/issues/6296

JMoises-XCode:
> Do you agree

Yes, I agree.

JMoises-XCode:
> @JMoises-XCode see #6296

On it.

PR #6287, opened 2026-02-10
Update mobile custom dashboard - treasury list 

### LINKED ISSUE #6198: [treasuries dashboard] improve transaction list on mobile
Currently on mobile this is the view we get, this is not very useful

As suggested by wiz, it would be more useful to show the amounts instead of the TXID, and the amount would be a link to the transaction.

Image

algodondev:
# PR Created to resolve #6198 

## Changes to Custom Dashboard - Treasury Transactions plugin
- Removed TXID column on mobile display. 
- Added Amount and Currency/Fiat columns on mobile display
- Made the Amount value to be the link to the transaction information
- Truncate overflow amounts of currency (limit the rendering of too long numbers)
- Added a currency tooltip whenever and only if the amount is being truncated (test data shown in media)

#### Standard mobile view
image

#### Smaller mobile devices view
image

#### Desktop view
image

### DIFF
--- a/frontend/src/app/components/address-transactions-widget/address-transactions-widget.component.scss
+++ b/frontend/src/app/components/address-transactions-widget/address-transactions-widget.component.scss
@@ -19,8 +19,8 @@
   }
   .table-cell-satoshis {
     text-align: right;
-    @media (max-width: 395px) {
-      padding-right: 2.5rem ;
+    @media (max-width: 1100px) {
+      padding-right: 3rem ;
     }
     .satoshis-link {
       display: inline;
@@ -38,17 +38,14 @@
   .table-cell-fiat {
     display: none;
     text-align: right;
-    text-overflow: ellipsis;
-    white-space: nowrap;
-    overflow: hidden;
-    @media (min-width: 395px) {
+    @media (min-width: 1100px) {
       display: table-cell;
     }
   }
   .table-cell-date {
     text-align: right;
-    @media (max-width: 395px) {
-      padding-right: 2.5rem ;
+    @media (max-width: 1100px) {
+      padding-right: 3rem ;
     }
   }
 }


algodondev:
Thanks for the feedback @nymkappa . For the first observation the fixed result should look like this: 
image

I feel like I under estimated the impact of the currency column and may be better to leave it out for this change. If you look at the third image you sent, the amounts are being affect by that column's styles and one of them is overflowing.

What do you think?

nymkappa:
> I feel like I under estimated the impact of the currency column and may be better to leave it out for this change. If you look at the third image you sent, the amounts are being affect by that column's styles and one of them is overflowing.
> 
> What do you think?

Yes it's probably better to only show the fiat value on desktop. There was an issue to add support for power of tens for fiat value https://github.com/mempool/mempool/issues/3214 and I made an attempt at it back then but it did not work out.

If you feel like it you can work on this in a separate PR.

algodondev:
If we decide to keep the currency column, I was thinking of a way to track the "," and, if the value is overflowing, display the suffix depending on how much currency is being displayed e.g.

$100M (100,000,000)
$123B (123,000,000,000)
$100k (100,000 in case thousands get to overflow)

Ideally, I think it would be better to keep the cell's value at 6-7 characters to avoid amounts and/or date overflowing.

algodondev:
Looks good, I'll go and push the most recent changes and take a look at that PR too. Thank you!

Do you think there's anything else to check in this PR? I see a test failed but am not sure if that is expected.

algodondev:
Final result:

iPhone 14 Pro Max preset:
image

iPad Mini preset:
image

iPad Pro preset:
image

nymkappa:
Thanks for giving this a shot but this is not well polished for all screen size. I've made an alternative implementation here https://github.com/mempool/mempool/pull/6291

algodondev:
I appreciate your feedback, thanks for your time during this PR and issue.

Issue #6296, opened 2026-02-11
[search] should not show a "go to" if the typed address is invalid


nymkappa:
<img width="506" height="189" alt="Image" src="https://github.com/user-attachments/assets/d803ab9c-b90c-439d-ba20-93796605d8f4" />

<img width="1196" height="213" alt="Image" src="https://github.com/user-attachments/assets/b805a247-91a8-42c6-bd71-42d3c71ec88e" />

nymkappa:
@JMoises-XCode this is the actual bug, do you want to take a shot at it ?

JMoises-XCode:
> [@JMoises-XCode](https://github.com/JMoises-XCode) this is the actual bug, do you want to take a shot at it ?

Yes @nymkappa , I’ll take a shot at it.

PR #6298, opened 2026-02-11
Search: hide "go to" for invalid addresses

JMoises-XCode:
This PR improves address validation in the search bar by using
checksum-based decoding instead of regex-only matching.

As a result, invalid Bitcoin and Liquid addresses no longer show
a "go to" option in search results.

### Demo
- Before: https://github.com/user-attachments/assets/c107faab-e2d7-4f92-87b5-5750660cf900  
- After: https://github.com/user-attachments/assets/4b28e09c-8618-47e6-9165-dbdab2bd9444  

Fixes [#6296](https://github.com/mempool/mempool/issues/6296)

### DIFF
--- a/frontend/src/app/components/svg-images/svg-images.component.html
+++ b/frontend/src/app/components/svg-images/svg-images.component.html
@@ -138,6 +138,19 @@
       </defs>
     </svg>
   </ng-container>
+  <ng-container *ngSwitchCase="'mempoolSquare'">
+    <svg [class]="class" [style]="style" [attr.width]="width" [attr.height]="height" [attr.viewBox]="viewBox" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path d="M124.706 110.25C124.706 118.849 117.772 125.791 109.183 125.791H15.5236C6.93387 125.791 0 118.849 0 110.25V16.4837C0 7.88416 6.98561 0.942383 15.5236 0.942383H109.183C117.772 0.942383 124.706 7.88416 124.706 16.4837V110.25Z" fill="#2E3349"/>
+      <path d="M0 63.5225V110.25C0 118.849 6.98561 125.791 15.5753 125.791H109.183C117.772 125.791 124.758 118.849 124.758 110.25V63.5225H0Z" [attr.fill]="'url(#paint0_linear' + randomId + ')'"/>
+      <path opacity="0.3" d="M109.909 109.11C109.909 111.026 108.615 112.581 107.011 112.581H90.8665C89.2624 112.581 87.9688 111.026 87.9688 109.11V17.6232C87.9688 15.7065 89.2624 14.1523 90.8665 14.1523H107.011C108.615 14.1523 109.909 15.7065 109.909 17.6232V109.11Z" fill="white"/>
+      <defs>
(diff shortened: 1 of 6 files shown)


JMoises-XCode:
> @JMoises-XCode please use existing regexes. We don't accept additional npm dependencies unless there is absolutely no other choice.

@nymkappa , The current regexes mainly validate address prefixes and basic format, but not the checksum.
This leads to multiple false positives (e.g. truncated or corrupted bech32 addresses still showing "go to").

In practice, the regex mainly checks for a valid prefix (e.g. "tb1")
and a minimal length constraint (>= 9 characters),
but does not validate the checksum.
As a result, many incomplete or invalid addresses still match.

For example, all of the following pass regex validation, but are not valid addresses:

Valid:
tb1qr8xjkrx46yfsch7q2ts2g007haufq48n9pe6qc

Invalid (but still match regex):
tb1qr8xjkrx46yfsch7q2ts2g007haufq48n9pe6q
tb1qr8xjkrx46yfsch7q2thaufq48n9pe6q
tb1q8n9pe6q

I understand the dependency policy, but using regex alone cannot reliably validate
bech32 or base58 addresses. Real validation requires checksum decoding by definition.

That said, I understand the constraints and will keep the existing regex-based approach.
Removing the extra libraries effectively restores the previous behavior,
since the project already relied on regex-only validation before this PR.

JMoises-XCode:
Since regex-based validation cannot reliably validate bech32/base58 addresses by definition,
what would be the preferred long-term approach here?

My original idea was to use bitcoinjs-lib on the frontend for proper checksum validation,
especially since the project already uses it on the [backend](https://github.com/mempool/mempool/tree/master/backend)

I’m happy to stick with regex for now, but would appreciate guidance on whether
a shared validation approach (frontend + backend) would be acceptable in the future.

nymkappa:
Often when we need a specific feature of a library, we literally copy/paste the code. I think for address validation that might not be much code? Maybe @mononaut or @natsoni can jump in

JMoises-XCode:
> Often when we need a specific feature of a library, we literally copy/paste the code. I think for address validation that might not be much code? Maybe @mononaut or @natsoni can jump in

Sounds good — let’s see what @mononaut or @natsoni suggest,

mononaut:
the bech32 and bs58check libraries are both very compact, so it is feasible to copy-paste the implementations rather than adding as dependencies.

that said, the parent library bitcoinjs-lib is already a trusted dependency of the backend, and we probably need client-side address validation for some other upcoming features, so it wouldn't be unreasonable to add these dependencies to the frontend too.

mononaut:
let's complete this PR using the bech32 and bs58check dependencies (not the whole bitcoinjs-lib) for now, and then we can swap them out for an internal implementation later before merging if we decide that's necessary.

JMoises-XCode:
That makes sense 👍
I agree that pulling just bech32 and bs58check is a good middle ground for now.
I'm already updating the implementation so Bitcoin uses the same lightweight validation approach as Liquid.

mononaut:
closing this for now due to the difficulties with proper address validation. we may revisit the idea later if/when we decide how to approach the dependencies.

PR #6302, opened 2026-02-12
[enterprise] add support for testnets in enterprise landing page

nymkappa:
This pull request updates the navigation link for the Enterprise section to include the current network as a query parameter when navigating. This ensures that the selected network context is preserved when users access the Enterprise page.

- Navigation update:
  * In `master-page.component.html`, the Enterprise navigation link now appends the current network value as a query parameter if not on mainnet already

### DIFF
--- a/frontend/src/app/components/master-page/master-page.component.html
+++ b/frontend/src/app/components/master-page/master-page.component.html
@@ -116,7 +116,7 @@
         </a>
       </li>
       <li class="nav-item" routerLinkActive="active" id="btn-enterprise" *ngIf="officialMempoolSpaceBuild">
-        <a class="nav-link" [routerLink]="['/enterprise' | relativeUrl ]" (click)="collapse()">
+        <a class="nav-link" [routerLink]="['/enterprise']" [queryParams]="{network: network.val.length ? network.val : null}" (click)="collapse()">
           <div class="svg-wrapper"><app-svg-images name="nav-enterprise" width="21.59" height="100%"></app-svg-images></div>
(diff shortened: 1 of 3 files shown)


nymkappa:
@orangesurf

frontend/src/app/components/master-page/master-page.component.html
```diff
@@ -116,7 +116,7 @@
         </a>
       </li>
       <li class="nav-item" routerLinkActive="active" id="btn-enterprise" *ngIf="officialMempoolSpaceBuild">
-        <a class="nav-link" [routerLink]="['/enterprise' | relativeUrl ]" (click)="collapse()">
+        <a class="nav-link" [routerLink]="['/enterprise' | relativeUrl ]" [queryParams]="{network: network.val.length ? network.val : null}" (click)="collapse()">
```
mononaut: this still links to e.g. mempool.space/testnet/enterprise?network=testnet, which is not a valid route and so redirects to mempool.space/testnet.

but I think if you just remove the `relativeUrl` pipe it should work ok:
```suggestion
        <a class="nav-link" [routerLink]="['/enterprise']" [queryParams]="{network: network.val.length ? network.val : null}" (click)="collapse()">
```

nymkappa: Weird that I missed that 🤔

nymkappa:
Fixed the issue you've mentioned. I also updated the enterprise landing page CTA link from the doc navigation, that I previously missed.

PR #6308, opened 2026-02-13
Feat: add fiat shortener to currency to prevent overflow in dashboard UI

### LINKED ISSUE #3214: Fix price overflow

algodondev:
Fixes #3214 

### Changes
This PR adds the new "M" suffix to currency amounts greater than 1,000,000 e.g. now 15,940,000 should look like 15.9M. 
It aims to fix the overflow UI issue for large amounts of fiat currency. 

Decided to only use that shortener since it is more universal than using K for thousands, B for billions, etc.

### Post data
The current price formatting and unused fiat-shortener code was removed.

### Screenshots
https://github.com/user-attachments/assets/1e6a3eaa-4cf7-4614-97ba-413fd8af70d3

### DIFF
--- a/frontend/src/app/shared/pipes/fiat-currency.pipe.ts
+++ b/frontend/src/app/shared/pipes/fiat-currency.pipe.ts
@@ -24,7 +24,15 @@ export class FiatCurrencyPipe implements PipeTransform {
     const digits = args[0] || 1;
     const currency = args[1] || this.currency || 'USD';
 
-    if (Math.abs(num) >= 1000) {
+    if (Math.abs(num) >= 1_000_000) {
+      return new Intl.NumberFormat(this.locale, {
+        style: 'currency',
+        currency,
+        notation: 'compact',
+        compactDisplay: 'short',
+        maximumFractionDigits: 2,
+      }).format(num);
+    } else if (Math.abs(num) >= 1000) {
       return new Intl.NumberFormat(this.locale, { style: 'currency', currency, maximumFractionDigits: 0 }).format(num);
     } else {
       return new Intl.NumberFormat(this.locale, { style: 'currency', currency }).format(num);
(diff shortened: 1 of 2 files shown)


algodondev:
My fault on that one. Let me address it. 

I'll also implement the locale-based currency suffix.

algodondev:
Came up with this result, I'll push it in a bit

https://github.com/user-attachments/assets/f26ef3ed-1506-4d54-95f1-0a974fb1cc1c

nymkappa:
> code changes look good.
> 
> The `Intl.NumberFormat` output looks a bit off in some locales though, e.g USD in spanish: Image perhaps it's better with `currencyDisplay: 'narrowSymbol'`? Image
> 
> can we also please squash the PR down into one commit 🙏

I think the issue is that there should NOT be a space between the number and the power of ten. So it should look like `41,62M $US` AFAIK

mononaut:
yeah I agree, but unfortunately that's just how Intl.NumberFormat spits it out 🤷

e.g:
```
Intl.NumberFormat('es', {
  style: 'currency',
  currency: 'USD',
  notation: 'compact',
  compactDisplay: 'short',
  maximumFractionDigits: 2,
}).format(12_345_000)

> '12,35 MUS$'
```

algodondev:
> code changes look good.
> 
> The `Intl.NumberFormat` output looks a bit off in some locales though, e.g USD in spanish: Image perhaps it's better with `currencyDisplay: 'narrowSymbol'`? Image
> 
> can we also please squash the PR down into one commit 🙏

Not a problem, I'll test that approach of using the narrowSymbol parameter

algodondev:
> > code changes look good.
> > The `Intl.NumberFormat` output looks a bit off in some locales though, e.g USD in spanish: Image perhaps it's better with `currencyDisplay: 'narrowSymbol'`? Image
> > can we also please squash the PR down into one commit 🙏
> 
> I think the issue is that there should NOT be a space between the number and the power of ten. So it should look like `41,62M $US` AFAIK

I think there might be a way to manipulate the string result to remove that space so it is displayed like that, I'll give it a shot

mononaut:
thank you for the hard work on this, but considering the difficulties with `Intl.NumberFormat` I think we'll leave this issue for now and revisit it at a later date, so I'll be closing the PR for now.

PR #6312, opened 2026-02-14
Fix: Transaction's ETA changes on reload page

### LINKED ISSUE #6286: [transaction] ETA changes immediately after (re)loading the page
Seen on localhost but the bug is also visible in prod

https://github.com/user-attachments/assets/0563df3a-28ae-482e-80b6-5ce336c1deec

rodribp:
### Description: 
This PR fixes UI glitch where the ETA of a transaction would immediately change  shown in issue: 
Closes #6286.

There are two scenarios:

1. Navigation: When switching from tx A to tx B the ETA of tx A would persist for a moment before updating.
2. On page reload: A mismatch between the local fee-based estimation and the backend's websocket connection time caused an update as soon as the backend provided the actual value.

> It is programmed to use the fee-based estimation as a backup option if the backend value is null. In eta.service.ts line:124.

### The issue:
The issue was caused by the behaviour of `combineLatest` in the parent component. The `mempoolTxPosition$` held the previous transaction memory. This caused the component to render before the backend updated the position.

### The fixes:
I implemented a validation process within the `ETA$` pipeline in the parent component:

- TX IDs Check: If there's a validation mismatch we return null, forcing the Skeleton loader to be kept until there's a match between the backend and client.
- `debounceTime(50)`: I added a 50ms buffer to the pipeline, this gives the backend the necessary time to update position via WebSocket , without this buffer, the local fee based estimation is flashed briefly before the backend value is received.
- `distinctUntilChanged()`: Added to prevent unnecessary re-renders if the ETA output remains the same value.

### Reproduction:
**Before:**
1. Open any transaction with a high ETA. For example in 3rd block at the mempool.
2. Reload the page, you would see the fee-based estimation first.
3. Click in another block > transaction.
4. You would see the ETA flashing the previous value.

**After:**
1. Navigate between transactions.
2. Reload the page and the backend will have enough buffer time to provide the real value.

### Video
https://github.com/user-attachments/assets/4b9c9f8d-b6cb-4d23-b495-52c963624d9c

https://github.com/user-attachments/assets/3e890c90-e9a3-4d14-84a8-e0113b181ccf


frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -833,7 +835,10 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
       this.isAccelerated$,
       this.txChanged$,
     ]).pipe(
+      debounceTime(50),
```
mononaut: I don't think the debounce helps here - better to just ensure all of the necessary data is present inside the `map`, either by explicitly checking, or removing `.pipe(startWith(null))` from the source observables (`combineLatest` will then wait for each observable to emit at least once).

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -833,7 +835,10 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
...
       map(([position, mempoolBlocks, da, isAccelerated]) => {
+        if (!this.tx) return null;
+        if (position && position.txid !== this.tx.txid) return null;
```
mononaut: style nit: we use the [eslint curly rule](https://eslint.org/docs/latest/rules/curly)

also these could be combined into a single if statement

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -844,6 +849,11 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
           isAccelerated,
           this.accelerationPositions,
         );
+      }),
+      distinctUntilChanged((prev: ETA | null, curr: ETA | null) => {
+        if (prev === curr) return true;
+        if (!prev || !curr) return false;
+        return prev.time === curr.time && prev.blocks === curr.blocks;
```
mononaut: as above, all if statements should use curly braces, but actually this can be simplified into a single expression:

```suggestion
        return prev === curr || (prev && curr && prev.time === curr.time && prev.blocks === curr.blocks);
```

nymkappa (approved):
Tested ACK @ 0fd2d42df4978fdd5f02de241045ca5ae9ca6952

Seems to have fixed the reported bug.
But I'll @mononaut decide on the merge based on the diff.

mononaut:
merged 🚀

congratulations, and thank you for the contribution!

Issue #6310, opened 2026-02-13
[docs] new Electrum RPC doc tab is missing on mobile


nymkappa:
### Desktop

<img width="1147" height="164" alt="Image" src="https://github.com/user-attachments/assets/555d5219-98f7-4658-a51d-f262f33b6406" />

### Smaller screen

<img width="400" height="320" alt="Image" src="https://github.com/user-attachments/assets/dfc8a9df-89e3-47ad-a641-4295246ced1e" />

PR #6317, opened 2026-02-14
Fix Electrum RPC tab not showing on mobile

JMoises-XCode:
### What was happening
The Electrum RPC documentation tab was hidden on mobile because it had the `hide-on-mobile` class applied.

### What I changed
I removed this class so the tab is now visible on mobile devices.

### Notes
It's not clear what the original purpose of hiding this tab was, but removing the class fixes the issue and restores access to the documentation on mobile.

On small screens the tabs wrap into a second line, so it might be a good idea to consider a horizontal scroll for the tabs instead of hiding them.
image

fixes: https://github.com/mempool/mempool/issues/6310


mononaut:
> It's not clear what the original purpose of hiding this tab was

good question - it was hidden by @hunicus in #3196, presumably at @wiz's request (since it was tested and merged within 7 minutes). perhaps they can remember why?

nymkappa:
> On small screens the tabs wrap into a second line, so it might be a good idea to consider a horizontal scroll for the tabs instead of hiding them.

Indeed the way it's currently done in your PR (based on your screenshot) is not acceptable UI. Maybe a drop-down would be more adequate for very small screens?

hunicus:
Must have been because of they way the stacked tabs looked, and there wasn't much doc content there anyway until recently.

Maybe a dropdown could work better, or maybe a horizontal scroll?

JMoises-XCode:
Thanks for the feedback! You're absolutely right that the stacked tabs don't look good on mobile.

I'll work on implementing a better solution. I'm thinking of adding a horizontal scroll for the tabs on small screens to keep them in a single line, which seems like the most straightforward approach and maintains consistency with the desktop experience.

Let me know if you have any preferences between horizontal scroll vs dropdown, or if there's a specific pattern you'd like me to follow that's already used elsewhere in the project.

JMoises-XCode:
> Must have been because of they way the stacked tabs looked, and there wasn't much doc content there anyway until recently.
> 
> Maybe a dropdown could work better, or maybe a horizontal scroll?

I’ve changed the approach to rely on the browser’s native scrolling behavior via scrollIntoView, so the active tab is always brought into view when navigating on mobile.

This avoids extra JS logic and keeps the behavior encapsulated in the template.

[demo.webm](https://github.com/user-attachments/assets/d4e0fc0f-ec39-4b17-a744-19a9bd854ce7)

nymkappa:
On a iPhone mini the 3rd tab is out of the screen by default and there are no indication that horizontal scroll is possible

image

---

The smooth scrolling animation does not work if you are on `FAQ` and click on `API - Websocket` (Firefox), the UI just jumps without animation.


https://github.com/user-attachments/assets/f2762632-2345-4783-8872-fffa321253c0

---

Finally, while it was not the goal of this PR, I'm a bit surprised that there was no mention of the actual documentation content being completely broken on mobile. Only the titles are showings and clicking on them does not expend the content.

image

---

Thank you for trying this out but this is not going to work out.

Issue #5614, opened 2024-10-26
[graphs] Block Fees Vs Subsidy does not keep legend selection when switching timeframe


nymkappa:
- open https://mempool.space/graphs/mining/block-fees-subsidy
- click on `Fees` legend to hide hide from the chart
- change the timeframe or reload the page
- `Fees` is now showing again

https://mempool.space/graphs/mining/block-fee-rates is a good example of the expected behavior when preferences are saved in local storage

algodondev:
Good day @nymkappa . I'll give this a shot

algodondev:
Question on this one @nymkappa , if the user has Fee selected in normal mode but then want to see the fiat category, should the previous individual selection be kept? e.g. the fee selection would remain on every mode (normal, fiat, percentage) until the user selects subsidy or want to see both.

The other behavior would be that each category has its own "memory" of selections.

nymkappa:
> Question on this one @nymkappa , if the user has Fee selected in normal mode but then want to see the fiat category, should the previous individual selection be kept? e.g. the fee selection would remain on every mode (normal, fiat, percentage) until the user selects subsidy or want to see both.
> 
> The other behavior would be that each category has its own "memory" of selections.  

Most graph have their own legends so it's probably better and easier to simply have one localstorage entry per graph

algodondev:
I couldn't get what you meant to, could you explain it, please? 

At a first glance I get that it would be the second behavior, "each category has its own memory".

nymkappa:
https://mempool.space/graphs/mining/block-fees-subsidy will have its own local storage key for legends
https://mempool.space/graphs/mining/block-fee-rates will have its own local storage key for legends
... etc

So yes, each ~category~ graph/chart will have it's own legends settings saved in local storage

algodondev:
Got you, thank you!

PR #6318, opened 2026-02-14
Feat: Fees vs subsidy persistence

algodondev:
Fixes: #5614

### Changes
Each graph in /graphs/mining/block-fees-subsidy now has its own key value in local storage. Resulting in, if the user chooses to see only fees graph in the fiat group, on reload the selection will be kept. 

#### Old behavior:
https://github.com/user-attachments/assets/e0846207-3dc4-4eca-958d-c609d96b91c9

#### New behavior:
https://github.com/user-attachments/assets/242b1d54-f0f2-4738-a467-411785cdc751

### DIFF
--- a/frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
+++ b/frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
@@ -50,6 +50,7 @@ export class BlockFeesSubsidyGraphComponent implements OnInit {
   timespan = '';
   chartInstance: any = undefined;
   displayMode: 'normal' | 'fiat' | 'percentage' = 'normal';
+  legendState: { [mode: string]: { subsidy: boolean, fees: boolean } };
   updateZoom = false;
   zoomSpan = 100;
   zoomTimeSpan = '';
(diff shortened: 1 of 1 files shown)


nymkappa (changes requested):
Sorry I must have no been clear in the issue description. Each graph should have its own localstorage entry, which means that

- https://mempool.space/graphs/mining/block-fee-rates has its own legend saved in localstorage
- https://mempool.space/graphs/mining/block-fee-rates has its own legend saved in localstorage
- https://mempool.space/graphs/mining/block-fee-rates has its own legend saved in localstorage

etc...

For example https://mempool.space/graphs/mining/block-fee-rates legends is saved as such in localstorage, and its should be the same for all other charts which are using a similar type of legend, just booleans

image

`{"Min":true,"10th":true,"25th":true,"Median":true,"75th":true,"90th":true,"Max":false}`

The `displayMode` does not need to stored in localstorage.

As suggested by @mononaut, when we switch mode just reset both related legends to true.

algodondev:
Thanks for the clarification guys, I'll work on it

algodondev:
The expected behavior should be like this, right?


https://github.com/user-attachments/assets/6090756f-4d41-4326-9a3f-ed6b45546b77

nymkappa:
> The expected behavior should be like this, right?
> Screen.Recording.2026-02-16.115928.mp4

Hard to stay because your video keeps cutting off but "that seems" to what we're expecting.

algodondev:
Sorry for that. In words, the behavior is the following:
- Store last graph's display mode before reloading (fiat, percentage, btc/normal), that way user is not sent automatically to the first mode (btc/normal) but the one they were looking at
- Store mode's display type before reloading (subsidy, fee), this, within the active mode
- If user changes to another mode, the previous mode goes back to default graph display (subsidy and fees)

nymkappa (approved):
The diff is not super straightforward to follow but it works well and the approach is good enough for me.

Tested ACK @ 366e1f4bede9b05d2473b6271c07ce2dab801947

_Can't merge because PR is pending @mononaut re-review_

mononaut:
it doesn't seem necessary to store separate localStorage items for each mode, especially when we always ignore the saved state and replace with true/true when switching mode:

Screenshot 2026-02-18 at 4 24 14 AM

all we really need is to persist the *current* state when refreshing the page and changing timespans, so instead I'd suggest a single `fees_subsidy_legend` entry like 
```
{ mode: "fiat", state: { "subsidy": true, "fees": false } }
```

nymkappa:
See https://github.com/mempool/mempool/pull/6327

Issue #6289, opened 2026-02-10
[Feature] Address chunking


orangesurf:
Sparrow recently added address chunking to make it easier to compare addresses. 

<img width="1538" height="428" alt="Image" src="https://github.com/user-attachments/assets/f2c18883-bc40-4096-971f-92828271fc4f" />

https://github.com/sparrowwallet/sparrow/issues/1890#issuecomment-3841434674

This works best with
- A monospace font
- High contrast between the chunks
- Small gap between the chunks
- Copy button to the right of addresses (on desktop) 
- Consistent concatenation string lengths if used 

I've made an extension to test this out - you can find it [here](https://github.com/orangesurf/mempool-chunker/tree/main)

<img width="2072" height="1092" alt="Image" src="https://github.com/user-attachments/assets/baad5f1d-1752-4c0d-9d46-23bc4fd69d2e" />

mononaut:
resolved by #6319

PR #6319, opened 2026-02-15
feat: Address chunking by color & spaces

rodribp:
### Summary
This PR implements Address Chunking as shown in issue #6289 to improve the readability of Bitcoin addresses throughout the whole explorer. Breaking addresses into groups of 4 characters using colors and optional spacing.

### Changes

- **New service**: `AddressFormattingService` to manage user preferences.
- **New component**: `AddressFormattingSelectorComponent`, a dropdown in the footer next to themeSelector to toggle between the formatting modes.
- **Custom pipe**: `AddressChunkPipe`, a reactive pipe that handles the logic of the reformatting.

### Modes
1. **Default**: Standard addresses format.
2. **Color only**: Alternating colors every 4 characters.
3. **Color + spacing**: Adds a visual spacing every 4 characters.
4. **Color + spacing + clipboard**: Adds the `app-clipboard` component next to the address in list.

### Notes

- The pipe is impure to ensure instant UI updates when changing settings without requiring a page reload.
- Implemented user-select: none on chuncked addresses to prevent messy manual selections.
- Optimized to be integrated with the existing `app-truncate` logic.

### Videos

https://github.com/user-attachments/assets/ed4fd7f3-01f4-4b55-92e9-208dc8a6ca21

### DIFF
--- a/frontend/src/app/components/transactions-list/transactions-list.component.html
+++ b/frontend/src/app/components/transactions-list/transactions-list.component.html
@@ -93,7 +93,7 @@
                     </ng-container>
                     <ng-container *ngSwitchCase="vin.prevout && vin.prevout.scriptpubkey_type === 'p2pk'">
                       <span>P2PK <a class="address p2pk-address" [routerLink]="['/address/' | relativeUrl, vin.prevout.scriptpubkey.slice(2, -2)]" title="{{ vin.prevout.scriptpubkey.slice(2, -2) }}">
-                        <app-chunked-address [address]="vin.prevout.scriptpubkey.slice(2, -2)" [showClipboard]="true"></app-chunked-address>
+                        <app-chunked-address [address]="vin.prevout.scriptpubkey.slice(2, -2)"></app-chunked-address>
                       </a>
                       <app-clipboard [text]="vin.prevout.scriptpubkey.slice(2, -2)" class="ml-2" size="normal"></app-clipboard></span>
                     </ng-container>


mononaut (changes requested):
The address chunking itself looks good, but there are a few issues with the approach:

1) The pattern of passing *every* bit of truncated text through the new `addressChunk` pipe isn't great for code separation - the truncate component is supposed to be generic, but now every instance carries an extra overhead of injecting and subscribing to `AddressFormattingService`.

2) The impure pipe approach isn't appropriate for this feature - we often have to render a very large number of addresses, and this way every single one will have to re-run `transform()` on every single change detection cycle.

3) constructing the address chunks as raw unsanitized innerHTML has some potential XSS concerns, but mainly just isn't very readable/maintainable.

A better approach which should resolve each of these problems would be to make a new dedicated "chunked-address" component with its own truncation logic, and construct the chunks using native Angular template syntax.

---

separately, although the footer dropdown is a great demo tool for quickly comparing the different formatting options, it doesn't belong in the final PR

rodribp:
Hi @mononaut , thanks for the feedback, understood for your addresses, I will work in your approach and keep the dropdown in the footer temporarily, if you already have the UI to switch the config figured out I can work on it as well.

I have one concern: I noticed in the address-text component that you have an if statement that checks if (similarity), do you have an example (link or screenshot) when this is true? I want to se the UI when it's the case and also implement the chunk address if necessary.

mononaut:
"similarity" is used to indicate possible address poisoning attacks, which we display with special underlining:

e.g: https://mempool.space/tx/eb692afae43f8246f1b60657271d37e4a4ab183ae542aa7452fa49b2359cf839
Screenshot 2026-02-18 at 4 37 04 AM

mononaut:
this looks much better with the ChunkedAddressComponent approach, but the PR diff is quite hard to follow interspersed with all of the address formatting selector logic - please remove.

rodribp:
Hi, thanks for the feedback, just did remove the formatting selector, let me know if there are more changes to make!

mononaut:
thanks! let's fix the display mode as "copy", then you can also rip out the `AddressFormattingService` and all of the conditional templating that uses it

rodribp:
Got it! Will do it shortly, I want to clarify something to be 100% sure, how would the user activate the chunking without selector and service? Do you plan to integrate this in another PR?

Or by fixing in 'copy' mode do you mean to entirely replace the default display with my new component as default? 🤔

mononaut:
yes, if we merge this PR your new chunked view would be the default (or rather the *only* option) for all addresses

rodribp:
Hi @mononaut , just removed the service and its trace from all components, additionally added some support for mobile devices in the address-text.component.html, passing less lastchars when is mobile and changed the spacing between chunks to 2px.

frontend/src/app/shared/components/address-text/address-text.component.html
```diff
@@ -11,7 +11,24 @@
...
+      class="d-none d-md-block"
+      style="flex: 0 1 auto; min-width: 0;">
+    </app-chunked-address>
+    <app-chunked-address 
+      [address]="address" 
+      [lastChars]="4" 
+      [link]="['/address/' | relativeUrl, address]"
+      class="d-md-none"
+      style="flex: 0 1 auto; min-width: 0;">
+    </app-chunked-address>
```
mononaut: do we need both versions of the chunked address component here? iirc, we always used `lastChars=8` in this context before for all breakpoints?

frontend/src/app/shared/components/address-text/address-text.component.html
```diff
@@ -11,7 +11,24 @@
...
-    <app-truncate [text]="address" [lastChars]="8"></app-truncate>
-  </a>
+  <div class="d-flex align-items-center" style="max-width: 100%; min-width: 0;">
+    <app-chunked-address 
+      [address]="address" 
+      [lastChars]="8" 
+      [link]="['/address/' | relativeUrl, address]"
+      class="d-none d-md-block"
+      style="flex: 0 1 auto; min-width: 0;">
+    </app-chunked-address>
```
mononaut: the mix of two different address display styles look a bit clumsy:

Image

perhaps we can just combine the two components? i.e. add support for the "similarity" input and underlining in `ChunkedAddressComponent`, and use that directly to replace the `AddressTextComponent` in the transactions list template?

rodribp: Hey! You are right, that's a good idea as address-text component would become useless if we add support to the poisined address case, I will do it!

frontend/src/app/shared/components/address-text/address-text.component.html
rodribp: So, the reason being is because in mobile the addresses shorten so much because of the spacing and it didn't look good because many times you could see just 1 char at the beginning of the address, and then it'd jump to the last chars

"b... Hj 8fj4 Sy" like this

So my solution for that was the combination of calling the component twice to send less lastChars in small devices, shorten the spacing 1px in general and putting less margin between the address and clipboard btn in small display

So, looking at the issue I just saw that @orangesurf mentions the clip btn would not appear in small display, and as I will move it to the transaction-list I will try to simplify it

rodribp:
Just made the following changes:

- Set the font-family to monospace.
- Added support for poisoned addresses to the `chunked-address` component.
- Hid the clipboard btn from small and medium breakbopints. (Desktop only as described in issue)
- Changed the `address-text` components in `transaction-list` for `chunked-address` components.
- Removed the `address-text` component. (Different commit)

Desktop:
image

Mobile
image

rodribp:
Just fixed duplicated clipboard button:

before:
image
image

After: 
image
image

frontend/src/app/shared/components/chunked-address/chunked-address.component.html
```diff
@@ -0,0 +1,90 @@
...
+    </ng-container>
+
+    <app-clipboard 
+    *ngIf="showClipboard"  
+    [text]="address" size="normal" 
+      class="d-none d-lg-block ml-lg-2">
+    </app-clipboard>
+    <ng-content></ng-content>
+  </div>
+}
```
mononaut: it would still be good to unify these two display modes, so that we can show addresses with *both* chunking *and* similarity underlining.

frontend/src/app/shared/components/chunked-address/chunked-address.component.ts
```diff
@@ -0,0 +1,112 @@
+import { Component, Input, ChangeDetectionStrategy, OnChanges, OnInit, OnDestroy, LOCALE_ID, Inject } from '@angular/core';
+import { AddressMatch } from '@app/shared/address-utils';
+
+interface AddressChunk {
+  text: string;
+  color: string;
+  hasMargin: boolean;
+}
```
mononaut, line 8: prefer CSS and angular template syntax to javascript for styling tasks

rodribp:
Hi @mononaut thanks for the PR to my branch, your solutions looks a lot cleaner so yes I took it, additionally I added the `<ng-content>` missing in the component as the `address` component needs it for the qr code and clipboard button

before: 
image

after:
image

mononaut:
@orangesurf - any thoughts?

Image
Image
Image

orangesurf:
for reference, my feedback was that the colors are too similar to be easily distinguished, personally I have been using #05ddff #6c92f9 but mononaut feels it's too much.

Perhaps you can explore a few different options and share them here for review 

image

rodribp:
### Other options I explored with:

Subtle purple #83abf5
image

Slightly darker blue #01a9f7
image

Green (similar to mononaut's but a little more of contrast) #00fdcf
image

orangesurf:
Ready for review or merge @mononaut

Issue #6181, opened 2025-12-23
[UI] softsimon theme selected button color


softsimon:
The color for Selected buttons/toggles are wrong and now too similar to the primary color on the softsimon "classic" theme.

<img width="408" height="106" alt="Image" src="https://github.com/user-attachments/assets/04f815ce-c593-46cd-80b4-185de221859a" />

alemanjosh7:
I don't understand this issue, I think it has already been fixed but still is open? Should not be like the following image?

<img width="674" height="98" alt="Image" src="https://github.com/user-attachments/assets/ec546472-7a93-48ce-a3c7-6b558bd8587f" />

nymkappa:
> I don't understand this issue, I think it has already been fixed but still is open? Should not be like the following image?

You're checking the wrong theme. Select `softsimon` theme in the footer

PR #6323, opened 2026-02-17
fix: adjust selected button color in softsimon theme

net-blip:
## Description
This Pull Request resolves issue #6181, where the color of selected buttons/toggles in the "softsimon classic" theme was too similar to the primary color.

## Changes Made
- The variable `$selected-button-color` was added with the value `#177ad8` to define a clearly distinguishable color for the selected buttons.

- The variable was exposed in `:root` as `--selected-button-color`.

- A specific CSS rule was added for the "softsimon" theme that applies this color to the selected buttons.

## Files Modified
- `Frontend/src/theme-softsimon.scss`: The variable and CSS rule for the selected buttons were added.

## Screenshots
- **Before:** 
Softsimon Before

- **After:** 
Softsimon after

### DIFF
--- a/frontend/src/theme-softsimon.scss
+++ b/frontend/src/theme-softsimon.scss
@@ -23,6 +23,7 @@ $nav-tabs-link-active-bg: $active-bg;
 $primary: #105fb0;
 $secondary: #2d3348;
 $tertiary: #653b9c;
+$selected-button-color: #177ad8; 
 $success: #1a9436;
 $info: #1bd8f4;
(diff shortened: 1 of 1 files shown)


nymkappa:
PR diff is full of tabs/whitespace/line changes

I've tested it locally and **I see no changes**, probably because of what copilot said about wrapping the new css in a unnecessary `.theme-softsimon` class.

Probably AI generated without actually testing it. Closing.

PR #6326, opened 2026-02-18
Update from transifex

mononaut:


wiz:
(as long as the tests pass)

Issue #6292, opened 2026-02-10
[UI] transaction details tooltip in block overview sometimes jumps


nymkappa:
https://github.com/user-attachments/assets/66f1bad9-91d7-48d7-9b10-c86eeaee11fd

algodondev:
Seems like it happens because a new row is added to that card component, specifically the Effective fee rate field. First solution that comes to my mind is to add some sort of animation to diminish the visual impact of the jump. 

https://github.com/user-attachments/assets/87952023-64ec-4178-bce0-869d2dded783

PR #6329, opened 2026-02-18
fix: transaction details tooltip (#6292)

OscarG673:
### Summary
the transaction details tooltip in the block overview sometimes jumped when moving the mouse. This happened mainly when the tooltip had more properties, meaning when the content was taller or wider

### Results
the tooltip stops jumping when it has more properties: the position is adjusted only once after the layout, using the correct size, and then it follows the cursor using the same logic, 

result video:
https://github.com/user-attachments/assets/2a0f306d-1cb0-41d5-a5ed-c0f037042e0f

fixes: #6292

### DIFF
--- a/frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
+++ b/frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
@@ -5,14 +5,13 @@ import { TransactionStripped } from '@interfaces/node-api.interface.js';
 import { Filter, FilterMode, TransactionFlags, toFilters } from '@app/shared/filters.utils';
 import { Block } from '@interfaces/electrs.interface.js';
 
-
 @Component({
   selector: 'app-block-overview-tooltip',
   templateUrl: './block-overview-tooltip.component.html',
   styleUrls: ['./block-overview-tooltip.component.scss'],
   standalone: false,
 })
-export class BlockOverviewTooltipComponent implements OnChanges, AfterViewChecked {
+export class BlockOverviewTooltipComponent implements OnChanges {
   @Input() tx: TransactionStripped | void;
   @Input() relativeTime?: number;
   @Input() cursorPosition: Position;
@@ -21,7 +20,6 @@ export class BlockOverviewTooltipComponent implements OnChanges, AfterViewChecke
   @Input() blockConversion: Price;
   @Input() filterFlags: bigint | null = null;
   @Input() filterMode: FilterMode = 'and';
-
 
   txid = '';
   time: number = 0;
@@ -36,29 +34,19 @@ export class BlockOverviewTooltipComponent implements OnChanges, AfterViewChecke
   filters: Filter[] = [];
   activeFilters: { [key: string]: boolean } = {};
 
-
-
-
   tooltipPosition: Position = { x: 0, y: 0 };
-  private pendingPositionRecalc = false;
-
-
-
 
   @ViewChild('tooltip') tooltipElement: ElementRef<HTMLCanvasElement>;
-
 
   constructor(
     private cd: ChangeDetectorRef,
     private appRef: ApplicationRef,
   ) { }
 
-
   ngOnChanges(changes): void {
     if (changes.cursorPosition && changes.cursorPosition.currentValue) {
       this.updateTooltipPosition(changes.cursorPosition.currentValue, false);
     }
-
 
     if (this.tx && (changes.tx || changes.filterFlags || changes.filterMode)) {
       this.txid = this.tx.txid || '';
@@ -79,8 +67,6 @@ export class BlockOverviewTooltipComponent implements OnChanges, AfterViewChecke
           this.activeFilters[filter.key] = true;
         }
       }
-
-
       if (!this.relativeTime) {
         this.timeMode = 'mempool';
       } else {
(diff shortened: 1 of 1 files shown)


mononaut:
this seems to improve the jumping a bit, but introduces a lateral shaking effect instead:

https://github.com/user-attachments/assets/b95e0358-add0-4f62-8f26-7c17a1f19903

OscarG673:
> this seems to improve the jumping a bit, but introduces a lateral shaking effect instead:
> 
>  Screen.Recording.2026-02-19.at.6.20.34.AM.mov

thanks for the feedback @mononaut. 
that lateral shaking issue has been resolved in the last commit.

https://github.com/user-attachments/assets/4495734a-7bc7-4d85-9d16-0dbdac378e44

frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
```diff
@@ -94,10 +96,64 @@ export class BlockOverviewTooltipComponent implements OnChanges {
...
     }
   }
 
+
+  ngAfterViewChecked(): void {
+    if (this.pendingPositionRecalc && this.cursorPosition && this.tx) {
+      this.pendingPositionRecalc = false;
+      requestAnimationFrame(() => {
+        this.updateTooltipPosition(this.cursorPosition, true);
+        this.appRef.tick();
```
mononaut: `this.appRef.tick()` is overkill here, `this.cd.detectChanges();` should be sufficient
```suggestion
        this.cd.detectChanges();
```

frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
```diff
@@ -94,10 +96,64 @@ export class BlockOverviewTooltipComponent implements OnChanges {
...
+
       this.cd.markForCheck();
+      if (this.cursorPosition) {
+        this.pendingPositionRecalc = true;
+      }
     }
   }
 
+
+  ngAfterViewChecked(): void {
```
mononaut: likewise, running these checks in `ngAfterViewChecked()` is excessive (it'll run on every change detection cycle across the whole application).

it should be fine to make this a simple class method and invoke it explicitly from inside `ngOnChanges` (when we know something relevant has actually changed)

and then you probably don't need the `pendingPositionRecalc` field either.
```suggestion
  private schedulePositionRecalc(): void {
```

frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
```diff
@@ -34,31 +36,30 @@ export class BlockOverviewTooltipComponent implements OnChanges {
   filters: Filter[] = [];
   activeFilters: { [key: string]: boolean } = {};
 
+
+
+
```
mononaut: there are a lot of stray newlines everywhere, please remove

frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
```diff
@@ -94,10 +96,64 @@ export class BlockOverviewTooltipComponent implements OnChanges {
...
+      if (belowY <= maxYInParent) {
+        y = belowY;
+      } else if (aboveY <= maxYInParent) {
+        y = aboveY;
+      } else {
+        y = maxYInParent;
+      }
+      y = Math.min(y, maxYInParent);
```
mononaut: this seems redundant? I don't think it's possible for `y` to be greater than `maxYInParent` by this point
```suggestion
```

frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
```diff
@@ -36,6 +36,8 @@ export class BlockOverviewTooltipComponent implements OnChanges {
 
   tooltipPosition: Position = { x: 0, y: 0 };
 
+  positionRafId: number | null = null;
+
   @ViewChild('tooltip') tooltipElement: ElementRef<HTMLCanvasElement>;
```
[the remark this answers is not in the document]
mononaut: preexisting issue

OscarG673:
> much smoother!
> 
> just a few minor issues to address

thank you for the feedback, @mononaut.

i have implemented the suggested changes in `block-overview-tooltip.component.ts` and verified them locally. 

Please let me know if there's anything else to adjust!

frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
```diff
@@ -93,9 +81,53 @@ export class BlockOverviewTooltipComponent implements OnChanges {
...
+      if (this.cursorPosition) {
+        this.schedulePositionRecalc();
+      }
+    }
+  }
+
+  private schedulePositionRecalc(): void {
+    requestAnimationFrame(() => {
+      this.updateTooltipPosition(this.cursorPosition, true);
+      this.appRef.tick();
```
mononaut, line 94: looks like we're still using `appRef.tick()` - did it not work properly with `cd.detectChanges()` instead?

frontend/src/app/components/block-overview-tooltip/block-overview-tooltip.component.ts
```diff
@@ -1,4 +1,4 @@
-import { Component, ElementRef, ViewChild, Input, OnChanges, ChangeDetectionStrategy, ChangeDetectorRef } from '@angular/core';
+import { Component, ElementRef, ViewChild, Input, OnChanges, AfterViewChecked, ChangeDetectorRef, ApplicationRef } from '@angular/core';
```
mononaut, line 1: `AfterViewChecked` is no longer used

Issue #4273, opened 2023-09-22
Feature suggestion: Display CPFP child fee suggestion


softsimon:
When creating a CPFP transaction to bump a transaction, I need to know what fee to set in order to bump the fee and still end up as high priority.

On the transaction page we should automatically calculate what sat/vB is required in a child transaction in order to end up in the current "high priority" fee rate band.

I am not sure where to put this, but I was considering on an info button with a tooltip next to the Fee Rate number with a text saying "A child transaction must pay 25 sat/vB to bump this transaction to high priority"

PR #6330, opened 2026-02-19
feat: suggested child fee rate for cpfp

rodribp:
### Description
This PR introduces a new feature decribed in issue #4273 to help users bump their unconfirmed transactions by calculating and displaying the fee rate a child pays for parent transaction needs to reach high priority.

### Motivation
Currently users who want to accelerate their tx with CPFP have to manually calculate the fee rate needed for the child tx.

### Key changes

- Added a calculation logic into the transaction component to determine the fee rate required for a standard 140 vB child transaction.
- The indicator is not shown if the transaction is in block 0 (next block).
- An icon is shown next to fee rate, on tooltip the user can see "A child ~140 vBytes transaction must pay x sats/vBytes to bump this transaction to high priority"

### Video

https://github.com/user-attachments/assets/a055ea31-3c32-407a-ae8a-f49b30b6a259

### DIFF
--- a/frontend/src/app/components/transaction/transaction-details/transaction-details.component.html
+++ b/frontend/src/app/components/transaction/transaction-details/transaction-details.component.html
@@ -228,7 +228,15 @@
 <ng-template #feeRateRow>
   @if (!isLoadingTx) {
     <tr>
-      <td i18n="transaction.fee-rate|Transaction fee rate">Fee rate</td>
+      <td>
+        <span i18n="transaction.fee-rate|Transaction fee rate">Fee rate</span>
+        @if (!tx?.status?.confirmed && recommendedChildFeeRate) {
+          <fa-icon class="info-badges ml-1" [icon]="['fas', 'angles-up']" style="color:gray" [fixedWidth]="true" 
+                   i18n-ngbTooltip="tx-details.recommended.fee.rate.child|Tooltip recommended fee rate child" 
+                   ngbTooltip="A child ~140 vBytes transaction must pay ~{{ recommendedChildFeeRate }} sat/vB to bump this transaction to high priority" 
+                   placement="bottom"></fa-icon>
+        }
+      </td>
       <td>
         <app-fee-rate [fee]="tx.feePerVsize"></app-fee-rate>
         @if (tx?.status?.confirmed && tx.fee && !hasEffectiveFeeRate && !accelerationInfo) {
--- a/frontend/src/app/components/transaction/transaction-details/transaction-details.component.ts
+++ b/frontend/src/app/components/transaction/transaction-details/transaction-details.component.ts
@@ -30,6 +30,7 @@ export class TransactionDetailsComponent implements OnInit {
   @Input() hasEffectiveFeeRate: boolean;
   @Input() cpfpInfo: CpfpInfo;
   @Input() hasCpfp: boolean;
+  @Input() recommendedChildFeeRate: number | null = null;
   @Input() accelerationInfo: Acceleration;
   @Input() acceleratorAvailable: boolean;
   @Input() accelerateCtaType: string;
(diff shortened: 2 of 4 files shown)


mononaut:
recommending accurate child rates for CPFP is a much trickier problem than it might first seem, because the set of ancestor transactions that will need to be boosted varies with the target fee rate.

for example, consider a transaction `tx1` with weight 100 vb and fee 100 sats (1 sat/vb), with one unconfirmed parent `tx2` with weight 10000 vb and fee 12000 sats (1.2 sat/vb).

currently, the `tx2` has a higher fee rate than `tx1`, so it pays for itself and won't be included in the `cpfpInfo` for `tx1`.

```mermaid
flowchart LR
    tx2["tx2<br/>10000 vb<br/>12000 sats<br/>1.2 sat/vb"]
    tx1["tx1<br/>100 vb<br/>100 sats<br/>1.0 sat/vb"]

    tx2 --> tx1

    subgraph CPFP["CPFP cluster"]
        tx1
    end
```

at a target rate of 1.1 sat/vb, `tx2` still pays for itself, so a child transaction only needs to pay an extra 10 sats to boost `tx1` by 0.1 sat/vb.

```mermaid
flowchart LR
    tx2["tx2<br/>10000 vb<br/>12000 sats<br/>1.2 sat/vb"]
    tx1["tx1<br/>100 vb<br/>100 sats<br/>1.0 sat/vb"]
    child["child<br/>+10 sats<br/>target 1.1 sat/vb"]

    tx2 --> tx1
    tx1 --> child

    subgraph CPFP["CPFP cluster"]
        tx1
        child
    end
```

at a target rate of 1.5 sat/vb, now any child needs to boost both `tx1` *and* `tx2` up to 1.5 sat/vb, at a cost of `(0.5 × 100 = 50 sats)` for `tx1` and `(0.4 × 10000 = 4000 sats)` for `tx2`.

```mermaid
flowchart LR
    tx2["tx2<br/>10000 vb<br/>12000 sats<br/>1.2 sat/vb"]
    tx1["tx1<br/>100 vb<br/>100 sats<br/>1.0 sat/vb"]
    child["child<br/>+4050 sats<br/>target 1.5 sat/vb"]

    tx2 --> tx1
    tx1 --> child

    subgraph CPFP["CPFP cluster"]
        direction LR
        tx2
        tx1
        child
    end
```

crucially, since `tx2` isn't *currently* CPFP'd by `tx1`, the frontend won't even be aware of `tx2`.

to make things even more difficult, `tx2` might have other unconfirmed children that already CPFP it up to some other rate, so you'd need to traverse the entire set of unconfirmed relatives and run the transaction selection algorithm to figure out exactly how the CPFP clusters break down and therefore exactly how much to pay to achieve a certain effective fee rate.

rodribp:
Hi @mononaut thanks for the feedback, I didn't think about it, as of now the code takes the total weight and fees of the family at the moment of making the calculation, how does exactly the family cluster works? If a tx1 is paid by a child at a target rate that is higher than tx1 and tx2 fee rate, then the tx2 is also included in the family?

So if I've understood well, your concern is, it can happen that the suggested fee rate, does not count that once bumping tx1, tx2 can be included in the family cluster for not considering it in the cpfpInfo, and so the target rate would bump it to a less priority than expected because after the child tx is broadcasted, the tx2 does get included in the relatives.

Did I understand your address correct?

mononaut:
yes, that's more or less the problem.

unfortunately the transaction selection / effective fee rate algorithm is currently rather complex, and depends on the context of the entire mempool, which is unavailable in the front end.

we have previously discussed adding a generic "unconfirmed ancestors/relatives" endpoint or data field which might help.

and the effective fee calculation may also change a bit with the new [cluster mempool design](https://delvingbitcoin.org/t/an-overview-of-the-cluster-mempool-proposal/393).

I suggest we leave this for now and revisit the feature later once those questions are resolved.

rodribp:
Alright, so you have way more context than me about this so I'm closing this PR

Issue #6227, opened 2026-01-20
[referral] show when a code is being applied to an acceleration


nymkappa:
A small indication of the referral code being used is needed

rodribp:
Hello, I want to work on this, I want to know if there's a valid referralCode I could use to see the behaviour in production

nymkappa:
> Hello, I want to work on this, I want to know if there's a valid referralCode I could use to see the behaviour in production

Being directly business related I'd rather implement this myself ✅

PR #6331, opened 2026-02-19
[accelerator] show referral code when activated

nymkappa:
Fixes https://github.com/mempool/mempool/issues/6227

When an accelerator referral code is active in local storage, we now display it in the details view.

I also tweaked some unrelated alignment issue I noticed _(some hard to read inline code has been split into multiple lines)_

Screenshot 2026-02-19 at 11-09-03 Transaction ca79302a13593b4fbf7375a4082ac40cebb4f5e58610d772a130bdb5b491ba3b - mempool - Bitcoin Explorer

--- 

**Accelerate Checkout UI Enhancements:**

* Added a new section to display the referral code in the summary row, using a dedicated `referralCodeDisplay` template for clearer visibility. [[1]](diffhunk://#diff-15906d96be0946bd51135ba6dd4e62ef8fa81778dbfe9dfd2d92e2722ae98c5aL247-R252) [[2]](diffhunk://#diff-15906d96be0946bd51135ba6dd4e62ef8fa81778dbfe9dfd2d92e2722ae98c5aR649-R654)
* Improved layout and alignment for the "accelerate to" and "customize" buttons by wrapping them in a flex container, ensuring consistent spacing and alignment.

### DIFF
--- a/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.html
+++ b/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.html
@@ -244,8 +244,12 @@
                     </tr>
                   </ng-container>
 
-                  <tr class="group-first group-last" style="border-top: 1px dashed grey">
-                    <td class="item"></td>
+                  <tr class="group-first group-last dashed-top">
+                    <td class="item" style="vertical-align:middle">
+                      @if (referralCode) {
+                        <ng-container *ngTemplateOutlet="referralCodeDisplay; context: {referralCode: referralCode}"></ng-container>
+                      }
+                    </td>
                     <td colspan="2">
                       <div class="d-flex">
                         <ng-container *ngTemplateOutlet="accelerateButton"></ng-container>
@@ -318,10 +322,8 @@
         <div>
           <div class="row summary-row">
             <div>
-              <div class="mb-2">
-                <div class="d-flex flex-column" for="accel">
-                  <ng-container *ngTemplateOutlet="accelerateOption; context: {etaInfo}"></ng-container>
-                </div>
+              <div class="d-flex flex-column" for="accel">
+                <ng-container *ngTemplateOutlet="accelerateOption; context: {etaInfo}"></ng-container>
               </div>
             </div>
             <div class="pie d-none d-lg-flex">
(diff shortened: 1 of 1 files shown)


mononaut:
would it fit better over here by the accelerate button where there's more space?

552043565-e7145e3e-1498-4606-aee1-559d6cd6d511
552043549-9de0602e-cee5-4fdc-bc90-5bb63ee575ae

mononaut:
also if we're going to display the referralCode directly in the client, we should probably do some validation instead of accepting any arbitrary strings:

Screenshot 2026-02-19 at 11 22 52 AM

(no XSS risk afaict, since we're using native angular template interpolation, but still)

nymkappa:
> would it fit better over here by the accelerate button where there's more space?

I'll try but I think it will look odd

nymkappa:
@mononaut (not pushed, example)

image

image

nymkappa:
@orangesurf thinks we should only show the code in the details view to not confused normal users so I'll clean this up

mononaut:
yeah I think that looks better, but I'll leave it to your judgement

nymkappa:
> yeah I think that looks better, but I'll leave it to your judgement

For now it actually does not really make sense to show the code directly (since the user won't get any kind of discount). So I'll just show it in the details view only. We can revisit the design at a later time **if** we give a discount upon using a code in the future.

frontend/src/app/components/transaction/transaction.component.html
```diff
@@ -86,11 +86,17 @@ <h1 i18n="shared.transaction">Transaction</h1>
     <ng-container *ngIf="!tx?.status?.confirmed && showAccelerationSummary && (ETA$ | async) as eta;">
       <span id='accelerate' #accelerate class="anchor"></span>
       <br>
-      <div class="title float-left mb-1">
-        <h2><a [href]="[ isMempoolSpaceBuild ? '/accelerator' : 'https://mempool.space/accelerator']" [target]="isMempoolSpaceBuild ? '' : 'blank'"><app-svg-images name="accelerator" [height]="isMobile ? '35px' : '45px'"></app-svg-images></a></h2>
+      <div class="d-flex flex-row justify-content-between align-items-center">
```
mononaut: what does this extra wrapper div do?

nymkappa: I'm not really sure anymore, it moves the `Hide accelerator` button a few pixels higher. Maybe I was experimenting with something else. I'll revert that change and merge.

Image

nymkappa:
Updated the PR to only update accelerate-checkout.component.html and remove unnecessary html/css changes to the transaction component

Issue #6339, opened 2026-02-25
[ONBTC] improve mobile responsiveness for ONBTC wallet view


OscarG673:
### Description

https://bitcoin.gob.sv/wallet/ONBTC

on the OBNBTC wallet page, some layout elements are not properly adapting to smaller screen sizes, causing overflow and spacing issues

view from iphone 12 pro:

<img width="384" height="780" alt="Image" src="https://github.com/user-attachments/assets/45ab3aed-d888-404a-927a-9581970e0e12" />

### Problem to be solved

improve the mobile layout of the wallet stats table and addresses treemap.

on small screens:
- stack table labell and values vertically instead of side by side
- make the treemap span the full width

PR #6340, opened 2026-02-25
[ONBTC] mobile view on onbtc wallet page

OscarG673:
### Summary

improves the mobile layout of the wallet stats table and addresses treemap by adjusting their structure for better responsiveness and readability on small screens

### Result

on mobile devices:

table rows are displayed in a stacked (vertical) layout instead of side by side
the addresses treemap is positioned below the table
the treemap now spans the full screen width

screenshot (iphone 12 pro):

Screenshot_20260224_231847

fixes: #6339

### DIFF
--- a/frontend/src/app/components/wallet/wallet.component.scss
+++ b/frontend/src/app/components/wallet/wallet.component.scss
@@ -46,31 +46,6 @@
   max-width: 200px;
 }
 
-@media (max-width: 575.98px) {
-  .address-table tr {
-    display: block;
-
-    td {
-      display: block;
-      width: 100%;
-    }
-
-    td:first-child {
-      padding-bottom: 0.25rem;
-      font-weight: 600;
-    }
-
-    td:last-child {
-      padding-top: 0;
-      text-align: left;
-    }
-  }
-
-  .address-table tr.addresses-row .address-list {
-    max-width: none;
-  }
-}
-
 h1 {
   margin: 0px;
   padding: 0px;
--- a/frontend/src/app/components/addresses-treemap/addresses-treemap.component.ts
+++ b/frontend/src/app/components/addresses-treemap/addresses-treemap.component.ts
@@ -75,6 +75,7 @@ export class AddressesTreemap implements OnChanges {
           progressive: 100,
           tooltip: {
             show: true,
+            confine: true,
             backgroundColor: 'rgba(17, 19, 31, 1)',
             borderRadius: 4,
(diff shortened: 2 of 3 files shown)


OscarG673:
> Concept ACK for the treemap fix, which is long overdue.
> 
> but the table layout intended to remain as two columns on mobile in order to match the many other pages with similar tables. if we do decide to redesign this, we should apply the new layout everywhere, but I think best to leave that for a future PR for now.

thank you for your feedback, @mononaut. I reverted the changes made to the table, I consider that at least on mobile it would work better, perhaps in the future it could be implemented, so i'm just leaving it as a suggestion :)

I also noticed that the treemap had a tooltip that was not constrained to the component. I've added `confine: true` to the treemap's tooltip options to fix this.

let me know, thanks! 

https://github.com/user-attachments/assets/156937c7-42d4-4225-8fc5-36a2f283ef95

frontend/src/app/components/addresses-treemap/addresses-treemap.component.scss
```diff
@@ -1,5 +1,12 @@
-.node-channels-container {
+:host {
+  display: block;
+  width: 100%;
+}
+
+.node-channels-container,
```
mononaut, line 6:
```suggestion
```

frontend/src/app/components/addresses-treemap/addresses-treemap.component.scss
```diff
@@ -1,5 +1,12 @@
-.node-channels-container {
+:host {
+  display: block;
+  width: 100%;
+}
```
mononaut, line 4: this applies the same styles to the same element as the 

```
@media (max-width: 767.98px) {
  .treemap-col app-addresses-treemap {
    ...
  }
}
```

in `wallet.component.scss`, except without the `@media (max-width: 767.98px)` limitation.

decide if you need these rules to apply everywhere, or just in the wallet context on mobile, or not at all, and remove one or both accordingly.

frontend/src/app/components/wallet/wallet.component.scss
```diff
@@ -8,6 +8,18 @@
 .treemap-col {
   width: 45%;
   height: 300px;
+
+  @media (max-width: 767.98px) {
+    width: 100%;
+    min-width: 0;
+  }
+}
```
mononaut, line 16: it would probably be cleaner to invert the media rule here, since I *think* the default `col-md` style is already `width: 100%`.

or there might be a smarter way to do this directly in bootstrap classes 🤔

PR #6343, opened 2026-02-25
[Docs] Add API - Accelerator documentation tab

orangesurf:
Add a dedicated Accelerator tab to the API docs page with three sub-tabs (Public API, Pro API, Auto Accelerator) showing endpoint flow cards, parameters, curl examples, and example responses.

All endpoint content is data-driven via acceleratorDocsData in api-docs-data.ts. Replace duplicate accelerator entries in the REST API tab with a link to the dedicated Accelerator tab.

image

image

### DIFF
--- a/frontend/src/app/docs/api-docs/api-docs-data.ts
+++ b/frontend/src/app/docs/api-docs/api-docs-data.ts
@@ -11951,40 +11951,50 @@ export const restApiDocsData = [
   },
   {
     type: 'category',
-    category: 'accelerator-public',
-    fragment: 'accelerator-public',
-    title: 'Accelerator (Public)',
+    category: 'accelerator',
+    fragment: 'accelerator',
+    title: 'Accelerator',
     showConditions: [''],
-    options: { officialOnly: true },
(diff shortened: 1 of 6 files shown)


mononaut:
Concept ACK

could use some polish, but the flow cards idea looks great!

---

I'd lean towards separating the high-level usage illustration with the flow cards etc from the list of API methods.

Maybe keep the sidebar on the left with all of the available API methods like we have on other API tabs, showing/hiding stuff as necessary as you switch between free & pro? Then the flow cards can just jump to the relevant section.

---

for checking acceleration status, you could use the individual endpoint `/api/v1/services/accelerator/accelerations/:txid` instead.

---

if we add another API docs tab, we'll need to finally figure out how to handle overflow on mobile

Screenshot 2026-02-27 at 3 57 17 AM

---

some of the text is a little hard to read with the subtle background colors and low contrast:

Screenshot 2026-02-27 at 3 57 44 AM
Screenshot 2026-02-27 at 3 57 54 AM

PR #6349, opened 2026-02-26
Use unique run ids for e2e tests

knorrium:
Should fix clashing run ids:

```
The run you are attempting to access is already complete and will not accept new groups.

The existing run is: https://cloud.cypress.io/projects/***/runs

When a run finishes all of its groups, it waits for a configurable set of time before finally completing. You must add more groups during that time period.

The --tag flag you passed was: pull_request
The --group flag you passed was: Tests on Chrome (testnet4)
The --parallel flag you passed was: true
The --ciBuildId flag you passed was: 70d1874cfdc577263d2947e8a173c95774168710-CI Pipeline for the Backend and Frontend-pull_request
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -382,7 +382,7 @@ jobs:
             cypress/e2e/signet/*.spec.ts
           group: Tests on Chrome (${{ matrix.module }})
           browser: "chrome"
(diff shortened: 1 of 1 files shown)


knorrium:
This was not needed. The run completion delay was increased to 180s, per the [docs](https://docs.cypress.io/cloud/features/smart-orchestration/parallelization#Run-completion-delay).

PR #6354, opened 2026-02-28
Feat: log scales

### LINKED ISSUE #5360: Charts log scale
The following graphs would benefit from a yaxis log scale

https://mempool.space/graphs/mining/block-fee-rates
https://mempool.space/graphs/mining/block-fees
https://mempool.space/graphs/acceleration/fees

https://x.com/SteveSimple/status/1815704621614080086

rodribp:
### Summary
This PR implements a new feature as shown in issue #5360 , allowing the user to see a switch to a log scale in the Y Axis of the following graphs:

1. [Block fee rates](https://mempool.space/graphs/mining/block-fee-rates)
2. [Block fees](https://mempool.space/graphs/mining/block-fees)
3. [Acceleration fees](https://mempool.space/graphs/acceleration/fees)

### Changes
1. Implemented a `toggleScale` method in the three graph components.
2. Added clamping functions to prevent echart errors when log(0), which usually occurs in early dates of Bitcoin fees.
3. Added toggle checkbox in the three graphs to easily switch from normal steps-graph to log scale.
4. When log is true changed from `bar` to `line` in block fee rates as eChart does not support stacked bars in a log scale, and unstacked bars cover each other by overlapping.

### Screenshots
image
image
image

Closes #5360

### DIFF
--- a/frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.html
+++ b/frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.html
@@ -7,33 +7,43 @@
       <button class="btn p-0 pl-2" style="margin: 0 0 4px 0px" (click)="onSaveChart()">
         <fa-icon [icon]="['fas', 'download']" [fixedWidth]="true"></fa-icon>
       </button>
-    </div>  
+    </div>
 
     <form [formGroup]="radioGroupForm" class="formRadioGroup" *ngIf="daysAvailable">
       <div class="btn-group btn-group-toggle" name="radioBasic">
         <label class="btn btn-primary btn-sm" *ngIf="daysAvailable >= 3" [class.active]="radioGroupForm.get('dateSpan').value === '1w'">
-          <input type="radio" [value]="'1w'" fragment="1w" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 1W
+          <input type="radio" [value]="'1w'" [fragment]="getFragment({time: '1w'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 1W
         </label>
         <label class="btn btn-primary btn-sm" *ngIf="daysAvailable >= 7" [class.active]="radioGroupForm.get('dateSpan').value === '1m'">
-          <input type="radio" [value]="'1m'" fragment="1m" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 1M
+          <input type="radio" [value]="'1m'" [fragment]="getFragment({time: '1m'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 1M
         </label>
         <label class="btn btn-primary btn-sm" *ngIf="daysAvailable >= 30" [class.active]="radioGroupForm.get('dateSpan').value === '3m'">
-          <input type="radio" [value]="'3m'" fragment="3m" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 3M
+          <input type="radio" [value]="'3m'" [fragment]="getFragment({time: '3m'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 3M
         </label>
         <label class="btn btn-primary btn-sm" *ngIf="daysAvailable >= 90" [class.active]="radioGroupForm.get('dateSpan').value === '6m'">
-          <input type="radio" [value]="'6m'" fragment="6m" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 6M
+          <input type="radio" [value]="'6m'" [fragment]="getFragment({time: '6m'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 6M
         </label>
         <label class="btn btn-primary btn-sm" *ngIf="daysAvailable >= 180" [class.active]="radioGroupForm.get('dateSpan').value === '1y'">
-          <input type="radio" [value]="'1y'" fragment="1y" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 1Y
+          <input type="radio" [value]="'1y'" [fragment]="getFragment({time: '1y'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 1Y
         </label>
         <label class="btn btn-primary btn-sm" *ngIf="daysAvailable >= 360" [class.active]="radioGroupForm.get('dateSpan').value === '2y'">
-          <input type="radio" [value]="'2y'" fragment="2y" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 2Y
+          <input type="radio" [value]="'2y'" [fragment]="getFragment({time: '2y'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 2Y
         </label>
         <label class="btn btn-primary btn-sm" *ngIf="daysAvailable >= 720" [class.active]="radioGroupForm.get('dateSpan').value === '3y'">
-          <input type="radio" [value]="'3y'" fragment="3y" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 3Y
+          <input type="radio" [value]="'3y'" [fragment]="getFragment({time: '3y'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> 3Y
         </label>
         <label class="btn btn-primary btn-sm" [class.active]="radioGroupForm.get('dateSpan').value === 'all'">
-          <input type="radio" [value]="'all'" fragment="all" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> ALL
+          <input type="radio" [value]="'all'" [fragment]="getFragment({time: 'all'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="dateSpan"> ALL
+        </label>
+      </div>
+    </form>
+    <form [formGroup]="scaleForm" class="formRadioGroup mr-0 mr-md-2">
+      <div class="btn-group btn-group-toggle" name="radioScale">
+        <label class="btn btn-primary btn-sm" [class.active]="scaleForm.get('scaleFunction').value === 'linear'">
+          <input type="radio" [value]="'linear'" [fragment]="getFragment({scale: 'linear'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="scaleFunction"> <span i18n="charts.linear-scale">Linear</span>
+        </label>
+        <label class="btn btn-primary btn-sm" [class.active]="scaleForm.get('scaleFunction').value === 'log'">
+          <input type="radio" [value]="'log'" [fragment]="getFragment({scale: 'log'})" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="scaleFunction"> <span i18n="charts.log-scale">Log</span>
         </label>
       </div>
     </form>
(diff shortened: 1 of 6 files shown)


mononaut:
Concept ACK, but I think it would look better with a toggle button (like the timespan selector) than a naked checkbox.

frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.ts
```diff
@@ -300,10 +306,10 @@ export class AccelerationFeesGraphComponent implements OnInit, OnChanges, OnDest
           name: 'Accelerated',
           yAxisIndex: 1,
           data: data.map(h =>  {
-            return [h.timestamp * 1000, h.count, h.avgHeight];
+            return [h.timestamp * 1000, Math.max(h.count, 1), h.avgHeight];
           }),
           type: 'bar',
-          barWidth: '90%',
+          barWidth: this.logScale ? undefined : '90%',
         },
```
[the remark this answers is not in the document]
rodribp: Clamping address does not apply because `https://mempool.space/api/v1/services/accelerator/accelerations/history/aggregated?timeframe=all` does not return values with 0, unlike the other two graphs

rodribp:
Hi @mononaut , I would like your feedback if would you consider good to have to include the scale function in the URL, just so if somebody wants to share the log view with someone else he's able to directly through URL, otherwise, the other person would have to click on the toggle button and the conditional setting of the options inside the `prepareChartOptions` would actually make sense, as now for default is set to steps, not log. I can implement it if you think so as well.

rodribp:
Latest UI:

image

image

To handle shareable URL view I added params to the fragment, (e.g):

http://localhost:4200/graphs/mining/block-fee-rates#time=all&scale=log

By adding a function called `updateUrlFragment()`, called inside statsObservable pipe to look at the form values, and merge them into a single fragment with two params.

mononaut:
looks good!

better toggle labels would be "Linear"/"Log"

yes, fragment params do make sense for this.

frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.ts
```diff
@@ -287,7 +320,7 @@ export class AccelerationFeesGraphComponent implements OnInit, OnChanges, OnDest
         {
           name: 'Total bid boost',
           data: data.map(h =>  {
-            return [h.timestamp * 1000, h.sumBidBoost, h.avgHeight];
+            return [h.timestamp * 1000, Math.max(h.sumBidBoost, 0.1), h.avgHeight, h.sumBidBoost];
```
mononaut: should the `Math.max(...)` clamping be conditional on `this.logScale`, like it is on the other charts?

frontend/src/app/components/block-fee-rates-graph/block-fee-rates-graph.component.ts
```diff
@@ -144,11 +174,16 @@ export class BlockFeeRatesGraphComponent implements OnInit {
                 for (const percentile in seriesData) {
                   series.push({
                     zlevel: 0,
-                    stack: 'Total',
+                    stack: this.logScale ? undefined : 'Total',
                     name: percentile,
                     data: seriesData[percentile],
-                    type: 'bar',
-                    barWidth: '100%',
+                    type: this.logScale ? 'line' : 'bar',
```
mononaut: changing the chart type is beyond the scope of this PR.

(true, the stacked bar chart has never really made much sense for block fee rates, but replacing it with something better is a job for a future PR)

frontend/src/app/components/block-fee-rates-graph/block-fee-rates-graph.component.ts
```diff
@@ -81,8 +84,33 @@ export class BlockFeeRatesGraphComponent implements OnInit {
...
+          } else {
+            if (['24h', '3d', '1w', '1m', '3m', '6m', '1y', '2y', '3y', 'all'].includes(fragment)) {
+              timeVal = fragment;
+            }
+            if (['linear', 'log'].includes(fragment)) {
+              scaleVal = fragment;
+            }
+          }
+
+          if (timeVal && ['1m', '3m', '6m', '1y', '2y', '3y', 'all'].includes(timeVal)) {
```
mononaut: mismatch between this list of acceptable time values and the one above (`['24h', '3d', '1w', '1m', '3m', '6m', '1y', '2y', '3y', 'all']`)?

frontend/src/app/components/block-fee-rates-graph/block-fee-rates-graph.component.ts
```diff
@@ -388,4 +424,49 @@ export class BlockFeeRatesGraphComponent implements OnInit {
...
+
+    (this.chartOptions.series as any[]).forEach(seriesItem => {
+      if (seriesItem.name !== 'Moving average') {
+        seriesItem.type = this.logScale ? 'line' : 'bar';
+        seriesItem.stack = this.logScale ? undefined : 'Total';
+        seriesItem.barWidth = this.logScale ? undefined : '100%';
+      }
+
+      seriesItem.data.forEach(point => {
+        point [1] = this.logScale ? Math.max(point[3], 0.1) : point[3];
```
mononaut: stray whitespace

```suggestion
        point[1] = this.logScale ? Math.max(point[3], 0.1) : point[3];
```

frontend/src/app/components/block-fees-graph/block-fees-graph.component.ts
```diff
@@ -317,4 +352,50 @@ export class BlockFeesGraphComponent implements OnInit {
...
+
+    yAxes[1].type = this.logScale ? 'log' : 'value';
+    yAxes[1].min = this.logScale ? 0.01 : undefined;
+
+    (this.chartOptions.series as any[]).forEach((seriesItem, index) => {
+      seriesItem.data.forEach(point => {
+        if (index === 0) {
+          point [1] = this.logScale ? Math.max(point[3], 0.001) : point[3];
+        } else if (index === 1) {
+          point [1] = this.logScale ? Math.max(point[3], 0.01) : point[3];
```
mononaut: stray whitespace

```suggestion
          point[1] = this.logScale ? Math.max(point[3], 0.001) : point[3];
        } else if (index === 1) {
          point[1] = this.logScale ? Math.max(point[3], 0.01) : point[3];
```

frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.html
```diff
@@ -37,6 +37,16 @@
         </label>
       </div>
     </form>
+    <form [formGroup]="scaleForm" class="formRadioGroup mr-0 mr-md-2">
+      <div class="btn-group btn-group-toggle" name="radioScale">
+        <label class="btn btn-primary btn-sm" [class.active]="scaleForm.get('scaleFunction').value === 'linear'">
+          <input type="radio" [value]="'linear'" fragment="linear" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="scaleFunction"> <span i18n="accelerations.linear-scale-function">Linear</span>
+        </label>
+        <label class="btn btn-primary btn-sm" [class.active]="scaleForm.get('scaleFunction').value === 'log'">
+          <input type="radio" [value]="'log'" fragment="log" [routerLink]="['/graphs/acceleration/fees' | relativeUrl]" formControlName="scaleFunction"> <span i18n="accelerations.log-scale-function">Log</span>
```
mononaut: prefer to reuse the same i18n ids across all charts, to avoid having separate translation items for each instance of "Linear" and "Log"

e.g. use "charts.linear-scale" and "charts.log-scale"

frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.ts
```diff
@@ -365,4 +398,46 @@ export class AccelerationFeesGraphComponent implements OnInit, OnChanges, OnDest
...
+    const time = this.radioGroupForm.controls.dateSpan.value;
+    const scale = this.scaleForm.controls.scaleFunction.value;
+    const newFragment = `time=${time}&scale=${scale}`;
+
+    if (this.route.snapshot.fragment !== newFragment) {
+      this.router.navigate([], {
+        relativeTo: this.route,
+        fragment: newFragment,
+        replaceUrl: true
+      })
```
mononaut: missing semicolon
```suggestion
      });
```

mononaut (changes requested):
Sorry for the delayed review on this!

I've left some comments inline that need to be addressed.

There are also a few broader issues:

1) When you click "Linear" or "Log", we end up navigating twice - first when the routerLink on the form input applies the new individual fragment, and then again when `updateUrlFragment` updates the fragment again with both parameters.

this is partly tech debt left over from the previous implementation of fragments for the timespan selector, but it becomes much more obvious with multiple fragment parameters.

I suggest refactoring so that the `routerLink`s on all of the form inputs set the *complete* fragment, using a helper function like:

```
getFragment(setParams: Record<string, string> = {}): string {                                                                                                 
  const params = new URLSearchParams(this.route.snapshot.fragment || '');                                                                                       
  params.set('time', setParams.time || this.radioGroupForm.controls.dateSpan.value);                                                                            
  params.set('scale', setParams.scale || this.scaleForm.controls.scaleFunction.value);                                                                          
  return params.toString();                                                                                                                                     
}

// usage: [fragment]="getFragment({scale: 'log'})"
```

and then *only* updating the chart configuration from the `fragmentSubscription`.

2) There are quite a few minor linting problems, which suggests you either don't have eslint configured in your editor/IDE yet or it's failing to pick up our linting rules from `frontend/eslint.config.js` or `frontend/.eslintrc`

mononaut:
also, on the acceleration fees chart, if there are very few bid boost values (or perhaps only zero values?) the axes glitch out:

Screenshot 2026-03-11 at 9 54 08 AM

frontend/src/app/components/block-fee-rates-graph/block-fee-rates-graph.component.ts
rodribp: Hey, that's true, is not the goal of this PR, the reason being is Echarts does not allow stacked bars in a log scale chart, so that's the solution I came up with, so I will revert this and look into it for a solution keeping the bars (I think I will just unstack it in log). Or maybe it's better to wait for the future PR, to apply the log function.

frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.ts
rodribp: It was because unlike the other charts this data is not set through a pipe but when init the chart, so if I make the clamping conditional I have to also make the clamping when the scale changes. And since the shortest linear chart has a first step of 500 sats, I thought the difference of changing zeros to 0.1 it's too minimal to worry about it, (also the tooltip shows the right amount). Either way ofc it's possible to put the condition, I will push it and let me know your feedback!

rodribp:
Hi @mononaut thanks a lot for the feedback!

Just pushed all the changes:

1. Fixed linting errors.
2. Fixed mismatches.
3. Implemented your fragment navigation idea, it's perfect.
4. Fixed the glitch of the axis in acceleration, the error was a mismatch between the clamping and min(floor), of the yAxis, if you set a clamped value below the floor, the graph axis glitches out.
5. Changed the type of the fee rates back to bar. Getting stacked bars to work with native EChart type='log' turned into a headache because of these limitation:

    - If you set the yAxis below 1 the stacked bars breaks as log10(1)=0 so echart anchors the bars at 1 (image below).
    - If you try to fix 0 sat/vB values using null, undefined, NaN or '-', it breaks the render because of the large: true optimization.
    - To solve it, I kept the axis linear and manually faked the log scale using Math.log10, this way setting values below 1, to an strict 0 (by condition).


image

mononaut (approved):
Tested ACK @ [fbaed240]

but can we please squash this down to clean up the commit history before merging.

rodribp:
Just checked there's conflicts, resolving them rn

rodribp:
Should I close or unstuck this pull request?

frontend/src/app/components/block-fee-rates-graph/block-fee-rates-graph.component.ts
```diff
@@ -123,32 +153,35 @@ export class BlockFeeRatesGraphComponent implements OnInit {
...
+                let zIndex = 10;
                 for (const percentile in seriesData) {
                   series.push({
-                    zlevel: 0,
-                    stack: 'Total',
+                    z: this.logScale ? zIndex-- : 0,
+                    stack: this.logScale ? undefined : 'Total',
                     name: percentile,
                     data: seriesData[percentile],
                     type: 'bar',
```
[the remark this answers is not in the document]
rodribp: Not anymore

PR #6357, opened 2026-03-01
feat: add network enabled guard to routing for testnet, testnet4, and signet paths

### LINKED ISSUE #5496: Disabled network still routable
Disable the angular routing to for example `/testnet` if the network is disabled in the frontend config.

kayyrod21:
Description
	•	Adds NetworkEnabledGuard and applies it as canMatch to the testnet, testnet4, and signet route roots.
	•	When a network is disabled via window.__env flags (TESTNET_ENABLED, TESTNET4_ENABLED, SIGNET_ENABLED), navigation to that network route is prevented and redirected to /.
	•	Keeps existing routing behavior unchanged for enabled networks.
	•	Closes #5496

How I tested
	•	Ran locally with TESTNET_ENABLED=false → Navigating to /testnet redirects to / (expected behavior)
	•	Ran locally with TESTNET_ENABLED=true → /testnet loads normally
Repeated the same validation for:
	•	TESTNET4_ENABLED
	•	SIGNET_ENABLED

### DIFF
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -9,11 +9,11 @@ import { StatusViewComponent } from '@components/status-view/status-view.compone
 import { AddressGroupComponent } from '@components/address-group/address-group.component';
 import { TrackerGuard } from '@app/route-guards';
 
-const browserWindow = window || {};
+const browserWindow = window as typeof window & { __env?: any };
 // @ts-ignore
 const browserWindowEnv = browserWindow.__env || {};
 
-let routes: Routes = [
+const testnetRoutes: Routes = browserWindowEnv.TESTNET_ENABLED ? [
   {
     path: 'testnet',
     children: [
@@ -52,6 +52,9 @@ let routes: Routes = [
       },
     ]
   },
+] : [];
+
+const testnet4Routes: Routes = browserWindowEnv.TESTNET4_ENABLED ? [
   {
     path: 'testnet4',
     children: [
@@ -90,6 +93,9 @@ let routes: Routes = [
       },
     ]
   },
+] : [];
+
+const signetRoutes: Routes = browserWindowEnv.SIGNET_ENABLED ? [
   {
     path: 'signet',
     children: [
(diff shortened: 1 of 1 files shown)


kayyrod21:
> Concept ACK, but approach NACK
> 
> Route guards are necessary when a route is conditionally available based on some dynamic state, like the tracker guard.
> 
> but here, we know exactly which networks will be available for the entire lifetime of the page at load-time, so it would be cleaner to just conditionally include them in the route definitions, e.g:
> 
> ```
> let routes: Routes = [
>   ...(env.TESTNET_ENABLED ? [{
>     path: 'testnet',
>     children: [ ... ]
>   }] : []),
> ```

Thanks, that makes sense. I’ll rework this to conditionally include the routes at definition time instead of using a guard.

kayyrod21:
Reworked this to conditionally include the testnet, testnet4, and signet route roots at definition time instead of using a guard, per review feedback.

Rebased onto latest master and squashed to one commit.

Tested locally:
- with TESTNET_ENABLED=false, navigating to /testnet falls back to the default route and /testnet is removed from the URL
- with TESTNET_ENABLED=true, /testnet loads normally

Repeated the same validation for TESTNET4_ENABLED and SIGNET_ENABLED.

mononaut:
thanks! we should apply the same logic to regtest and liquid testnet, but otherwise this looks great

kayyrod21:
Applied the same route-definition logic to regtest and liquid testnet as well.

Tested locally:
- with REGTEST_ENABLED=true, /regtest remains on the regtest route and loads normally at the routing level
- with REGTEST_ENABLED=false, navigating to /regtest falls back to the default route and /regtest is removed from the URL
- in liquid mode with LIQUID_TESTNET_ENABLED=true, /testnet loads normally
- in liquid mode with LIQUID_TESTNET_ENABLED=false, navigating to /testnet falls back to the default route and /testnet is removed from the URL

mononaut (approved):
Tested ACK @ [cfa6df6e]

works great, thank you!

PR #6360, opened 2026-03-03
fix(a11y): add accessible names to network selector and toggle buttons

### LINKED ISSUE #6359: Accessibility: Icon-only buttons missing accessible names (network selector, block display, time toggle)
Bug: Icon-only interactive buttons missing accessible names

Lighthouse reports the following accessibility issue:

"Buttons do not have an accessible name"

The affected elements are:

1. Network dropdown toggle (master-page.component.html)
2. Network dropdown toggle (liquid-master-page.component.html)
3. Block display toggle button (blockchain.component.html)
4. Time direction toggle button (blockchain.component.html)

Image

These buttons are icon-only and currently lack accessible names, which causes screen readers to announce them generically as "button".

Proposed fix:
Add aria-label (and title) attributes to each affected button without changing layout or behavior.

This change is accessibility-only and does not alter UI/UX behavior.

kayyrod21:
Closes #6359

## Summary

This PR resolves a Lighthouse accessibility failure:

> Buttons do not have an accessible name

Several icon-only buttons lacked accessible names, which caused screen readers to announce them generically as “button”.

## Changes

Added `aria-label` (and matching `title`) attributes to the following buttons:

- Network dropdown toggle in:
  - `master-page.component.html`
  - `liquid-master-page.component.html`
- Block display toggle in:
  - `blockchain.component.html`
- Time direction toggle in:
  - `blockchain.component.html`

Also added `type="button"` to the toggle buttons in `blockchain.component.html` to prevent unintended form submission behavior.

## Scope

- Accessibility-only changes
- No layout or styling changes
- No behavioral changes
- No logic changes

## How I tested

- Ran the app locally using `serve:local-prod`
- Verified via Chrome DevTools that affected buttons now expose accessible names
- Re-ran Lighthouse and confirmed the “Buttons do not have an accessible name” issue is resolved
- Accessibility score improved from 76 → 85 after changes

### DIFF
--- a/frontend/src/app/components/blockchain/blockchain.component.html
+++ b/frontend/src/app/components/blockchain/blockchain.component.html
@@ -1,4 +1,4 @@
-<div class="text-center" class="blockchain-wrapper" [class.time-ltr]="timeLtr" [class.ltr-transition]="ltrTransitionEnabled" #container>
+<div class="text-center blockchain-wrapper" [class.time-ltr]="timeLtr" [class.ltr-transition]="ltrTransitionEnabled" #container>
   <div #positionContainer class="position-container" [ngClass]="network ? network : ''" [style]="positionStyle">
     <span>
       <div class="blocks-wrapper">
@@ -10,8 +10,31 @@
         </ng-container>
       </div>
       <div id="divider" [hidden]="pageIndex > 0">
-        <button class="block-display-toggle" (click)="toggleBlockDisplayMode()"><fa-icon [icon]="['fas', 'exchange-alt']" [fixedWidth]="true"></fa-icon></button>
-        <button class="time-toggle" (click)="toggleTimeDirection()"><fa-icon [icon]="['fas', 'exchange-alt']" [fixedWidth]="true"></fa-icon></button>
+        <button
+          type="button"
+          class="block-display-toggle"
+          (click)="toggleBlockDisplayMode()"
+          [attr.aria-label]="blockDisplayMode === 'fees'
+            ? 'Switch to displaying block sizes'
+            : 'Switch to displaying total fees per block'"
+          [attr.title]="blockDisplayMode === 'fees'
+            ? 'Switch to displaying block sizes'
+            : 'Switch to displaying total fees per block'">
+          <fa-icon [icon]="['fas', 'exchange-alt']" [fixedWidth]="true"></fa-icon>
+        </button>
+
+        <button
+          type="button"
+          class="time-toggle"
+          (click)="toggleTimeDirection()"
+          [attr.aria-label]="timeLtr
+            ? 'Switch to displaying newest blocks on the right'
+            : 'Switch to displaying newest blocks on the left'"
+          [attr.title]="timeLtr
+            ? 'Switch to displaying newest blocks on the right'
+            : 'Switch to displaying newest blocks on the left'">
+          <fa-icon [icon]="['fas', 'exchange-alt']" [fixedWidth]="true"></fa-icon>
+        </button>
       </div>
     </span>
   </div>
--- a/frontend/src/app/components/liquid-master-page/liquid-master-page.component.html
+++ b/frontend/src/app/components/liquid-master-page/liquid-master-page.component.html
@@ -45,7 +45,13 @@
   </a>
 
   <div ngbDropdown (window:resize)="onResize()" class="dropdown-container" *ngIf="env.TESTNET_ENABLED || env.SIGNET_ENABLED || env.REGTEST_ENABLED || env.LIQUID_ENABLED || env.LIQUID_TESTNET_ENABLED">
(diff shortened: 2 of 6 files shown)


frontend/src/app/components/blockchain/blockchain.component.html
```diff
@@ -10,8 +10,25 @@
...
       <div id="divider" [hidden]="pageIndex > 0">
-        <button class="block-display-toggle" (click)="toggleBlockDisplayMode()"><fa-icon [icon]="['fas', 'exchange-alt']" [fixedWidth]="true"></fa-icon></button>
-        <button class="time-toggle" (click)="toggleTimeDirection()"><fa-icon [icon]="['fas', 'exchange-alt']" [fixedWidth]="true"></fa-icon></button>
+        <button
+          type="button"
+          class="block-display-toggle"
+          (click)="toggleBlockDisplayMode()"
+          aria-label="Toggle block display mode"
+          title="Toggle block display mode"
+          [attr.aria-pressed]="isBlockDisplayToggled">
```
mononaut: `aria-pressed` might not be the best approach for this button - iirc it's intended for boolean toggles where a single static label + yes/no state is sufficient for the user to know both what the current state is and what clicking the button will do.

here, a screen reader will announce something like "toggle block display mode, toggle button, not pressed", which doesn't give the user enough information to know what to do with it.

better to just make the aria-label itself conditional, e.g:

`[attr.aria-label]="blockDisplayMode === 'fees' ? 'switch to displaying sizes of blocks' : 'switch to displaying total fees in blocks'"`

frontend/src/app/components/blockchain/blockchain.component.ts
```diff
@@ -68,6 +68,12 @@ export class BlockchainComponent implements OnInit, OnDestroy, OnChanges {
     return item.index;
   }
 
+  get isBlockDisplayToggled(): boolean {
+    // Return true when the alternate block display mode is active.
+    // Derive this directly from the same state variable used in toggleBlockDisplayMode().
```
mononaut: the code is self-explanatory, no need for the comments here

frontend/src/app/components/liquid-master-page/liquid-master-page.component.html
```diff
@@ -45,7 +45,13 @@
...
 
   <div ngbDropdown (window:resize)="onResize()" class="dropdown-container" *ngIf="env.TESTNET_ENABLED || env.SIGNET_ENABLED || env.REGTEST_ENABLED || env.LIQUID_ENABLED || env.LIQUID_TESTNET_ENABLED">
-    <button ngbDropdownToggle type="button" class="btn btn-secondary dropdown-toggle-split d-flex justify-content-center align-items-center" aria-haspopup="true">
+    <button 
+    ngbDropdownToggle 
+    type="button" 
+    class="btn btn-secondary dropdown-toggle-split d-flex justify-content-center align-items-center" 
+    aria-haspopup="true" 
+    [attr.aria-label]="'Select network. Current: ' + networkDisplayName" 
+    [attr.title]="'Select network. Current: ' + networkDisplayName">
```
mononaut: misaligned indentation

```suggestion
    <button 
      ngbDropdownToggle 
      type="button" 
      class="btn btn-secondary dropdown-toggle-split d-flex justify-content-center align-items-center" 
      aria-haspopup="true" 
      [attr.aria-label]="'Select network. Current: ' + networkDisplayName" 
      [attr.title]="'Select network. Current: ' + networkDisplayName">
```

frontend/src/app/components/liquid-master-page/liquid-master-page.component.ts
```diff
@@ -52,6 +52,20 @@ export class LiquidMasterPageComponent implements OnInit, OnDestroy {
...
+      'signet': 'Signet',
+      'testnet': 'Testnet3',
+      'testnet4': 'Testnet4',
+      'regtest': 'Regtest',
+      'liquid': 'Liquid',
+      'liquidtestnet': 'Liquid Testnet',
+    };
+    return labels[val] ?? val;
+  }
+
```
mononaut, line 58: it would be good to extract this helper out somewhere (perhaps the `StateService`?) to avoid repeating ourselves in both master page components

mononaut (changes requested):
Concept ACK, subject to a few minor changes.

let's also rebase on current master and squash this down to one clean commit please 🙏

frontend/src/app/components/blockchain/blockchain.component.html
kayyrod21: Thanks for the guidance here.

frontend/src/app/components/liquid-master-page/liquid-master-page.component.ts
kayyrod21: Good call. I’ve refactored the networkDisplayName helper into StateService so both master page components can consume the same logic and we avoid duplication.

Thanks for pointing that out, definitely cleaner this way.

mononaut (approved):
Tested ACK @ [fc611e03]

works great, thank you!

PR #6379, opened 2026-03-12
[calculator] allow user to get historical closing day rate

nymkappa:
User can now pick a past date to get historical price point _(should return the price at midnight UTC on the next day after the selected one)_

If today's date is selected, we keep the real time websocket price update enabled.

image

### DIFF
--- a/frontend/src/app/components/calculator/calculator.component.ts
+++ b/frontend/src/app/components/calculator/calculator.component.ts
@@ -237,7 +237,7 @@ export class CalculatorComponent implements OnInit {
   }
 
   updatePrice(): void {
-    this.timestamp$.next(Date.UTC(this.dateModel.year, this.dateModel.month - 1, this.dateModel.day, 23, 59, 59) / 1000);
+    this.timestamp$.next(Date.UTC(this.dateModel.year, this.dateModel.month - 1, this.dateModel.day + 1) / 1000);
   }
 
   get todaySelected() {
--- a/backend/src/repositories/PricesRepository.ts
+++ b/backend/src/repositories/PricesRepository.ts
@@ -330,7 +330,7 @@ class PricesRepository {
       const [rates] = await DB.query(`
         SELECT ${ApiPriceFields}
         FROM prices
-        WHERE UNIX_TIMESTAMP(time) < ?
+        WHERE UNIX_TIMESTAMP(time) <= ?
           AND USD >= 0
         ORDER BY time DESC
         LIMIT 1`,


frontend/src/app/components/calculator/calculator.component.ts
```diff
@@ -46,17 +58,31 @@ export class CalculatorComponent implements OnInit {
...
+    }).pipe(
+      switchMap(({ currency, timestamp }) => {
+        this.currentCurrency = currency;
         this.updateCurrencyDecimals();
-        return this.stateService.conversions$.asObservable();
+
+        return this.todaySelected
+          ? this.stateService.conversions$.asObservable()
+          : this.apiService.getHistoricalPrice$(timestamp, currency).pipe(
+            map((p: any) => {
```
mononaut, line 72: the `: any` typecast probably isn't necessary

nymkappa:
> Tested ACK @ [[b5ac25e](https://github.com/mempool/mempool/commit/b5ac25edbe6e0a3fa11916474bb10c38472fefcd)]
> 
> some of the copilot comments raise good points, but nothing super important.
> 
> also, what's the reasoning behind 11pm UTC as the daily close time instead of midnight?

Programming laziness. I would have to manipulate a Date object to get "the next day opening price at 00:00" instead of the "selected day", which involves manipulating a Date object to make sure we don't try to get price on the 32th.

But thinking about it, since we use timestamp, I could just add 3600 seconds and that might work. I also need to change our backend where clause to `UNIX_TIMESTAMP(time) <= ?` for selecting the price and that should be fine. Let me try it out.

frontend/src/app/components/calculator/calculator.component.ts
nymkappa: Something is odd with the FOSS project, I only get partial autocompletion and a lot of stuff underligned in red in VSCode, which is probably why I added some `: any` here and there.

@knorrium can you take a look at this or do I have some weird local settings?

mononaut:
I think you can just pass day + 1 to the Date constructor and it'll handle the overflows for you

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#:~:text=If%20any%20parameter-,overflows,-its%20defined%20bounds

nymkappa:
> I think you can just pass day + 1 to the Date constructor and it'll handle the overflows for you
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#:~:text=If%20any%20parameter-,overflows,-its%20defined%20bounds

Thanks yeah that work, coupled with the `<=` fix it picks the correct closing price at midnight now.

https://github.com/mempool/mempool/pull/6379/commits/6b778a9fee8af7d197c36358dcd24c244774e126

PR #6395, opened 2026-03-17
fix(frontend): localize mining and acceleration chart legend and tooltip labels

### LINKED ISSUE #6391: Missing i18n localization in additional mining and acceleration graphs
Following #3334, there are additional graph components with hardcoded English strings that are not localized.

Affected graphs:

1. Block Fees vs Subsidy graph
   - "Subsidy"
   - "Fees"
   - "Subsidy (USD)"
   - "Fees (USD)"
   - "Subsidy (%)"
   - "Fees (%)"
   
Image

2. Block Fees graph
   - "Fees BTC"
   - "Fees USD"
   - Tooltip: "On average around block"
   - Legend
Image

3. Acceleration Commissions graph
   - "Total bid boost"
   - "Accelerated"
   - Legend
Image

These appear to be hardcoded and not using $localize.

This likely requires applying the same approach used in #3334.

kayyrod21:
## Summary

Localizes remaining hardcoded legend and tooltip labels in:

- Block Fees graph
- Block Fees vs Subsidy graph
- Acceleration Commissions graph

## Changes

- Wrapped hardcoded legend and tooltip strings in `$localize`
- Introduced shared localized label constants where legend labels are also used as logic keys
- Updated legend.data, legend.selected, series names, tooltip comparisons, and dispatchAction calls to use localized constants

No functional changes to chart behavior.
Closes #6391

### DIFF
--- a/frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
+++ b/frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
@@ -55,6 +55,13 @@ export class BlockFeesSubsidyGraphComponent implements OnInit {
   zoomTimeSpan = '';
   legend: { mode: 'normal' | 'fiat' | 'percentage', subsidy: boolean, fees: boolean } = { mode: 'normal', subsidy: true, fees: true };
 
+  private readonly subsidyLabel = $localize`:@@graphs.blockFeesSubsidy.subsidy:Subsidy`;
+  private readonly feesLabel = $localize`:@@graphs.blockFeesSubsidy.fees:Fees`;
+  private readonly subsidyUsdLabel = $localize`:@@graphs.blockFeesSubsidy.subsidyUsd:Subsidy (USD)`;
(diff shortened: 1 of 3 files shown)


frontend/src/app/components/block-fees-subsidy-graph/block-fees-subsidy-graph.component.ts
natsoni:
```diff
-       if (params.name.includes('USD')) {
+       if ([this.subsidyUsdLabel, this.feesUsdLabel].includes(params.name)) {
```

natsoni:
```diff
-      } else if (params.name.includes('%')) {
+      } else if ([this.subsidyPercentLabel, this.feesPercentLabel].includes(params.name)) {
```

kayyrod21:
> Tested this PR using placeholder values for the new translations and it worked nicely overall, thanks!
> 
> There's one regression on the Block fees / subsidy graph: switching modes breaks legends and is not properly processed (the underlying bug was already there to be fair, the introduction of localization just triggered it)

Thanks, good catch on the regression bug! I updated the mode detection to use the localized label constants instead of checking for English substrings like USD and %.

PR #6397, opened 2026-03-18
bump frontend npm deps

mononaut:
equivalent of https://github.com/mempool/mempool.space/pull/2208

- @angular-devkit/build-angular and @angular/cli: 20.3.16 → 20.3.20
- @angular/*: 20.3.17 → 20.3.18
- ajv: 6.12.6 → 6.14.0
- copy-webpack-plugin: 13.0.1 → 14.0.0
- flatted: 3.3.1 → 3.4.2
- immutable: 3.8.2 → 3.8.3
- serialize-javascript: 6.0.2 → 7.0.4
- terser-webpack-plugin: 5.3.16 → 5.4.0

PR #6399, opened 2026-03-19
[npm] regenerate frontend package-lock.json

nymkappa:
Looks like the package-lock.json is outdated on master, which might be why https://github.com/mempool/mempool/pull/6397 CI does not work

Might be related to a `npm` bug ? https://github.com/npm/cli/issues/8726


nymkappa:
Not needed anymore, issue was fixed by bumping npm in https://github.com/mempool/mempool/pull/6400

PR #6414, opened 2026-03-29
Revert "feat: Address chunking by color & spaces"

wiz:
Reverts mempool/mempool#6319


mononaut:
you'll need to revert #6410 at the same time if you want to remove this feature

PR #6424, opened 2026-03-31
add scripts for checking for common supply chain attack vectors

mononaut:
Adds utility scripts to help detect and reduce the risk of supply-chain attacks from dependency updates, such as https://github.com/mempool/mempool/pull/6421/changes#diff-f89763fc11c1bc48ac3c7ce8a148fbbbc5c2062393ee78664a70273d714daf5cR5954-R5959

- `backend/meta/scripts/check-install-scripts.sh`
- `backend/meta/scripts/safe-install.sh`
- `frontend/meta/scripts/check-install-scripts.sh`
- `frontend/meta/scripts/safe-install.sh`

---

`check-install-scripts.sh` audits the npm lockfiles for packages with `hasInstallScript: true` and fails unless every such package is explicitly whitelisted.

For frontend, it checks `frontend/package-lock.json`.

For backend, it checks both:
- `backend/package-lock.json`
- the install chain used by the backend `preinstall` step in `rust/gbt`, by generating a temporary lockfile for the pinned `@napi-rs/cli` dependency and auditing that as well

Use this to verify that a dependency update PR, such as a Dependabot PR, is not introducing unexpected install scripts.

---

`safe-install.sh` refreshes `package-lock.json` with `npm install --ignore-scripts --package-lock-only`, audits the resulting lockfile with `check-install-scripts.sh`, and only then runs `npm ci` if the audit passes.

If the audit phase fails or is interrupted, the script restores the original `package-lock.json`. Because the audit phase is lockfile-only, it does not leave newly installed packages behind in `node_modules`.

Use this instead of `npm install` when evaluating an untrusted dependency upgrade.

### DIFF
--- a/frontend/src/app/components/liquid-master-page/liquid-master-page.component.html
+++ b/frontend/src/app/components/liquid-master-page/liquid-master-page.component.html
@@ -54,7 +54,7 @@
       [attr.title]="'Select network. Current: ' + networkDisplayName">
       <app-svg-images class="d-flex justify-content-center align-items-center current-network-svg" [name]="network.val === '' ? 'liquid' : network.val" width="20" height="20" viewBox="0 0 125 125"></app-svg-images>
     </button>
-    <div ngbDropdownMenu [ngClass]="{'dropdown-menu-right' : isMobile}">
+    <div ngbDropdownMenu [ngClass]="{'dropdown-menu-end' : isMobile}">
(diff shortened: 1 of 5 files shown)


mononaut:
@knorrium it might be good to add a CI check for this, so that we don't accidentally merge dependabot PRs like https://github.com/mempool/mempool/pull/6421 which are compromised in this way.

another red flag was the lack of attestation mentioned in dependabot's PR description:

> Attestation changes
> This version has no provenance attestation, while the previous version (1.13.5) was attested. Review the [package versions](https://www.npmjs.com/package/axios?activeTab=versions) before updating.

maybe we can automatically reject dependabot PRs with this warning too?

PR #6419, opened 2026-03-30
Light mode ◻️

natsoni:
Builds on #6022

This PR adds light mode to the list of color themes.

PR #6426, opened 2026-03-31
Turn on the lights

natsoni:


mononaut:
we should turn the lights off again once April 1st is over (unless the user voluntarily selected the nymkappa theme themselves)

mononaut:
I merged #6419 already, so let's rebase this on master for easier review

natsoni:
Rebased, and updated the logic to only force nymkappa theme on April 1st, and don't persist it in local storage so that on April 2nd it goes back to whatever user-selected theme.

This means that on April 1st we'll apply nymkappa theme on every initial load, even if user changed the theme in a previous session, is that fine? That might be a bit pushy lol

mononaut:
no, that'll get super annoying 😂

can we keep the local storage marker so it only runs once, but separately track when users manually change the theme. then, from April 2nd onwards, reset to default/previous preference if:
 - if the marker *is* set
 - theme is still nymkappa
 - but the user didn't choose it manually

mononaut:
and ofc we'll revert all of this logic on the next deploy after April 1st

mononaut:
reworked the logic slightly.

now:
 - enable the theme once on first load during April 1st.
 - save the user's previous preference to localStorage.
 - restore the saved preference on first load *after* April 1st.
 - unless the user manually chose a theme first.
 - nothing happens on custom dashboards with configured themes

PR #6432, opened 2026-04-01
[accelerator] use new invoice api

nymkappa:
Switch to the new accelerator invoice API, the legacy one is still kept (forever?) for umbrel users (the same new code is executed regardless).

Requires https://github.com/mempool/mempool.space/pull/2265


mononaut (approved):
utACK @ [148fcd9f]

PR #6445, opened 2026-04-03
Update GH actions workflows

knorrium:


wiz:
Setting the token on the command line is even less secure than setting it in the environment

PR #6446, opened 2026-04-05
Recent transactions page

knorrium:
Adding a new recent transactions page, linked in the Recent Transactions widget to show more than 5 recent transactions.

The menu bar allows capping the total number of txs, and the pause icon allows pausing the updates without unsubscribing from the websocket data.

This is useful to find transactions of interest without misclicking on the main page.

E2E tests are included to verify that the page updates, pauses updating and caps the total of transactions.

Link on the widget
Screenshot 2026-04-05 at 4 21 21 PM

Desktop page
Screenshot 2026-04-05 at 4 21 14 PM

Mobile view
Screenshot 2026-04-05 at 5 38 21 PM

Screenshot 2026-04-05 at 5 38 05 PM


knorrium:
Changed to pause and resume icons:

Screenshot 2026-04-05 at 5 59 00 PM
Screenshot 2026-04-05 at 5 58 57 PM

nymkappa (dismissed):
Since we have a component for this, we should use it in the dashboard and cleanup all the duplicated code from the dashboard component ?

nymkappa:
This UI behavior is also weird. I can imagine this being super frustrating on mobile haha.

https://github.com/user-attachments/assets/daa38257-54d5-4ceb-8c8c-d42ff4c65d9f

knorrium:
Cleaned up the duplicated code and added a twitter timeline like behavior when scrolling to the bottom:

Screenshot 2026-04-08 at 3 41 23 PM

Transactions are paused and the table stops updating until scrolling back up manually or clicking on the pill.

nymkappa (commented):
Tested ACK @ 0a1471355f7089bf69d4e0835597cfe6d4ee4a95.

Copilot has decent comments to review before merging.

frontend/src/app/components/recent-transactions-list/recent-transactions-list.component.ts
```diff
@@ -0,0 +1,172 @@
...
+            this.pausedNewTxids.clear();
+            this.bufferedCount = 0;
+            const result = [...newTxs, ...acc].slice(0, limit);
+            this.pausedTransactions = result;
+            return result;
+          }, this.pausedTransactions),
+          startWith(this.pausedTransactions),
+        );
+      }),
+    );
```
mononaut: we could translate a lot of this into more idiomatic rxjs, to avoid holding and managing so much state manually

e.g, something like:

```
// deduplicated FIFO queue of live transactions
const accumulated$ = this.stateService.transactions$.pipe(
  filter((txs): txs is TransactionStripped[] => txs !== null && txs !== undefined),
  scan((acc, txs) => {
    // insert into map in chronological order
    for (let i = txs.length - 1; i >= 0; i--) {
      const tx = txs[i];
      if (!acc.has(tx.txid)) {
        acc.set(tx.txid, tx);
      }
    }
    while (acc.size > 1000) {
      acc.delete(acc.keys().next().value);
    }
    return acc;
  }, new Map<string, TransactionStripped>()), // ES6 maps preserve ordering
  map((acc) => Array.from(acc.values()).reverse()), // newest-first for rendering
  shareReplay(1), // multicast to transactions$ + bufferedCount$
);

const paused$ = combineLatest([this.manualPaused$, this.autoPaused$]).pipe(
  map(([m, a]) => m || a),
  distinctUntilChanged(),
);

this.transactions$ = combineLatest([accumulated$, this.limit$, paused$]).pipe(
  filter(([, , paused]) => !paused), // stop updating while paused
  map(([live, limit]) => live.slice(0, limit)), // otherwise update in real time and enforce the limit
);

// keep track of how many new txs have arrived since the last pause for the pill button
this.bufferedCount$ = combineLatest([accumulated$, paused$]).pipe(
  scan((state, [live, paused]) => {
    if (!paused) {
      return { frozen: null as number | null, count: 0 };
    }
    const frozen = state.frozen ?? live.length;
    return { frozen, count: Math.max(0, live.length - frozen) };
  }, { frozen: null as number | null, count: 0 }),
  map((state) => state.count),
  distinctUntilChanged(),
);
```

frontend/src/app/components/recent-transactions-list/recent-transactions-list.component.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  trackByTxid(index: number, tx: TransactionStripped): string {
+    return tx.txid;
+  }
+
+ngOnDestroy(): void {
+    this.currencySubscription?.unsubscribe();
+  }
+}
```
mononaut: lost some indentation
```suggestion
  ngOnDestroy(): void {
      this.currencySubscription?.unsubscribe();
    }
  }
```

frontend/src/app/components/recent-transactions-list/recent-transactions-list.component.ts
```diff
@@ -0,0 +1,172 @@
...
+  isLoading = true;
+  isManualPaused = false;
+  isAutoPaused = false;
+  bufferedCount = 0;
+  txLimit = 50;
+  limitOptions = [10, 50, 100, 500, 1000];
+  private pausedTransactions: TransactionStripped[] = [];
+  private pausedNewTxs: TransactionStripped[] = []; // buffered while paused, newest first
+  private pausedNewTxids = new Set<string>();
+  private limit$ = new BehaviorSubject<number>(50);
```
mononaut: we could defer initializing this observable until ngOnInit to avoid immediately replacing the value in the widget case

frontend/src/app/components/recent-transactions-list/recent-transactions-list.component.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  constructor(
+    public stateService: StateService,
+    private websocketService: WebsocketService,
+    private seoService: SeoService,
+    private ogService: OpenGraphService,
```
mononaut: unused?
```suggestion
```

frontend/src/app/components/recent-transactions-list/recent-transactions-list.component.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  transactions$: Observable<TransactionStripped[]>;
+  network$: Observable<string>;
+  currency: string;
+  currencySubscription: Subscription;
+  isLoading = true;
+  isManualPaused = false;
+  isAutoPaused = false;
+  bufferedCount = 0;
+  txLimit = 50;
```
mononaut: probably don't need *both* a `txLimit` field *and* a `limit$` Subject - just bind to the subject in the template directly with `(limit$ | async)` and/or feed it into the other observable pipelines where required.

mononaut (commented):
Tested Concept ACK @ [0a147135]

looks and works great, and comes with e2e tests 😍

I've left a few code suggestions inline, but it's mostly a matter of preference and style.

knorrium:
@mononaut it should be good to go now, thanks for reviewing it

mononaut (approved):
Tested ACK @ [0ea93a40]

mononaut:
seems like the recent txs page e2e tests often time out when mempool activity is low, because it waits for the trailing skeleton loader to disappear, which only happens after accumulating 50 recent txs

knorrium:
> seems like the recent txs page e2e tests often time out when mempool activity is low, because it waits for the trailing skeleton loader to disappear, which only happens after accumulating 50 recent txs

@mononaut I noticed that too after bumping the timeout. I think I'll just mock the websocket here instead of relying on real data.

PR #6454, opened 2026-04-07
[bitcoin] BIP21 invoices QR codes - remove LBTC

nymkappa:
Requires https://github.com/mempool/mempool.space/pull/2300

You should be able to test the frontend change by proxying `/api/v1/services` to `mempool.space/api/v1/services`, so you don't need your own local instance of btcpay.


mononaut (approved):
Tested ACK @ [ee7e97f3]

BIP21 QR scans nicely with various wallets, in both lightning-only and dual formats

PR #6455, opened 2026-04-07
Fix page jump when toggling tx details

natsoni:
Fixes page jump when toggling transaction details: 

https://github.com/user-attachments/assets/b7a88773-bd83-4314-967d-a4b6e4e026cf

### DIFF
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -376,7 +376,7 @@ if (!window['isMempoolSpaceBuild']) {
 @NgModule({
   imports: [RouterModule.forRoot(routes, {
     initialNavigation: 'enabledBlocking',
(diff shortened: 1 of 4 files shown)


mononaut (changes requested):
we generally want to avoid manipulating the URL/history state outside of Angular Router, otherwise stuff can get out of sync.

e.g with this PR:
- load the tx page with showDetails=true
- toggle showDetails via the Location API
- trigger another navigation with `queryParamsHandling: 'merge'`
- Angular Router still thinks showDetails is set and so updates the URL state & fires the `queryParams` observable incorrectly.

https://github.com/user-attachments/assets/afa8d593-883a-49e8-9d92-f2e169e59e6a

mononaut (approved):
Tested ACK @ [311204a9]

PR #6456, opened 2026-04-07
Fix: Bump Rust dependencies

junderw:
Ran `cargo update tokio bytes tracing-subscriber` to update specific dependencies that reported some vulnerabilities.

This is only a Rust lockfile update.


mononaut (approved):
Tested ACK @ [80d589ac]

PR #6459, opened 2026-04-08
[partner] set code directly from subdomain

nymkappa:
Resolves https://github.com/mempool/mempool.space/issues/2228


frontend/src/app/components/tracker/tracker.component.ts
```diff
@@ -870,5 +863,26 @@ export class TrackerComponent implements OnInit, OnDestroy {
...
+      if (info?.name) {
+        this.partnerCode = info.name;
+      } else {
+        const fragmentsParams = new URLSearchParams(window.location.hash.slice(1));
+        this.partnerCode = fragmentsParams.get('partnerCode') ?? this.storageService.getValue('partnerCode') ?? undefined;
+      }
+      if (this.partnerCode) {
+        this.storageService.setValue('partnerCode', this.partnerCode);
+        // Prevents the same partner code from being re-applied upon reloading the page
+        window.location.hash = window.location.hash.replace(`&partnerCode=${this.partnerCode}`, '').replace(`#partnerCode=${this.partnerCode}`, '');
```
wiz, line 884: @mononaut is this ok?

nymkappa: @mononaut is this ok? ❤️

mononaut: I realize this pattern was pre-existing, but no not really - directly manipulating the `window.location.hash` conflicts with Angular Router state. also, as written, if partnerCode is the first of multiple fragment params this breaks the URL.

I think the correct approach requires a slightly deeper refactor, so I've opened a separate PR which also includes this code-from-subdomain feature: #6520

mononaut:
closing in favor of #6520

PR #6460, opened 2026-04-08
Fix apple-mobile-web-app-capable deprecation warning

knorrium:
Fixing the following warning:

`<meta name="apple-mobile-web-app-capable" content="yes"> is deprecated. Please include <meta name="mobile-web-app-capable" content="yes">`


frontend/src/index.mempool.strategy.html
nymkappa, line 1: @mononaut I don't think we use those index anymore right we migrated everything to custom dashboards?

mononaut: I still use them for local dev builds, but there's no reason they need to be checked in to git

mononaut (approved):
utACK @ [42f6a93e]

PR #6461, opened 2026-04-09
Fix footer social icons

natsoni:


nymkappa (approved):
Tested ACK @ ad9c046b8b8150980c55b9a036bc472ed546d2bf

PR #6463, opened 2026-04-10
alternative scroll jumping fix

mononaut:
alternate approach to #6455, using a custom ViewportScroller to exempt same-page navigations from the default scroll reset behavior


natsoni:
Tested ACK @ [73be92d933015f957b7a369057e106ca253c9706]

Thanks that's cleaner! I assume we need to apply the same approach to the hybrid frontend

PR #6467, opened 2026-04-10
Fix currency pipe

knorrium:
Fixing the currency pipe which affected the Recent Transactions and Recent Treasury Transactions widgets.

Note that this does not fix the truncation issue for long numbers, especially in the case of JPY values:

Screenshot 2026-04-10 at 3 14 26 PM

Before:

Screenshot 2026-04-10 at 3 08 04 PM

Screenshot 2026-04-10 at 3 13 51 PM

After:

Screenshot 2026-04-10 at 3 08 11 PM

Screenshot 2026-04-10 at 3 13 45 PM


nymkappa:
The digits should be shown for small number eg $14 should be $14.69. I think it makes sense to not show decimal if the value is >= 1000

This is more for UI balance that actually be useful

nymkappa (approved):
Tested ACK @ 445392be37248f937b46f50363b7147ffc1a530f

PR #6473, opened 2026-04-17
Guard against undefined block

knorrium:
Noticed these while looking at Umbrel logs.

If `blocksByHash[hash].height` is undefined, the second part (`|| (await bitcoinApi.$getBlock(hash))?.height;`) does not run and the error is thrown. Needs to be tested on device.

Screenshot 2026-04-17 at 10 25 10 AM


backend/src/repositories/BlocksRepository.ts
```diff
@@ -687,6 +688,10 @@ class BlocksRepository {
         FROM blocks
         ORDER BY height DESC
       `);
+      if (!blocks || blocks.length === 0) {
+        logger.warn('Cannot validate chain: no indexed blocks in database');
+        return true;
```
mononaut: I think we want to throw here instead.

- returning `true` means "chain validated successfully"
- returning `false` as copilot suggests means "chain divergence detected"
- but an absence of any indexed blocks is actually an error state since it implies `generateBlockDatabase()` hasn't run successfully and so we can't even attempt to validate the chain.
```suggestion
        throw new Error('Cannot validate chain: no indexed blocks in database'); 
```

backend/src/repositories/BlocksRepository.ts
```diff
@@ -703,7 +708,12 @@ class BlocksRepository {
       // ensure that indexed blocks are correctly classified as stale or canonical
       // iterate back to genesis, resetting canonical status where necessary
       let hash = tip;
-      const tipHeight = blocksByHash[hash].height || (await bitcoinApi.$getBlock(hash))?.height;
+      const indexedTip = blocksByHash[hash];
+      const tipHeight = indexedTip?.height ?? (await bitcoinApi.$getBlock(hash))?.height;
+      if (tipHeight == null || typeof tipHeight !== 'number') {
+        logger.warn('Cannot validate chain: could not resolve tip block height from index or node');
+        return true;
+      }
```
mononaut: first null check is redundant.

again, we want to throw in this case:
- we can't validate the chain if the tip isn't indexed
- but that doesn't mean the chain has diverged (as copilot's suggestion to return `false` would imply)
- so we should throw and try again on the next run once block indexing has caught up.
```suggestion
      if (typeof tipHeight !== 'number') {
+       throw new Error(`Cannot validate chain: could not resolve tip block height for ${hash} from index or node`);
      }
```

backend/src/repositories/BlocksRepository.ts
```diff
@@ -741,7 +752,9 @@ class BlocksRepository {
 
       if (firstBadBlockHeight != null) {
         logger.warn(`Chain divergence detected at block ${firstBadBlockHeight}`);
-        await HashratesRepository.$deleteHashratesFromTimestamp(blocksByHash[firstBadBlockHeight].timestamp - 604800);
+        if (firstBadBlockTimestamp != null) {
+          await HashratesRepository.$deleteHashratesFromTimestamp(firstBadBlockTimestamp - 604800);
+        }
```
mononaut, line 755: good catch

mononaut (changes requested):

mononaut (approved):
Tested ACK @ [b84df41c]

PR #6478, opened 2026-04-22
Generate config array support

knorrium:
Add proper support for array-based config overrides.

Screenshot 2026-04-22 at 11 48 06 AM


wiz:
I don't think we actually need array support, we should probably just keep the comma separated list

knorrium:
No longer needed

PR #6502, opened 2026-05-08
Update provenance check json keys

knorrium:
Update the keys so the provenance check workflow can pull the updated dependency version again


mononaut (approved):
ACK @ [2fe8a4ce]

PR #6503, opened 2026-05-08
Skip docker step for ops prs

knorrium:


mononaut (approved):
ACK @ [7e2a25e1]

the docker step seems to be correctly skipped on this PR with the ops label, so I assume this is working as intended?

mononaut:
the four pending Backend/Frontend checks seem to be stuck, but I'm guessing that's unrelated?

knorrium:
> the four pending Backend/Frontend checks seem to be stuck, but I'm guessing that's unrelated?

@mononaut yes, because they are marked as required to merge, but as an admin can you bypass and merge it anyway?

we could also make the steps a no-op but it could allow anyone to skip them if they use the ops tag or branch prefix, so I’d rather make admins decide when to skip the protections

wiz (approved):

wiz:
can't merge because tests are stuck

wiz:
feel free to re-open if still needed

PR #6520, opened 2026-05-21
refactor partnerCode handling

mononaut:
This PR extracts partnerCode handling into a small self-contained service so that we have a single source-of-truth and more consistent/predictable behavior across the various components that rely on this state.

also resolves https://github.com/mempool/mempool.space/issues/2228 and supersedes #6459

### DIFF
--- a/frontend/src/app/components/tracker/tracker.component.ts
+++ b/frontend/src/app/components/tracker/tracker.component.ts
@@ -27,12 +27,12 @@ import { BlockExtended, CpfpInfo, RbfTree, MempoolPosition, DifficultyAdjustment
 import { PriceService } from '@app/services/price.service';
 import { ServicesApiServices } from '@app/services/services-api.service';
 import { EnterpriseService } from '@app/services/enterprise.service';
+import { PartnerCodeService } from '@app/services/partner-code.service';
 import { ZONE_SERVICE } from '@app/injection-tokens';
 import { TrackerStage } from '@components/tracker/tracker-bar.component';
 import { MiningService, MiningStats } from '@app/services/mining.service';
 import { ETA, EtaService } from '@app/services/eta.service';
 import { getTransactionFlags, getUnacceleratedFeeRate } from '@app/shared/transaction.utils';
-import { StorageService } from '@app/services/storage.service';
 
 
 interface Pool {
@@ -84,6 +84,8 @@ export class TrackerComponent implements OnInit, OnDestroy {
   blocksSubscription: Subscription;
   miningSubscription: Subscription;
   currencyChangeSubscription: Subscription;
+  urlFragmentSubscription: Subscription;
+  partnerCodeSubscription: Subscription;
   rbfTransaction: undefined | Transaction;
   replaced: boolean = false;
   latestReplacement: string;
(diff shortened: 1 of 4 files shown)


frontend/src/app/components/tracker/tracker.component.ts
```diff
@@ -869,6 +866,36 @@ export class TrackerComponent implements OnInit, OnDestroy {
...
+    if (partnerCode) {
+      this.partnerCodeService.setFragmentPartnerCode(partnerCode);
+    }
+    fragmentParams.delete('partnerCode');
+    this.router.navigate([], {
+      relativeTo: this.route,
+      fragment: fragmentParams.toString() || null,
+      queryParamsHandling: 'merge',
+      replaceUrl: true,
+    });
```
nymkappa, line 899: I think we can also wrap this in the partner code service?

nymkappa: All components which needs to read the partner code follow the same logic right.

1. Get partner code (from subdomain || from url || from local storage)
2. Clean up the url

nymkappa: So I would go as far as saying that consumers of the partner-code service should only have one call it to it `getPartnerCode()` and the partner-code service would just take care of everything maybe.

mononaut: ideally yes, but the awkward part is how that interacts with the other pages' fragment handling and navigation.

it's nice for e.g. the transaction page component to fully "own" the url and internal navigation for its own route, then you can consider that logic in isolation without worrying about side-effects leaking in from elsewhere.

frontend/src/app/services/partner-code.service.ts
```diff
@@ -0,0 +1,52 @@
...
+    // enterprise partner code takes precedence, and remains applicable for all requests on this instance
+    // one-off code applies to the first successful request after being set, iff no enterprise code is present
+    this.partnerCode$ = combineLatest([
+      enterprisePartnerCode$,
+      this.storedPartnerCode$,
+    ]).pipe(
+      map(([enterpriseCode, storedCode]) => enterpriseCode || storedCode),
+      distinctUntilChanged(),
+      shareReplay(1),
+    );
```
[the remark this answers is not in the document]
nymkappa: I think this is fine because if the $info call fails, the user will probably be redirected to the main website anyway so all this does not matter anymore?

frontend/src/app/services/partner-code.service.ts
```diff
@@ -0,0 +1,52 @@
+import { Injectable } from '@angular/core';
+import { BehaviorSubject, combineLatest, Observable } from 'rxjs';
+import { distinctUntilChanged, filter, map, shareReplay } from 'rxjs/operators';
+import { EnterpriseService } from '@app/services/enterprise.service';
```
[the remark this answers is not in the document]
nymkappa: See my comment above

nymkappa:
I've noticed that http://localhost:4200/tx/cad8ef7199c14bd9d887241e7d6f27d6767dfb3d1526e09e5c0e7e16b1491785#accelerate&partnerCode=sHDE72yN would result in http://localhost:4200/tx/cad8ef7199c14bd9d887241e7d6f27d6767dfb3d1526e09e5c0e7e16b1491785#accelerate= with a trailing `=` which has been fixed in https://github.com/mempool/mempool/pull/6520/commits/5aecc87d9c13afe57136e01051b640d4f2f0b44d

nymkappa (approved):
Otherwise my local tests looks good to me 👍🏻

nymkappa (approved):
Re-tested ACK @ 361fbbe63b869cad7a0cadf69f807d67ea9185cf. Thanks for simplifying the string manipulation.

The extra `=` does not seem to be a an issue anymore, and the subdomain code still properly replaces the one passed in the url (if any).

I see some doc updated sneaked in but that's cool haha.

mononaut:
thanks!

I just rebased on master and fixed the trailing '=' issue - the doc change you saw was probably just the diff between the original branch and the rebased one with recent commits included, not actually part of this PR.

PR #6548, opened 2026-06-09
fix: False positive coinjoin (#6464)

### LINKED ISSUE #6464: False positive coinjoin flag in goggles
Transactions re-using same address in 1000 inputs are incorrectly flagged as coinjoins, see:

- https://mempool.space/tx/d956c419e95adafcc39584a41a232bcecd2428051dde04132cdb469731cf60ce
- https://mempool.space/tx/03159c7534e3b82d6452c6174fc4cffce25019e34b00777c902d679639c27ea3
- https://mempool.space/tx/d339704a51e85280a293b28e0a391f08ed888fc07c2f0a545014fe65f7470b69

etc.

rodribp:
fixes #6464

In the case of the following txs:
https://mempool.space/tx/d956c419e95adafcc39584a41a232bcecd2428051dde04132cdb469731cf60ce
https://mempool.space/tx/03159c7534e3b82d6452c6174fc4cffce25019e34b00777c902d679639c27ea3
https://mempool.space/tx/d339704a51e85280a293b28e0a391f08ed888fc07c2f0a545014fe65f7470b69

They are flagged as coinjoin, but we can easily recognize they are not, as there's not as many outputs with the same amount.

So I added a condition that checks if a fifth of the inValues count at least 5 inputs with the same amount. The same for the outputs, but individually.

I was able to test inside the tx page only, not in block overview graph, as I'm still not able to run the backend.

Feedback is appreciated!

### DIFF
--- a/backend/src/api/common.ts
+++ b/backend/src/api/common.ts
@@ -749,7 +749,15 @@ export class Common {
     // fast but bad heuristic to detect possible coinjoins
     // (at least 5 inputs and 5 outputs, less than half of which are unique amounts, with no address reuse)
     const addressReuse = Object.keys(reusedOutputAddresses).reduce((acc, key) => Math.max(acc, (reusedInputAddresses[key] || 0) + (reusedOutputAddresses[key] || 0)), 0) > 1;
-    if (!addressReuse && tx.vin.length >= 5 && tx.vout.length >= 5 && (Object.keys(inValues).length + Object.keys(outValues).length) <= (tx.vin.length + tx.vout.length) / 2 ) {
+    const tokenRelated = (flags & (TransactionFlags.inscription | TransactionFlags.op_return)) !== 0n;
+    if (!addressReuse &&
+        tx.vin.length >= 5 &&
+        tx.vout.length >= 5 &&
+        (Object.keys(inValues).length + Object.keys(outValues).length) <= (tx.vin.length + tx.vout.length) / 2 &&
+        !tokenRelated &&
+        tx.vin.length / tx.vout.length < 5 &&
+        tx.vin.length / tx.vout.length > 0.2
+      ) {
       flags |= TransactionFlags.coinjoin;
     }
     // more than 5:1 input:output ratio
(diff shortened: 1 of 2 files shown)


mononaut (changes requested):
I suspect this new heuristic is too strict, and might make us miss a lot of genuine coinjoins:

[wabisabi-type coinjoins](https://mempool.space/tx/c178152b5379bd416e83bd843f5780455b36c3363cbfbc1c8e63b86b9533e3b3) mix different denominations with common factors at the same time, so there aren't necessarily always very many inputs or outputs of the same amount.

and [joinmarket-type coinjoins](https://mempool.space/tx/5771b3bde6ff14bf0f45e3472700900fd8f20fd3e3e9546d9aa0b1e4f2fcfab9) bake fees & change into the protocol, so input amounts are almost always all unique, and many of the outputs are unique amount change outputs. although tbf we already fail to detect most joinmarket coinjoins anyway.

for the false positives you identified, it seems to me that the main reasons they're obviously not coinjoins are:
1) it's clearly something token related (runestone output, dust amounts)
2) the repeated amount is just dust.
    - unlikely anyone is wasting money coinjoining such small amounts;
    - being dust is sufficient to explain why the amounts are all the same, which makes "non-unique values" much weaker evidence of anything else, including being a coinjoin.
3) it's clearly primarily a *consolidation*.
4) relatedly, there are only two same-value outputs vs ~a thousand same-value inputs, so 99.8% of those inputs can't be gaining any anonymity.

one or more of those reasons would probably make for a better heuristic. e.g. we could:
1) exclude likely "token related" transactions (anything with op_returns, inscriptions, or weird sighash flags).
2) exclude dust amounts from the input/output analysis (or all amounts below a certain value).
3) exclude transactions already matched by the "consolidation" or "batch payment" heuristics.
4) do something more sophisticated to identify how many inputs/outputs gain any meaningful extra anonymity.

mononaut:
longer term, it might be interesting to do a more rigorous analysis of past coinjoins from all of the major protocols (whirlpool, wabisabi, joinmarket etc) and properly backtest different heuristics to find one with the best false positive & false negative rates.

rodribp:
Just added your proposed heuristics 1 and 3 as it's not too strict but solves the issue

backend/src/api/common.ts
```diff
@@ -657,7 +658,11 @@ export class Common {
           case 'v1_p2tr': {
             flags |= TransactionFlags.p2tr;
             if (vin.witness?.length) {
-              flags = Common.isInscription(vin, flags);
+              const newFlags = Common.isInscription(vin, flags);
+              if (flags !== newFlags) {
+                tokenRelated = true;
+              };
+              flags = newFlags;
```
mononaut: this construction is a bit awkward.

instead of maintaining `tokenRelated` separately in several different places, I'd suggest calculating it once, immediately before the coinjoin section, like:

```
const tokenRelated = (flags & (TransactionFlags.inscription | TransactionFlags.op_return)) !== 0n;
```

so that we have a single source of truth (the `flags` bitarray), and all of the logic for the coinjoin heuristic stays together in one place.

mononaut (approved):
Tested ACK @ [bf60cf72]

works great, good job!

PR #6549, opened 2026-06-09
Increase demo cycle time from 15s to 5mins

orangesurf:


wiz (changes requested):
how about making it configurable as another query param

PR #6559, opened 2026-06-11
Add detailed data formats and verification process to Taptree doc

natsoni:
Adds the following the Taptree doc entry:
- details on what data exactly can be used to add new leaves to the Taptree
- step by step verification process of the passed data

Screenshot 2026-06-11 at 12 21 21


mononaut (approved):
Tested ACK @ [dc34f617]

PR #6561, opened 2026-06-12
add: Inifinite scrolling to blocks/stale page

### LINKED ISSUE #6555: paginate stale blocks
The stale blocks page currently fetches and displays up to the latest 50 stale chain tip known by the backend.

This is a much larger response than necessary for casual browsing, but also insufficient to explore older stale blocks (especially on more fork-happy test networks).

Instead,

1) The underlying `/stale-tips` API should be properly paginated, and support paging back through all stale blocks rather than just the latest 50 in memory.
2) The frontend /blocks/stale page should be updated to use this paginated API, either via explicit page buttons like on the /blocks page, or lazy-loading infinite scroll like transactions on the /address or /block pages.

Bear in mind that the set of stale blocks isn't guaranteed to be stable - items can change near the tip due to reorgs, and in future new stale blocks may be added via replication from other servers - so we probably want to use a height-based cursor rather than one based on blockhashes or page numbers. Maybe a similar shape to the regular paginated blocks API `/api/v1/blocks/:height`.

rodribp:
Closes #6555

### Summary

This PR implements a lazy loader infinite scroll in the blocks/stale page, it previously rendered a list of the latest 50 items in memory with no chance to see older stale blocks. This PR adds height based pagination served from the database history where stale = 1, falling back to cache when indexing is disabled and prioritizing current cache stale blocks.

### Backend

- Added new query for stale blocks in BlocksRepository.$getStaleBlocks.
- Added ChaintTips.$getStaleTipsPage(fromHeight, limit) to assemble the ChainTips from BlockExtended type.
- $getStateTipsPage returns cachedTips if indexing is disabled and if the stale block from the database matches a cachedBlock hash.
- Route added `stale-tips/:height`

### Frontend

- Added height parameter to ApiService.getStaleTips$.
- Stalelist component refactored for infinite scrolling with loading spinner at the bottom.

### Considerations
- DB-only stale blocks have no status, returning an undefined status to the frontend, so maybe is a good idea to either include the statuses in the database or make a default badge for these cases.
- Branchlen (Depth) of DB stale blocks is inferred using previousblockhash.
- Tested with a regtest node.

https://github.com/user-attachments/assets/ba186f2d-d5d0-4cea-acef-6a4418022a11

### DIFF
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -61,6 +61,7 @@ class BitcoinRoutes {
       .get(config.MEMPOOL.API_URL_PREFIX + 'blocks-bulk/:from/:to', this.getBlocksByBulk.bind(this))
       .get(config.MEMPOOL.API_URL_PREFIX + 'chain-tips', this.getChainTips.bind(this))
       .get(config.MEMPOOL.API_URL_PREFIX + 'stale-tips', this.getStaleTips.bind(this))
+      .get(config.MEMPOOL.API_URL_PREFIX + 'stale-tips/:height', this.getStaleTips.bind(this))
       .post(config.MEMPOOL.API_URL_PREFIX + 'prevouts', this.$getPrevouts)
       .post(config.MEMPOOL.API_URL_PREFIX + 'cpfp', this.getCpfpLocalTxs)
       // Temporarily add txs/package endpoint for all backends until esplora supports it
@@ -628,14 +629,18 @@ class BitcoinRoutes {
   private async getStaleTips(req: Request, res: Response) {
     try {
       if (['mainnet', 'testnet', 'signet', 'testnet4', 'regtest'].includes(config.MEMPOOL.NETWORK)) { // Bitcoin
-        res.setHeader('Expires', new Date(Date.now() + 1000 * 60).toUTCString());
-        const tips = await chainTips.getStaleTips();
-        if (tips.length > 0) {
-          res.json(tips);
-        } else {
-          handleError(req, res, 503, `Temporarily unavailable`);
-          return;
+        let fromHeight: number | undefined;
+        if (req.params.height !== undefined) {
+          fromHeight = parseInt(req.params.height, 10);
+          if (isNaN(fromHeight) || fromHeight < 0) {
+            handleError(req, res, 400, `Parameter 'height' must be a block height (integer)`);
+            return;
+          }
         }
+
+        res.setHeader('Expires', new Date(Date.now() + 1000 * 60).toUTCString());
+        const tips = await chainTips.$getStaleTipsPage(fromHeight, 10);
+        res.json(tips);
       } else { // Liquid
         handleError(req, res, 404, `This API is only available for Bitcoin networks`);
         return;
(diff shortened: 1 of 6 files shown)


mononaut (commented):
thanks, this is a good first pass, and the general pagination/infinite-scroll approach looks reasonable.

I'm realizing as I review this that my issue description was confusing: the `/blocks/stale` page and `/api/v1/stale-tips` endpoint are currently designed to list stale *chain tips* rather than *all stale blocks*.

some of my existing implementation is a bit confused about the distinction there too, which doesn't help 😅

on mainnet where forks are very rarely more than one block long the stale tips and stale blocks are basically equivalent, but on the testnets there are often much longer forks with many stale *blocks* leading to each stale *tip*, so it's worth figuring out exactly what we're trying to achieve and then making the implementation consistent with that.

I think we have two options:

1) keep the page + endpoint focused on stale *tips* only.

then the implementation becomes a lot simpler, since we already have the height, hash, branchlen, and fork status in memory, direct from the Bitcoin Core RPC, and all we need to do is take a slice and enrich it with copies of the canonical and stale blocks from either the in-memory caches or the DB.

-- or --

2) redesign the `/blocks/stale` page to show *all* stale blocks.

- change the page title
- move the paginated list of stale blocks to a new endpoint with a more appropriate name (perhaps `/api/v1/blocks/stale/:height`)
- think a bit more deeply about what data we actually need from that endpoint.
    - e.g. `branchlen` and `status` are properties of a fork/chain-tip, rather than an individual block.

rodribp:
Hey @mononaut , thanks for your feedback, just changed it to query only stale chain tips from the DB and I get the depth by querying the block with the previousBlockHash and counting until the block.stale == 0.

Additionally, testing, I noticed the depth of cached chain tips was not showing the branchlen from core but it was always 0 (depth 1), because the tip and the block were always the same height (at least in my network), so I added it in updateOrphanedBlocks.

mononaut:
thanks! I'll review again properly later today.

but just from skimming the diff: are you sure we need that nested DB query to identify the stale tips? If I recall correctly, we already have a list of all known chain tips from the bitcoin RPC in `this.chainTips` with their tip block hashes and `branchlen`s, right?

rodribp:
Ok, I see that I should have researched first... I assumed Bitoin's RPC `getchaintips` didn't return all chain tips the node knew about, but it was limited, and I thought that was reason of the field `stale` in the db... So no, no need to query the DB, your option 1, enriching `this.chainTips` with the block's info is way simpler, I want to ask you, then, by using `this.chainTips` we'd no longer use cached `staleTips` so I need to remove it right? Or shall I keep it as it is and query canonical and stale block using db if not available in cache? I guess keep it?

mononaut:
yeah, sorry the terminology was a bit confusing!

I think it's worth keeping the `staleTips` cache for now, so that we don't have to constantly hit the DB for the first couple of pages of results. but we'll want to clean up the implementation so that we're caching the right data (for `branchlen`, etc).

mononaut:
looks great!

I noticed a couple of small things during review that would make it work even better though. rather than spend too much time going back and forth with reviews and requested changes, I've opened a PR (#6579) onto this branch with some fixes and suggestions. feel free to merge the whole thing into your PR, cherry-pick individual commits, or just use as a reference.

rodribp:
Thank you, just did merge your PR!

mononaut (approved):
Tested ACK @ [f6547041]

PR #6564, opened 2026-06-12
Only show accelerator checkout on transaction page with #accelerate

natsoni:
First shot at implementing https://github.com/mempool/mempool.space/issues/2327

This PR improves focus on the accelerator checkout by hiding all transaction elements when displaying the checkout after user clicks `Accelerate` from the transaction details, or `#accelerate` is in the URL.

For consistency we don’t show the accelerator checkout on the transaction page by default anymore, meaning that the accelerator checkout will only appear (alone on the page) if `#accelerate` fragment or user clicks `Accelerate` in the transaction details. but I'm not sure this is actually what we want @orangesurf? 

https://github.com/user-attachments/assets/a0628396-7148-4b7a-ac11-033746ed8e5a

https://github.com/user-attachments/assets/02052474-fe96-4c94-bf88-e4c6902d468f


mononaut:
nice work, but Concept NACK (see my comment on the issue)

orangesurf:
I don't think we should remove the scroll to widget step when clicking accelerate in the tx details.

Clicking the accelerate button **within the widget** should activate the focused view.

mononaut:
@orangesurf can we close this in favor of #6583?

orangesurf:
Yep

PR #6565, opened 2026-06-15
fix: remove indexing block's badge from unnecesary graphs

### LINKED ISSUE #6556: Remove "indexing blocks" message where not relevant
We show an "Indexing blocks" badge on some pages and charts to explain why data isn't yet available.

Image

However, currently it's also shown on several pages and charts which do not rely on any indexed block data, such as the address graph, utxo graph, price chart, etc.

The badge should be removed from these pages.

rodribp:
closes #6556 

This PR removes the `app-indexing-indicator` from the components it is unnecessary, and hides specific loading indicators depending on the component dependency of the progress, following the next table.

| Component | Blocks | Pools hashrate (weekly) | Hashrate (daily) |
|---|---|---|---|
| `block-fee-rates-graph` | ✅ | ❌ | ❌ |
| `block-fees-graph` | ✅ | ❌ | ❌ |
| `block-fees-subsidy-graph` | ✅ | ❌ | ❌ |
| `block-health-graph` | ✅ | ❌ | ❌ |
| `block-rewards-graph` | ✅ | ❌ | ❌ |
| `block-sizes-weights-graph` | ✅ | ❌ | ❌ |
| `blocks-list` | ✅ | ❌ | ❌ |
| `pool-ranking` | ✅ | ❌ | ❌ |
| `pool` | ✅ | ✅ | ❌ |

This is the list of components I ripped off the indexing-indicator completely as they don't depend on any of the three loading indicators:

- `acceleration-fees-graph`
- `address-graph`
- `price-chart`
- `treasuries-graph`
- `utxo-graph`

### DIFF
--- a/frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.html
+++ b/frontend/src/app/components/acceleration/acceleration-fees-graph/acceleration-fees-graph.component.html
@@ -1,4 +1,3 @@
-<app-indexing-progress *ngIf="!widget"></app-indexing-progress>
 
 <div [class.full-container]="!widget">
   <div *ngIf="!widget" class="card-header mb-0 mb-md-4">
--- a/frontend/src/app/components/address-graph/address-graph.component.html
+++ b/frontend/src/app/components/address-graph/address-graph.component.html
@@ -1,4 +1,3 @@
-<app-indexing-progress *ngIf="!widget"></app-indexing-progress>
 
 <div [class.full-container]="!widget">
   <ng-container *ngIf="!error">
--- a/frontend/src/app/components/block-fee-rates-graph/block-fee-rates-graph.component.html
+++ b/frontend/src/app/components/block-fee-rates-graph/block-fee-rates-graph.component.html
@@ -1,4 +1,4 @@
-<app-indexing-progress *ngIf="!widget"></app-indexing-progress>
+<app-indexing-progress *ngIf="!widget" [showNetworkHashrate]="false" [showPoolsHashrate]="false"></app-indexing-progress>
 
 <div [class.full-container]="!widget">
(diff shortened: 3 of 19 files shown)


mononaut (changes requested):
looking at this again, I don't think these particular charts actually need *any* of the indexing progress indicators (blocks, networkHashrate *or* poolsHashrate), so it would be cleaner to just delete the `<app-indexing-progress>` element from their templates entirely.

but your changes to the `IndexingProgressComponent` itself should still be useful, if there are any other charts where some of the indexing labels are relevant but not others.

I suppose the pools and hashrate charts need all three labels, but charts like the block fees, block sizes & weights etc only use block data, not indexed hashrates, and so might not need the hashrate indexing progress labels?

mononaut (changes requested):
logic works great now, but when multiple indicators apply, they're all rendered stacked on top of each other in the same absolute position:

Screenshot 2026-08-01 at 5 44 11 PM

(transparency and border added to illustrate)

rodribp:
I found this bug on scroll as well:
Screenshot 2026-08-03 at 2 39 33 PM

I assume the intended UI is to keep the indicators along with the navbar in an absolute position (53020c05b4e6a04ab89f6cc13aa27fe530880ab7):
Screenshot 2026-08-03 at 2 40 54 PM

OR as part of the router-outlet container (e88c983685103157221644c0ee6c583a8c2644ce):

https://github.com/user-attachments/assets/acbff14f-f063-4d24-aeb2-ffcad5788e14

LMK which UI to keep.

rodribp:
Just squashed down to fixed position at the bottom edge of the navbar (my prefered):
image
mobile (flex-wrap):
image

mononaut (approved):
Tested ACK @ [9b01cb39]

PR #6585, opened 2026-06-23
fix: goggles overflow on Y

### LINKED ISSUE #6581: Goggles filters overflow
### Description

When selecting many filters with the goggles in the block-overview component the list of filters overflow.
#### Version

node202.va1.mempool.space (v3.4-dev) [[1ce9f8b3](https://github.com/mempool/mempool/commit/1ce9f8b3)][14dc90b2]

### Screenshots

Image

#### Device or machine

Happens in both desktop and mobile

rodribp:
fixes #6581 

I propose one of these 2 solutions to avoid the active flags wrapping to push the filter menu out of block-overview-graph.component.ts

https://github.com/user-attachments/assets/a67945ec-74d8-4888-9712-c165c67f564a

https://github.com/user-attachments/assets/8586a894-2741-48d2-af77-c33098656ebf

### DIFF
--- a/frontend/src/app/components/block-filters/block-filters.component.html
+++ b/frontend/src/app/components/block-filters/block-filters.component.html
@@ -6,7 +6,7 @@
     <button class="menu-toggle" (click)="menuOpen = !menuOpen" title="Mempool Goggles&reg;">
       <app-svg-images name="goggles" width="100%" height="100%"></app-svg-images>
     </button>
-    <div class="active-tags">
+    <div class="active-tags" [class.menu-open]="menuOpen">
       <ng-container *ngFor="let filter of activeFilters;">
         <button class="btn filter-tag selected" (click)="toggleFilter(filter)">{{ filters[filter].label }}</button>
       </ng-container>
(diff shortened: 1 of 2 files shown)


mononaut:
looks great! I think the horizontal scrolling solves this very nicely.

but it would be even slicker if you could figure out some kind of truncation/overflow indicator, to show more clearly that there are items hidden offscreen.

rodribp:
Just added this scrollbar


https://github.com/user-attachments/assets/1f76007f-3bfa-45af-aac8-ddd3b0449554

mononaut:
hmm, the scrollbar doesn't look great, I see now why you removed it initially 🤔

how about a subtle CSS mask on the edge of the container, when items overflow in that direction?
Screenshot 2026-06-25 at 4 54 35 AM
Screenshot 2026-06-25 at 4 55 26 AM

rodribp:
Just did, however I used `animation-timeline: scroll()`, which under [caniuse.com](https://caniuse.com/?search=scroll%28%29) seems to be used by 82% of global users.

Otherwise I could use angular conditional class and (scroll) event to show each edge fade accordingly, but is not as clean as `animation-timeline: scroll()`


https://github.com/user-attachments/assets/41296c0b-5a1d-4dbc-b902-44cb94599002

mononaut:
looks great! this is just an extra visual enhancement, so less-than-perfect compatibility is probably fine as long as it degrades gracefully.

mononaut (approved):
Tested ACK @ [d928feb6]

PR #6586, opened 2026-06-23
Mempool goggles graph

### LINKED ISSUE #5608: Graph Mempool Goggles Tx Tag Data
### Description

Add a graph view of the count of transactions tagged with Mempool goggles tags (Y) vs block height (X). 
- abs count 
- % of block by count
- % of block by weight

Allow same selection functionality as we currently have with Mempool goggles selection (AND/OR multiple selection)

### Problem to be solved

Mempool goggles data is currently shown at a tx level only. Block level data may reveal interesting trends over time.

rodribp:
# Summary

This PR implements a new graph `Transactions per block` to TMOSP, which as described in issue #5608, shows the abs tx count per block height, and it's connected to the mempool goggles. 

| No filters | Filters active | Goggles menu open |
| - | - | - |
| image | image | image |

## The chart has two modes:

1. If there's no filters selected it paints a chart under the interval selected (24h default), of the tx count per block.
2. If there are filters selected it paints a chart where y-axis is the txCount per block height, of the txs matching the filters, the tooltip shows: a) Block or startBlock - endBlock (In the case of intervals that use block ranges). b) Matched transaction abs count. c) Share of all txs in the block (%). d) (If the interval uses block ranges) It shows the average of transactions per block in the block range.

| No filters (per block) | No filters (Block ranges) | Filters selected (per block) | Filters selected (Block ranged intervals) |
|-|-|-| -|
| image | image  | image | image |

### Database and query

In order to achieve this I decided based on a benchmarking done and discussed about in #5608, that the best approach was to precompute and index to the database the "Transactions' count per unique bit array bigint value" per block height, so the database counts with a new table with the following structure:

Column | Type | Null | Key | Default | Notes
-- | -- | -- | -- | -- | --
blocks_count | enum('1','36','72','144','432','720','1008') | NO | PRI | — | Bucket size (blocks per aggregation window); part of composite PK
start_height | int unsigned | NO | PRI | — | First block height of the window; part of composite PK
flag_value | bigint unsigned | NO | PRI | — | Distinct transaction flags bitfield value; part of composite PK
tx_count | int unsigned | NO | — | — | Number of transactions in the window that had this exact flag_value

This way, initial indexing takes care of precomputing these values to the database next to `blocks.$generateBlocksSummariesDatabase()`. And once all the flag values have been indexed, the frontend would send to the backend three parameters:
1. Interval. (Required) If it's the only parameter it fallbacks to pure txCount.
2. op. Means "operation" and tells the backend the type of operation ( 'and' | 'or' | 'nor'), gotten from the goggles.
3. mask. The unique bitarray bigint value, that will be used in order to query to the database the tx_counts that meet the operation type against this mask.

So the where clause that validates the txCount to the matching filters are as following:

|and|or|nor|
|-|-|-|
|(flag_value & [mask]) = [mask] AND|(flag_value & [mask]) > 0 AND|(flag_value & [mask]) = 0 AND|

### Indexing

There's three indexing paths that satisfy the flag_values table:

#### 1. On initial sync.

If the instance of mempool has BLOCKS_SUMMARIES_INDEXING set to true, on initial sync it would execute right after indexing the blockSummary, inside `blocks.$generateBlocksSummariesDatabase()`. The function is `blocks.$indexFlagValues()` which keeps a cache memory of the buckets depending if the block is within the height range of the bucket, computes the transaction count per flag appending each tx count to its corresponding flag value already cached or not, and once the bucket size has been reached it inserts it to the table and deletes any out of block span oldest records from the database if applicable.

#### 2. On new block.

Inside `blocks.$saveBlockData` right after `blocks.$indexBlockSummary`, just like 1st path.

#### 3. Reconciliation from blocks_summaries

This path has its own function `blocks.$generateFlagValuesDatabase`, which is a similar algorithm to `blocks.$generateBlocksSummariesDatabase` but its source of truth is the database blocks_summaries, instead of other functions data flows. When flag_values lack records in comparison with blocks_summaries, this function makes sure to backfill the missing flag_values. 

## Video

https://github.com/user-attachments/assets/f2792c7c-616c-45a6-b5b8-ba2402ad0b0e

---

# Notes

1. This PR does not implement % of block weight yet.
2. All tests have been done using a fully synced signet node + esplora.
3. The frontend uses URL params for shareable graph view.
4. The start heights managed across three index paths are fixed calculated based on the height of the block to be processed by `Math.floor(height / blocksCount) * blocksCount`, just so overlaps are not an issue because of ON DUPLICATE KEY when insert.
5. Insert is done using single query with BATCH INSERT instead of many queries iterated.
6. The feature does not introduce an additional config, it is gated under BLOCKS_SUMMARIES_INDEXING and when it is enabled the graph shows a title "Blocks summaries indexing is required for this graph".
7. Each tier based on blocks_count stores only on blockSpan window, deleting everything that's left out of the blockSpanWindow.

Closes #5608

### DIFF
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -1176,16 +1176,18 @@ class BitcoinRoutes {
       const op = (req.params.op) as 'and' | 'or' | 'nor' | undefined;
       const mask = BigInt(req.params.mask ?? 0n);
 
-      const { tip }  = await FlagValueRepository.$getTipAndTailIndexedByBucketSize(bucketSize) || { tip: undefined };
+      const { tip, tail }  = await FlagValueRepository.$getTipAndTailIndexedByBucketSize(bucketSize) || { tip: undefined, tail: undefined };
 
-      if (!tip) {
+      if (tip === undefined || tail === undefined) {
         handleError(req, res, 400, `Failed to get latest indexed flag values for ${interval}`);
         return;
       }
 
+      const totalCount = await FlagValueRepository.$getTotalBlocksIndexedByBucketSize(bucketSize === 1 ? 1008 : bucketSize) ?? tip - tail;
+
       const startHeight = presets[interval].retentionSpan !== -1 ? (tip - presets[interval].retentionSpan) : -1;
       const txsCount = await FlagValueRepository.$queryTxCountBasedOnMask(mask, bucketSize, op, startHeight);
-      res.header('X-total-count', tip.toString());
+      res.header('X-total-count', totalCount.toString());
       res.header('Expires', new Date(Date.now() + 1000 * 3600 * 24 * (presets[interval].bucketSizes[0] / 144)).toUTCString());
       res.send(txsCount);
     } catch (e: any) {
--- a/backend/src/api/bitcoin/bitcoin.routes.ts
+++ b/backend/src/api/bitcoin/bitcoin.routes.ts
@@ -23,12 +23,14 @@ import { calculateMempoolTxCpfp } from '../cpfp';
 import { handleError } from '../../utils/api';
 import poolsUpdater from '../../tasks/pools-updater';
 import chainTips from '../chain-tips';
+import FlagValueRepository, { INTERVAL_PRESETS } from '../../repositories/FlagValueRepository';
 
 const TXID_REGEX = /^[a-f0-9]{64}$/i;
 const BLOCK_HASH_REGEX = /^[a-f0-9]{64}$/i;
 const ADDRESS_REGEX = /^[a-z0-9]{2,120}$/i;
 const SCRIPT_HASH_REGEX = /^([a-f0-9]{2})+$/i;
 const MAX_TRANSACTION_TIMES = 100;
+const JUST_NUMBERS_REGEX = /^[1-9]\d*$/;
 
 class BitcoinRoutes {
   public initRoutes(app: Application) {
@@ -70,6 +72,10 @@ class BitcoinRoutes {
       .get(config.MEMPOOL.API_URL_PREFIX + 'internal/blocks/definition/list', this.getBlockDefinitionHashes)
       .get(config.MEMPOOL.API_URL_PREFIX + 'internal/blocks/definition/current', this.getCurrentBlockDefinitionHash)
       .get(config.MEMPOOL.API_URL_PREFIX + 'internal/blocks/:definitionHash', this.getBlocksByDefinitionHash)
+
+      .get(config.MEMPOOL.API_URL_PREFIX + 'goggles/:interval/', this.getTxCountPerFlagValue)
+      .get(config.MEMPOOL.API_URL_PREFIX + 'goggles/:interval/:bucketSize', this.getTxCountPerFlagValue)
+      .get(config.MEMPOOL.API_URL_PREFIX + 'goggles/:interval/:bucketSize/:op/:mask', this.getTxCountPerFlagValue)
       ;
 
       if (config.MEMPOOL.BACKEND !== 'esplora') {
@@ -1129,6 +1135,61 @@ class BitcoinRoutes {
       }
     } catch (e) {
       handleError(req, res, 500, 'Failed to get difficulty change');
+    }
+  }
+
+  private async getTxCountPerFlagValue(req: Request, res: Response) {
+    try {
+      if (!Common.blocksSummariesIndexingEnabled()) {
+        handleError(req, res, 404, `Block summaries indexing is required for this API`);
+        return;
+      }
+
+      const presets = INTERVAL_PRESETS;
+      const operations = ['and', 'or', 'nor', undefined];
+      const intervals = Object.keys(presets);
+      const interval = req.params.interval;
+
+      if (!intervals.includes(interval)) {
+        handleError(req, res, 400, `Invalid interval, must be one of ${intervals.toString()}`);
+        return;
+      }
+
+      const validBucketSizes = presets[interval].bucketSizes;
+      const rawBucketSize = req.params.bucketSize;
+      const bucketSize: number = rawBucketSize === undefined ? validBucketSizes[0] : Number(rawBucketSize);
+      if (!Number.isInteger(bucketSize) || !validBucketSizes.includes(bucketSize)) {
+        handleError(req, res, 400, `Invalid bucket size, must be ${validBucketSizes.toString()}`);
+        return;
+      }
+
+      if (!operations.includes(req.params.op)) {
+        handleError(req, res, 400, `Invalid operation, must be 'and', 'or', 'nor' or undefined.`);
+        return;
+      }
+
+      if (req.params.mask && !JUST_NUMBERS_REGEX.test(req.params.mask)) {
+        handleError(req, res, 400, `Invalid mask value, must be a positive integer`);
+        return;
+      }
+
+      const op = (req.params.op) as 'and' | 'or' | 'nor' | undefined;
+      const mask = BigInt(req.params.mask ?? 0n);
+
+      const { tip }  = await FlagValueRepository.$getTipAndTailIndexedByBucketSize(bucketSize) || { tip: undefined };
+
+      if (!tip) {
+        handleError(req, res, 400, `Failed to get latest indexed flag values for ${interval}`);
+        return;
+      }
+
+      const startHeight = presets[interval].retentionSpan !== -1 ? (tip - presets[interval].retentionSpan) : -1;
+      const txsCount = await FlagValueRepository.$queryTxCountBasedOnMask(mask, bucketSize, op, startHeight);
+      res.header('X-total-count', tip.toString());
+      res.header('Expires', new Date(Date.now() + 1000 * 3600 * 24 * (presets[interval].bucketSizes[0] / 144)).toUTCString());
+      res.send(txsCount);
+    } catch (e: any) {
+      handleError(req, res, 400, e instanceof Error ? e.message : 'Failed to get flag values');
     }
   }
 
--- a/backend/src/api/blocks.ts
+++ b/backend/src/api/blocks.ts
@@ -40,6 +40,7 @@ import CpfpRepository from '../repositories/CpfpRepository';
 import { parseDATUMTemplateCreator, parseDMNDTemplateCreator } from '../utils/bitcoin-script';
 import database from '../database';
 import { getBlockFirstSeenFromLogs, getOldestLogTimestampFromLogs, scanLogsForBlocksFirstSeen } from '../utils/file-read';
+import FlagValueRepository, { INDEXING_PRESETS } from '../repositories/FlagValueRepository';
 
 class Blocks {
   private blocks: BlockExtended[] = [];
@@ -54,6 +55,8 @@ class Blocks {
   private oldestCoreLogTimestamp: number | undefined | null = undefined;
 
   private mainLoopTimeout: number = 120000;
+  private indexingFlagValues: boolean = false;
+  private flagValuesDeleteQueue: number[]= [];
 
   constructor() { }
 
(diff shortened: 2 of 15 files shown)


rodribp:
Hello, here's a quick video of how's the feature going so far:

https://github.com/user-attachments/assets/bed3f3c7-5927-40be-a608-5dfcc12bdd34

Overall it's going well, I managed to resolve a heap out of memory issue due to the first approached I made for the on-startup indexer of flag values.

Next steps are:
1. Validation of params in the backend.
2. Normalization of names (The names of the variables are quite a mess right now).
3. On new block indexing. (Currently I've only done on startup).
4. More improvements and QA.
5. Moving the logic to a dedicated class to have a single source of truth (also to have a single variable of rollup and interval presets).
6. URL parameters on the frontend for boolean and flags based for URL sharing.

For now the on startup indexer indexes by polling from blocks_summaries per rolled blocks amount, depending on the interval sent by the frontend, if no flags are sent it just returns the txCount on the interval selected. If there are flags gotten from the frontend tx count based on those filters. After polling blocks_summaries, it processes and makes a Object of type { flag_value: string, txCount: number} to process per flag bit array value txCounts on the blocks gotten from blocks_summaries, and then indexes per block start_height, blocks_count (how many blocks are rolled up in this row) and unique bit array flag value.

Feedback is highly appreciated! Thank you

mononaut:
great progress!

re. indexing, wherever possible I think we want to index these flag values at the *same time* as we index the block summaries, to avoid the roundtrip of loading the summary back out of the DB.

I think we used a similar pattern when added CPFP indexing and the original Mempool Goggles indexing (`$classifyBlocks`):

https://github.com/mempool/mempool/blob/419fe73edf2b7cbf38353f2e5bb5a9e660b4b468/backend/src/indexer.ts#L215-L230

so the three points at which we could index this data for a block (or range of blocks) are:
1) when a new block arrives, immediately after indexing the block summary, while those classified transactions are still in memory.
2) while indexing historic block summaries.
3) a separate, slow, "catching up" process for instances that just upgraded to this PR, and have *already* indexed the historic block summaries and per-transaction goggles flags.
    - fetch the list of block heights for which we *have* indexed block summaries and goggles flags, but *have not* indexed this new data
    - index just those heights

rodribp:
Hi @mononaut I agree on indexing at the same time blocks_summaries is being indexed, so I'd refactor my function (or create another one), just so is called alongside:

https://github.com/mempool/mempool/blob/bd2b0c6645f251dca120219446b09f250bc83a26/backend/src/api/blocks.ts#L1545-L1547

in getStrippedTransaction, and the function would:

1. Instantly compute and index the flag_values if the height is above (current block height from core) - (blockspan of 1 week).
2. Compute, and append to a cached property the flag_values if height is between the block span of the aggregated intervals, if the height is above (current block height from core) - (N block span of the interval). And index to the database, once the amount of blocks appended to the property has reached the group of blocks size.
3. One cached property per group of blocks (except per block) (property names: flagValuesPer36, flagValuesPer144, ...) following structure: { startHeight, lastBlockComputed, flag_values: {'flag': txCount}}, so once lastBlockComputed reaches the size of the group of blocks, it indexes and resets.

Does this approach sound good for you? Or do you have a different one in mind? I'd be concerned by the size of the properties. Also I think I can refactor the DB query to batch insert the flag_values instead of iterating over flag_values and inserting per row.

For new instances that upgraded to this PR I think the current function solves it but need to double check.

Secondly: I noticed you do Common.sleep() on classify transactions, would you recommend me to do it as well? I'm not sure how much is considered indexing too fast, maybe not for the on new block indexing, but on the upgrade to this PR.

mononaut:
hmm, `getStrippedBlockTransactions` is called in some other contexts where it wouldn't be appropriate to try to index this aggregated flag data.

I'd suggest doing it in `blocks.$generateBlocksSummariesDatabase` for the "catch up" indexing (might have to alter `blocks.$indexBlockSummary` to return the data you need), and at the end of `blocks.$saveBlockData` for new blocks (data is already available in `processingResult.blockSummary`).

otherwise your approach sounds good, provided we mostly try to make progress in order of blockheight (so the cache doesn't fill with too many partially completed block spans) and we respect `config.MEMPOOL.INDEXING_BLOCKS_AMOUNT`.

---

let's leave the sleeps out for now, we can always add some later if the new indexing affects server performance.

mononaut (changes requested):
this is awesome, great work!

just a few thoughts:

1) we should make sure this supports reorgs.

we explicitly delete other stale indexed block data when a reorg is detected, and then let the regular periodic indexing tasks re-index the new blocks automatically.

https://github.com/mempool/mempool/blob/8429948d2c90b2a2759ccc3bbbf9a283af5b6911/backend/src/api/blocks.ts#L1407-L1413

this is probably a good place to do the same for the new flags index.

2) you need to be careful to filter out stale blocks when you fetch lists of blocks to index. I think at the moment this just indexes all block summaries in the database, including stale blocks? I suspect that might throw off your bucket indices too.

3) better to create the db table for all networks, and then just exempt liquid from the indexing itself (if we do want to exclude it).

4) the graph should always show the full selected range, rather than truncating to just the non-zero values. otherwise it's difficult to compare different combinations of flags.

5) the partially-completed flag buckets and related methods probably belong in the `FlagValuesRepository` class rather than `blocks.ts`.

although, since indexing is a lot quicker than I expected, it might actually be feasible to skip that complicated real-time indexing and just rely on `$generateFlagValuesDatabase` to keep the index up to date.

what do you think?

6) there seems to be a lot of parsing `blocksCount` back and forth between integer and string. I'd suggest converting once as close to the database as possible, and keeping it a consistent integer type everywhere else.

I also found `blocksCount` vs `blockSpan` vs the various blocks sums and totals and lengths etc quite hard to follow. perhaps these could be renamed to something more intuitive, like `bucketSize` and `retentionSpan`?

7) how difficult would it be to index vsize/weight as well as the transaction counts? it would also be cool to have an option see the charts in terms of the percentages of all transactions - and it looks like you have the data for this already, since it's shown in the tooltip?

rodribp:
Hi @mononaut , thanks for the feedback.

1. Ok
2. Agree, no I didn't filter out stale blocks I forgot, good catch.
3. Ok
4. Agree. I think the easiest is to modify the DB query to return 0 in the bucket rows where the mask does not meet the bitwise operation, how about you?
5. So, I think I didn't understand well, do you mean not to check if a flagValues obj has been completed inside blocks.ts, but in the $saveBatchFlagValues? Or do you mean to index either way, and use UPDATE if the bucket has already been indexed?

About if using just `$generateFlagValuesDatabase()`: I think is easier and simpler to just rely on `$generateFlagValuesDatabase()`, and it's true is impressive how fast it is. As I can't run mainnet on my end yet, I could just try it on signet, and it lasts about 2 mins to index the 4 configs (bucket sizes), and tbh I haven't measured how faster it is the diff between `blocks.$generateBlocksSummariesDatabase()` in master vs in this branch (with the extra step) but if we prefer indexing time, using the three paths I think is ok, but if we are ok to sacrifice more initial indexing time for an improved dev ux in the future (if we ever need to change something in the workflow or add more columns to the table) using a single function for the indexing of these flag values seems to be the right scope to me. I would keep it simple, as the time is reasonable and if we don't want to slow down the indexers right after `blocks.$generateFlagValuesDatabase()` we can just move it to be the last indexer inside `indexer.$run()`

6. Agree.
7. I thought it was not possible in the beginning but It's easy actually, it's just summing up the vsize per block/bucket, index it, query to it, and get its % = vsize_total(DB) / (1M*bucketSize)

mononaut:
(4) yeah that works. or you could append/prepend dummy values to the data in frontend chart code.

(5) I meant some of the new code and class members in `blocks.ts` feel like they don't belong in that file, because they're more closely related to the flags repository than the main `blocks.ts` state & functionality.

e.g. the flag bucket interface:

https://github.com/mempool/mempool/blob/5cfa660cd985e86fce79e8b593af4e7b9c211fe7/backend/src/api/blocks.ts#L45-L51

and flag cache:

https://github.com/mempool/mempool/blob/5cfa660cd985e86fce79e8b593af4e7b9c211fe7/backend/src/api/blocks.ts#L56
https://github.com/mempool/mempool/blob/5cfa660cd985e86fce79e8b593af4e7b9c211fe7/backend/src/api/blocks.ts#L932-L941

and related methods

https://github.com/mempool/mempool/blob/5cfa660cd985e86fce79e8b593af4e7b9c211fe7/backend/src/api/blocks.ts#L943-L956

for example, that last method makes lots of references to members of `FlagValueRepository`, but never accesses any of the `blocks.ts` context via `this` or `Blocks`.

---

yeah, the more I think about it the better it sounds to simplify the indexing code and just rely on `$generateFlagValuesDatabase()`.

this data is just for the charts, so unlike the blocks summaries or CPFP indexes it's not super important to index new blocks immediately.

frontend/src/app/components/block-goggles-graph/block-goggles-graph.component.ts
```diff
@@ -0,0 +1,443 @@
...
+              tooltip += `${matchedMarker} ` + $localize`Matched transactions` + `: ${formatNumber(matchedCount, this.locale, '1.0-0')}<br>`;
+            }
+            if (totalCount > 0) {
+              tooltip += `${matchedMarker} ` + $localize`Share of all txs` + `: ${formatNumber(matchedCount / totalCount * 100, this.locale, '1.0-2')}%<br>`;
+            }
+            const matchedVSize = matched ? matched.data[3] : 0;
+            if (matchedVSize > 0) {
+              const weightShare = matchedVSize / (bucketSize * MAX_BLOCK_VSIZE) * 100;
+              tooltip += `${matchedMarker} ` + $localize`Share of block weight` + `: ${formatNumber(weightShare, this.locale, '1.0-2')}%<br>`;
+            }
```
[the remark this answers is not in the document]
rodribp: Expected behaviour

rodribp:
Just did the following changes:

1. Removed all indexing logic of 2 paths: a) Inside `blocks.$generateBlocksSummariesDatabase()`. b) Inside `blocks.$saveBlockData()`. Just leaving `blocks.$generateFlagValuesDatabase()`.
2. Changed the names of the rows and variables to `bucketSize` and `retentionSpan`.
3. Added support for reorgs, deleting flagValues above startHeight whose tailFork.height fell within existing indexed buckets.
4. Moved parsing of bucketSize into `FlagValuesRepository`.
5. Added support for % of block weight.
6. Visual enhancements for a greater visual comparison of matched tx against total txs per block in general.


https://github.com/user-attachments/assets/bc7df6aa-e2e6-474c-8d43-5987bd769f95

It is possible to show instead of % of block weight against absolute block capacity, against actual total block weight. and also it is possible to switch the view from abs txCount to % of block weight with a select just like intervals. Maybe @orangesurf could give us feedback on this?

orangesurf:
Great, I think it should have selection for selecting the measure, bucket and scale (ABS/rel)

Here is a screenshot from something I'm working on which uses the same logic (ignore range selection)

image

rodribp:
Just added this selector of unit:

Screenshot 2026-07-06 at 10 51 43 PM

Mobile:
Screenshot 2026-07-06 at 10 52 53 PM

So the tooltip shows depending on the unit:

vBytes:
- vsize ||  totalVsize if bucketed data
- if bucketed data: avg vsize

if goggles used:
- matched vsize || avg matched vsize if bucketed.
- share of vsize (%) (this time in relation to the actual vsize of the block, not block capacity)

And the same for tx count.

| bytes | tx count |
| - | - |
| Screenshot 2026-07-06 at 10 58 30 PM | Screenshot 2026-07-06 at 10 59 17 PM |

backend/src/repositories/FlagValueRepository.ts
```diff
@@ -0,0 +1,123 @@
...
+      default: throw new Error(`Invalid op '${op}', expected 'and' | 'or' | 'nor' | undefined`);
+    }
+    params.push(bucketSize.toString());
+    params.push(startHeight);
+    try {
+      const [rows]: any[] = await DB.query(`
+        SELECT bucket_size as bucketSize, start_height as startHeight, ${statFields} FROM flag_values
+        WHERE bucket_size = ? AND start_height >= ?
+        GROUP BY start_height ORDER BY start_height DESC
+        `, params);
```
mononaut: I think moving the bitwise predicates to the `WHERE` clause should be more efficient, e.g:

```suggestion
    let flagPredicate = '';
    let params: any[]= [];
    switch (op) {
      case 'and': {
        flagPredicate = 'AND (flag_value & ?) = ?';
        params = [bucketSize.toString(), startHeight, mask, mask];
      } break;
      case 'or': {
        flagPredicate = 'AND (flag_value & ?) > 0';
        params = [bucketSize.toString(), startHeight, mask];
      } break;
      case 'nor': {
        flagPredicate = 'AND (flag_value & ?) = 0';
        params = [bucketSize.toString(), startHeight, mask];
      } break;
      case undefined: { // op not passed, no boolean operations
        params = [bucketSize.toString(), startHeight];
        break;
      }
      default: throw new Error(`Invalid op '${op}', expected 'and' | 'or' | 'nor' | undefined`);
    }
    try {
      const [rows]: any[] = await DB.query(`
        SELECT bucket_size as bucketSize, start_height as startHeight,
          SUM(tx_count) as txCount, SUM(vsize_total) as vSizeTotal
        FROM flag_values
        WHERE bucket_size = ? AND start_height >= ? ${flagPredicate}
        GROUP BY start_height ORDER BY start_height DESC
        `, params);
```

since then you're only evaluating the bitwise mask once per row in the bucket size & height range instead of twice (separately in each `CASE WHEN ...`), and you only need to aggregate the rows with matching flags instead of every row in the range.

frontend/src/app/components/block-goggles-graph/block-goggles-graph.component.ts
```diff
@@ -0,0 +1,466 @@
...
+                : `${m} ` + $localize`Matched transactions` + `: ${fmtCount(matchedTxCount)}<br>`;
+              if (matchedTxCount > 0) {
+                tooltip += `${m} ` + $localize`Share of all txs` + `: ${fmtPct(matchedTxCount / baseTxCount * 100)}<br>`;
+              }
+            } else {
+              tooltip += rolledUp
+                ? `${m} ` + $localize`Avg matched vSize per block` + `: ${fmtVSize(matchedVSize / bucketSize)}<br>`
+                : `${m} ` + $localize`Matched vSize` + `: ${fmtVSize(matchedVSize)}<br>`;
+              if (matchedVSize > 0) {
+                tooltip += `${m} ` + $localize`Share of block vSize` + `: ${fmtPct(matchedVSize / (bucketSize * baseVSize) * 100)}<br>`;
```
mononaut: this calculation is wrong - I think `baseVSize` is already summed over the width of the bucket?

Image

backend/src/api/blocks.ts
```diff
@@ -689,16 +690,170 @@ class Blocks {
...
+      throw(e);
+    }
+  }
+
+  /**
+   * Indexes a block summary and returns its classified transactions so the caller can roll up flag values.
+   *
+   * @asyncUnsafe
+   */
+  public async $indexBlockSummary(hash: string, height: number, stale?: boolean, blockSummary?: BlockSummary): Promise<TransactionClassified[]> {
```
mononaut: do we still need this refactored version of `$indexBlockSummary`, now that we're only indexing flag values in `blocks.$generateFlagValuesDatabase()` ?

mononaut (changes requested):
almost there!

I've left a few more inline comments.

---

I did a bit of profiling, and although the indexing does complete quickly, it's very CPU-intensive, which could cause issues in production.

we could add some artificial throttling and run it as a background task, like we do in `$classifyBlocks`, but this is more of an ops decision that I need to think about some more.

---

also, it's not very clear in the codebase, but `blocks_summaries` is versioned, and version 0 summaries aren't guaranteed to have flag data (the `flag` field could either be defaulted to 0, or missing).

so we need to handle the edge case where an instance with block summaries already indexed at version 0 (i.e. without flags) upgrades to a version after this PR.

in that case, I think `$generateFlagValuesDatabase()` will run *before* `$classifyBlocks` adds the transaction flag data to all of the legacy block summaries, and either throw errors due to missing flag fields, or produce results from incorrectly zeroed flag values.

the two approaches I can think of to fix this are:

A) defer `$generateFlagValuesDatabase()` until *after* `$classifyBlocks` finishes.

that's slightly awkward because then it has to run as a background task instead of being `await`ed in the main indexer loop.

B) modify `$generateFlagValuesDatabase()` to only fetch block summaries with version >= 1 (which should all include goggles flag data). buckets containing version 0 summaries would then be skipped as incomplete and picked up on a later run once they've been properly classified.

backend/src/repositories/FlagValueRepository.ts
rodribp: Agree, since the graph always show the total of txs the chart no longer truncates 0 values.

backend/src/api/blocks.ts
rodribp: Not really, I would only keep this quick fix in `$indexCPFP()`:

https://github.com/mempool/mempool/blob/6c27fc91ec0a2792f74f1ec6ec7abff41950f204/backend/src/api/blocks.ts#L1928-L1930

rodribp:
> A) defer $generateFlagValuesDatabase() until after $classifyBlocks finishes.
> 
> that's slightly awkward because then it has to run as a background task instead of being awaited in the main indexer loop.
> 
> B) modify $generateFlagValuesDatabase() to only fetch block summaries with version >= 1 (which should all include goggles flag data). buckets containing version 0 summaries would then be skipped as incomplete and picked up on a later run once they've been properly classified.

What I'd be concerned of is average instances, since I doubt self hosted instances run esplora, that would make instances with blocks_summaries versioned to 0 have the `all` interval truncated forever.

mononaut:
that's fine. version 0 block summaries don't have flag data anyway, and if the relevant indexing is enabled they'll get automatically migrated to version 1 over time by the background `$classifyBlocks` task.

rodribp:
> that's fine. version 0 block summaries don't have flag data anyway, and if the relevant indexing is enabled they'll get automatically migrated to version 1 over time by the background `$classifyBlocks` task.

But esplora is required to classify
https://github.com/mempool/mempool/blob/5b633d8f4d08c06a5845393725756ff747d51f07/backend/src/api/blocks.ts#L807-L810

orangesurf:
I think the following would be useful additional selection options
- Selection for Day/week/month/year buckets (in addition to the existing range selection).
- Selection for scale (Absolute / %) instead of this just being in the on-hover details

Additionally
- If goggles filters are enabled, allow hiding of the total data (blue bars) 
- Name, could be "Mempool Goggles" instead of Transactions per Block

rodribp:
Updates:

- Added of abs / rel selector when goggles activated.
- Added week / month bucket size selector.
- Narrowed intervals to just `24h`, `6m`, `1y`, `2y`, `3y` and `all`.
- Added legend selector on top following the design of other graphs.
- Standardized tooltip with the rest of the graphs.

Relative - vBytes - ANY filter - both series
Screenshot 2026-08-05 at 1 58 10 AM

Absolute - count - AND filter - just matched serie
image

24h (per block) just matched selected
Screenshot 2026-08-05 at 2 04 12 AM

Tooltip (all transactions is not the sum of txs in the bucket but the avg, and below specified *around block):
image
The tooltip depends on the series showed / hidden
Screenshot 2026-08-05 at 2 07 44 AM

No filters at all hide legends and scale selector:
image


SVG export:
image

rodribp:
Pending: Incrementalize the computation of heavy buckets to avoid heap out of memory when querying block summaries for long bucket sizes on mainnet such as months and weeks.

orangesurf:
Looks great! Please can you add Y axis titles.

rodribp:
>Ranged buckets are identified only by their midpoint, but the PR explicitly requires the tooltip to show startBlock - endBlock. A 4,032-block bucket cannot be identified precisely from “Around block”, so display its actual inclusive bounds.

Agree just with this suppressed comment

rodribp:
Hey @mononaut I just updated this PR with the following changes from your last review:

### Frontend:

- Suggestions from OS:
  - Added abs / rel selector when goggles activated.
  - Added week / month bucket size selector.
  - Narrowed intervals to just 24h, 6m, 1y, 2y, 3y and all.
  - Added legend selector on top following the design of other graphs (To hide the totals).
  - Added yAxis title to show the unit and mode (abs/rel).

- Standardized tooltip with the rest of the graphs.
- Improved the .svg export with title on x and y axis.

### Backend:

- Narrowed the indexing buckets sizes to just per block, 1008 (1w) and 4032 (1m).
- 1w and 1m bucket sizes store all the blockchain and per block stores the latest 144 blocks.
- Refactored the backend to be ran by `void`.
- Refactored the function to index tip-down to fill intervals from shortest to longest.
- Made the function query `blocks_summaries` at a maximum of 144 blocks to avoid heap OOM that could be caused if we query 1008 or 4032 blocks (previous per bucket querying implementation), and incrementalized `dataPerFlag`.
- Defined `Expires` header to the interval equivalent minimum bucketSize.
- Throttled `$generateFlagValuesDatabase()` with `Common.sleep$(250)` (matching `$classifyBlocks()`)

Please let me know if there are more changes to make that you could spot, I'm happy to polish!

Merging this PR will cause conflicts to #6648 due to `database-migration.ts` and these changes have been tested on Signet only.

Please find in this video the latest demo
https://drive.google.com/file/d/1Z-A-X-Mhn7fUYiTIxeA7MIamBsxCzSbx/view?usp=sharing

backend/src/api/bitcoin/bitcoin.routes.ts
```diff
@@ -1132,6 +1138,61 @@ class BitcoinRoutes {
...
+      const { tip }  = await FlagValueRepository.$getTipAndTailIndexedByBucketSize(bucketSize) || { tip: undefined };
+
+      if (!tip) {
+        handleError(req, res, 400, `Failed to get latest indexed flag values for ${interval}`);
+        return;
+      }
+
+      const startHeight = presets[interval].retentionSpan !== -1 ? (tip - presets[interval].retentionSpan) : -1;
+      const txsCount = await FlagValueRepository.$queryTxCountBasedOnMask(mask, bucketSize, op, startHeight);
+      res.header('X-total-count', tip.toString());
```
mononaut: the frontend relies on this header to be an accurate count of the number of blocks worth of data actually available from this endpoint, in order to set the visibility of the dateSpan selectors correctly.

setting it to `tip` is correct if we've definitely indexed all the way to height 0, but not if we're running with a limited `config.MEMPOOL.INDEXING_BLOCKS_AMOUNT` or still working through the initial indexing.

`tip - tail` would be more accurate. (although we'd want the actual indexed block heights, rather than simply the bucket boundaries, to be totally accurate.

mononaut (changes requested):
Looks and works great, really nice job!

I'm waiting for a full index to finish running before I give this a final ACK, but in the meantime, there's one small issue I noticed left to fix

mononaut (approved):
Tested ACK @ [a0e74fcf]

Really nice work on this, it's a great addition to the mempool charts, and I'm very pleased how it turned out!

I tested a full mainnet index, which is pretty fast on good hardware (less than an hour iirc).

The final `flag_values` table only occupies 69MB, and typical queries range from 5-40ms, worst case around 100ms on my machine, and all very cacheable.

PR #6587, opened 2026-06-23
FEAT: Cost to spend utxos in address page

### LINKED ISSUE #6557: Address page cost to spend UTXOs
It would be useful and informative to display how much it would cost to spend an address's UTXOs on the address page, e.g:
- vsize of an input from this address
- minimum cost to spend (number of utxos × vsize per input × prevailing feerate)
- maximum cost to spend ((vsize of 1 input + typical transaction overhead) × prevailing feerate)
- minimum/maximum "effective balance" of the address (balance - cost to spend)

vsize per input would need to be estimated for some addresses (e.g. p2sh, p2wsh, p2tr), or extracted from previous spends.

jramos0:
Fix #6557

Adds a "Cost to Spend" section to the address page showing what it would cost in fees
to spend an address's UTXOs, plus the resulting effective balance. 

### Cost model

- **Spend all together (min):** all UTXOs consolidated into a single transaction, so
  the tx overhead is paid once. Theoretical lower bound (a real tx of this many inputs
  would exceed the standard weight limit).
- **Spend separately (max):** each UTXO spent in its own transaction (input + overhead),
  rounding each spend up to whole sats before summing.
- **Min / Max effective balance:** balance minus the max / min cost respectively.

### Open questions 

- **What should I do about the Dust :** the UTXO count is currently raw (not filtered for
  outputs that cost more to spend than they're worth). Wanted your call on whether
  "effective balance" should exclude these. @mononaut
- p2tr script-path spends without a prior spend fall back to the key-path estimate
  (underestimate); resolved automatically once the address has a measurable spend.
- The display labels/copy are still rough and could be improved.
- The vsize functions are written as pure functions so they're straightforward to cover once
  one exists. Should I create unit test for this?

https://github.com/user-attachments/assets/1dab771a-fe67-4c12-ae50-c66e76bef768

https://github.com/user-attachments/assets/84b104c8-7678-49ec-aa91-38c957bfa2a5

### DIFF
--- a/frontend/src/app/components/address/address.component.html
+++ b/frontend/src/app/components/address/address.component.html
@@ -135,15 +135,17 @@
       </div>
     </ng-container>
 
-    <ng-container *ngIf="network !== 'liquid' && network !== 'liquidtestnet' && !address.electrum && addressTypeInfo && (chainStats.utxos + mempoolStats.utxos) > 0">
+    <ng-container *ngIf="network !== 'liquid' && network !== 'liquidtestnet' && !address.electrum && addressTypeInfo && spendableUtxoCount > 0">
       <br>
       <div class="title-tx">
-        <h2 class="text-start cost-to-spend-toggle" (click)="showCostToSpend = !showCostToSpend">
-          <span i18n="address.cost-to-spend">Cost to Spend</span>
-          <fa-icon [icon]="['fas', showCostToSpend ? 'angle-up' : 'angle-down']" [fixedWidth]="true"></fa-icon>
-        </h2>
+        <h2 *ngIf="showCostToSpend" class="text-start float-start" i18n="address.cost-to-spend">Cost to Spend</h2>
+        <button type="button" class="btn btn-outline-info btn-sm float-end" (click)="showCostToSpend = !showCostToSpend">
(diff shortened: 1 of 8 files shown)


jramos0:
https://github.com/user-attachments/assets/f3c58893-02f4-4e44-b82c-b0e37a142c98

I made some updates to visual part so the display could be easier to read and understand

orangesurf:
Very nice! Would be nice if the user can adjust the feerate, perhaps explore a few options including a slider (like sparrow), low/medium/high feerate buttons, input box

jramos0:
Hey @orangesurf I've been improving the feature and I took some of your ideas  today and anothers from Mononaut so I'd like you to check this version of the feature when the UTXO's in the address are less than 500 , is taking every utxo to convert it into a pie chart showing the amount of the fee can eat for any utxo according its size

https://github.com/user-attachments/assets/73dcac21-9605-4a91-95eb-ececfe58b308

orangesurf:
Very cool! It would be interesting to see a different color than red, perhaps grey or background blue. 

Additionally it would be interesting to see the % spent on fees for both the individual and aggregate spend options

PR #6589, opened 2026-06-24
[accelerator] update checkout component and documentation

nymkappa:
This is the companion PR to https://github.com/mempool/mempool.space/pull/2462. Both PRs needs to be tested and reviewed together.

This PR updates the accelerator checkout flow to offer accelerator credit payment over lightning when applicable.

It also updates the documentation to reflect that all account are now able to use accelerator credits to accelerate.

### DIFF
--- a/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.html
+++ b/frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.html
@@ -32,7 +32,6 @@
         <app-accelerate-fee-graph
           [tx]="tx"
           [estimate]="estimate"
-          [showEstimate]="hasAccessToBalanceMode"
           [maxRateOptions]="maxRateOptions"
           [maxRateIndex]="selectFeeRateIndex"
           (setUserBid)="setUserBid($event)"
@@ -112,7 +111,7 @@
                 <tbody>
                   <!-- ESTIMATED FEE -->
                   <ng-container *ngIf="showDetails">
-                    @if (hasAccessToBalanceMode) {
+                    @if (estimate.isProUser) {
                       <tr class="group-first">
                         <td class="item" i18n="accelerator.next-block-rate">Next block market rate</td>
                         <td class="amt" style="font-size: 16px">
@@ -187,7 +186,7 @@
                   </ng-container>
 
                   <!-- NEXT BLOCK ESTIMATE -->
-                  <ng-container *ngIf="hasAccessToBalanceMode">
+                  <ng-container *ngIf="estimate.isProUser">
                     <tr class="group-first">
                       <td class="item">
                         <b style="background-color: #5E35B1" class="p-1" i18n="accelerator.estimated-cost">Estimated acceleration cost</b> ~{{ estimate.targetFeeRate | number : '1.0-0' }} sat/vB
@@ -208,7 +207,7 @@
                   <ng-container>
                     <tr class="group-first group-last">
                       <td class="item">
-                        @if (hasAccessToBalanceMode) {
+                        @if (estimate.isProUser) {
                           <b style="background-color: var(--primary);" class="p-1" i18n="accelerator.maximum-cost">Maximum acceleration cost</b>
                         } @else {
                           <b style="background-color: var(--primary);" class="p-1" i18n="accelerator.cost">Acceleration cost</b>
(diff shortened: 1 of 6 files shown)


frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.html
```diff
@@ -371,63 +371,62 @@ <h1 style="font-size: larger;"><ng-content select="[slot='cta-title']"></ng-cont
...
+            <div class="col-sm text-center flex-grow-0 d-flex flex-column justify-content-center align-items-center">
+              <span class="text-nowrap mb-3 mt-3" [class.mt-md-0]="!forceMobile">&mdash;&mdash;<span i18n="or"> OR </span>&mdash;&mdash;</span>
+            </div>
+          }
+          @if (canPayWithBalance || canPayWithBitcoin) {
+            <div class="col-sm text-center d-flex flex-column justify-content-center align-items-center">
+              @if (canPayWithBalance) {
+                <p><ng-container i18n="accelerator.your-account-will-be-debited">Use accelerator credits for </ng-container>&nbsp;<small style="font-family: monospace;">{{ cost | number }}</small>&nbsp;<span class="symbol" i18n="shared.sats">sats</span></p>
+                <div class="d-flex justify-content-center" [class.disabled]="quoteError || accelerateError || showSuccess">
+                  <div class="paymentMethod mx-2 d-flex justify-content-center align-items-center" style="width: 200px; height: 55px; font-size: 30px" (click)="accelerateWithMempoolAccount()">
```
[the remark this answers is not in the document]
nymkappa: This is not a possible scenario, `canPay` cannot be `false` while `canPayWithBalance` is `true`.

frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.ts
```diff
@@ -1095,19 +1094,19 @@ export class AccelerateCheckout implements OnInit, OnDestroy {
...
   get canPay(): boolean {
-    return this.canPayWithBalance || this.canPayWithBitcoin || this.canPayWithCashapp || this.canPayWithApplePay || this.canPayWithGooglePay;
+    return this.canPayWithBalance || this.canPayWithBitcoin || this.canPayWithCashapp || this.canPayWithApplePay || this.canPayWithGooglePay || this.canPayWithCardOnFile;
   }
 
-  get hasAccessToBalanceMode(): boolean {
-    return this.isLoggedIn() && this.estimate?.hasAccess;
+  get canUseBalance(): boolean {
+    return this.isLoggedIn() && this.estimate?.userBalance > 0;
   }
```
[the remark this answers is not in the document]
nymkappa: That's fine, in case the authenticated user does not have enough balance (and is non pro) then we fallback to unauthenticated payment methods.

We don't want to add friction at checkout.

frontend/src/app/components/accelerate-checkout/accelerate-checkout.component.html
```diff
@@ -371,63 +370,69 @@ <h1 style="font-size: larger;"><ng-content select="[slot='cta-title']"></ng-cont
...
-            }
-            @if ((canPayWithBitcoin && (canPayWithCashapp || canPayWithApplePay || canPayWithGooglePay || canPayWithCardOnFile))) {
-              <div class="col-sm text-center flex-grow-0 d-flex flex-column justify-content-center align-items-center">
-                <span class="text-nowrap mb-3 mt-3" [class.mt-md-0]="!forceMobile">&mdash;&mdash;<span i18n="or"> OR </span>&mdash;&mdash;</span>
-              </div>
-            }
-            @if (canPayWithBitcoin) {
-              <div class="col-sm text-center d-flex flex-column justify-content-center align-items-center">
+                </div>
+              } @else {
```
mononaut, line 427: this makes the "pay with balance" option mutually exclusive with the "pay with bitcoin" option, which means logged in users with non-zero balance can no longer accelerate using lightning.

is that intended?

nymkappa: The balance payment option takes precedence over the lightning option only if the user balance is enough to cover the cost of acceleration, at least that's the idea. I will double check the actual behavior again.

nymkappa: Confirmed it works as expected. It's fair to assume that if a user bought accelerator credits he wants to use them.

Plus some less technical user might even get confused by the bitcoin button thinking it would use their balance but then see a QR code.

This can be updated easily in the future regardless.

mononaut (approved):
Tested ACK @ [c4debdd1]

Issue #6600, opened 2026-06-29
limit websocket init data to default initial subscriptions


mononaut:
the backend currently includes some unnecessary data in the initial `serializedInitData` sent when a client first initializes a websocket connection.

e.g. `stratumJobs` is included even if the client has not subscribed to stratum data.

instead, `serializedInitData` should only hold the essential data that is *always* required when the frontend first loads.

mononaut:
resolved by #6603

PR #6603, opened 2026-06-29
Send stratum jobs only when requested on stratum page

natsoni:
`StratumJobs` is currently added to every init response, this PR fixes that to only send these on `track-stratum` subscriptions.


mononaut:
resolves #6600 ?

mononaut (approved):
Tested ACK @ [dc42a46b]

PR #6611, opened 2026-07-02
FEAT: Focused address view in transaction page

### LINKED ISSUE #6162: [Feature] Focused view for transaction payment to a specified address
When a bitcoin transaction recipient provides a URL to the sender to track payments, the destination address is known - the intention when linking to mempool.space is to inform the user of the status (pending/confirmed, and the ETA).

For this use case much of the tx page can be hidden, and made accessible under a `details` button as with the mobile pizza tracker (e.g. hide input address, transaction features, goggles tags, bowtie, change address etc)

By providing the destination address as an anchor tag the recipient could trigger the page to load in a very simplified view, without the server learning that this is the focus. 

An example use case is a bitcoin exchange linking to a pending deposit. In this context, the exchange policy for confirmations before a transaction is "settled" could be provided as an anchor tag, providing additional useful context (a step beyond "confirmed" in the progress bar)

Image

rodribp:
This PR implements a new address-focused view in `/tx/<txid>` page, for payments. As mentioned in #6162 , the scope of this PR is to show a different view when the URL specifies the destination of the payment with the param `#destination=<address>`.
| unconfirmed | confirmed (not settled) | settled |
|-|-|-|
|unconfirmed|confs pending|settled|

Full behaviour (from unconfirmed to settled):

https://github.com/user-attachments/assets/4522de44-64a0-4c91-9d31-867acff3181d

Replacement:

https://github.com/user-attachments/assets/89a823bb-42e5-4ae0-972c-3636fe3bae4e

# Implementation

For this I made a new `payment.component` and `PaymentGuard`, that routes to the new component if the URL Fragment includes the `destination` param. When loading the tx, the component verifies that `destination` is in the list of outputs, otherwise it routes to the normal tx view.

The `confs` param is used to specify the amount of confirmations required to "settle" the payment.

Mobile views:

| confirmed (not settled) | settled |
|-|-|
|confs pending mobile|settled mobile|

The view is inspired in the `tracker.component` to minimize the amount of info displayed and show just the relevant for a user awaiting for a payment. In which the tracking is done by just showing the `trackerBar` and `ETA` if unconfirmed and `Confirmed At <height> <N confirmations>` if confirmed.

### Addition
I added a toggle of the Bitcoin unit in the top-right corner of the amount box as I doubt users know the footer selector exist, especially the mobile users that expect a small payment. Which can also be switched from BTC to sats back and forth by clicking the amount in Bitcoin (Some wallets do this and I think is a good UX and we could use this UX as the ONLY toggle, or hide the top-right toggle in mobile. I would personally strip it out completely).

|desktop|mobile|
|-|-|
|selector view desktop|selector view mobile|

https://github.com/user-attachments/assets/11292fe6-7b81-44a2-a09b-d9c7ce4e46ec

closes #6162


frontend/src/app/components/payment/payment.component.html
```diff
@@ -0,0 +1,149 @@
...
+    <td class="td-width" i18n="transaction.status|Transaction Status">Status</td>
+    <td>
+      @if (replaced) {
+        <span class="badge badge-warning" i18n="transaction.rbf.replaced|Replaced">Replaced</span>
+      } @else if (!tx) {
+        <span class="skeleton-loader"></span>
+      } @else if (!tx.status?.confirmed) {
+        <ng-container *ngIf="(ETA$ | async) as eta; else pending">
+          @if (eta.blocks >= 7) {
+            <span i18n="transaction.eta.not-any-time-soon|Transaction ETA mot any time soon">Not any time soon</span>
```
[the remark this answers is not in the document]
rodribp: same i18n property in `transaction-details.component` and `tracker.component`

orangesurf:
Awesome work

I can't recall why I added the address QR in my issue mockup, seeing it implemented I think this was a mistake and it can be removed.

I think it is worth playing with the design of the confirmation target counter and where it makes sense to display that.

frontend/src/app/route-guards.ts
```diff
@@ -14,10 +14,19 @@ class GuardService {
...
     const preferredRoute = this.router.getCurrentNavigation()?.extractedUrl.queryParams?.mode;
     const path = this.router.getCurrentNavigation()?.extractedUrl.root.children.primary.segments;
-    return (preferredRoute === 'status' || (preferredRoute !== 'details' && this.navigationService.isInitialLoad())) && window.innerWidth <= 767.98 && !(path.length === 2 && ['push', 'test', 'preview'].includes(path[1].path));
+    return (!this.paymentGuard() && (preferredRoute === 'status' || (preferredRoute !== 'details' && this.navigationService.isInitialLoad())) && window.innerWidth <= 767.98 && !(path.length === 2 && ['push', 'test', 'preview'].includes(path[1].path)));
+  }
+
+  paymentGuard(route?: Route, segments?: UrlSegment[]): boolean {
+    const fragmentParams = new URLSearchParams(this.router.getCurrentNavigation()?.extractedUrl.fragment || '');
+    return (fragmentParams.has('destination'));
   }
```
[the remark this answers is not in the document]
rodribp: I don't think this is necessary, the transaction.component doesn't do this neither.

orangesurf:
Screenshots of the latest version for review:

## Unconfirmed
image

## 1 Conf
image

## 5 Confs
image


Notes:
-  IMO `settled` is too strong to describe 1 confirmation_
- I tried to test the confirmation target but couldn't get it to work, I tried [this formatting](http://localhost:4200/tx/fa13dfe11f37465cc12648087e00edf1b421e1d33366bdfba08bf136dd219f6d#destination=bc1qx5hzu45u3r5jvzlfhg64l3fuxdejrhdhw7clck&confs=2)

rodribp:
Latest demo:

https://github.com/user-attachments/assets/403b3c1b-0795-4d1a-8e26-c6b8f9c8ac16

I explored different options for the conf target indicator as shown below but none of them really convinced me rather than this yellow circular progress bar as it let's us increase the cap of `confs` if needed.

|prev design| option (new settled stage in bar) | option |
|-|-|-|
|image|image|image|

| option | option | option |
|-|-|-|
|image|image|image|

orangesurf:
Looks good. What motivated
- The default confirmation target of 3
- The enforcement of a cap at 6

rodribp:
> * The default confirmation target of 3

Hey OS actually on this I just researched on the most common confs in different exchanges and in reality the mode in a research of 9 exchanges I made is 2. being 4 exchanges out of 9. (binance, coinbase, crypto.com and bybit). Minimum is 1 by bitfinex (weird to me) max is 6 from unchained trading desk and "Kucoin".

> * The enforcement of a cap at 6

It was just a UI decision due to the prev space consuming "confs" indicator, I think we can bump it to an "extreme" value such as 24, the exchange/service will control this indicator either way, but the user can change it from the URL and I don't think is good to let the UI display ANY value.

rodribp:
Latest demo (again):
<video src="https://private-user-images.githubusercontent.com/134734899/637353861-403b3c1b-0795-4d1a-8e26-c6b8f9c8ac16.mov?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODcyMTI0NjEsIm5iZiI6MTc4NzIxMjE2MSwicGF0aCI6Ii8xMzQ3MzQ4OTkvNjM3MzUzODYxLTQwM2IzYzFiLTA3OTUtNGQxYS04ZTI2LWM2YjhmOWM4YWMxNi5tb3Y_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwODIwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDgyMFQwNzQ5MjFaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT00N2U0OWNiNDkzYmY1NzkyOTE3ZDBlZDE5ZjUyYzU4MDQ4YmQ5Y2NhZjZjMTZhNjMwNWEzMGM2MmQyODBiOGRlJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZyZXNwb25zZS1jb250ZW50LXR5cGU9dmlkZW8lMkZxdWlja3RpbWUifQ.vcg7TNNAu_FdytH6fEXzV-ICOu22aOec4M_DgTvIkfQ" data-canonical-src="https://private-user-images.githubusercontent.com/134734899/637353861-403b3c1b-0795-4d1a-8e26-c6b8f9c8ac16.mov?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODcyMTI0NjEsIm5iZiI6MTc4NzIxMjE2MSwicGF0aCI6Ii8xMzQ3MzQ4OTkvNjM3MzUzODYxLTQwM2IzYzFiLTA3OTUtNGQxYS04ZTI2LWM2YjhmOWM4YWMxNi5tb3Y_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwODIwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDgyMFQwNzQ5MjFaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT00N2U0OWNiNDkzYmY1NzkyOTE3ZDBlZDE5ZjUyYzU4MDQ4YmQ5Y2NhZjZjMTZhNjMwNWEzMGM2MmQyODBiOGRlJlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZyZXNwb25zZS1jb250ZW50LXR5cGU9dmlkZW8lMkZxdWlja3RpbWUifQ.vcg7TNNAu_FdytH6fEXzV-ICOu22aOec4M_DgTvIkfQ" controls="controls" muted="muted" class="d-block rounded-bottom-2 border-top width-fit" style="max-height:640px; min-height: 200px">
  </video>
  
  But its default confs required is 2, max allowed is 24.

rodribp:
I just researched a bit more to increase sample to 22 and these are the stats that I got:

Stat | #
-- | --
Mode | 3
Average | 3
Minimum | 1
Maximum | 6
25th percentile | 2
Median | 3
75th percentile | 3

This is a bit tricky because most of the big exchanges target to 2 and less confs, within those you find OKX(1), binance(1), bitfinex(1), bitget(1), HTX(1), coinbase (2), crypto.com (2), byBit(2), gemini(3), bitstamp(3), kraken(3), KuCoin(6), .

So if we just consider these "big" exchanges (by trading volume) then default should be 1 if we just consider mode, 2 if we consider average.

Here's the research with sources: https://docs.google.com/spreadsheets/d/1xniN6av_g9O_mYC0lad61-EAv_9Nxfr04EAB_RKviCQ/edit?usp=sharing

orangesurf:
Great research 

6 confirmations was the number from the original bitcoin whitepaper reflecting the number required for an attacker controlling ~10% of hashrate. Today we have mining pools with much more than 10%.

I am very surprised that large exchanges like OKX & Binance use just 1 and that coinbase uses just 2!

I've made a poll out of interest: https://x.com/OrangeSurfBTC/status/2091888716461617437 but I would be tempted to stick with 6 as a sane default.

rodribp:
Binance and other large ones require just one to show the user their balance but two or more to let the user withdraw the one I don't understand is Bitfinex, I haven't tested the exchanges though, just trough documentation

orangesurf:
In my poll 1 is the most common but still only 41.7% of votes and 20% of votes are using 6 confs.

Personally I think we should use a default value of 6 confs, and if app developers want they can change the number of confs to be fewer when providing a link for their customer from their service. 

image

rodribp:
Accelerator demo:

https://github.com/user-attachments/assets/14bd1a03-2138-4b4d-bb78-caefdb16b8bb

https://github.com/user-attachments/assets/a52da47e-69bd-453e-84b8-0370bcc65519

rodribp:
For accelerator demo see: https://github.com/mempool/mempool/pull/6611#issuecomment-5565181214

## Option A:

### Confirmed:
|Desktop|Mobile|
|-|-|
|image|image|

### In mempool:
image

### Replaced: 
|Desktop|Mobile|
|-|-|
|image|image|

---
## Option B:

### Confirmed:
|Desktop|Mobile|
|-|-|
|image|image|

### In mempool:
image

### Replaced: 
|Desktop|Mobile|
|-|-|
|image|image|

---

If the address is not in the list of outputs we redirect to /tx/ page and show up this alert:
image

HEAD is in option B (my preffered)

rodribp:
I made default confs target 20 for a6d697a25266ce3c78774fd1d75f896b7af522ada209b0f6228ea497bc49a46d (liquid return of funds) and bumped max confs to 48

frontend/src/app/shared/common.utils.ts
```diff
@@ -160,6 +160,17 @@ export function seoDescriptionNetwork(network: string): string {
...
+export function addressesMatch(a: string, b: string): boolean {
+  if (!a || !b) {
+    return false;
+  }
+  if (a === b) {
+    return true;
+  }
+  return /^(bc1|tb1|bcrt1|ex1|lq1|tex1|tlq1)/i.test(a) && a.toLowerCase() === b.toLowerCase();
+}
+
```
mononaut: I'd prefer to normalize the user-provided address when we receive it, like we do in e.g. https://github.com/mempool/mempool/blob/8158a230ce44265b14a44d13a785a0ac092ec478/frontend/src/app/components/address/address.component.ts#L201-L203

frontend/src/app/route-guards.ts
```diff
@@ -14,10 +14,19 @@ class GuardService {
...
 }
 
 export const TrackerGuard: CanMatchFn = (route: Route, segments: UrlSegment[]): boolean => {
   return inject(GuardService).trackerGuard(route, segments);
 };
+
+export const PaymentGuard: CanMatchFn = (route: Route, segments: UrlSegment[]): boolean => {
+  return inject(GuardService).paymentGuard(route, segments);
+};
```
mononaut, line 37: ideally this would mirror the implementation of the existing `trackerGuard`.

for example, it should check the route being accessed and explicitly exclude `/tx/status`, `/tx/details`, etc.

frontend/src/app/master-page.module.ts
```diff
@@ -93,6 +94,15 @@ const routes: Routes = [
       {
         path: 'tx',
         component: StartComponent,
+        canMatch: [PaymentGuard],
+        runGuardsAndResolvers: 'always',
+        data: { preload: false, networkSpecific: true },
+        loadChildren: () => import('@components/payment/payment.module').then(m => m.PaymentModule),
```
mononaut: I think this route probably belongs in the `TransactionModule`, scoped to the `/tx/:id` path, rather than as a second `/tx` route up here in the master page module.

That fixes some route precedence issues, and I think also eliminates some of the awkward module imports in the payment module.

frontend/src/app/components/payment/payment.component.ts
```diff
@@ -0,0 +1,952 @@
...
+import { StorageService } from '@app/services/storage.service';
+import { TransactionFlags } from '@app/shared/filters.utils';
+import { Pool } from '@components/transaction/transaction.component';
+
+const DEFAULT_CONFS = 6;
+const MAX_CONFS = 48;
+const TXID_CONFS_DEFAULTS: { [txid: string]: number } = {
+  // liquid attack return of funds
+  'a6d697a25266ce3c78774fd1d75f896b7af522ada209b0f6228ea497bc49a46d': 20,
+};
```
mononaut: is this supposed to be hardcoded? I'd prefer rules based on e.g. output amount thresholds if we really want to vary the default number of confs.

frontend/src/app/components/transaction/transaction.component.ts
```diff
@@ -1317,4 +1329,18 @@ export class TransactionComponent implements OnInit, AfterViewInit, OnDestroy {
...
+  showDestinationAlert(): void {
+    if (!this.tx) {
+      this.destinationAlert = undefined;
+      return;
+    }
+    const destinationAlert = (new URLSearchParams(this.route.snapshot.fragment || '')).get('destinationAlert');
+    // the fragment is user supplied, so only alert on an address this
+    // transaction genuinely does not pay, and never echo anything else
+    if (getRegex('address', (this.network || 'mainnet') as any).test(destinationAlert) &&
+        !this.tx.vout?.some((vout) => addressesMatch(destinationAlert, vout.scriptpubkey_address))) {
```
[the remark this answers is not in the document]
mononaut: I agree with this.

frontend/src/app/components/transaction/transaction.component.html
```diff
@@ -6,6 +6,14 @@
       <app-truncate [text]="rbfTransaction.txid" [lastChars]="12" [link]="['/tx/' | relativeUrl, rbfTransaction.txid]"></app-truncate>
     </div>
 
+    <div *ngIf="this.destinationAlert !== undefined && !(rbfTransaction && !tx?.status?.confirmed)" class="alert alert-mempool" role="alert">
+      <span i18n="transaction.destination.not-in-outputs|Destination alert">The destination address is not in any of the outputs:</span>
```
mononaut: this message could be a bit more straightforward. maybe something like "This transaction does not pay <address>"

frontend/src/app/components/transaction/payment/payment.component.ts
```diff
@@ -0,0 +1,952 @@
...
+
+    this.isValidView = this.destinationInOutputs(tx);
+    return this.isValidView;
+  }
+
+  setAmount(): void {
+    this.amount = (this.tx?.vout || []).reduce((total, vout) => {
+      return this.destinationMatches(vout.scriptpubkey_address) ? total + vout.value : total;
+    }, 0);
+  }
```
mononaut, line 637: I wonder if this should subtract spent amounts from the input side? I can imagine this page being used for *outgoing* payments or self-sends too.

frontend/src/app/components/transaction/payment/payment.component.ts
```diff
@@ -0,0 +1,952 @@
...
+
+  isValidDestination(tx: Transaction): boolean {
+    if (!this.destinationIsAddress()) {
+      return false;
+    }
+
+    this.isValidView = this.destinationInOutputs(tx);
+    return this.isValidView;
+  }
```
mononaut, line 613: it's a bit confusing for an `is...` method to have side effects like this.

but if it *does* set `isValidView` it needs to set it back to false in the early return branch.

frontend/src/app/components/payment/payment.component.ts
```diff
@@ -0,0 +1,952 @@
...
+    this.seoService.logSoft404();
+    this.isLoadingTx = false;
+    return of(false);
+  }
+
+  resetTransaction(): void {
+    const fragmentParams = new URLSearchParams(this.route.snapshot.fragment || '');
+    this.destination = fragmentParams.get('destination') || '';
+    const parsedConfs = Math.min(Math.ceil(parseInt((fragmentParams.get('confs') || this.defaultConfs.toString()), 10)), MAX_CONFS);
+    this.confsRequired = (!isNaN(parsedConfs) && parsedConfs >= 1) ? parsedConfs : this.defaultConfs;
```
mononaut: is this different to how we parse the confs param up in `updateFragmentParams`?

frontend/src/app/components/transaction/payment/payment.component.ts
```diff
@@ -0,0 +1,952 @@
...
+    // spell the target out in links we generate, the txid they point at may
+    // not be the one the default was picked for
+    if (this.confsRequired !== DEFAULT_CONFS) {
+      parts.push(`confs=${this.confsRequired}`);
+    }
+    if (anchor) {
+      parts.push(anchor);
+    }
+    return parts.join('&');
+  }
```
mononaut, line 910: prefer using `URLSearchParams` to build these param strings, like we do in the transaction component etc.

frontend/src/app/components/transaction/payment/payment.component.ts
```diff
@@ -0,0 +1,952 @@
...
+        if (this.replaced) {
+          this.trackerStage = 'replaced';
+        } else if (eta?.blocks === 0) {
+          this.trackerStage = 'next';
+        } else if (eta?.blocks < 3){
+          this.trackerStage = 'soon';
+        } else {
+          this.trackerStage = 'pending';
+        }
+      })
```
mononaut, line 553: a `null` eta (as produced by, e.g. a confirmed transaction in the map above) looks like it falls all the way through and sets `this.trackerStage = 'pending'`?

frontend/src/app/components/transaction/payment/payment.component.ts
```diff
@@ -0,0 +1,952 @@
...
+        },
+        fragment: 'destinationAlert=' + this.destination,
+      });
+    } else {
+      this.router.navigate(baseRoute, {
+        queryParams: {
+          mode: 'details',
+        }
+      });
+    }
```
mononaut, line 725: when this is used for automatic redirects, we need to pass `replaceUrl: true` to avoid polluting the browser history.

frontend/src/app/components/payment/payment.component.scss
```diff
@@ -0,0 +1,263 @@
...
+
+    &.wrap-cell {
+      white-space: normal;
+    }
+  }
+}
+
+.address ::ng-deep .truncate {
+  justify-content: end !important;
+}
```
mononaut: is there any way to achieve this without `::ng-deep`?

frontend/src/app/components/payment/payment.component.scss
```diff
@@ -0,0 +1,263 @@
...
+
+  @media (min-width: 768px) {
+    flex: 0 0 50%;
+    max-width: 50%;
+    margin-bottom: 0;
+  }
+
+  app-truncate ::ng-deep span {
+    word-wrap: normal;
+  }
```
mononaut: is there any way to achieve this without `::ng-deep`?

mononaut (changes requested):
The end result looks great, but the implementation needs a few fixes and a bit more polish.

frontend/src/app/components/payment/payment.component.ts
rodribp: Yeah it was just for this case, but I will get rid of this logic as it's too late, but I like your idea of defining amount thresholds, I think it'd be interesting to research if exchanges vary the confs target based on the amount

frontend/src/app/components/transaction/payment/payment.component.ts
rodribp: Hmmm, I personally cannot think in any use case, like, maybe just change but really most (decent) wallets don't do address reuse, any thoughts @orangesurf ?

orangesurf: Unfortunately the vast majority of transactions on the network currently do reuse addresses, so I think @mononaut has a good point here, could have

amount = Σ (outputs to address) − Σ (inputs spent from address)

Consider transaction with #destination=Addr1 
Addr1 (1 BTC) -> Addr2 (0.25 BTC) & Addr1 (0.75 BTC) 

With the current code the amount shown is 0.75 BTC, but actually the balance of this address decreased by 0.25 BTC, so this could be confusing. 

I think using the net is a good idea. Possibly you could change the accompanying text depending on the net value

- **Normal payment-tracking (Net > 0)** is a pure receive. 
Show "Received X BTC"  sum of outputs to the address

- **Outgoing with change (Net < 0)** is a spend from a reused address. 
Show "Sent X BTC, received Y change back" 

- **Self-send (all outputs return to the address)** is a self-transfer. 
Show "Self-transfer" (net ≈ −fee) 

WDYT @mononaut ?

orangesurf:
@rodribp I agree - option B is better

frontend/src/app/components/transaction/payment/payment.component.ts
```diff
@@ -0,0 +1,951 @@
...
+          let transactionObservable$: Observable<Transaction>;
+          const cached = this.cacheService.getTxFromCache(this.txId);
+          if (cached && cached.fee !== -1) {
+            transactionObservable$ = of(cached);
+          } else {
+            transactionObservable$ = this.electrsApiService
+              .getTransaction$(this.txId)
+              .pipe(catchError(this.handleLoadElectrsTransactionError.bind(this)));
+          }
+          return merge(transactionObservable$, this.stateService.mempoolTransactions$);
```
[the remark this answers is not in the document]
rodribp: Decided to just exclude liquid

rodribp:
Just applied the net-address-balance logic and excluded liquid.network from the #destination route

This address...
|Sent|Received|Self-transfer|
|-|-|-|
|image|image|image|

Just the text above the amount changes

frontend/src/app/components/transaction/payment/payment.component.ts
mononaut: be careful *not* to set `replaceUrl: true` for manual navigation events though (like clicking the "details" button), otherwise the user can't use the browser back button to return to the page they were on before.

mononaut:
~~is there any need to exclude liquid?~~

I guess unblinding would be complicated to handle actually - fine to leave this for a future PR then.

PR #6615, opened 2026-07-03
Fix effective fee row on mobile

natsoni:
Effective fee row value is not correctly right aligned on mobile when transaction is part of a cluter:

| Before  | After |
| Screenshot 2026-07-03 at 12 11 18 | Screenshot 2026-07-03 at 12 11 14 |


mononaut (approved):
Tested ACK @ [5a44d53f]

PR #6624, opened 2026-07-08
Export isWitnessProgram and add PSBT_GLOBAL

natsoni:


mononaut (approved):
utACK @ [4019c0b5]

looks fine

PR #6625, opened 2026-07-08
Update spiral logo

natsoni:
Screenshot 2026-07-08 at 14 38 48

Screenshot 2026-07-08 at 14 39 13


wiz (dismissed):
The SVG should be inline to prevent loading delay

mononaut (approved):
Tested ACK @ [f634b7ce]

PR #6637, opened 2026-07-16
fix: empty incoming transactions chart

rodribp:
This PR solves a rare bug in `incoming-transactions-graph`.

Sometimes it shows an empty graph, like this:
image

`statistics-api.ts` was returning `[]` EVEN when the DB query failed, causing a false 200 [] response cached by NGINX as `warm-cache` expired every 10s. So the false `[]` would be persistent for 10s on mempool.space.

https://github.com/mempool/mempool/blob/87f14d25f5f635b018c777df5cd83bba0b8db510/backend/src/api/statistics/statistics-api.ts#L291-L300

In case of error the frontend would refetch the api three times more every 0.5s, and if still it fails, the pipe inside `dasboard.component.ts` would reassign the response to `[]`.

So I changed the `return []` inside the catch block of the queries for `throw e` and let the handleError inside `statistics.routes.ts` return error.

### DIFF
--- a/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.html
+++ b/frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.html
@@ -1,6 +1,6 @@
 <div class="echarts" *browserOnly echarts [initOpts]="mempoolStatsChartInitOption" [options]="mempoolStatsChartOption" (chartRendered)="rendered()"
   (chartInit)="onChartInit($event)" [style]="{opacity: isLoading ? 0.5 : 1}">
 </div>
-<div class="text-center loadingGraphs" *ngIf="!stateService.isBrowser || isLoading">
+<div class="text-center loadingGraphs" *ngIf="!stateService.isBrowser || (isLoading && !error)">
   <div class="spinner-border text-light"></div>
 </div>
--- a/frontend/src/app/components/statistics/statistics.component.ts
+++ b/frontend/src/app/components/statistics/statistics.component.ts
@@ -1,8 +1,8 @@
 import { Component, OnInit, LOCALE_ID, Inject, ViewChild, ElementRef } from '@angular/core';
 import { ActivatedRoute } from '@angular/router';
 import { UntypedFormGroup, UntypedFormBuilder } from '@angular/forms';
-import { of, merge} from 'rxjs';
-import { switchMap } from 'rxjs/operators';
+import { of, merge, Observable} from 'rxjs';
+import { catchError, retry, scan, switchMap, tap } from 'rxjs/operators';
 
 import { OptimizedMempoolStats } from '@interfaces/node-api.interface';
 import { WebsocketService } from '@app/services/websocket.service';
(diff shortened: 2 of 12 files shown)


nymkappa (approved):
It's been a while I did not work on this part of the code but changes looks sensible to me.
Locally, I've tested the dashboard behavior by manually throwing an error in `statistics-api.ts:$list2H()` and it looks good. API call is retried 3 times, then the loading spinner shows until we get some websocket data. The caching logic change looks good too.

So that would be a tested ACK @ 0a522c9434659af52c4dfdb4dde8585754bd35fc for me

rodribp:
The HEAD that kappa tested was left untouched @ 74155f06f5c43b3251d7e2d4b92e69ae5596b8a5, just pushed the error handling of the custom-dashboard component and rebased.

frontend/src/app/components/statistics/statistics.component.ts
mononaut: this slicing logic interacts badly with defaulting to an empty array of results after multiple failures.

in that case, every subsequent data point received from the `live2Chart` subscription gets `unshifted` into the (initially empty) mempoolStats array, and then immediately sliced back out 😅.

instead, we should either enforce a fixed known number of data points, or borrow the `scan`-based approach from the dashboard component.

frontend/src/app/components/incoming-transactions-graph/incoming-transactions-graph.component.html
```diff
@@ -1,6 +1,6 @@
 <div class="echarts" *browserOnly echarts [initOpts]="mempoolStatsChartInitOption" [options]="mempoolStatsChartOption" (chartRendered)="rendered()"
-  (chartInit)="onChartInit($event)" [style]="{opacity: isLoading ? 0.5 : 1}">
+  (chartInit)="onChartInit($event)" [style]="{opacity: isLoading || isEmpty ? 0.5 : 1}">
 </div>
-<div class="text-center loadingGraphs" *ngIf="!stateService.isBrowser || isLoading">
+<div class="text-center loadingGraphs" *ngIf="!stateService.isBrowser || isLoading || isEmpty">
```
mononaut: we should differentiate `isLoading` (waiting for an API response etc) from `isEmpty`, and both from any explicit error response.

frontend/src/app/components/statistics/statistics.component.ts
```diff
@@ -171,6 +172,13 @@ export class StatisticsComponent implements OnInit {
     document.location.reload();
   }
 
+  errorHandlerPipe(obs: Observable<OptimizedMempoolStats[]>) {
+    return obs.pipe(
+      retry({ count: 3, delay: 1000 }),
+      catchError(() => of([])),
+    );
+  }
+
```
mononaut, line 201: the retry approach is good, but we should throw an explicit error if we still can't get a good response instead of falling back to an empty array, and then handle that appropriately where the observables are consumed.

otherwise it's difficult to distinguish API errors from legitimately empty 200 OK responses.

mononaut (changes requested):
a few minor issues and edge cases to fix up, but otherwise looks good!

rodribp:
On widget in both, error of API and Empty array (Awaiting for websocket):
Screenshot 2026-08-03 at 1 20 11 AM

On error on `statistics.component.ts`:

Screenshot 2026-08-03 at 1 25 05 AM

Empty array:

Screenshot 2026-08-03 at 1 25 48 AM

mononaut (approved):
Tested ACK @ [581eeaca]

confirmed this fixes the issue, and code looks clean and well thought out. LGTM!

PR #6647, opened 2026-07-21
Fix: aligned 'why is this block empty'

### LINKED ISSUE #6602: mis-aligned "why is this block empty" link
when "audit mode" is toggled off, the empty block "why is this block empty" button/link is mis-aligned. 

Image

it should be positioned inside the (empty) block visualization component:

Image

jramos0:
Fix  #6602 

This change adds the existing `block-graph-wrapper` class to the graph container that is rendered when Audit is disabled. The “Why is this block empty?” link uses absolute positioning, but in this branch it did not have a positioned ancestor, so its coordinates were calculated relative to the block’s overall container and it appeared misaligned. By reusing the same wrapper with `position: relative` which the view already uses when Audit is enabled the link is correctly positioned within the visualization without affecting the logic or design of the branch with Audit enabled.


mononaut (approved):
Tested ACK @ [a1022cbb]

PR #6657, opened 2026-07-25
Prevent horizontal page overflow

natsoni:
This PR fixes an horizontal overflow on Liquid base module (see horizontal scroll bar at the bottom of the screen):

https://github.com/user-attachments/assets/aa113cd8-634f-4100-8837-35d9f28b7ef4

It was introduced by the migration to bootstrap 5, but it is not visible on mempool base module because its master page uses `overflow: clip`.


mononaut (approved):
Tested ACK @ [57707181]

PR #6665, opened 2026-07-28
Add sv2 logo to about page

natsoni:
Screenshot 2026-07-28 at 20 20 42

Screenshot 2026-07-28 at 20 20 20

Screenshot 2026-07-28 at 20 19 47


wiz (approved):
looks kinda huge but SV2 is very important for bitcoin so i guess it should be huge
@pavlenex wdyt

PR #6680, opened 2026-08-04
Fix page jump on mobile when scrolling to footer

natsoni:
Fixes a page jump happening in a few places in the app, easily reproducible by scrolling to the footer on the main dashboard. See how the browser is anchoring to a moving row of the Recent Transactions widget, causing the jumps:

https://github.com/user-attachments/assets/fb0f5fcd-ca12-4aeb-a5dc-9e83b683d154

This PR adds `overflow-anchor: none;` property to affected widgets to exclude the moving rows from anchor selection.

`overflow-anchor` seems to be [not yet supported on webkit](https://webkit.org/css-status/#?search=overflow-anchor), but when testing on Safari iOS the issue is fixed.

### DIFF
--- a/frontend/src/app/components/acceleration/accelerations-list/accelerations-list.component.scss
+++ b/frontend/src/app/components/acceleration/accelerations-list/accelerations-list.component.scss
@@ -97,6 +97,10 @@ tr, td, th {
 }
 
 .widget {
+  .acceleration-list {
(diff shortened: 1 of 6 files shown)


nymkappa (approved):
Thank you!

Tested ACK @ 35399957178c586e11f0e25037c449099cf1c279

PR #6686, opened 2026-08-07
Fix Liquid URL from block height search

natsoni:
When searching for a block by height on Liquid, the address bar URL is wrongly rewritten to `https://liquid.network/liquid/block/<hash>` because the block component rewrite the URL with the raw network prefix `'/' + this.network` instead of using `RelativeUrlPipe`.


mononaut (approved):
utACK @ [38102aa3]

PR #6700, opened 2026-08-15
Fix fee delta display on pending acceleration

natsoni:
This PR fixes an issue where the out-of-band fee delta is not displayed anymore in the transaction details of pending accelerated transactions, but still included in the fiat value:

Before:

Screenshot 2026-08-15 at 12 41 47

After:

Screenshot 2026-08-15 at 12 41 49


mononaut (approved):
Tested ACK @ [1c937141]

I had some difficulty reproducing the original bug, because the fee delta shows up fine both in master and on the current prod deploy commit when running the angular dev server locally.

It seems like the original `@if` clause was actually already accidentally correct if we assume `bidBoost`s and `feeDelta`s can't be negative - `accelerationInfo?.bidBoost` is nullish when bidBoost isn't available, so the clause correctly resolves to `(tx.feeDelta > 0)` and the `@if` condition is satisfied.

*but* in the production build there's [a bug in terser](https://github.com/terser/terser/pull/1721) that incorrectly simplifies the LHS to a boolean true or false, which changes the logic and makes the `@if` condition fail, hiding the fee delta.

I've confirmed that the fix in this PR survives terser and so should resolve the issue in prod builds.

PR #6708, opened 2026-08-20
fix: winning block link overflows in stale block view

### LINKED ISSUE #6667: Winning block link overflows in stale block view in mobile
### Description

The block hash of the winning block in the alert of a stale block overflows to the right on mobile.

#### Version

https://github.com/mempool/mempool/commit/e9d6cf8c

### Screenshots

Image

rodribp:
|Before|After|
|-|-|
|image|image|

fixes #6667

### DIFF
--- a/frontend/src/app/components/block/block.component.scss
+++ b/frontend/src/app/components/block/block.component.scss
@@ -13,6 +13,10 @@
   .alert-mempool {
     flex-direction: row;
     flex-wrap: wrap;
+
+    app-truncate {
+      min-width: 0;
+    }
   }
 
   .container-button {


mononaut:
we use this component in a *lot* of different places, so I'm wary of overriding the host layout rules like this.

`min-width: 0;` is probably safe, and I think sufficient to fix the bug.

but `max-width: 100%` feels like it might have unintended consequences elsewhere.

rodribp:
Just moved the style to `block.component.scss` instead as I think is safer

mononaut (approved):
Tested ACK @ [8e36be06]

works great!

PR #6710, opened 2026-08-21
fix: calculate ISP top country by node count

### LINKED ISSUE #6709: Bug on the calculate ISP top country by node count
Image

Image

## The Problem

The **Top country** field on Lightning ISP pages does not show the country with the most geolocated nodes. Instead, it ends up selecting the country of the **first geolocated node** returned by the API.

This affects both the normal ISP page and its preview component.

### Examples

**DigitalOcean — AS14061**

https://mempool.space/lightning/nodes/isp/14061

The UI shows **Netherlands 🇳🇱**, while the API response contained:

* US: 70 nodes
* DE: 30 nodes
* SG: 12 nodes
* NL: 9 nodes

Should be: **United States 🇺🇸**

**Google Cloud — AS396982**

https://mempool.space/lightning/nodes/isp/396982

The UI shows **Switzerland 🇨🇭**, which is the country of the first geolocated node returned by the API.

The response contained:

* US: 40 nodes
* CH: 8 nodes
* BE: 7 nodes
* NL: 4 nodes

Should be: **United States 🇺🇸**

## Root Cause

Both components currently use:

```ts
countries[node.geolocation.iso] =
  countries[node.geolocation.iso] ?? 0 + 1;
```

Because `+` has higher precedence than `??`, this is evaluated as:

```ts
countries[node.geolocation.iso] =
  countries[node.geolocation.iso] ?? 1;
```

Each country is therefore set to `1` on its first occurrence but never incremented, so all countries have the same count and the first one wins.

Affected files:

* `frontend/src/app/lightning/nodes-per-isp/nodes-per-isp.component.ts`
* `frontend/src/app/lightning/nodes-per-isp/nodes-per-isp-preview.component.ts`

jramos0:
Fix #6709 

Fixed the top country calculation by correctly incrementing the node count for each country using nullish coalescing with the proper operator precedence. The same fix was applied to both the standard ISP component and its preview.

image
image


mononaut (approved):
Tested ACK @ [01d5365a]

PR #6712, opened 2026-08-21
fix correct 1-year block capacity calculation

### LINKED ISSUE #6711: Wrong Acceleration stat for (1 year) block capacity %
## The problem

On the [Accelerator Dashboard](https://mempool.space/acceleration), the **Acceleration stats (1 year)** card understates the `Total vSize` **% of blocks** value.

With the `1y` period selected, the UI currently displays `0.00% of blocks`, and always gonna show `0.00% of blocks` even though the current data should round to `0.01%`. **It's insignificant right now**, but once the accelerator starts being used more widely, it will become noticeable because its value will never change since its formula is incorrect and, instead of basing the calculation on one year, it uses ~30 years cause

Image

The frontend estimates the number of blocks in the selected period before calculating:

```ts
stats.totalVsize / (1_000_000 * blocksInPeriod) * 100

For 1y, it currently uses:

30.5 * 144 * 365

144 is the estimated number of Bitcoin blocks per day, based on the 10-minute target block interval. While 30.5 * 144 is a reasonable approximation for one month, multiplying it by 365 makes the 1-year denominator equivalent to approximately 30.5 years of blocks.

So I'll make the pr to fix it

jramos0:
## What changed

Fixed the estimated block count used to calculate the `Total vSize` percentage for the `1y` acceleration stats period.

The previous calculation used an extra `30.5` multiplier, effectively treating one year as 30.5 years of blocks and causing the value to round down to `0.00%`.

closes #6711

### DIFF
--- a/frontend/src/app/components/acceleration/acceleration-stats/acceleration-stats.component.ts
+++ b/frontend/src/app/components/acceleration/acceleration-stats/acceleration-stats.component.ts
@@ -69,7 +69,7 @@ export class AccelerationStatsComponent implements OnInit, OnChanges {
         this.blocksInPeriod = 30.5 * 144;
         break;
       case '1y':
-        this.blocksInPeriod = 30.5 * 144 * 365;
+        this.blocksInPeriod = 144 * 365;
         break;
       case 'all':
         this.blocksInPeriod = Infinity;


mononaut (approved):
Tested ACK @ [b5e9802c]

PR #6716, opened 2026-08-24
Fix calculator 🧮

natsoni:
Fixes a (new?) slight issue on the calculator: it gets on a broken state whenever the live price updates:

https://github.com/user-attachments/assets/61e82490-736a-4fb5-84fe-78e4385e58ac


mononaut (approved):
Tested ACK @ [fc54d4a9]

good catch! I was able to reproduce the issue by faking conversions events, and then confirmed that the PR fixes it.

PR #6719, opened 2026-08-26
[OPS] add freebsd + arm support to the install script

rodribp:
The problem with the freebsd + arm setup is caused when installing node through nvm. It tries to build the OpenSSL bundled dependency on Optimized assembly for linux x86, therefore it throws an err. By forcing openssl to be built in C it completes.

The other issue it solves is an upstream `napi-rs` FreeBSD + ARM not suported, because the `index.js` file that is generated to export rust-gbt gates freebsd OS to support `x64` only, excluding all other archs in freebsd. e.g.

``` JavaScript
  case 'freebsd':
    if (arch !== 'x64') {
      throw new Error(`Unsupported architecture on FreeBSD: ${arch}`)
    }
    localFileExisted = existsSync(join(__dirname, 'gbt.freebsd-x64.node'))
    // ....
```

However cargo does compile the *.node file so my solution is to make an `index.js` wrapper that first tries to import a `<os>-<arch>.node` and fallbacks to `index.napi.js` if not found.

``` JavaScript
const local = join(__dirname, `gbt.${process.platform}-${process.arch}.node`)
if (existsSync(local)) {
  nativeBinding = require(local)
}
if (!nativeBinding) {
  nativeBinding = require('./index.napi.js')
}
```

I verified bumping `napi-rs` doesn't solved the compatibility check.


rust/gbt/wrapper.js
```diff
@@ -0,0 +1,11 @@
...
+ *  even though napi-rs *.node builds and runs so we wrap it here
+ */
+const { existsSync } = require('fs')
+const { resolveBinding } = require('./binding-name.js')
+
+const nativeBinding = require(resolveBinding(__dirname, existsSync))
+
+const { GbtGenerator, GbtResult } = nativeBinding
+module.exports.GbtGenerator = GbtGenerator
+module.exports.GbtResult = GbtResult
```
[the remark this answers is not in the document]
rodribp: Replicating the same pattern as the generated `index.js` file

``` JavaScript
const { GbtGenerator, GbtResult } = nativeBinding

module.exports.GbtGenerator = GbtGenerator
module.exports.GbtResult = GbtResult
```

production/install
```diff
@@ -26,6 +26,13 @@ case `uname -s` in
 
 esac
 
+NODE_CONFIGURE_FLAGS="--shared-zlib"
+case `uname -m` in
+    arm64|aarch64)
+        NODE_CONFIGURE_FLAGS="${NODE_CONFIGURE_FLAGS} --openssl-no-asm --dest-cpu=arm64"
```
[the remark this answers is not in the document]
rodribp: To be tested nvm install `version` works on debian arm64 | aarch64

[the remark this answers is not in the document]
rodribp: The debian path doesn't work

production/install
```diff
@@ -1620,8 +1596,32 @@ fi
 echo "[*] Installing crontabs"
 case $OS in
     FreeBSD)
-        osSudo "${ROOT_USER}" crontab -u "${BITCOIN_USER}" "${MEMPOOL_HOME}/${MEMPOOL_REPO_NAME}/production/bitcoin.crontab"
```
wiz, line 1623: if you're no longer using these files why not delete them?

production/install
```diff
@@ -26,6 +26,13 @@ case `uname -s` in
 
 esac
 
+NODE_CONFIGURE_FLAGS="--shared-zlib"
+case "${OS}:$(uname -m)" in
+    FreeBSD:arm64|FreeBSD:aarch64)
+        NODE_CONFIGURE_FLAGS="${NODE_CONFIGURE_FLAGS} --openssl-no-asm --dest-cpu=arm64"
```
wiz: you set arm64 even for aarch64 ?

wiz: you set arm64 even for aarch64 ?

wiz (approved):
ACK on the prod install script with a few comments, need @junderw to review rust bits

production/install
rodribp: Just tested a rebuild without the --dest-cpu flag and I confirm is redundant, node maps aarch64 to arm64 automatically

junderw (changes requested):
https://github.com/mempool/electrs/blob/9b14438a96e7064de4c0da73fcfdf1bdc782136c/.github/workflows/ci.yml#L44-L85

Please add a CI job that actually runs the full test suite in freebsd arm64.

(The job in electrs does a lot of weird stuff to move the cache layer into the VM, your job probably won't need to do that).

rodribp:
Hi @junderw , I added the CI Job you requested

However the node installation doesn't match the install script path using nvm, but FreeBSD's packages `node24` and `npm-node24` which installs node v24.18.0 (FreeBSD's `pkg` doesn't offer version selection), and the install script installs v24.13.0.

The nvm build takes too long otherwise because it builds from source in an emulated `aarch64` VM inside `mempool-ci's linux x86 host`, [see this run took 3h](https://github.com/mempool/mempool/actions/runs/33835277236/job/100906398533?pr=6719#logs), so I don't think it would finish within the 6h job limit.

junderw (approved):
Very hacky, and will require a lot of manual maintenance... but it works.

And it allows us to support a platform that NAPI-RS doesn't support, so as long as we ensure that the Rust code isn't mis-compiling (which we probably are checking well enough by doing the Unit tests inside the ARM FreeBSD VM) it should be fine.

LGTM

PR #6725, opened 2026-08-29
Remove clipboard and ngx-infinite-scroll dependencies

knorrium:
The `clipboard` dependency is not needed as we already call the native `navigator.clipboard.writeText` function

`ngx-infinite-scroll` can be refactored using native listeners while retaining the same logic to fetch more transactions and mined pool blocks that is used in production. 

Bonus: uuid was being imported for Cypress tests for no reason


wiz (approved):
removing npm deps, hell yeah
bet you can't do that again

mononaut (approved):
Tested ACK @ [fed93c1a]

PR #6741, opened 2026-09-04
Fix Liquid reserves history

natsoni:
This PR fixes an issue in the Liquid audit monthly reserves logic.

The previous query grouped federation UTXOs by their creation month so reserves created in a previous month were not counted into the monthly balance. This causes a false positive in the `Reserves Deficit` section.
The issue did not show up before because, up until recently, federation UTXOs had a 4032 blocks or 28 days timelock, meaning that unspent UTXOs at end of month were definitely created in the same month. Since the new federation UTXOs timelock has increased to 8064 the issue can show up more frequently.

The new query uses monthly events to compute the reserve history, by adding UTXOs in their creation month and subtracting them in their spend / expiry month.

Tested against a local Liquid database: the correct reserves balance for August is 4216.61835901 BTC, not 3983.08377182 BTC as currently shown on prod.

### DIFF
--- a/backend/src/api/liquid/elements-parser.ts
+++ b/backend/src/api/liquid/elements-parser.ts
@@ -539,7 +539,7 @@ class ElementsParser {
 
   /** @asyncUnsafe */
   public async $getPegDataByMonth(): Promise<any> {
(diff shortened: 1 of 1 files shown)


mononaut (approved):
Tested ACK @ [ba6e707a]

PR #6745, opened 2026-09-07
Fix OP_RETURN toggle on address page

natsoni:
On the address page, expanding an OP_RETURN also expands OP_RETURNs at the same output index in other transactions:

https://github.com/user-attachments/assets/9202ad36-0604-4bcc-a845-3476833f3763

This PR fixes this by using the same approach as inscriptions and runestones (include the txid in the toggle key).


mononaut (approved):
Tested ACK @ [20429168]

PR #6747, opened 2026-09-08
fix: skip fee calculation for coinbase txs in $convertTransaction

### LINKED ISSUE #6706: bitcoinCoreApi instance fails when trying to get coinbaseTx of stale blocks
If you want to get the coinbaseTx of an stale block using `await bitcoinCoreApi.$getCoinbaseTx(block.id)`, it fails because `bitcoin-api.ts $getRawTransaction()` calls `$convertTransaction()`.

https://github.com/mempool/mempool/blob/3bf0d1ae76ee5335d3d358014d7726595d4e2400/backend/src/api/bitcoin/bitcoin-api.ts#L54

So it passes a transaction with `confirmations = 0` and guards positive to execute `$appendMempoolFeeData()` in:

https://github.com/mempool/mempool/blob/3bf0d1ae76ee5335d3d358014d7726595d4e2400/backend/src/api/bitcoin/bitcoin-api.ts#L349-L351

Which throws the error: `No such mempool or blockchain transaction. Use gettransaction for wallet transactions.`

---
I haven't verified it but I suspect this is also the reason why mempool.space shows "transaction not found" when trying to see at the coinbaseTx of an stale block.

To verify click on the bottom-left tx of the stale block: https://mempool.space/block/00000000000000000001a7b86759d1ef4795cbebf29c2a480c2dd6db3c99c253

rodribp:
closes #6706 

In #6648 I tried to use bitcoinCoreApi to get the coinbaseTx of a stale block, which bitcoin core does recognize, but `$convertTransaction` fee calculation failed on `getMempoolEntry` from `appendMempoolFeeData` since the tx is not confirmed nor in mempool ofc, so if the transaction is coinbase we skip the fee calculation at all.

### DIFF
--- a/backend/src/api/bitcoin/bitcoin-api.ts
+++ b/backend/src/api/bitcoin/bitcoin-api.ts
@@ -367,6 +367,10 @@ class BitcoinApi implements AbstractBitcoinApi {
       };
     }
 
+    if (esploraTransaction.vin[0].is_coinbase) {
+      return esploraTransaction;
+    }
+
     if (addPrevout) {
       try {
         esploraTransaction = await this.$calculateFeeFromInputs(esploraTransaction, false, lazyPrevouts);


mononaut (approved):
Tested ACK @ [ac3ff41b]

I reproduced the scenario on regtest and confirmed this fixes the issue.

PR #6748, opened 2026-09-08
Feat: Group address tx by block

jramos0:
## What changed

This PR changes the address transaction history to group confirmed transactions by block instead of displaying the history as a single flat list.

### Transaction grouping

- Confirmed transactions are grouped by block height.
- Each block group shows:
  - the block height, linked to the block page;
  - the block timestamp;
  - the number of address transactions included in that block.
- Unconfirmed transactions remain in a separate group.
- Transactions inside each group keep their existing transaction-level information and behavior.

### Activity by block

A new **Activity by block** visualization provides a compact overview of the blocks in which the address was active.

Each block displays address-specific information instead of the usual block metrics:

- number of transactions involving the address;
- whether the address sent, received, or both sent and received funds in that block;
- the net value change for the address;
- block timestamp.

Unconfirmed activity is represented separately from confirmed blocks.

Gaps between blocks are intentionally represented instead of implying that the displayed blocks are consecutive, since the address may have no activity for many blocks between two entries.

The block activity data is derived from the same address transaction history so the visual summary and transaction list remain consistent as the history changes.

The current address transaction list makes it difficult to tell which transactions were confirmed together without opening individual transactions or comparing timestamps manually.

Grouping the history by block makes this relationship explicit.

The block visualization also provides a higher-level view of the address activity, making it easier to quickly identify:

- which blocks contain activity for the address;
- how many transactions occurred in each block;
- whether funds were primarily sent or received;
- the net change associated with each block.

The transaction list remains available below for inspecting the individual transactions.

The main functionality is implemented, but some UI details and interactions may still be refined before it is ready for  merge


jramos0:
@orangesurf I Added another iteration of the address transaction grouping more visual as you suggested.

The main addition is a new **Activity by block** overview above the transaction list. Instead of only separating the transaction history with block headers, blocks are now represented visually using an adaptation of the existing mempool block design.

Each block shows address-specific activity:

- number of transactions in the block;
- sent / received activity;
- net value change;
- block timestamp.

Unconfirmed transactions are represented separately, and gaps between displayed blocks indicate that the address had no activity in the omitted block range.

I also kept the grouped transaction list below so the block overview works as a summary while the individual transactions remain available for inspection.

There are still some UI details I want to refine, particularly around the collapse/expand behavior, but the main block-based direction is now implemented

https://github.com/user-attachments/assets/d7c10309-d85c-40e3-b4d2-7c56ce776478


https://github.com/user-attachments/assets/5ad4bc84-d672-4a96-85ea-38e63b5e60af

orangesurf:
Nice work, I think it's a direction worth continuing to explore! I think the design of the blocks could do with a little polishing (size and position of block numbers and text elements on blocks) but it adds a nice visual element to what is otherwise quite long text list, and it looks like a convenient way to explore the history

jramos0:
@orangesurf I made few improvents to look cleaner about the text and the size, besides this changes what do you think we need to improve more?

image

image

orangesurf:
I think the block design still looks cramped with too small text in the latest screenshot, which you may have fixed in subsequent commit? Compare the existing blocks design with this one

image
image

PR #6753, opened 2026-09-09
Hide liquid incident alert when network restarts

natsoni:
This PR hides the Liquid incident banner once the tip is > 4050335 and the network resumes.


mononaut (approved):
utACK @ [788f2cfc]

looks correct

PR #6762, opened 2026-09-10
optimize liquid audit

mononaut:
significantly speeds up the bulk phase of the liquid federation audit by writing a single UPDATE at periodic checkpoints instead of updating every tracked UTXO separately on every block, and holding tracked UTXOs in memory instead of querying everything again on every block.

### DIFF
--- a/backend/src/api/liquid/elements-parser.ts
+++ b/backend/src/api/liquid/elements-parser.ts
@@ -18,6 +18,18 @@ interface FederationAddress {
   blocknumber: number;
 }
 
+interface AuditUtxo {
+  unspent: number;
+  txid: string;
+  txindex: number;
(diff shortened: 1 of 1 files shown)


natsoni:
Tested ACK @ [01e7231]

Changes look good, and I successfully indexed a liquid instance from scratch at least ~30% faster than on master 👍

natsoni (approved):

mononaut:
Thanks for review + testing!

I'm surprised it was only a 30% speedup, but I guess the gains are concentrated towards recent history where there were the most outstanding UTXOs?

wiz:
Thanks for fixing this. While it might only make a 30% faster sync on a macbook pro, this is probably a huge improvement for our prod intel servers

PR #6769, opened 2026-09-11
Fix false stale block banner on latest blocks opened by height

### LINKED ISSUE #6768: A canonical block shown as stale when opened by height
## the problem

It's a tiny bug when you opening a recent block by height (`/block/<height>`) can incorrectly mark it as stale after the next block is mined.

The stale banner says the block was replaced, but the replacement hash is the block's own hash. Reloading the page or navigating to the block by hash shows it correctly.

Image

Image

## Steps to reproduce

1. Open a recent block by searching for its height, so the URL remains `/block/<height>`.
2. Wait for the next block to be mined.
3. The current block may be marked as stale even though it is still part of the main chain.

This only affects recent blocks loaded from the `latestBlocks` cache.

## The  cause

When a block is loaded from cache by height, `this.blockHash` is not updated.

The websocket reorg check then compares incoming blocks against a stale or undefined hash. When it finds the current block at the same height, it incorrectly treats it as a replacement and marks the block as stale.

jramos0:
## What changed

Fixes stale-block detection when a recent block is opened by height.

- Keeps `this.blockHash` in sync with the loaded block id across all block-loading paths.
- Only treats a same-height websocket block as a replacement when its id differs from the currently displayed block.

## Why

Blocks loaded from `latestBlocks` by height did not update `this.blockHash`. On the next block update, the reorg detector could therefore compare against a stale or undefined hash and incorrectly mark the canonical block as stale.

This could also make the block height link reference a previously loaded block.

Closes #6768

### DIFF
--- a/frontend/src/app/components/block/block.component.ts
+++ b/frontend/src/app/components/block/block.component.ts
@@ -174,7 +174,7 @@ export class BlockComponent implements OnInit, OnDestroy {
                 this.fees = block.extras.reward / 100000000 - this.blockSubsidy;
               }
             }
-          } else if (block.height === this.block?.height) {
+          } else if (block.height === this.block?.height && block.id !== this.block?.id) {
             this.block.stale = true;
             this.block.canonical = block.id;
             this.fetchCanonicalBlock();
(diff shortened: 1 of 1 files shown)


mononaut (approved):
Tested ACK @ [d124446a]

